PR #16, opened 2017-03-22
created typescript definitions file

taljacobson:
for your consideration a typescript type definition file.
i put together a d.ts file for your module API, using your documentation.  not all is fully flushed out like database and storage methods.
this should not affect this project's flow type checking in any way but expand its usefulness to any typescript user


Salakar:
@taljacobson thanks for this, it's a good start; would it be worth copying the firebase web sdk type def files over and modifying to match, would certainly be easier as the majority of the api has been built to match the web SDK with only minor changes/additions here and there, much less work :)

https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/firebase

Salakar:
Although having said that, the firebase one I linked doesn't look like a complete definition

index.d.ts
```diff
@@ -0,0 +1,634 @@
...
+       * Sets a key/value pair of data on the current user.
+       */
+      setUserProperty(name: string, value: string): void
+    }
+
+    interface User {
+      /**
+       * The user's display name (if available).
+       */
+      isplayName: string | null
```
Salakar: Typo here :)

Salakar:
@taljacobson is this ready to go?

taljacobson:
still playing around with it, mostly good but didn't have much time to fully flush it out yet, added some un blocking any types,
but should be usable

PR #19, opened 2017-03-24
Make database on method return successCallback

greena13:
As per https://firebase.google.com/docs/reference/js/firebase.database.Reference, the `on` method should return the `successCallback`.

### DIFF
--- a/lib/modules/database/reference.js
+++ b/lib/modules/database/reference.js
@@ -99,16 +99,21 @@ export default class Reference extends ReferenceBase {
    * @param successCallback
    * @param failureCallback
    * @param context TODO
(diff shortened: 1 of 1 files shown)


greena13:
On second thought, this may not work with the primary purpose for returning the callback: ease of unbinding it later. 

Will require more testing.

Salakar:
@greena13 this is as it should be though and doesn't break anything existing so i'm happy to merge this as is to be honest as currently this returns the promises from db.on - which is wrong

Salakar:
Closing this as I have added this in my local copy -  i'm working on transaction support atm.

PR #30, opened 2017-04-04
add version to .flowconfig

Leeds-eBooks:
The consequence of this is that **flow will not run in this project** unless you have `flow@0.40.x` installed. One solution to this is adding `flow-bin` to this package's devDependencies.


Leeds-eBooks:
I'm not 100% sure what will happen if someone's react native project is using a different version of flow. Maybe we need to think further about the consequences of this PR. One alternative would be to install `flow-bin@0.40.x` and *not* merge this PR... then contributors to this package would just use the local `flow-bin` to dev against. What do you think?

chrisbianca:
I'd be inclined to follow what has been done in the main RN project and add `flow-bin` to RN-Firebase's devDependencies.  At least you ensuring that when developing with the library, everybody is using the same version of Flow and working to a consistent standard?

Leeds-eBooks:
Yep, sounds good, although the main RN project also has the `[version]` entry in `.flowconfig`. I've tried to look for advice for package authors on this subject but no luck. I think I'll close this PR and submit a new one with the `flow-bin` dependency (matching the `[version]` specified in the latest stable RN release. Sound ok?

Leeds-eBooks:
Ok 👍

Salakar:
@Leeds-eBooks thanks for the PR!

PR #34, opened 2017-04-04
Fixing various flow errors

Leeds-eBooks:
The second commit changes a few pieces of logic very slightly to appease flow; hopefully I've kept these changes local to their functions.

### DIFF
--- a/lib/modules/database/reference.js
+++ b/lib/modules/database/reference.js
@@ -157,19 +157,25 @@ export default class Reference extends ReferenceBase {
    * @param onComplete
    * @param applyLocally
    */
-  transaction(transactionUpdate, onComplete?: () => any, applyLocally: boolean = false) {
+  transaction(
+    transactionUpdate: Function,
+    onComplete?: (?Error, any, ?Snapshot) => any,
+    applyLocally: boolean = false
+  ) {
     if (!isFunction(transactionUpdate)) return Promise.reject(new Error('Missing transactionUpdate function argument.'));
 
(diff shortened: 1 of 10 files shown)


lib/modules/database/reference.js
```diff
@@ -157,19 +157,25 @@ export default class Reference extends ReferenceBase {
...
+          if (typeof onComplete === 'function') {
+            onComplete(error, committed, null);
+          }
           return reject(error);
         }
 
         const snapshot = new Snapshot(this, snapshotData);
 
-        if (isFunction(onComplete)) {
+        if (typeof onComplete === 'function') {
```
Salakar: Not nit-picking here but any reason you changed this one? There's one above also on 165

Salakar:
Generally looks good to me, just the one comment above I was unsure about, will merge in the morning and run the tests.

lib/modules/database/reference.js
Leeds-eBooks: Yeah, I didn't want to but flow just wouldn't pick up that `isFunction()` guarantees that `onComplete` is a function. `0.40` is several weeks old now so it could be fixed in the latest version of flow, but we have to follow what's used in rn. Sometimes you have to be a bit more explicit and a bit less DRY to make flow happy.

Salakar: Make sense. 👍

Salakar:
All the tests seem to be passing 👍

Leeds-eBooks:
Ace!

PR #36, opened 2017-04-04
Remove redundant platform-specific index files

Leeds-eBooks:
Allows flow and other code intelligence tools to follow the import trail into the lib.


Salakar:
@Leeds-eBooks danke

PR #59, opened 2017-04-19
fixed CrashReporting on iOS

alexkuttig:
CrashReporting was not working on iOS, because the CrashReporting ios files were not included in the project and the build target. firebase.crash().log resulted in an "cannot read property 'log' of undefined" error. 

I added RNFirebaseCrash.h and RNFirebaseCrash.m to ios project and build target.
Renamed lib/modules/crash class name from Analytics to Crash

Again, since I'm no iOS dev, please have a look before you merge :)

### DIFF
--- a/ios/RNFirebase.xcodeproj/project.pbxproj
+++ b/ios/RNFirebase.xcodeproj/project.pbxproj
@@ -7,6 +7,7 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
+		29C199451EA7A851007B6BF8 /* RNFirebaseCrash.m in Sources */ = {isa = PBXBuildFile; fileRef = 29C199441EA7A851007B6BF8 /* RNFirebaseCrash.m */; };
 		D90882D61D89C18C00FB6742 /* RNFirebaseMessaging.m in Sources */ = {isa = PBXBuildFile; fileRef = D90882D51D89C18C00FB6742 /* RNFirebaseMessaging.m */; };
 		D950369E1D19C77400F7094D /* RNFirebase.m in Sources */ = {isa = PBXBuildFile; fileRef = D950369D1D19C77400F7094D /* RNFirebase.m */; };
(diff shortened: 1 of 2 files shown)


Salakar:
Another one, awesome. Seems ok to me, @chrisbianca can you confirm?

chrisbianca:
Looks good to me too.  I'm importing using the Podfile which is why this hadn't been spotted previously.

Leeds-eBooks:
Sorry, could you guys explain – do I need to do something in my project to get this working? Or will an npm upgrade to `react-native-firebase` (when you've published an update) just work? Thanks!

alexkuttig:
It should work after an update is published. You then have to upgrade your installation of react-native-firebase and make a new build. Do you have any problems with it?

Leeds-eBooks:
@alexkuttig thanks, no it's just wherever a podfile is concerned, and native ios packages, I get totally confused. Should I also updated my pods?

alexkuttig:
@Leeds-eBooks I can totally understand :) 
I think when you are linking react-native-firebase via Podfile (https://github.com/invertase/react-native-firebase/blob/master/docs/installation.ios.md 2.1), this fix should not affect you in any way. Only if you are linking the project via npm link (https://github.com/invertase/react-native-firebase/blob/master/docs/installation.ios.md 2.2) you should update your react-native-firebase with the next npm release. 
@chrisbianca can you confirm this?

chrisbianca:
That's pretty much how I understand it too!  The Podfile loads all the files from the directory, so this fix shouldn't make any difference.

Leeds-eBooks:
Thanks guys! Just to be clear:

> Only if you are linking the project via npm link (https://github.com/invertase/react-native-firebase/blob/master/docs/installation.ios.md 2.2) you should update your react-native-firebase with the next npm release.

I used `react-native link`. So what do I need to do, just use the latest version of `react-native-firebase`, or do I need to run some `pod` command?

chrisbianca:
You should just be able to use the latest version of `react-native-firebase`, in theory :D

PR #67, opened 2017-04-23
[tests] Move test suite into same repo

greena13:
As discussed in gitter, moving the test suite into the main application (but excluding it from what's published to npm via an `.npmignore` file) presents many advantages:

* Allows keeping the version of `react-native-firebase` with the version of the test suite that should be run against it
* Prevents having to wait until a new release to npm before linking the test suite to a new version of the `react-native-firebase` code
* Allows pull requests to contain the additional or updated tests necessary to cover the accompanying code changes

I've tackled this pretty naively:

* copied the latest version of the test application code to `/tests/` directory
* Removed `.git` directory from the files copied across
* Added `tests` to `.npmignore` to prevent the tests being published to npm
* Updated any links or installation instructions that refer to the tests repo to point to the `/tests` directory of this repo

The test suite seems to build and run the same as when it was a separate repo. Although I am currently having problems build the Android app from the terminal - builds fine in Android Studio. Would be interested to see if anyone can reproduce the error that I am getting:

```
Failed to notify ProjectEvaluationListener.afterEvaluate(), but primary configuration failure takes precedence.
java.lang.IllegalStateException: buildToolsVersion is not specified.
```

**Warning:** There is some suggestion that yarn ignore the `.npmignore` file and copies the `test` directory into itself when running `yarn install` so `/test/node_modules/react-native-firebase/test` contains `test` again. This cause build failures. `npm install` works, though.


greena13:
**Update:** Building the test app from the command line has just started working. I suspect some broken build artefact or setting became invalidated in the cache and was rebuilt.

There's now no known problems with the migration of the tests in the main application (as long as you use `npm install` and **not** `yarn install` to install the dependencies of the test application.

Salakar:
Nice, makes sense, saves setting up weird watchers and stuff. Merging and will test later today while I do a few things.

greena13:
Unfortunately, I believe that watchers are still necessary as `npm install` copies directories when you use the `file:../` syntax. I'd be very interested to know if there is any alternative or a flag I am missing that means you don't have to keep two directories in sync.

Salakar:
So I got android and ios native sides correctly loading from the actual directories of RNFirebase rather than the copied version `file:..` does.

Just an issue getting the JS to require it, packager isn't liking it:
image

greena13:
Is there any reason you are using `./../../` instead of `../../`? Or is that just a find and replace artefact?

I'd suggest trying the latter to see if it works - if it's like webpack, the packager will use how a path starts to decide how it should be handled. I'm guessing here, though.

Issue #45, opened 2017-04-10
Crashing error while startup on Android (io.invertase.firebase.auth.RNFirebaseAuth.init)


alekseykarpenko:
Hi guys! Thanks for your work!

Please put me in the right direction.. Im trying to integrate Firebase, but when i'm done with all initializations, built my app and trying to start it on my device, ive got a crashing error at
  ```io.invertase.firebase.auth.RNFirebaseAuth.init```:
![screen shot 2017-04-10 at 6 05 11 pm](https://cloud.githubusercontent.com/assets/4648983/24868292/44ae067e-1e18-11e7-9100-af7aa263029c.png)

Appreciate if u could help me with this.

RN v0.41.0
react-native-firebase v1.0.0-alpha12

Ehesp:
Hi @alekseykarpenko

If you remove all JS firebase references does the app build? Can you also show where you initialise Firebase in your app please.

alekseykarpenko:
@Ehesp when i clear all JS related to firebase including ```import``` – the app still crashes.

Ehesp:
Are you using `firebase.auth()` anywhere?

alekseykarpenko:
@Ehesp Nope, that's the thing 😄  I was going to use only cloud messaging, but even when i remove all [it's setup](https://github.com/invertase/react-native-firebase/blob/master/docs/installation.android.md#3---cloud-messaging-optional), crashlog doesn't change

alekseykarpenko:
Auth is used in [RNFireBasePacPackage.java:38](https://github.com/invertase/react-native-firebase/blob/master/android/src/main/java/io/invertase/firebase/RNFirebasePackage.java#L38), and there is a crash

alekseykarpenko:
@Ehesp, **important update:** when i comment that line with calling ```new RNFirebaseAuth()```, the next line with ```new RNFirebaseDatabase()``` crashes with the same error ```java.lang.NoSuchMethodError```. Maybe i need to update RN, Android SDK, or other environment tools?

Ehesp:
cc @Salakar didn't we have this issue? 

@alekseykarpenko I'd recommend deleting your node modules and trying again, with a full app reinstall.

Salakar:
This normally tends to happen when react isn't correctly requiring its androik sdk correctly from node modules and is instead referencing the jcenter/maven version which was long deprecated (its on version .29 or something)

Make sure you're correctly including react in your gradle files, if you can show them that might be helpful

[...]

PR #69, opened 2017-04-26
remove firebase and gms dependency version numbers

v1kku:
Fix for #45 . See [explanation](https://github.com/invertase/react-native-firebase/issues/45#issuecomment-297338621)


Ehesp:
Are there any downsides to doing this? As I understand it, this will install the latest version of FB on a users device, which may have new features unsupported yet.

Don't really have enough experience to make a call on this tbh!

chrisbianca:
Personally, I'm not sure I like this approach as it leaves the library open to issues should there be any breaking changes in the Firebase SDK.  Having a version specified, shows that it has been tested and known to work with that version and whilst it is a pain to keep bumping this, it means we won't have unforeseen issues.

Dependency resolution/conflict is generally a pain as it is different for any app due to the vast number of dependencies out there.  In my mind, it should be up to the app author to resolve those issues, not to leave the project open-ended to future unexpected issues.

Salakar:
Agree with @chrisbianca on this, better to use a known working version that we have tested in our test apps rather than assume the latest versions of firebase / gms will just work.

End users can still decide what version to use should they need to via [`excludeGroup`](https://medium.com/@suchydan/how-to-solve-google-play-services-version-collision-in-gradle-dependencies-ef086ae5c75f) in their `build.gradle` file.

Closing this as do not merge.

PR #111, opened 2017-05-12
launchMode needs to be set on <activity> instead of <application>

fkoester:
`launchMode` is not a valid property on the [`<application>` element](https://developer.android.com/guide/topics/manifest/application-element.html).

It must be set on the [`<activity>` element](https://developer.android.com/guide/topics/manifest/activity-element.html) instead.


Salakar:
Well spotted, thanks.

Issue #105, opened 2017-05-10
downloadFile(...).then is a Promise instead of a function returning a Promise


Damnum:
This line seems to be wrong: https://github.com/invertase/react-native-firebase/blob/master/lib/modules/storage/task.js#L55

then should be a function not a promise right?

I changed it to

`this.then = promise.then.bind(promise);`

and it works

davidtrogers:
@Damnum I noticed the same thing yesterday when trying to upload files. The [type def](https://github.com/invertase/react-native-firebase/blob/80b519158fa0132d9126366fe1b07d8da8ed83b5/lib/modules/storage/task.js#L45) will need to be changed too to match up with catch:
``` javascript
  then: Promise<*>
  catch: () => Promise<*>
```

andfk:
+1

Salakar:
@Damnum @davidtrogers the fix for this is landing in v2. In the meantime you can use the callbacks instead of promises as a workaround.

andfk:
@Salakar FYI and for interested people, I ended resolving the promise that comes from: `firebase.storage().ref('/image/1').putFile(image.uri).then`. Notice the final **`.then`**. That one returns the promise we need with the upload info. Then as soon as this is fixed, we´ll have to remove all `.then`.

bennygenel:
@Salakar Hi! Can you please give an example of how to use callbacks please.
I tried this but it didn't work.

```
firebase.storage()
    .ref('/images/' + id)
    .putFile(image.path, (uploadedFile)=>{
      console.log('uploadedFile',uploadedFile);
    });
```

Salakar:
@bennygenel I believe this was fixed in v1.1.0, try updating the npm module

andfk:
@bennygenel yeah, I can confirm the storage works perfectly on v1.1.0

bennygenel:
@Salakar & @andfk Thanks for the heads up. I updated and it fixed the problem.

PR #115, opened 2017-05-14
fix for #105

alexkuttig:
I have included the fix discussed in #105 in my fork. Tested it on iPhone Simulator and it worked fine.


Salakar:
The problem with this is that the resolve part of the promise still needs to go through `_interceptSnapshotEvent` - it attaches on the additional props to match the Web SDK. 

I ideally the promise over the RN bridge needs changing into a callback instead and the .then and .catch are faked, as opposed to being a real promise - at the moment if you don't use a promise you get unhandled promise warnings.

alexkuttig:
Okay, I see...

Salakar:
It's a weird one ;p I do have it on my list of things to do though,

I guess I could merge it though considering the current one is broken as is - you'll just miss out on .task and .ref props for now.

PR #122, opened 2017-05-23
Bump the Firebase version

felixclack:
There are more recent versions of the Firebase SDKs available, so this provides
a simple version bump.

To make this easier in future, I extracted the number to a variable.


Salakar:
Thanks for the PR 👍

PR #176, opened 2017-06-16
Fixed auth function call in example.

BuffMcBigHuge:
Without calling auth as a function, the call fails on iOS.

### DIFF
--- a/docs/modules/authentication.md
+++ b/docs/modules/authentication.md
@@ -305,7 +305,7 @@ LoginManager
   })
   .then(data => {
     // create a new firebase credential with the token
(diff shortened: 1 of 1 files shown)


chrisbianca:
Which version of RNFirebase are you using? We've checked and this seems to be working fine without the function call and follows the Web SDK API.

BuffMcBigHuge:
`"react": "16.0.0-alpha.12"`
`"react-native": "0.45.1"`

You are correct Chris. However, I was unable to retrieve credentials any other way, specifically for iOS. I will review my code and see if it was something else.

https://firebase.google.com/docs/auth/web/facebook-login#top_of_page

PR #207, opened 2017-06-26
Support javascript plain error object in crash.report

ghsdh3409:
* I found that `crash().report(Error)` does not work, since it returns empty when error object does not contain `code`.
* `code` is only included in `FirebaseError`. Plain `Error` object does not contain `code`.
* Thus, I removed logic to check `code` in error object in `report()` function.


Ehesp:
Good catch! Thanks

PR #212, opened 2017-06-27
fixed documentation of putFile with progress error handling

nico1510:
Currently the docs are not in sync with the code here. When using the docs code, errors are swallowed (unhandled promise exception). I noticed that there is a [TODO](https://github.com/invertase/react-native-firebase/blob/master/android/src/main/java/io/invertase/firebase/storage/RNFirebaseStorage.java#L351) which would make the current docs correct but at the moment neither android nor iOS support the event error handler.


Ehesp:
Hey thanks for this. Storage docs do need some TLC! @Salakar was there any changes planned to storage around this?

Issue #215, opened 2017-06-28
c++ exception in nativemodules java.lang.noclassdeffounderror


mkayswork:
If I keep the transitive flag = false I get the following error message:

c++ exception in nativemodules java.lang.noclassdeffounderror: failed resolution of: Lcom/google/android/gms/common/GoogleApiAvailability

If I set the transitive flag to true no errors are shown and the application starts well. Why is this happening? Can you explain this to me please???

Ehesp:
Can we see your `build.gradle` file please.

ugiacoman:
I am getting this error as well. 

```
apply plugin: "com.android.application"

import com.android.build.OutputFile


project.ext.react = [
    entryFile: "index.js",
    bundleInDebug: false,
    bundleInRelease: true,
]

apply from: "../../node_modules/react-native/react.gradle"

/**
 * Set this to true to create two separate APKs instead of one:
 *   - An APK that only works on ARM devices
 *   - An APK that only works on x86 devices
 * The advantage is the size of the APK is reduced by about 4MB.
 * Upload all the APKs to the Play Store and people will download
 * the correct one based on the CPU architecture of their device.
 */
def enableSeparateBuildPerCPUArchitecture = false

/**
 * Run Proguard to shrink the Java bytecode in release builds.
 */
def enableProguardInReleaseBuilds = false

android {
    signingConfigs {
        config {
            keyAlias 'key0'
            keyPassword 'test123'
            storeFile file('/Users/ulises.giacoman/androidKeyStore.jks')
            storePassword 'test123'
        }
        config1 {
        }
    }
    compileSdkVersion 25
    buildToolsVersion '23.0.1'
    defaultConfig {
        applicationId "com.forrent.community"
        minSdkVersion 16
        targetSdkVersion 22
        versionCode 1
        versionName "1.0"
        ndk {
            abiFilters "armeabi-v7a", "x86", "armeabi", "mips"
        }
    }
    splits {
        abi {
            reset()
            enable enableSeparateBuildPerCPUArchitecture
            universalApk false  // If true, also generate a universal APK
            include "armeabi-v7a", "x86"
        }
    }
    buildTypes {
        release {
            minifyEnabled enableProguardInReleaseBuilds
            proguardFiles getDefaultProguardFile("proguard-android.txt"), "proguard-rules.pro"
            signingConfig signingConfigs.config
        }
    }
    // applicationVariants are e.g. debug, release
    applicationVariants.all { variant ->
[...]

PR #231, opened 2017-07-05
Add example project + Update android installation docs

ugiacoman:
1. Adds an example project that has a barebones implementation of `react-native-firebase` on both platforms. 
2. Update to the Android installation docs where `google-play-services` were not getting compiled. Associated issue #215.


Ehesp:
Could you add the `example` directory to the root `.npmignore` too please!

ugiacoman:
@Ehesp Good call, pushed!

Salakar:
@ugiacoman @Ehesp looks good now from my point of view, happy to merge if you are El

PR #264, opened 2017-07-20
Updated admob.md with sidenote about zIndex crash

bitfabrikken:
Just added a quick note that tells you not to use zIndex in style tag on Banner tag.

### DIFF
--- a/docs/modules/admob.md
+++ b/docs/modules/admob.md
@@ -63,6 +63,9 @@ render() {
 
 ```
 
+Side note: If you use a style on the Banner, do not use zIndex. It will crash on release versions of your app (App Store review, TestFlight, adhoc distribution).
+
+
 ### Interstitial
 
 An interstitial is a full screen advert which creates a new activity on top of React. As they need to be controlled,


bitfabrikken:
@Ehesp 

First thank for a great project :)

I just wanted my custom-sized banner (screen width x 50) to be at the bottom at all times. 

So I styled it absolute with bottom: 0. It's probably not the way to do it correctly, but I'm still very new to RN. 

But having zIndex in there gave the crashes. 

I've posted some crash logs in the issue below, but never really saw anything in those logs directly related to this. Had to narrow it down the hard way. It's odd.

https://github.com/invertase/react-native-firebase/issues/253

chrisbianca:
@bitfabrikken Absolutely styling the banner certainly does make sense, but you don't need to use zIndex to make it appear on top of everything else.  If you take a look at https://facebook.github.io/react-native/docs/layout-props.html#zindex you'll see that you can essentially ensure that something is on top by rendering it last in a tree.  e.g.

```
<View>
  <View /*inner view with stuff in*/>

  </View>
  <Banner /* absolute styling to show at bottom */ />
</View>
```

The banner will show over the inner view, though note that potentially, you'll cut off the bottom of the inner view in this case.

bitfabrikken:
@chrisbianca thanks, looks like I've done it the correct way then :)

PR #292, opened 2017-07-31
Added info about security rules

Leeds-eBooks:


chrisbianca:
Thanks for this @Leeds-eBooks, very useful!

PR #297, opened 2017-08-02
Fix android compilation (react-native v0.47)

brapifra:
According to the [changelog](https://github.com/facebook/react-native/releases/tag/v0.47.0) of react-native v0.47, `createJSModules` calls should be removed.


chrisbianca:
@brapifra Thanks for this.  We went with a slightly different approach to maintain compatibility with earlier versions of React Native.  v2.1.0 of `react-native-firebase` now supports RN 0.47 and earlier versions.

PR #299, opened 2017-08-02
Remove createJSModules @overide marker - RN 0.47 compatibility

radko93:
[`createJSModules` is now not required on Android](https://github.com/facebook/react-native/commit/ce6fb337a146e6f261f2afb564aa19363774a7a8) from RN 0.47. I actually can't build an app on RN 0.47 without removing this method/@ovveride marker. #297 is similar but seems that this one is better for backwards compatibility.

### DIFF
--- a/android/src/main/java/io/invertase/firebase/RNFirebasePackage.java
+++ b/android/src/main/java/io/invertase/firebase/RNFirebasePackage.java
@@ -38,7 +38,7 @@ public class RNFirebasePackage implements ReactPackage {
    * listed here. Also listing a native module here doesn't imply that the JS implementation of it
    * will be automatically included in the JS bundle.
    */
-  @Override
+  // Deprecated RN 0.47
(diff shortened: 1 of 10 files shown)


Ehesp:
https://github.com/invertase/react-native-firebase/pull/297

radko93:
@Ehesp this is a bit different approach.

Ehesp:
Yeah sorry was more a reference for others looking and reviewing :)

radko93:
I tested and it works on RN 0.46.4 with https://github.com/tinycreative/react-native-intercom/pull/116.

chrisbianca:
@radko93 Thanks for this! We've used the approach you suggested, but slightly tweaked so we can keep track and remove the unnecessary methods in the future.

v2.1.0 of react-native-firebase has been released which supports v0.47.0 as well as earlier versions.

PR #327, opened 2017-08-10
Fix wrong RNFirebaseDatabase implementation

zoontek:
`RNFirebaseDatabase` => `RNFirebaseDatabaseReference` in `RNFirebaseDatabaseReference.m`

screen shot 2017-08-10 at 22 26 51

### DIFF
--- a/ios/RNFirebase/database/RNFirebaseDatabaseReference.m
+++ b/ios/RNFirebase/database/RNFirebaseDatabaseReference.m
@@ -190,6 +190,6 @@
 @end
 
 #else
-@implementation RNFirebaseDatabase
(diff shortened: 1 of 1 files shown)


egunsoma:
Can this bug cause something like DB events not firing on iOS device?

zoontek:
No, this bug only appear when you don't use the Firebase Database SDK

phips28:
please merge!

chrisbianca:
Just released v2.1.2 which addresses this issue more thoroughly
@zoontek thanks for raising - most people have the database enabled which is why it's not come up previously.

PR #346, opened 2017-08-19
Make to work on RN 0.48

jozan:


Ehesp:
@Salakar @chrisbianca Well, carrying on the EventEmitter fun....

jozan:
Just noticed that in 3.0 replacing EventEmitter entirely with your own solution. Have you been experiencing issues with it other than performance?

Salakar:
v3 uses the RN EE even more so, it's not been replaced @jozan 

Problem with this change is it's not backwards compatible and there's no way to make it either as dynamic module requires aren't possible in RN, I'm working on something though as we're already aware of this issue.

Will close this PR as we're already working on something to resolve the issue :) appreciate you taking the time to PR though, we don't seem to get many!

PR #348, opened 2017-08-21
moved EventTypes to new file to avoid enumerable key in AdMobComponen…

blrichards:


Ehesp:
I still don't get how this fixes the issue haha. I'll just double check everything still works with the tests tomorrow.

blrichards:
The issue is that react-native is throwing an error/performance complaint because an enumerable key was used for the initial implementation of propTypes in AdMobComponent. I just reimplemented it without the enumerable key. Everything else should be the same

PR #367, opened 2017-08-28
react-redux-firebase example

prescottprue:
Initial pass at adding a redux example with `react-redux-firebase`. Open to input about improvements.


Salakar:
Thanks for doing this, 

They're live at https://invertase.io/react-native-firebase/#/redux

prescottprue:
Looks like I forgot to include the creation of the reducer. Just a one line fix, but I will open another PR.

PR #374, opened 2017-08-30
Update installation-android.md

ckalpit:
updated version of com.google.firebase:firebase-perf from 11.0.0 to 11.0.4

### DIFF
--- a/docs/v2/installation-android.md
+++ b/docs/v2/installation-android.md
@@ -179,6 +179,6 @@ In the same file, add the `firebase-perf` module to your dependencies:
 ```
 dependencies {
   ...
-  compile "com.google.firebase:firebase-perf:11.0.0"
+  compile "com.google.firebase:firebase-perf:11.0.4"
 }
 ```


Ehesp:
@ckalpit This should be 11.2.0 actually, mind updating?

ckalpit:
I am not sure if it will be compatible, my simple logic to set 11.0.4 is because other firebase libraries are also using 11.0.4 hence to avoid build time issues made it 11.0.4, also its working on my project.

we need to change all firebase libraries to 11.2.0 and run the project then we can do that too.

Ehesp:
Yeah for new users though, 11.2.0 is what we want to target. Hopefully existing users like yourself will know!

ckalpit:
So, I hope it will be accepted and merged soon because if someone is trying to use v2 they do not have any other build issues regarding this library.

Salakar:
Merging as this is for the v2 docs, v3 only is 11.2.x+ - so this is fine.

Issue #119, opened 2017-05-18
Firebase Phone Number Auth integration


jasan-s:
Is the new Firebase Phone number Auth on the roadmap?

Ehesp:
Yeah it will be. Will look at seeing what's possible to do

Salakar:
@jasan-s @Ehesp - might make sense to wait until android support is out for this at the end of may: https://firebase.google.com/docs/auth/android/phone-auth

DavidKuennen:
They will support this? Awesome 😄

jeveloper:
I believe its already available as they have open sourced their SDKs

Salakar:
@jeveloper it's only available in ios at the moment

jeveloper:
ohh thats odd, i thought they were just reusing fabrics digits which works on anything

thanks man!

Froelund:
Maybe _now_ is too early June :)
Phone number sign-in is coming soon to Firebase on Android. Check this page in early June 2017 for details.

chrisbianca:
Yesterday it said late May! :)

perrosnk:
+1 I am also really interested in this feature!

wbattel4607:
+1 We're anxiously awaiting this feature after Google finishes implementation for Android

Battelman2:
https://firebase.google.com/support/release-notes/android#20170607

Now available on Android SDK!

wbattel4607:
https://firebase.google.com/support/release-notes/android#20170607

Now available on Android SDK!

Sparragus:
Is anyone assigned or already working a PR to this issue? If not, I could try to do it myself.

Ehesp:
@Sparragus no one specifically. I'll take a look at Android on Monday if theres no PR. IOS looks a bit too complicated for my object c skills

Ehesp:
I've been looking into this today. I'll have to speak internally about how to actually implement this (more best practice). The aim of this lib is to try and replicate the web SDK, however I don't think it'll be possible here.

1. On the Web [docs](https://firebase.google.com/docs/auth/web/phone-auth) there's no need for steps 1-3. 
2. `signInWithPhoneNumber` takes a phoneNumber & appVerifier (where appVerifier is redundant on the device).
3. The native SDKs (at least on Android) takes a `verifyPhoneNumber`. It can also do things like auto detection. The response of which, can be directly passed into `signInWithCredential`.

[...]

PR #382, opened 2017-09-01
[auth][phone] Allow for linkWithCredential() for PhoneAuthProvider.

brianhiss:
@chrisbianca Pull request for `linkWithCredential()` capability with `PhoneAuthProvider`. (re: https://github.com/invertase/react-native-firebase/issues/119#issuecomment-326510174)

This allows someone to link a phone to a currently logged in user, or sign the user in.

To test:

```
firebase.auth().signInWithPhoneNumber(phoneNumber)
      .then(confirmResult => this.setState({ confirmResult }))
```

Set the credential object with the six-digit SMS code (`codeInput`):
```
let credential = {
   provider: 'phone',
   token: this.state.confirmResult._verificationId,
   secret: codeInput
};
```

- If there is a logged-in/anonymous user, call: `firebase.auth().currentUser.linkWithCredential(credential);`

- If no one is logged in, call: `firebase.auth().signInWithCredential(credential);`

### DIFF
--- a/android/src/main/java/io/invertase/firebase/auth/RNFirebaseAuth.java
+++ b/android/src/main/java/io/invertase/firebase/auth/RNFirebaseAuth.java
@@ -917,6 +917,8 @@ class RNFirebaseAuth extends ReactContextBaseJavaModule {
         return TwitterAuthProvider.getCredential(authToken, authSecret);
       case "github":
         return GithubAuthProvider.getCredential(authToken);
+      case "phone":
+        return PhoneAuthProvider.getCredential(authToken, authSecret);
       case "password":
         return EmailAuthProvider.getCredential(authToken, authSecret);
       default:
--- a/index.d.ts
+++ b/index.d.ts
@@ -455,6 +455,10 @@ declare module "react-native-firebase" {
        */
       reauthenticate(credential: Credential): Promise<void>
       /**
+       * Link the user with a 3rd party credential provider.
+       */
+      linkWithCredential(credential: Credential): Promise<User>
+      /**
        * Refreshes the current user.
        */
       reload(): Promise<void>
--- a/ios/RNFirebase/auth/RNFirebaseAuth.m
+++ b/ios/RNFirebase/auth/RNFirebaseAuth.m
@@ -825,6 +825,8 @@ RCT_EXPORT_METHOD(fetchProvidersForEmail:
         credential = [FIREmailAuthProvider credentialWithEmail:authToken password:authTokenSecret];
     } else if ([provider compare:@"github" options:NSCaseInsensitiveSearch] == NSOrderedSame) {
         credential = [FIRGitHubAuthProvider credentialWithToken:authToken];
+    } else if ([provider compare:@"phone" options:NSCaseInsensitiveSearch] == NSOrderedSame) {
+        credential = [[FIRPhoneAuthProvider provider] credentialWithVerificationID:authToken verificationCode:authTokenSecret];
     } else {
         NSLog(@"Provider not yet handled: %@", provider);
     }


Salakar:
@brianhiss nice, thanks for this, its a good start so will merge it, will break though on android if auto verify - as it signs in straight after. Will update it a bit today to expand on your PR.

Also there is firebase.auth.PhoneAuthCredential.getCredential(verifyToken, smsCode); to get the credential

brianhiss:
@Salakar Thanks for merging. I somehow missed `firebase.auth.PhoneAuthCredential.getCredential(verifyToken, smsCode);`.

As you said, it will break on Android with auto-verify. I think a few of us are stumped on the best practice for auto-verify, sign-in and linkWithCredential. The workflow seems to contradict itself.

nicholasstephan:
@brianhiss. Thanks everyone for putting in the work on this. 

I'm trying to implement, but just get the error:

```
{ 
    error_name: 'ERROR_MISSING_APP_TOKEN',
    NSLocalizedDescription: 'There seems to be a problem with your project\'s Firebase phone number authentication set-up, please make sure to follow the instructions found at https://firebase.google.com/docs/auth/ios/phone-auth' },
    domain: 'FIRAuthErrorDomain' 
}
```

when I try `signInWithPhoneNumber`. Can anyone point me in the right direction to get this working? 

Any idea when this might make it into production?

kevando:
@brianhiss thanks this PR works great. However, I'm experiencing a small issue..

When the current user is anon, and I link it to the phone user, the resulting user is all good, except its still `user.isAnonymous == true`

But if I sign the user in, it disregards the original user, but the resulting phone `user.isAnonymous == false`


``` javascript
var credential = firebase.auth.PhoneAuthProvider.credential(verificationId, codeInput)

// Links correctly but resulting user is still anonymous
firebase.auth().currentUser.linkWithCredential(credential)

// Signs in correctly not anonymous, but disregards data from anon user
firebase.auth().signInWithCredential(credential)
```

Am I missing something?

brianhiss:
@kevando I don't have time to dig into it today, but off the top of my head, I'm just listening to `firebase.auth().onAuthStateChanged()` and don't believe tracking `user.isAnonymous`, so you could be right. That could be a firebase bug, or intended behavior, not too sure right now.

I'm tracking `providerData` and `providerId === "phone"` when they have a valid phone number attached, or something of that sort.

I'll try to look deeper in the next couple of days when I get a chance, report back.

chrisbianca:
@kevando Is this on iOS or Android? I was doing some work last week and noticed that the user doesn't reload correctly on iOS when using `linkWithCredential` - this is resolved on the latest version of the 3.0.x branch which we'll hopefully get a release of soon.

If this is an iOS issue and you're in a position to test whether this is now fixed, you can install the latest by doing the following:

`npm install --save https://github.com/invertase/react-native-firebase#v3.0.x`

kevando:
Thanks @chrisbianca I am on iOS and installing that branch did not trigger onAuthStateChanged after calling linkWithCredential. I did not update the pod or anything so maybe I have to do that?

@brianhiss Thanks, I will use the providerData for determining whether or not a user is anon or not. thanks

chrisbianca:
@kevando Ah, so `onAuthStateChanged` will only ever fire when the user is either logged in or logged out.  This is by design in the Firebase SDK.  However, the firebase.auth().currentUser will have been updated at the end of the `linkWithCredential` call.

I've found this to be confusing too and not very helpful when it comes to how you integrate easily into an app...

So, as of v3.0.6 there's a very experimental `firebase.auth().onUserChanged()` method which gets called every time the current user is changed with the `User` object.  By experimental, I mean that it fundamentally works, but there may be some edge cases which we'd like to flush out.  It's also not something that's in the web or native SDKs so technically violates are policy of sticking to these where possible.  However, we will discuss this with Firebase directly as I feel it's something that would be of benefit for the core SDKs as well...

PR #384, opened 2017-09-02
Change import path to support RN 0.48.0

chaitanya0bhagvan:


Salakar:
This will break support for all versions prior to RN 0.48 - there's currently no way to support all versions without copying over the EE library itself into our project - which we've done in v3, so consider this fixed. 😛

PR #403, opened 2017-09-12
Set opened_from_tray for local_notification

phips28:
Set opened_from_tray true if a local notification is clicked from notification center/tray.

Currently its only set for a remove_notification line 131.


avinashwigroup:
Is this merged ?

PR #406, opened 2017-09-13
Update with Google Maven reference

dannyshisler:
I ran into this problem trying to follow your installation instructions.  The build kept failing with errors stating that it was unable to resolve the firebase packages. 

I could see that the 11.2.x packages weren't included in the latest Google Repository (58) and the latest release notes (https://developers.google.com/android/guides/releases) state that "Google Play services dependencies are now available via maven.google.com", so I added the Google Maven reference to the root build.gradle and it resolved the problem.

### DIFF
--- a/docs/installation-android.md
+++ b/docs/installation-android.md
@@ -60,6 +60,19 @@ dependencies {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


Salakar:
Thanks for this, this was mentioned in the http://invertase.link/v3 change log, just not been mentioned in the docs... until now :)

PR #416, opened 2017-09-16
link fix

mertnesvat:
ios and android was broken, fixed!


Salakar:
Thabk for this, however the docs are fine, they're not made to be used from the repo, we have a docs site generated off of them: https://invertase.io/react-native-firebase/#/initial-setup - both links work fine there.

PR #429, opened 2017-09-20
Android fix: Duplicate dependency error.

IvRRimum:
Had a rather large app with a lot of dependencies and this duplicated caused it to fail at compile time.

Thanks!


Salakar:
Not sure where the duplicate is here? There's only one instance of config in the file.

IvRRimum:
https://github.com/invertase/react-native-firebase/blob/master/android/build.gradle

Scroll to the bottom and click Ctrl+f and type config. It will light up

Salakar:
Wow i am blind ha, it was late, thays my excuse, merging :)

PR #432, opened 2017-09-21
Fix syntax error in messaging

JackThomson2:
App wouldn't compile before.


chrisbianca:
Thanks for this - looks like we didn't properly pick up on an invalid PR.  I've just released 3.0.0-alpha.5

PR #439, opened 2017-09-23
[auth][unlink] Call native function instead of 'FirebaseAuth'

hjumeau:
screenshot_20170923-233751


Salakar:
Woops! Thanks for this :)

PR #443, opened 2017-09-26
deleteInstanceId feature to unregister from push notifications until next getToken call.

bm-software:


Salakar:
Looking good, thanks for this - can you also add it to the messaging docs also please, https://invertase.io/react-native-firebase/#/v2/modules/cloud-messaging

android/src/main/java/io/invertase/firebase/messaging/RNFirebaseMessaging.java
```diff
@@ -73,6 +74,18 @@ public void getToken(Promise promise) {
   }
 
   @ReactMethod
+  public void deleteInstanceId(Promise promise){
+    try {
+      Log.d(TAG, "Deleting instance id");
+      FirebaseInstanceId.getInstance().deleteInstanceId();
+      promise.resolve(null);
+    } catch (IOException e) {
+      e.printStackTrace();
```
chrisbianca: Can you change this to log out the exception using `Log` rather than printStackTrace?

bm-software: Done

PR #480, opened 2017-10-08
add note for installing GoogleService-Info.plist on iOS

neilmock:
referencing https://github.com/firebase/firebase-ios-sdk/issues/340#issuecomment-334825911

I had similar issues getting started with react-native-firebase on iOS, so I thought some kind of documentation on this point was necessary.  Apologies if this is already addressed somewhere, but I looked through the issues and didn't notice anything.


neilmock:
oops, just saw another PR doing the exact same thing.  closing.

Salakar:
Added this earlier, thanks

Issue #304, opened 2017-08-03
Dynamic Links


g4laura:
I saw it mentioned in another issue that dynamic links were a possible long term goal but not a priority right now, but I was wondering if anyone could give me any tips about how to set up dynamic links with react native, especially for iOS. 

I have receiving dynamic links working well on android but only sometimes on iOS. It works if the app is installed, on some iphones, it doesn't work on others. Also, when the app isn't installed, the dynamic link takes the user to the apple app store, but after the app is open, it doesn't take the user to the screen corresponding to the link.
Also, right now I'm creating the dynamic links with javascript so, they are very long.

If anyone could guide me on how to get dynamic links working, I'd appreciate it. 

I don't know Objective C, but I have experience with java.

chrisbianca:
@g4laura This is still very much on the roadmap and I did start taking a look, but unfortunately got pulled away onto other priorities - multi app support being the main thing we're focusing on at the moment.

The best way to currently get Dynamic Links working on iOS is to make use of a combination of Firebase and the built in React Native linking library (https://facebook.github.io/react-native/docs/linking.html).  Follow the setup guide for the Linking library, then:

In your AppDelegate.m, you need to add the following methods:

```
- (BOOL)application:(UIApplication *)app
            openURL:(NSURL *)url
            options:(NSDictionary<NSString *, id> *)options {
  return [self application:app openURL:url sourceApplication:nil annotation:@{}];
}

- (BOOL)application:(UIApplication *)application
            openURL:(NSURL *)url
  sourceApplication:(NSString *)sourceApplication
         annotation:(id)annotation {
  FIRDynamicLink *dynamicLink =
  [[FIRDynamicLinks dynamicLinks] dynamicLinkFromCustomSchemeURL:url];
  
  if (dynamicLink) {
    return [RCTLinkingManager application:application
                                  openURL:dynamicLink.url
                        sourceApplication:sourceApplication
                               annotation:annotation];
  }
  
  return [RCTLinkingManager application:application
                                openURL:url
                      sourceApplication:sourceApplication
                             annotation:annotation];
}
```
Then in your JS code you can add the following when the app mounts:
```
componentDidMount() {
  // Check if the app was opened from a link
  Linking.getInitialURL()
    .then(url => this.onLink({ url }))
    .catch(err => console.error('An error occurred getting the initial page', err));
  // Register Linking listener
  Linking.addEventListener('url', this.onLink);
}

onLink = ({ url }) => {
[...]

PR #481, opened 2017-10-09
Dynamic Links Support

omer88:
This PR includes Dynamic Links support.

Creation API  mimics Firebase's REST API for Dynamic Links creation.
There are options to create both long and short links.

Receiving is done like in React Native Linking Library.

For more information see docs.

relates to #304.


android/src/main/java/io/invertase/firebase/links/RNFirebaseLinks.java
```diff
@@ -0,0 +1,294 @@
...
+  public void getInitialLink(final Promise promise) {
+    if (mInitialLinkInitialized) {
+      promise.resolve(mInitialLink);
+    } else {
+      Activity activity = getCurrentActivity();
+      if (activity != null) {
+        resloveLink(activity.getIntent(), new ResolveHandler() {
+          @Override
+          public void onResolved(String url) {
+            if (url != null) {
```
omer88: It was done because of this: https://firebase.google.com/docs/reference/android/com/google/firebase/dynamiclinks/FirebaseDynamicLinks.html#getDynamicLink(android.content.Intent)

The getDynamicLink method usually takes the url from the server. Therefore in the second time getInitialLink invoked, getDynamicLink will return null. In order to be consistent to the iOS implementation where this value is kept in userActivityDicionary I store the initial link in a member, and cleans it on host destroy lifecycle method.

Salakar:
Woop, thanks guys, we'll get this properly reviewed and get back to you next day or two. 

In the mean time what's your discord usernames on our server - we'll add you to the private contrib channel so we can chat there if needed.

lib/modules/links/index.js
```diff
@@ -0,0 +1,100 @@
...
+   */
+  createShortDynamicLink(parameters: Object = {}): Promise<String> {
+    if (!this._validateParameters(parameters)) {
+      return Promise.reject(new Error('Invalid Parameters.'));
+    }
+    return this._native.createShortDynamicLink(parameters);
+  }
+
+  _validateParameters(parameters: Object): boolean {
+    const suportedParametersObject = {
```
Salakar: Could this not be outside of the class?

Salakar: Also are all these parameters required?

omer88:
Thanks! my discord username is olevy.

lib/modules/links/index.js
omer88: Yes, it could be outside of the class.
Just put it outside of the class on the same file?

In addition, these parameters are all supported. Not all of them required. I just validate that the input parameters object is contained in this object, to make sure there is no invalid parameters or parameters with typo.
The required parameters are checked in the native code.

Salakar: Aye just at the top outside class I guess, saves making a new instance of it every time.

And ok, makes sense :)

Ehesp:
Looks good!

android/src/main/java/io/invertase/firebase/links/RNFirebaseLinks.java
```diff
@@ -0,0 +1,294 @@
...
+    getReactApplicationContext().addLifecycleEventListener(this);
+  }
+
+  @Override
+  public String getName() {
+    return "RNFirebaseLinks";
+  }
+
+
+  private void resloveLink(Intent intent, final ResolveHandler resolveHandler, final ErrorHandler errorHandler) {
```
chrisbianca: Typo: `resolveLink` ;)

android/src/main/java/io/invertase/firebase/links/RNFirebaseLinksPackage.java
```diff
@@ -0,0 +1,51 @@
...
+
+  /**
+   * @return list of JS modules to register with the newly created catalyst instance.
+   * <p/>
+   * IMPORTANT: Note that only modules that needs to be accessible from the native code should be
+   * listed here. Also listing a native module here doesn't imply that the JS implementation of it
+   * will be automatically included in the JS bundle.
+   */
+  // TODO: Removed in 0.47.0. Here for backwards compatability
+  public List<Class<? extends JavaScriptModule>> createJSModules() {
```
chrisbianca: This method can be removed as in 3.0.X we removed backwards compatibility for anything prior to 0.48.0

docs/_sidebar.md
```diff
@@ -26,6 +26,7 @@
  - [Crash Reporting](/modules/crash)
  - [Database](/modules/database)
  - [Firestore (Beta)](/modules/firestore)
+ - [Dynamic Links](/modules/links)
```
chrisbianca: Can this be moved above Firestore?  The rest are all in alphabetical order - I'm assuming this was a merge thing because we added it at the same time.

docs/modules/links.md
```diff
@@ -0,0 +1,125 @@
...
+  // ...
+});
+```
+### Parameters
+
+Only the following parameters are currently supported:
+
+```javascript
+{
+  dynamicLinkInfo: {
```
chrisbianca: Do you think that it's necessary to break suffix out of the main info object?  I'm just thinking that it might be simpler from a user implementation point of view to not need the `dynamicLinkInfo` key at all, and just have the object as:

```
{
  dynamicLinkDomain,
  link,
  androidInfo,
  iosInfo,
  socialMetaTagInfo,
  suffix,
}
```

chrisbianca: I realise that this goes against the REST API, but in the interest of keeping it simpler for the user to implement it seems like an unnecessary separation.

The only point where it would make sense to keep the separation is if we were to support the long to short link conversion, which I can't see being necessary on a mobile app.

docs/_sidebar.md
omer88: Actually, it because I called it Links in the beginning (the name of the library - which in every other place it's after) - but sure, here it has to be before.

docs/modules/links.md
omer88: I have change the API to be that way.

royisch:
when can we expect to see it merge?

Ehesp:
@chrisbianca @Salakar All good?

Salakar:
Good to me @Ehesp

chrisbianca:
@omer88 Thanks again for this - I've just merged it in for our upcoming 3.1.x branch of releases.  No firm date on when that will be yet, but it's now there so that people can start testing with it

JackThomson2:
@omer88  thanks for this great PR! I am not getting the deep link parameters on android. React Native's Linking library is finding it, however.

omer88:
@JackThomson2, what exactly do you get? Does it get any url?
In addition, where does it happen? in the initial url or the subscription?

omer88:
I just found an issue with iOS. Please don't mess with it, I'll create a PR soon

adamgins:
Thanks for this capability - Do we assume that using https://rnfirebase.io/docs/v4.3.x/links/reference/links and the `onLink` callbacks/events are preferred vs using https://facebook.github.io/react-native/docs/linking

PR #487, opened 2017-10-10
ViewPropTypes is not undefined in React-native 0.42 and View.propType…

rahal:
ViewPropTypes is not undefined in React-native 0.42 and View.propType is still accessible in React-Native 0.47

This caused : `AdMobComponent` has no propType for native prop `RNFirebaseAdMobBanner.accessibilityLabel` of native type `String`


rahal:
Well, I checked again in react-native repo, and they don't recommend accessing View.propType cause it's available in DEV mode only.. So, I'm not sure if it's the best solution

Ehesp:
`View.propType` is deprecated in the latest I believe. Weird issue this one... Documentation is really lacking.

Issue #505, opened 2017-10-13
Timezone ignored in server timestamp value


urajat:
### Issue

In Firestore, when I use `FieldValue.serverTimestamp`, the timestamp value is set correctly on Firestore including the timezone. But when I read it back, the timezone information sent by Firestore is discarded and interpreted as UTC without the value itself being converted. This results in the timestamp value being offset by the local timezone again.

**Code to reproduce the issue**

```
import firebase from 'react-native-firebase';

firebase.firestore().collection('users')
  .doc('user1')
  .set({
    createdAt: firebase.firestore.FieldValue.serverTimestamp()
  })
  .then(() => {
    firebase.firestore().collection('users')
      .doc('user1')
      .get()
      .then((documentSnapshot) => {
         const value = documentSnapshot.data();
         console.log(value);
      });
```

Saved value on Firestore (checked from console): `October 13, 2017 at 12:36:56 AM UTC+5:30`

console.log output => `createdAt: Fri Oct 13 2017 06:06:56 GMT+0530 (IST)` which is wrong - an extra offset of local timezone (+5:30) has been added.

Added Log.d statements in the`buildTypeMap` method in `FirestoreSerialize.java`:
```
} else if (value instanceof Date) {
    Log.d(TAG, "raw timestamp value:" + value.toString());
    Log.d(TAG, "formatted timestamp value:" + DATE_FORMAT.format((Date) value));
    typeMap.putString("type", "date");
    typeMap.putString("value", DATE_FORMAT.format((Date) value));
} else {
```
With this logging, I got the following results:

Raw timestamp value: `Fri Oct 13 00:36:56 GMT+05:30 2017` (which is correct)
Actual formatted value printed currently: `2017-10-13T00:36:56.158Z` (which is wrong)
Expected formatted value: Either `2017-10-13T00:36:56.158+0530` or `2017-10-12T19:06:56.158Z`

Basically, either retain the timezone info or convert the value returned by Firestore into UTC and then format it. Retaining is preferable I think.

### Environment

1. Application Target Platform: **Android**

2. Development Operating System: **macOS Sierra**

3. Build Tools: **Android 7.1.1, API 25, Nexus 5X**

4. React Native version: **0.48.3**

5. RNFirebase Version: **3.0.3**

6. Firebase Module: **Firestore**

chrisbianca:
PR has been merged pending release

PR #506, opened 2017-10-13
#505 Retain timezone information while formatting serverTimestamp value

urajat:
Resolves issue #505


chrisbianca:
@urajat Thanks for this.  Are you also setting up the equivalent iOS app? Any chance you'd be able to check this is working there too?

urajat:
@chrisbianca Will likely get to the iOS app in a couple of weeks. But I don't have any experience working with Objective-C/Swift - so unlikely to be very helpful on that front :-(

chrisbianca:
I've merged and will update the iOS date format to match

PR #535, opened 2017-10-20
Update authentication.md

ziadalame:
```Promise.resolve``` allows the promise chain to continue unlike the needed flow which is to break the chain and jump to the catch statement. Currently this code is problematic and crashes the app.


Ehesp:
I'm not even sure the example worked before this - the promise resolve would pass cancelled onto the next promise chain which would crash. We've got new docs, so need to remove these actually.

Salakar:
Hi, this is correct on the new docs over here: https://rnfirebase.io/docs/v3.0.*/auth/social-auth#Facebook

As the docs no longer exist in this repo I'll close this. Thanks though :)

PR #554, opened 2017-10-27
[firestore][ios] fix date formatter

mikesurowiec:
The previous format seems to only have worked when the chrome debugger was open due to different JS runtime environments (see: https://stackoverflow.com/questions/41874676/react-native-code-doesnt-work-without-remote-debugger-enabled).

By formatting it this way and setting the timezone to UTC, it appears to work properly with the firestore server date in both debug and non-debug modes.


chrisbianca:
Thanks for this @mikesurowiec, Dates have certainly proved to be troublesome as you've noted due to the difference in behaviour between JS runtimes when debugging. We've actually switched to using UTC milliseconds as this was most consistent across iOS and Android as it doesn't rely on `new Date()` on the JS side correctly interpreting a string.

See the commit here: https://github.com/invertase/react-native-firebase/commit/403a7eefa0108113b74e59d9e3b5fbf404cde669

We're just finishing up a few other things before releasing a new version with this fix in.

mikesurowiec:
👍 awesome, that seems like a more sustainable fix!

PR #570, opened 2017-11-01
Fix compilation issue on iOS on `master`

erezrokah:
If you install the library from `master` it causes compilation issues (missing `RNFirebaseStorage.m`).

`storage` and `Products` had the same id of `134814211AA4EA7D00B7C361`.
I re-added the `storage` directory to the project in order to fix this issue.


chrisbianca:
@erezrokah Thanks for this - it was probably caused by the merge when we brought back the changes from the 3.0.x branch

erezrokah:
@chrisbianca I figured this was a merge issue. Thanks for the quick reply

chrisbianca:
Yeah... Xcode projects don't make merges easy for you with how they structure their file and folder mappings! No idea why they went down the approach they went for...

Issue #572, opened 2017-11-02
Sending file straight from camera failing


peterdivvito:
### Issue

I integrated with https://github.com/react-community/react-native-image-picker to add image upload to my app. When trying to upload a file to from a newly taken photo it failed with:

`ReactNativeJS: 'error uploading image', { [Error: An unknown error has occurred.] framesToPop: 1, code: 'storage/unknown' }`

When I changed the image picker options to store the image (added 'storageOptions' option) the upload worked.

I noted that he differences from failing to working was that the URI passed to RNFireBase was:

FAILS:
content://XXX/app_images/Android/data/XXX/files/Pictures/image-a8721787-9b0b-44d5-a11c-9850c6db7ca8.jpg

WORKS:
file:///storage/emulated/0/DCIM/image-fe3b561f-db6d-43df-b847-458414a40f29.jpg

I debugged the code and found it failing in RNFirebaseStorage.getRealPathFromURI() with an error along the lines of 'column _data not found'.

I made some changes to use putStream instead of putFile and this fixed it. Pull request to follow.

### Environment

1. Application Target Platform: Android 7.0 API 24

2. Development Operating System: macOS Sierra

3. Build Tools: Android Studio 3

4. React Native version: 0.48.4

5. RNFirebase Version: Git v3.0.x 270279551aaf8061f64a3d7c47cbc4c10de03458

6. Firebase Module: Storage

chrisbianca:
This is now resolved on the 3.0.x branch and ready for our 3.1.0 release

PR #573, opened 2017-11-02
Issue #572 - Fix for Storage.putFile() with content:// paths

peterdivvito:


chrisbianca:
@peterdivvito Thanks for flagging this and the PR.  All looks good and makes sense, but just want to check that you confirmed it still works for file:// paths as well?  

Out of curiosity, have you had a chance to try the same on iOS yet?

chrisbianca:
I've just given this a try with a file:// path using my own app, and it looks like by using `putStream` you lose access to the size of the file being uploaded and as such can no longer present a percentage uploaded indicator which isn't ideal.

chrisbianca:
With putStream, it will never be possible to know the size of the stream, so this makes sense.

I think switching this so that it goes back to using `putFile` for files, and `putStream` for `content://` URLs is the way to go.  I'm just making and testing the changes and will push up a commit shortly.

chrisbianca:
@peterdivvito See: https://github.com/invertase/react-native-firebase/commit/0391b2834d85c9681dad1171a0ff05c2e0a42fb4

If you're able to confirm that this works for your streams (it should do as I'm using your code) then I'll close this and the issue off.

Note, I've made the change on master rather than the 3.0.x branch.  We're currently prepping the 3.1.0 release which will go out in the next couple of days.

peterdivvito:
Hi @chrisbianca ,

I was just thinking about what you said concerning the upload progress and thought I'd check out the api docs and saw this under putFile().

The source of the upload. This can be a file:// scheme or any content URI. A content resolver will be used to load the data. 

Have you tried just passing the URI straight to this?

peterdivvito:
I just tried that out and it worked a treat.

chrisbianca:
Just tested this and all works well, thanks!
I'll get this merged over to master.

PR #603, opened 2017-11-09
Make sure the promise is resolved/rejected for transaction

gyfelton:
Right now the following code will be broken under v3.x.x:
```
async function() {
  await ref.transaction(
              function(foo) {
                 // deal with foo
              },
              function(error, committed, ss) { 
                 // additional work on complete callback
              },
              true     
          );
  // NOTE: Code from here will never execute because promise above never gets resolved
}
```
v2 will work because the promise is always resolved, v3 does not because it now depends on whether the promise is resolved in the completion callback

@kurtvan

### DIFF
--- a/lib/modules/database/reference.js
+++ b/lib/modules/database/reference.js
@@ -195,8 +195,11 @@ export default class Reference extends ReferenceBase {
     return new Promise((resolve, reject) => {
       const onCompleteWrapper = (error, committed, snapshotData) => {
         if (isFunction(onComplete)) {
-          if (error) return onComplete(error, committed, null);
-          return onComplete(null, committed, new Snapshot(this, snapshotData));
+          if (error)  { 
+            onComplete(error, committed, null); 
+          } else {
+            onComplete(null, committed, new Snapshot(this, snapshotData));
+          }
         }
 
         if (error) return reject(error);


Ehesp:
I think the reason this has been done is because if you pass in a func for onComplete, we assume you'd want to know when it's complete there rather than the promise telling you.

In general though I think it should work your way, however your PR removes the error check which is still needed. It's the `return` part which was causing the promise to not resolve.

lib/modules/database/reference.js
```diff
@@ -195,8 +195,7 @@ export default class Reference extends ReferenceBase {
     return new Promise((resolve, reject) => {
       const onCompleteWrapper = (error, committed, snapshotData) => {
         if (isFunction(onComplete)) {
-          if (error) return onComplete(error, committed, null);
-          return onComplete(null, committed, new Snapshot(this, snapshotData));
+          onComplete(error, committed, new Snapshot(this, snapshotData));
```
chrisbianca: @gyfelton To maintain the previous functionality, I believe this should be:
```
if (error) {
  onComplete(error, committed, null);
} else {
  onComplete(null, committed, new Snapshot(this, snapshotData));
}
```

gyfelton:
@chrisbianca another look?

Issue #263, opened 2017-07-19
Rewarded video not opening when called from modal on ios


CoderProject:
I'm using wix/react-native-navigation to open a modal and showing a rewarded ad from modal. Ad is loaded but never shown on ios. On Android it works perfectly.
This is the warning i get on console:
`Warning: Attempt to present <GADInterstitialViewController: 0x7ffefdd020a0> on <RCCDrawerController: 0x7ffefdc018a0> whose view is not in the window hierarchy!`

When i move the code to the root screen, the ad is shown as expected. So the problem is opening from modal. As a workaround I will close modal to show ad and then open again on onAdClosed event.

### Environment

1. Target Platform (e.g. iOS, Android): iOS
2. Development Operating System (e.g. macOS Sierra, Windows 10): macOS/ Windows 10
3. Build tools (Xcode or Android Studio version, iOS or Android SDK version, if relevant): Xcode
4. React Native version (e.g. 0.45.1): 0.45.1
5. RNFirebase Version (e.g. 2.0.2): 2.0.1

Ehesp:
When you say Modal, is that the React Native Modal or the navigation type Modal on Wixs end?

CoderProject:
its the navigation type Modal which is shown by calling `showModal()`

CoderProject:
>  As a workaround I will close modal to show ad and then open again on onAdClosed event.
When using this workaround I found another bug. I was using a class-wide variable as below:
```
export default class MyScreen extends Component {
    firebase = RNFirebase.initializeApp();
    advert = this.firebase.admob().rewarded('ca-app-pub-adunit...');
    
    constructor(props) {		
        super(props);
        this.advert.on('onAdClosed', () => {
			console.log('ad closed');
		});
    }
...
```

On first ad view it was working great, but on the second time that I open this modal, `onAdClosed` event was being called multiple times. On third time even more, I guess the events was not being removed.

So I solved it by making advert static and adding event only once:
```
export default class MyScreen extends Component {
    firebase = RNFirebase.initializeApp();
    static advert=null;
    
    constructor(props) {		
        super(props);
        if (!MyScreen.advert) {
		MyScreen.advert = this.firebase.admob().rewarded('ca-app-pub-adunit...');
		MyScreen.advert.on('onAdClosed', () => {
			console.log('ad closed');
		});
	}
    }
```

I did not understand why old `onAdClosed` events are also fired, even though a new instance of advert is created.

Salakar:
[...]

PR #615, opened 2017-11-13
Fixes bug when presenting Interstitial and RewardedVideo

ianzitow:
This pull request is intended to avoid problems with view hierarchy in iOS by finding out what is the top most view controller being shown.

### DIFF
--- a/ios/RNFirebase/RNFirebaseUtil.h
+++ b/ios/RNFirebase/RNFirebaseUtil.h
@@ -8,6 +8,8 @@
 
 + (void)sendJSEvent:(RCTEventEmitter *)emitter name:(NSString *)name body:(NSDictionary *)body;
 + (void)sendJSEventWithAppName:(RCTEventEmitter *)emitter appName:(NSString *)appName name:(NSString *)name body:(NSDictionary *)body;
++ (UIViewController*)topViewController;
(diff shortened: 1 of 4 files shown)


Ehesp:
@ianzitow thanks very much for this & sorry for the delay. Although this looks fine to me, what cases are you having where the view controller isn't the top one already? 

My iOS knowledge is fairly limited, but I was under the impression RN used a single VC?

ianzitow:
@Ehesp Looks like when using React Native Navigation by Wix, it really pushes new VC on top of another, which can cause this kind of issue: "Warning: Attempt to present VC1 on VC2 whose view is not in the window hierarchy!" #263. 

I noticed that the current code assumes the iOS app just needs a single VC since it tries to present always from a root VC. Thus, I think it need to be extended so we can avoid problems when the root itself is not the top.

However, let me run more tests to confirm everything is fine.

ovaris:
@ianzitow any progress on this? ;)

Salakar:
Closing as stale, no responses from original author for several months. Happy to re-open on request. We'll internally look into this as part of the AdMob re-write if not, sometime in the future.

Thanks

nenjamin2405:
Any update on this guys, currently I'm facing this issue :(

daheeahn:
I still faced same issue.

PR #633, opened 2017-11-21
[android][messaging] Handle SecuritException in cancelAllLocalNotific…

akshetpandey:
…ations


akshetpandey:
Context: https://stackoverflow.com/questions/36634008/why-notificationmanagercompatcancelall-gets-securityexception

akshetpandey:
I can see crashes for this in google play console but now in firebase crash reporting :|

Salakar:
@akshetpandey thanks for this, merging for now but, we'll probably look at making this a promise in v3.2 - then exceptions can be sent to JS to be handled as and when needed.

Issue #654, opened 2017-11-29
React-native-firebase supports only one language in the email template


StepanNaryshkov:
I use **react-native-firebase** for my project and I can not set a language in the email template. For example, I use [`sendPasswordResetEmail`](https://firebase.google.com/docs/auth/web/manage-users#send_a_password_reset_email) function to send an email, but I can not set language from my device (only in my [firebase console](https://console.firebase.google.com)). In the web, I can set it [`firebase.auth().languageCode = 'fr';`](https://firebase.google.com/docs/auth/web/manage-users#send_a_password_reset_email). Does somebody have any ideas?

Ehesp:
WIP PR created: https://github.com/invertase/react-native-firebase/pull/655

chrisbianca:
This will be released in our next version, either v3.2.3 or v3.3.0

scaralfred:
Was did fixed? What is the exact fucntion to get this working?
I tried auth().useDeviceLanguage().  and auth().setLanguage(appLanguage)
but they did not work

mikehardy:
I just tested this in my app @scaralfred and I'm glad you made me do so! I had an error and was *not* calling `auth().setLanguageCode('my code')` correctly after a re-factoring.

Now, before I send the verification email I call `auth().setLanguage` with either `en` or `es` (my two codes) as the language argument and I get an email template in either English or Spanish as I would expect.

It works for me?

PR #655, opened 2017-11-30
[WIP] Set auth language #654

Ehesp:
This PR gives the ability to set the language for the auth module, or use the device language. Usage is the same as the web sdk:

https://firebase.google.com/docs/auth/web/manage-users#send_a_password_reset_email

- [x] Set language code
- [x] Set language via device default
- [x] PR for documentation
- [x] Get language code
- [x] Updated typescript/flowtype
- [x] Handle dynamic apps
- [x] Tests

### DIFF
--- a/android/src/main/java/io/invertase/firebase/auth/RNFirebaseAuth.java
+++ b/android/src/main/java/io/invertase/firebase/auth/RNFirebaseAuth.java
@@ -6,6 +6,7 @@ import android.util.Log;
 import android.net.Uri;
 import android.support.annotation.NonNull;
 
+import java.util.Map;
 import java.util.concurrent.TimeUnit;
 import java.util.HashMap;
 import java.util.List;
@@ -1438,5 +1439,30 @@ class RNFirebaseAuth extends ReactContextBaseJavaModule {
     eventMap.putString("type", type);
     eventMap.putMap("state", state);
     Utils.sendEvent(mReactContext, "phone_auth_state_changed", eventMap);
+  }
+
+  /**
+   * Constants bootstrapped on react native app boot
+   *
+   * @return
+   */
+  @Override
+  public Map<String, Object> getConstants() {
+    Map<String, Object> constants = new HashMap<>();
+
+    List<FirebaseApp> firebaseAppList = FirebaseApp.getApps(getReactApplicationContext());
+    final Map<String, Object> appLanguage = new HashMap<>();
+
+    for (FirebaseApp app : firebaseAppList) {
+      String appName = app.getName();
+
+      FirebaseApp instance = FirebaseApp.getInstance(appName);
+      FirebaseAuth firebaseAuth = FirebaseAuth.getInstance(instance);
+
+      appLanguage.put(appName, firebaseAuth.getLanguageCode());
+    }
+
+    constants.put("APP_LANGUAGE", appLanguage);
+    return constants;
   }
 }
--- a/android/src/main/java/io/invertase/firebase/auth/RNFirebaseAuth.java
+++ b/android/src/main/java/io/invertase/firebase/auth/RNFirebaseAuth.java
@@ -1129,6 +1129,7 @@ class RNFirebaseAuth extends ReactContextBaseJavaModule {
   /**
    * fetchProvidersForEmail
    *
+   * @param appName
    * @param promise
    */
   @ReactMethod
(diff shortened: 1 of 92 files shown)


lib/modules/auth/index.js
```diff
@@ -322,6 +322,23 @@ export default class Auth extends ModuleBase {
     return this._native.fetchProvidersForEmail(email);
   }
 
+  /**
+   * Sets the language for the auth module using the device language
+   * @returns {*}
+   */
+  useDeviceLanguage() {
+    return this._native.useDeviceLanguage();
```
chrisbianca: No need to return here or on languageCode

Ehesp: Ah yes!

lib/modules/auth/index.js
```diff
@@ -322,6 +322,23 @@ export default class Auth extends ModuleBase {
     return this._native.fetchProvidersForEmail(email);
   }
 
+  /**
+   * Sets the language for the auth module using the device language
+   * @returns {*}
+   */
+  useDeviceLanguage() {
```
chrisbianca: Can you add some flow typings?

lib/index.d.ts
```diff
@@ -637,6 +637,11 @@ declare module "react-native-firebase" {
          */
         currentUser: User | null
 
+        /**
+         * Gets/Sets the language for the app instance
+         */
+        languageCode: string | null;
```
Ehesp: cc @chrisbianca not sure how to do a getter/setter in TypeScript...? This works for the getter but not sure if it's moan when trying to set it.

aMarCruz: @Ehesp it is ok, `languageCode` is a read/write property, the implementation is hidden.

compojoom:
@ecirish - I said this on discord, but I'll note it here as well. 
It seems that on ios there is an issue - probably not in our code, but it would be good to confirm where the error is.

Using auth.languageCode = 'en' - would sent a firebase email in english. But using auth().useDeviceLanguage() would default to firebase's project default and not the device language.

On Android this seems to behave correctly. It would be good if someone else tests useDeviceLanguage() on iOS as well before we merge this.

Ehesp:
If that is the case, it seems as though it's a Firebase issue? The code on our end is quite simple:

```
[[FIRAuth authWithApp:firApp] useAppLanguage];
```

I'll try and confirm this, but if it is different to Android then we have another discepancy.

compojoom:
I know it seems to be a firebase issue and if it really is - then we can add it to the notes and also submit a bug report to firebase. If we just ship it as it is and it doesn't work, then other people will waste time on this as well :(

Ehesp:
@compojoom We discussed this yesterday internally, for the time being we're just going to support `languageCode`, and add `useDeviceLanguage` as an unsupported method. We've flagged it with Google. 

A workaround would be to use another lib to grab the device language and set it via `languageCode`.

compojoom:
Ok, so when is  3.2.1 coming out :)

android/src/main/java/io/invertase/firebase/auth/RNFirebaseAuth.java
```diff
@@ -1177,6 +1179,31 @@ public void onComplete(@NonNull Task<ProviderQueryResult> task) {
...
+
+    firebaseAuth.setLanguageCode(code);
+  }
+
+  /**
+   * Use the device language
+   * @param appName
+   */
+  @ReactMethod
+  public void useDeviceLanguage(String appName) {
```
Ehesp: @Salakar @chrisbianca should I leave the native ones here for the time being? JS currently flags as unsupported due to Firebase issue, but no harm in them being here?

Ehesp:
@compojoom can confirm the iOS issue! Just need to sort one last thing and should be ready to get merged.

Ehesp:
cc @chrisbianca @Salakar All good to merge is you chaps are happy

lib/utils/index.js
```diff
@@ -340,6 +340,30 @@ export function nativeToJSError(code: string, message: string, additionalProps?:
   return error;
 }
 
+/**
```
chrisbianca: This has moved to lib/utils/native.js so doesn't need to be here anymore - I assume this was from a merge?

lib/modules/auth/index.js
```diff
@@ -342,6 +353,14 @@ export default class Auth extends ModuleBase {
     return this._user;
   }
 
+  get namespace(): string {
```
chrisbianca: Just noticed that this is no longer required either.

goleador0322:
I tried to use useDeviceLanguage function (like firebase.auth().userDeviceLanguage)

but, I got this error.
```
ExceptionsManager.js:65 _reactNativeFirebase2.default.auth(...).useDeviceLanguage is not a function
```

how can i fix this :( ?


Podfile.lock
```
- Firebase/Auth (4.8.1):
    - Firebase/Core
    - FirebaseAuth (= 4.4.2)
  - Firebase/Core (4.8.1):
    - FirebaseAnalytics (= 4.0.7)
    - FirebaseCore (= 4.0.14)
  - FirebaseAnalytics (4.0.7):
    - FirebaseCore (~> 4.0)
    - FirebaseInstanceID (~> 2.0)
    - GoogleToolboxForMac/NSData+zlib (~> 2.1)
    - nanopb (~> 0.3)
  - FirebaseAuth (4.4.2):
    - FirebaseAnalytics (~> 4.0)
    - GoogleToolboxForMac/NSDictionary+URLArguments (~> 2.1)
    - GTMSessionFetcher/Core (~> 1.1)
  - FirebaseCore (4.0.14):
    - GoogleToolboxForMac/NSData+zlib (~> 2.1)
  - FirebaseInstanceID (2.0.8):
    - FirebaseCore (~> 4.0)
```

yarn.lock
```
react-native-firebase@^3.2.2:
  version "3.2.2"
  resolved "https://registry.yarnpkg.com/react-native-firebase/-/react-native-firebase-3.2.2.tgz#768d0384de489ffcee106f53c9fbb41cee7eebb0"
  dependencies:
    bows "^1.6.0"
    opencollective "^1.0.3"
    postinstall-build "^5.0.1"
    prop-types "^15.6.0"
```

Issue #661, opened 2017-12-03
Not receiving Dynamic Link when app is installed from the appstore


seblau:
### Issue

I followed your setup guide and adapted `AppDelegate.m` accordingly. Dynamic Links are working if the app is already installed, but when the app is installed from the appstore, I do not receive the dynamic link. Instead I get the following link: `[CUSTOM_SCHEME]://google/link/?dismiss=1&is_weak_match=1` which indicates that FDL is working correctly but could not extract any dynamic link.

Probably closely related to: https://github.com/firebase/firebase-ios-sdk/issues/233

### Environment

I am currently testing this by uninstalling the app on the device -> clicking the dynamic link -> redirects to the appstore -> install the app from xcode.

1. Application Target Platform: 
* iOS 

2. Development Operating System: 
* macOS 10.11.6

3. Build Tools: 
* Xcode 8.2.1
* iOS 10.2.1

4. React Native version: 
* react-native@0.43.4

5. Firebase Module: 
* react-native-firebase@3.1.1

**Podfile.lock**

```
PODS:
  - Firebase/Core (4.7.0):
    - FirebaseAnalytics (= 4.0.5)
    - FirebaseCore (= 4.0.12)
  - Firebase/DynamicLinks (4.7.0):
    - Firebase/Core
    - FirebaseDynamicLinks (= 2.3.0)
  - Firebase/Messaging (4.7.0):
    - Firebase/Core
    - FirebaseMessaging (= 2.0.7)
  - FirebaseAnalytics (4.0.5):
    - FirebaseCore (~> 4.0)
    - FirebaseInstanceID (~> 2.0)
    - GoogleToolboxForMac/NSData+zlib (~> 2.1)
    - nanopb (~> 0.3)
  - FirebaseCore (4.0.12):
    - GoogleToolboxForMac/NSData+zlib (~> 2.1)
  - FirebaseDynamicLinks (2.3.0):
    - FirebaseAnalytics (~> 4.0)
  - FirebaseInstanceID (2.0.7)
  - FirebaseMessaging (2.0.7):
    - FirebaseAnalytics (~> 4.0)
    - FirebaseCore (~> 4.0)
    - FirebaseInstanceID (~> 2.0)
    - GoogleToolboxForMac/Logger (~> 2.1)
    - Protobuf (~> 3.1)
  - Flurry-iOS-SDK/FlurrySDK (7.10.0)
  - GoogleToolboxForMac/Defines (2.1.3)
[...]

PR #668, opened 2017-12-05
renamed installLink to initalLink

Codeseer:
Made it more clear this is just caching the initialLink since on first install on iOS the launchOptions are not set by Firebase. See #661


Salakar:
Have a manual patch based on the previous PR you did that I will merge at some point - just getting a couple people to test it.

Thanks again!

PR #671, opened 2017-12-06
Fix typescript

jsm:
Fixes typescript typedef

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,8 +4,9 @@
   "author": "Invertase <contact@invertase.io> (http://invertase.io)",
   "description": "A well tested, feature rich Firebase implementation for React Native, supporting iOS & Android. Individual module support for Admob, Analytics, Auth, Crash Reporting, Cloud Firestore, Database, Dynamic Links, Messaging (FCM), Remote Config, Storage and Performance.",
   "main": "dist/index.js",
+  "types": "dist/index.d.ts",
   "scripts": {
-    "build": "npm run build-lib && npm run build-flow",
+    "build": "npm run build-lib && npm run build-flow && cp lib/index.d.ts dist/index.d.ts",
     "build-flow": "flow-copy-source -i */__tests__* lib dist",
     "build-lib": "BABEL_ENV=publish babel lib -d dist --ignore __tests__",
     "clean": "rimraf dist/",


chrisbianca:
@jsm thanks for flagging this and for the PR. I've slightly tweaked your PR to use babel to copy the Typescript typings over rather than explicitly using a `cp` command as this will be better supported across different environments.

You can check out the change here: https://github.com/invertase/react-native-firebase/commit/4646c6723db15d2c73bd0901dc08a6db4332b2ab

If you're able to confirm that's working for you, I can schedule a release for later today

chrisbianca:
Bit too quick on the trigger - also corrected the path to the typings here: https://github.com/invertase/react-native-firebase/commit/f96e673683d9f090afb8a30c453341a6f448dd0d

jsm:
@chrisbianca works for me. Thanks!

tm-3:
@chrisbianca @jsm 

Hi...I've been working on my own set of type defs for a few days now. I do not think they are completely right, but I have gone through everything that is documented on the rnfirebase.io site as well as looked through code and compared to the firebase sdk's. I think they are more complete than anything else I've seen. Over the next few days I'm going to start using react-native-firebase heavily while building an app and I'll get to test out many of the definitions. I'll submit a PR when I feel that it is of higher quality and well tested, but I really wanted to just get this out in the wild to see if anyone else wanted to help out. I know I have quite a few gaps to fill and I'm sure I've misunderstood some of the shapes/interfaces/classes, but another set of eyes is always appreciated. 

If anyone out there wants to check them out, you can find them [here](https://github.com/tacosandtech/react-native-firebase-types). 

Thanks!

chrisbianca:
@tm-3 Thanks for letting us know about your work.  As neither @Salakar, @Ehesp or I are using Typescript in a project it means that it's hard for us to test and often gets left behind.  We'll definitely be happy for any PR you can put together!

ChristoRibeiro:
@chrisbianca this PR is included in the last release 3.1.1?

chrisbianca:
@ChristoRibeiro no, this wasn't part of the 3.1.1 release. We should be doing another release in the next couple of days which will include this fix.

ChristoRibeiro:
Nice, i'm on the waiting list. Meawhile how can I do to make this works with typescript?

PR #697, opened 2017-12-19
[iOS] Remove crash when calling crash().report()

Minishlink:
It doesn't have the same behaviour as on Android. (and it's weird that it makes the app crash as it's non-fatal...)


Minishlink:
After looking at the Firebase iOS documentation, it doesn't seem to be possible to log non fatal crashes on Firebase. Could you confirm this is correct? That's a shame :(

Salakar:
@Minishlink correct, iirc we added this assert to make android and iOS behaviour consistent. 

Having said that, we've recently added Crashlytics support which is much better suited to do the job and has support for logging non fatal errors. See the docs: https://rnfirebase.io/docs/v3.2.x/crashlytics/reference/crashlytics

Will close this - but thanks for the PR 👍

PR #724, opened 2018-01-04
[typings][typescript] Add missing types for Crashlytics

krystofcelba:
The typescript typings for Crashlytics related functions are missing now. So this PR adds them.

Thanks for this library!


Ehesp:
Thanks for this!

PR #730, opened 2018-01-07
Set NSNumber to nonnull for storage config calls

joserocha3:
Made change requested by @chrisbianca via Discord chat. New versions of react-native require that all NSNumber arguments are explicitly marked as `nonnull` to ensure compatibility with Android.

Only changed for a few storage methods. The other method exports with NSNumber will probably need to be updated as well, but I did not have a project setup to test them.


chrisbianca:
Thanks for this!

PR #740, opened 2018-01-10
Enable foreground notification on ios

hisokakei:
Enable `show_in_foreground` flag for local notification on ios


Salakar:
LGTM - thank you for the PR

drailanjohngss:
How can I use this ?

Ehesp:
@drailanjohngss `show_in_foreground: true` in the options object.

drailanjohngss:
It working now, hmm but when I tested it, notifications are sent every second.

hlbp:
hi @drailanjohngss  did you find a solution for this? 
thanks!

kholiavko-roman:
@Ehesp Could you please help me, is 'show_in_foreground' works now ? 
I try `"show_in_foreground": true` and `"show_in_foreground":1` but it not work in current version.

PR #757, opened 2018-01-16
Add Typescript definitions for Dynamic Links

pheromonez:
Typescript support for Dynamic Links

### DIFF
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -64,6 +64,14 @@ declare module "react-native-firebase" {
      */
     crash(): RNFirebase.crash.Crash;
 
+    /**
+     * Firebase Dynamic Links are links that work the way you want, on multiple
+     * platforms, and whether or not your app is already installed.
+     * See the official Firebase docs:
+     * https://firebase.google.com/docs/dynamic-links/
+     */
+    links(): RNFirebase.links.Links;
+
     static fabric: {
       crashlytics(): RNFirebase.crashlytics.Crashlytics;
     };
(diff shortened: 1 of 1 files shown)


lib/index.d.ts
```diff
@@ -895,5 +903,60 @@ declare module "react-native-firebase" {
...
+          iosIpadBundleId?: string,
+          iosIpadFallbackLink?: string,
+        },
+        socialMetaTagInfo?: {
+          socialTitle: string,
+          socialImageLink: string,
+          socialDescription: string,
+        },
+        suffix?: {
+          option: 'SHORT' | 'UNGUESSABLE',
```
Ehesp: Is `UNGUESSABLE` correct here? Shouldn't it just be a string unless I'm missing something...!

pheromonez: It is correct. As noted in [the Firebase docs](https://firebase.google.com/docs/reference/dynamic-links/link-shortener), if `suffix` is specified then it must either have value `{ option: "SHORT"
 }` or `{ option: "UNGUESSABLE" }`.

Sample code (works):
```typescript
import firebase from "react-native-firebase";
firebase.app().links().createDynamicLink({
    dynamicLinkDomain: "<app>.goo.gl",
    link: "https://data-to-pass-to-app",
    suffix: {
        option: "UNGUESSABLE",
    },
});
```
Sample code (causes compile error due to disallowed type for `option`):
```typescript
import firebase from "react-native-firebase";
firebase.app().links().createDynamicLink({
    dynamicLinkDomain: "<app>.goo.gl",
    link: "https://data-to-pass-to-app",
    suffix: {
        option: "MADE_UP_STRING",
    },
});
```

Salakar: Confirmed correctly typed as per docs. Will merge as all else seems good.

PR #761, opened 2018-01-17
[ios][links] Cache initial link for getInitialLink

Salakar:
See discord general chat for discussion and confirmation of this fix - pinned the conversation with a pr number tag.

Partially addresses https://github.com/invertase/react-native-firebase/issues/661

### DIFF
--- a/ios/RNFirebase/links/RNFirebaseLinks.m
+++ b/ios/RNFirebase/links/RNFirebaseLinks.m
@@ -130,7 +130,7 @@ RCT_EXPORT_METHOD(getInitialLink:(RCTPromiseResolveBlock)resolve rejecter:(RCTPr
         NSURL* url = (NSURL*)self.bridge.launchOptions[UIApplicationLaunchOptionsURLKey];
         [self handleInitialLinkFromCustomSchemeURL:url resolver:resolve rejecter:reject];        
     } else if(initialLink) {
-         [self handleInitialLinkFromCustomSchemeURL:initialLink resolver:resolve rejecter:reject];
+        [self handleInitialLinkFromCustomSchemeURL:initialLink resolver:resolve rejecter:reject];
     } else {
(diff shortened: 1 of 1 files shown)


ios/RNFirebase/links/RNFirebaseLinks.m
```diff
@@ -123,8 +128,9 @@ -(void)handleInitialLinkFromUniversalLinkURL:(NSDictionary *)userActivityDiction
 RCT_EXPORT_METHOD(getInitialLink:(RCTPromiseResolveBlock)resolve rejecter:(RCTPromiseRejectBlock)reject) {
     if (self.bridge.launchOptions[UIApplicationLaunchOptionsURLKey]) {
         NSURL* url = (NSURL*)self.bridge.launchOptions[UIApplicationLaunchOptionsURLKey];
-        [self handleInitialLinkFromCustomSchemeURL:url resolver:resolve rejecter:reject];
-        
+        [self handleInitialLinkFromCustomSchemeURL:url resolver:resolve rejecter:reject];        
+    } else if(initialLink) {
```
Salakar: @chrisbianca thoughts on this, was confirmed working on discord, however have a feeling this should be last resort, so after the final else block i.e. if `handleInitialLinkFromUniversalLinkURL` couldn't find a link then try the cached one - does mean `handleInitialLinkFromUniversalLinkURL` would need a bit of a rewrite to not use the promise - so that we can check it's return value.

Codeseer: When firebase is installed from the app store there is no "native" iOS open url... because it was installed from app store and opened manually by the user. Firebase "fakes" the open url by manually invoking the `application:openURL:options:` delegate.

Not sure if there is a better way to do this since we need to manage the state of the initialLink due to this firebase behavior.

PR #771, opened 2018-01-22
Add Typescript Definitions for Cloud Firestore

pheromonez:
Add the missing typings for the firestore module

### DIFF
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -72,6 +72,15 @@ declare module "react-native-firebase" {
      */
     links(): RNFirebase.links.Links;
 
+    /**
+     * Cloud Firestore is Firebase's new flagship database solution for mobile
+     * development, however as at the time of this writing (22 Jan 2018) it is
+     * still a beta product, and may not be as stable as Firebase Realtime
+     * Database. Comparison of the two products hers:
+     * https://firebase.google.com/docs/database/rtdb-vs-firestore
+     */
+    firestore(): RNFirebase.firestore.FirestoreModule;
+
     static fabric: {
       crashlytics(): RNFirebase.crashlytics.Crashlytics;
     };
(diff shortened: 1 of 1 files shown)


Salakar:
LGTM - cc @Ehesp 

@pheromonez thanks for this - out of interest did you pull these from the web sdk or write them up yourself?

pheromonez:
@Salakar I wrote these based on the code from this project. I'm not 100% sure that the 'statics' part of the typings are correct. I assumed that when instantiating a one of those types (e.g `FieldPath`) we have to access it through the `firestore()` module:
```typescript
import firebase from "react-native-firebase";
const firestore = firebase.app().firestore();
const fieldPath = new firestore.FieldPath("a", "path", "to", "something");
```

Is this the correct bahaviour?

lib/index.d.ts
```diff
@@ -1067,5 +1076,276 @@ declare module "react-native-firebase" {
...
+      class DocumentChange {
+        constructor(firestore: Firestore, nativeData: Types.NativeDocumentChange);
+        get doc(): DocumentSnapshot;
+        get newIndex(): number;
+        get oldIndex(): number;
+        get type(): string;
+      }
+
+      class DocumentReference {
+        constructor(firestore: Firestore, documentPath: Path);
```
Ehesp: Needed?

lib/index.d.ts
```diff
@@ -1067,5 +1076,276 @@ declare module "react-native-firebase" {
...
+        onSnapshot(documentListenOptions: DocumentReference.DocumentListenOptions, onNext: DocumentReference.ObserverOnNext, onError?: DocumentReference.ObserverOnError): () => void;
+        onSnapshot(documentListenOptions: DocumentReference.DocumentListenOptions, observer: DocumentReference.Observer): () => void;
+        set(data: object, writeOptions?: Types.WriteOptions): Promise<void>;
+        update(obj: object): Promise<void>;
+        update(key1: Types.UpdateKey, val1: any): Promise<void>;
+        update(key1: Types.UpdateKey, val1: any, key2: Types.UpdateKey, val2: any): Promise<void>;
+        update(key1: Types.UpdateKey, val1: any, key2: Types.UpdateKey, val2: any, key3: Types.UpdateKey, val3: any): Promise<void>;
+        update(key1: Types.UpdateKey, val1: any, key2: Types.UpdateKey, val2: any, key3: Types.UpdateKey, val3: any, key4: Types.UpdateKey, val4: any): Promise<void>;
+        update(key1: Types.UpdateKey, val1: any, key2: Types.UpdateKey, val2: any, key3: Types.UpdateKey, val3: any, key4: Types.UpdateKey, val4: any, key5: Types.UpdateKey, val5: any): Promise<void>;
+      }
```
Ehesp: `isEqual(other: DocumentReference): boolean;` - need to check our code here. [ref](https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore-types/index.d.ts#L441)

lib/index.d.ts
```diff
@@ -1067,5 +1076,276 @@ declare module "react-native-firebase" {
...
+          data: {
+            [key: string]: TypeMap;
+          };
+          metadata: SnapshotMetadata;
+          path: string;
+        }
+
+        interface SnapshotMetadata {
+          fromCache: boolean;
+          hasPendingWrites: boolean;
```
Ehesp: `isEqual(other: SnapshotMetadata): boolean;` [ref](https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore-types/index.d.ts#L600)

lib/index.d.ts
```diff
@@ -1067,5 +1076,276 @@ declare module "react-native-firebase" {
...
+      }
+
+      class DocumentSnapshot {
+        constructor(firestore: Firestore, nativeData: Types.NativeDocumentSnapshot)
+        get exists(): boolean;
+        get id(): string | null;
+        get metadata(): Types.SnapshotMetadata;
+        get ref(): DocumentReference;
+        data(): object | void;
+        get(fieldPath: string | FieldPath): any | undefined;
```
Ehesp: `isEqual(other: DocumentSnapshot): boolean;` [ref](https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore-types/index.d.ts#L668)

lib/index.d.ts
```diff
@@ -1067,5 +1076,276 @@ declare module "react-native-firebase" {
...
+        onSnapshot(observer: Observer): () => void;
+        onSnapshot(queryListenOptions: Query.QueryListenOptions, onNext: Query.ObserverOnNext, onError?: Query.ObserverOnError): () => void;
+        onSnapshot(queryListenOptions: Query.QueryListenOptions, observer: Query.Observer): () => void;
+        orderBy(fieldPath: string | FieldPath, directionStr?: Types.QueryDirection): Query;
+        startAfter(snapshot: DocumentSnapshot): Query;
+        startAfter(...varargs: any[]): Query;
+        startAt(snapshot: DocumentSnapshot): Query;
+        startAt(...varargs: any[]): Query;
+        where(fieldPath: string, op: Types.QueryOperator, value: any): Query;
+      }
```
Ehesp: `isEqual(other: Query): boolean;` [ref](https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore-types/index.d.ts#L874)

lib/index.d.ts
```diff
@@ -1067,5 +1076,276 @@ declare module "react-native-firebase" {
...
+
+      class QuerySnapshot {
+        constructor(firestore: Firestore, query: Query, nativeData: QuerySnapshot.NativeData);
+        get docChanges(): DocumentChange[];
+        get docs(): DocumentSnapshot[];
+        get empty(): boolean;
+        get metadata(): Types.SnapshotMetadata;
+        get query(): Query;
+        get size(): number;
+        forEach(callback: (snapshot: DocumentSnapshot) => any);
```
Ehesp: JS docs have `forEach(cb, thisArg?: any`) - applicable here? [ref](https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore-types/index.d.ts#L969)

lib/index.d.ts
```diff
@@ -1067,5 +1076,276 @@ declare module "react-native-firebase" {
...
+        FieldValue: typeof FieldValue;
+        GeoPoint: typeof GeoPoint;
+        enableLogging(enabled: boolean): void;
+      };
+
+      class CollectionReference {
+        constructor(firestore: Firestore, collectionPath: Path);
+        get firestore(): Firestore;
+        get id(): string;
+        get parent(): DocumentReference;
```
Ehesp: missing path? [ref](https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore-types/index.d.ts#L1036)

lib/index.d.ts
```diff
@@ -1067,5 +1076,276 @@ declare module "react-native-firebase" {
...
+        onSnapshot(observer: Observer): () => void;
+        onSnapshot(queryListenOptions: Query.QueryListenOptions, onNext: Query.ObserverOnNext, onError?: Query.ObserverOnError): () => void;
+        onSnapshot(queryListenOptions: Query.QueryListenOptions, observer: Query.Observer): () => void;
+        orderBy(fieldPath: string | FieldPath, directionStr?: Types.QueryDirection): Query;
+        startAfter(snapshot: DocumentSnapshot): Query;
+        startAfter(...varargs: any[]): Query;
+        startAt(snapshot: DocumentSnapshot): Query;
+        startAt(...varargs: any[]): Query;
+        where(fieldPath: string, op: Types.QueryOperator, value: any): Query;
+      }
```
Ehesp: `isEqual(other: CollectionReference): boolean;` missing? [ref](https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore-types/index.d.ts#L1064)

lib/index.d.ts
```diff
@@ -1067,5 +1076,276 @@ declare module "react-native-firebase" {
...
+
+      class FieldPath {
+        static documentId(): FieldPath;
+        constructor(...segments: string[]);
+      }
+
+      class FieldValue {
+        static delete(): FieldValue;
+        static serverTimestamp(): FieldValue;
+      }
```
Ehesp: `isEqual(other: FieldValue): boolean;` missing? [ref](https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore-types/index.d.ts#L1091)

lib/index.d.ts
```diff
@@ -1067,5 +1076,276 @@ declare module "react-native-firebase" {
...
+
+      class GeoPoint {
+        constructor(latitude: number, longitude: number);
+        get latitude(): number;
+        get longitude(): number;
```
Ehesp: `isEqual(other: GeoPoint): boolean;` missing? [ref](https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore-types/index.d.ts#L172)

Ehesp:
@pheromonez the correct behaviour is:

```
import firebase from "react-native-firebase";
const firestore = firebase.app().firestore;
const fieldPath = new firestore.FieldPath("a", "path", "to", "something");
```

Ehesp:
Also, seems there is `enableNetwork`  & `disableNetwork` which we should add as unsuppported.

lib/index.d.ts
```diff
@@ -1067,5 +1076,276 @@ declare module "react-native-firebase" {
...
+        update(key1: Types.UpdateKey, val1: any): Promise<void>;
+        update(key1: Types.UpdateKey, val1: any, key2: Types.UpdateKey, val2: any): Promise<void>;
+        update(key1: Types.UpdateKey, val1: any, key2: Types.UpdateKey, val2: any, key3: Types.UpdateKey, val3: any): Promise<void>;
+        update(key1: Types.UpdateKey, val1: any, key2: Types.UpdateKey, val2: any, key3: Types.UpdateKey, val3: any, key4: Types.UpdateKey, val4: any): Promise<void>;
+        update(key1: Types.UpdateKey, val1: any, key2: Types.UpdateKey, val2: any, key3: Types.UpdateKey, val3: any, key4: Types.UpdateKey, val4: any, key5: Types.UpdateKey, val5: any): Promise<void>;
+      }
+      namespace DocumentReference {
+        // JS code expects this value to be true.
+        interface DocumentListenOptions {
+          includeMetadataChanges: true;
```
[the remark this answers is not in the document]
pheromonez: Based on the following line [(ref)](https://github.com/invertase/react-native-firebase/blob/master/lib/modules/firestore/DocumentReference.js#L115):
```javascript
} else if (optionsOrObserverOrOnNext.includeMetadataChanges) {
```
if we specify the first argument of the `onSnapshot()` function to be a `DocumentListenOptions` then the `includeMetadataChanges` parameter must be defined and must be `true`, else the condition fails and we receive the runtime error on [this line](https://github.com/invertase/react-native-firebase/blob/master/lib/modules/firestore/DocumentReference.js#L144). Calling `onSnapshot({})`, `onSnapshot({ includeMetadataChanges: false })` will both result in a runtime error, whereas calling `onSnapshot({ includeMetadataChanges: true })` will not.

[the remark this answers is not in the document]
chrisbianca: @pheromonez You're right, this is actually a bug that came up recently, I just haven't had a chance to fix it yet :)

lib/index.d.ts
```diff
@@ -1067,5 +1076,276 @@ declare module "react-native-firebase" {
...
+        commit(): Promise<void>;
+        delete(docRef: DocumentReference): WriteBatch;
+        set(docRef: DocumentReference, data: object, options?: Types.WriteOptions): WriteBatch;
+        // multiple overrides for update() to allow strong-typed var_args
+        update(docRef: DocumentReference, obj: object): WriteBatch;
+        update(docRef: DocumentReference, key1: Types.UpdateKey, val1: any): WriteBatch;
+        update(docRef: DocumentReference, key1: Types.UpdateKey, val1: any, key2: Types.UpdateKey, val2: any): WriteBatch;
+        update(docRef: DocumentReference, key1: Types.UpdateKey, val1: any, key2: Types.UpdateKey, val2: any, key3: Types.UpdateKey, val3: any): WriteBatch;
+        update(docRef: DocumentReference, key1: Types.UpdateKey, val1: any, key2: Types.UpdateKey, val2: any, key3: Types.UpdateKey, val3: any, key4: Types.UpdateKey, val4: any): WriteBatch;
+        update(docRef: DocumentReference, key1: Types.UpdateKey, val1: any, key2: Types.UpdateKey, val2: any, key3: Types.UpdateKey, val3: any, key4: Types.UpdateKey, val4: any, key5: Types.UpdateKey, val5: any): WriteBatch;
```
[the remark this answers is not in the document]
pheromonez: The problem with the simpler approach is that it allows the method to be called with invalid arguments:
```javascript
update(docRef, 'field1', 2, 3)
```
Here we've got a second 'key-value pair' where the key is 3 (which is not valid, since only strings and `FieldPath` may be a key), and the value is not even in the method call (which is invalid since we must have an odd number of arguments when using key-value pairs [ref](https://github.com/invertase/react-native-firebase/blob/master/lib/modules/firestore/WriteBatch.js#L75).

The method signatures that I've written are certainly unwieldy and messy looking, and limited to only 5 key-value pairs, but they clearly document what parameters the user can pass to the function. 

I can re-write this method to use the simpler syntax if you like, however as a heavy Typescript user I would much prefer to have the longer unwieldy typings if they can guarantee that I am calling the method with the correct parameters.

lib/index.d.ts
```diff
@@ -1067,5 +1076,276 @@ declare module "react-native-firebase" {
...
+      namespace QuerySnapshot {
+        interface NativeData {
+          changes: Types.NativeDocumentChange[];
+          documents: Types.NativeDocumentSnapshot[];
+          metadata: Types.SnapshotMetadata;
+        }
+      }
+
+      class WriteBatch {
+        constructor(firestore: Firestore);
```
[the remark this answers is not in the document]
pheromonez: The constructor can be removed and `WriteBatch` can be an interface rather than a class

lib/index.d.ts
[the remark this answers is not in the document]
pheromonez: @chrisbianca Thanks for the info! The `Query` class has a similar issue: [ref](https://github.com/invertase/react-native-firebase/blob/master/lib/modules/firestore/Query.js#L204).
The TS definitions as they are now will prevent users from getting a runtime exception if they call the method with parameters that will trigger the bug, (e.g., `onSnapshot({})`)., but if these bugs are going to be addressed soon then we should change the definitions to match what you suggested earlier.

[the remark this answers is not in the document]
chrisbianca: @pheromonez Just pushed up fixes for both these issues. Thanks for the kick :)
https://github.com/invertase/react-native-firebase/commit/66a8ef6926783cbd5d1369efa71b710702628137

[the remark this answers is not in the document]
pheromonez: @chrisbianca Awesome, I'll update the TS definitions to reflect this

danielw93:
Hey hey! Unfortunately this PR doesn't compile with `tsc`

> lib/index.d.ts(1099,8): error TS1036: Statements are not allowed in ambient contexts
Lose the `;` after the interface declaration

> lib/index.d.ts(1114,30): error TS2304: Cannot find name 'Observer'.
Observer needs to be defined in this file or replaced by `any` 

> lib/index.d.ts(1186,13): error TS1086: An accessor cannot be declared in an ambient context.
> lib/index.d.ts(1187,13): error TS1086: An accessor cannot be declared in an ambient context.
> lib/index.d.ts(1193,13): error TS1086: An accessor cannot be declared in an ambient context.
> lib/index.d.ts(1194,13): error TS1086: An accessor cannot be declared in an ambient context.
> lib/index.d.ts(1195,13): error TS1086: An accessor cannot be declared in an ambient context.
> lib/index.d.ts(1196,13): error TS1086: An accessor cannot be declared in an ambient context.
Convert all getters from `get propertyName(): PropertyType` to `readonly propertyName: PropertyType`. (That would be functionally equivalent right?)

> lib/index.d.ts(1273,9): error TS7010: 'forEach', which lacks return-type annotation, implicitly has an 'any' return type.
Explicitly return either `void` or `any`

Maybe this can be fixed while resolving the merge conflicts in #788 ?

pheromonez:
@danielw93 Thanks for the heads up! I'll apply the changes in #788 as you suggested while resolving the conflict, which I'll get around to in a moment.

PR #784, opened 2018-01-26
TypeScript compiler validation of lib/index.d.ts

danielw93:
I've noticed that there were some lingering errors in the typescript declaration file and that it's really easy to make these small syntax errors when editing the file.
When you then try to use this project in a TypeScript application you will have a lot of trouble because your TypeScript compiler (`tsc`) will always throw them as errors. 

This PR introduces typescript as a devDependency and adds a `tsconfig.json` to configure the compiler to be fairly strict and only look at `lib/index.d.ts` without scanning for other files.

Further, a new npm task is added to `package.json` called `validate-ts-declarations` which just starts `tsc` with the options specified in tsconfig.json.

The validate-ts-declarations command is also added as a prerequisite for `build-lib`. This means that generating the `dist/` folder will fail if `lib/index.d.ts` is broken.

Note: This does not mean that the build will fail because the type declarations don't accurately reflect the actual code. It just checks that the present declarations are valid TypeScript.

Also the compiler will not produce any output files. It will either silently complete or throw compiler errors of the form:

>  tsc --project ./
>
> lib/index.d.ts(10,53): error TS2304: Cannot find name 'AuthCredential'.
> lib/index.d.ts(597,32): error TS8020: JSDoc types can only be used inside documentation comments.
> lib/index.d.ts(1035,27): error TS7006: Parameter 'url' implicitly has an 'any' type. 

I think it would really help to have this in the project in order to make it easier and safer to contribute missing TypeScript declarations.

### DIFF
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -5,6 +5,13 @@
 
 declare module "react-native-firebase" {
 
+  /** 3rd party provider Credentials */
+  type AuthCredential = {
+    providerId: string,
+    token: string,
+    secret: string
+  }
+
   type AuthProvider = {
     PROVIDER_ID: string,
     credential: (token: string, secret?: string) => AuthCredential,
@@ -594,7 +601,7 @@ declare module "react-native-firebase" {
        *
        * @param forceRefresh: boolean - default to false
        */
-      getIdToken(forceRefresh: boolean?): Promise<string>
+      getIdToken(forceRefresh?: boolean): Promise<string>
 
       /**
        * Link the user with a 3rd party credential provider.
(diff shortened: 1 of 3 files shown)


Salakar:
Thanks for the nice and precise PR here,

We've discussed this a little and I believe we're all in favour of having this validate as part of the build, maybe even part of the pre-commit hook.

Between us we've not really used TS, but we can see there's quite a few people relying on this so we have been pushing to update it in the past few weeks, so this will help.

cc @chrisbianca @ehesp both happy with this? TS validating as part of the build step

danielw93:
Nice! Glad to help out :)

chrisbianca:
@danielw93 another thanks, this is great!  As Mike says, we've been wanting to put some focus on TS for a while and this will definitely help.

I'm happy for this to be merged and we can look to incorporate in the pre-commit hook once we've checked it's working without issue as part of the build process.

lib/index.d.ts
```diff
@@ -5,6 +5,13 @@
 
 declare module "react-native-firebase" {
 
+  /** 3rd party provider Credentials */
+  type AuthCredential = {
```
KnisterPeter: @danielw93 All types here should better be interfaces

PR #788, opened 2018-01-29
[typings][Typescript] Fix TS definitions so modules can be accessed from the default export

### LINKED ISSUE #774: Typescript not working on `firebase.` unless used as `firebase.app().`
### Issue

I was using `firebase.app().auth` previously until discovering that `app()` bit is no longer needed. Switching to `firebase.auth` causes typescript error below (not only for auth, but in general)

example

> 
> [ts] Property 'auth' does not exist on type 'typeof FireBase'.
> any

### Environment

1. Application Target Platform: iOS, Android

2. Development Operating System: macOS High Sierra

3. Build Tools: 

4. React Native version: 0.52

5. RNFirebase Version: 3.2.1

6. Firebase Module:

pheromonez:
Fixes #774 

Have kept this commit small and only touched the parts of the code that needed to change in order to resolve the linked issue. Can now access `auth` (and other modules) using this syntax:
```
import firebase from 'react-native-firebase';
firebase.auth().signInAnonymously().then(........);
```
The current method of access is still supported:
```
import firebase from 'react-native-firebase';
firebase.app().auth().signInAnonymously().then(........);
```

### DIFF
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -583,7 +583,7 @@ declare module "react-native-firebase" {
        *
        * @param forceRefresh: boolean - default to false
        */
-      getIdToken(forceRefresh: boolean?): Promise<string>
+      getIdToken(forceRefresh?: boolean): Promise<string>
 
       /**
        * Link the user with a 3rd party credential provider.
@@ -1041,7 +1041,7 @@ declare module "react-native-firebase" {
          * Returns an unsubscribe function, call the returned function to
          * unsubscribe from all future events.
          */
-        onLink(listener: (url) => void): () => void;
+        onLink(listener: (url: string) => void): () => void;
       }
 
       /**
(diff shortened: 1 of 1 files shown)


chrisbianca:
This looks great, thank you!

@Ehesp @Salakar if you're happy we can get this merged in.

chrisbianca:
@pheromonez I've merged #771 which has thrown up a merge conflict here. Are you able to resolve and then we can get this merged too?  Thanks again!

chrisbianca:
@pheromonez Thanks for the tweaks. Has this resolved all of the compilation issues that were flagged by @danielw93 as part of #771?  If so, we can get this merged...

pheromonez:
Just pushed another change, somehow two of the fixes that @danielw93 put through in #784 got reverted. I probably merged incorrectly. Anyhow, my local Typescript compiler isn't complaining about this file so I think it's good to go :)

Ehesp:
Thanks again for this!

pheromonez:
@Ehesp Happy to help, thanks for the great library, it's saved us heaps on our current project

PR #792, opened 2018-01-30
WIP: Add generated types for TypeScript

meyer:
I converted `react-native-firebase` to TypeScript and generated type definition files with `tsc`. Currently testing in a side project. I’ll update this PR as I find and fix type bugs.

Hand-writing types is great and all, but in my (admittedly limited) experience, sidecar typedefs have always ended getting out of date pretty quickly. This branch is easy to rebase since Flow and TypeScript have a lot of similarities. It should hopefully be easy to keep up to date.

This PR is a work in progress and could do with a few extra sets of eyeballs. Most notably the `firebase.ts` and `firebase-apps.ts` have diverged significantly from the original files.

Source branch is here: https://github.com/meyer/react-native-firebase/tree/types-src

Some type information was lost in the transformation since Flow and TS are not 1:1.

If i get some time I’d like to automate the conversion of TypeScript to Flow, since there isn’t anything super wild going on in the Flow types. Alternately, if you all feel like switching to TypeScript (😜), there are quite a few tools out there for converting TS projects to Flow. Not a lot for going from Flow to TS, unfortunately.

This PR is kind of hard to review, sorry 😓

⚠ not OK to merge (yet) ⚠


types/modules/core/firebase.d.ts
```diff
@@ -0,0 +1,59 @@
...
+import Database from '../database';
+import Firestore from '../firestore';
+import Links from '../links';
+import Messaging from '../messaging';
+import Perf from '../perf';
+import Storage from '../storage';
+import Utils from '../utils';
+import { FirebaseOptions } from '../../types/index';
+export declare class Firebase {
+    admob: AdMob;
```
chrisbianca: I don't think these declarations are correct - it should export both the module constructor and the set of statics.

chrisbianca:
@meyer This looks great! We'd spoken about separating the types out into modules previously, but never considered doing it at the file level...

I'm interested in why you think there'd be benefit to automating the generation of Flow types from Typescript?  Given that the Flow types already live alongside the code (without causing anybody issues) they're much easier to maintain that way.  Unfortunately we can't have both typing libraries in the same file, and given that the RN ecosystem defaults to Flow this is our preference - so no switching to Typescript I'm afraid ;)

types/modules/core/firebase.d.ts
meyer: Heh, yeah, I found that out pretty quickly while testing (https://github.com/meyer/react-native-firebase/commit/c5e157e9f06275c0f7ac6892020ac003dcb2e49d). Still trying to figure out the best way to mix statics object with the function in an automated/inferred way. Worst case scenario would be specifying them manually, but that would lead to eventual mismatches.

meyer:
@chrisbianca it’s a pretty brute-force approach but I’m hoping it will lead to the best quality types. I can mess with some `tsconfig` settings and output a single file with module declarations if that’s what you’d prefer. I’ve got no strong feelings on the matter. My only caveat is that generating the output should be automated.

My suggestion for automating Flow types from TypeScript was mostly a joke, but there does seem to be quite a lot of tooling around converting Flow to TS. Maybe because a lot of browser React libraries are written in TypeScript? I’m not sure though. Automation of Flow types from TypeScript would be neat if Flow was the ecosystem underdog, but that’s not the case (yet, heh).

types/modules/admob/AdMobComponent.d.ts
```diff
@@ -0,0 +1,54 @@
+/// <reference types="react" />
+import * as React from 'react';
+export interface AdMobComponentProps {
```
pheromonez: This is missing the required `size: string` and `unitId: string` properties. It should also extend `ViewPropTypes` from `react-native`

types/modules/admob/Banner.d.ts
```diff
@@ -0,0 +1,4 @@
+/// <reference types="react" />
+import * as React from 'react';
+declare const Banner: React.SFC;
```
pheromonez: Not sure what happened here, but `Banner` should be a `React.SFC<AdMobComponentProps>` (or just a regular `React.Component` or `React.PureComponent`)

pheromonez:
This is cool!

There are a couple of functions that we'd probably want to manually write the definitions for (e.g. `DocumentReference.update(...)`), where the code inside the function itself requires particular argument types that aren't immediately visible simply by looking at the function signature. For instance, the signature in JS is provided as `update(...args: any[])`, and so the generated TS code also has the signature `update(...args: any[])`. If we dig into the code or read the documents, however, we see that we should call the function in the following ways:
`update({ key1: 'val1', key2: 2})` (1 arg = object)
`update('key1', 'val1', 'key2', 2)` (multiple args = alternating key/value pairs).

Is there any way to prevent 'private' variables and methods (ones prefixed with an underscore) from being emitted in the type definition files?

meyer:
@pheromonez good call with the private method suggestion. I’ve just updated the types to specify which methods and attributes are private. I ended up finding quite a bit of unused code as well. Bonus!

For functions that could do with more specific signatures (like the `update` example you provided), feel free to post documentation links here and I’ll update the types in this PR accordingly. I’ve done the closest thing to a direct translation from Flow to TS, so any place that Flow types are a bit too lenient will lead to TypeScript types being lenient as well.

Related: some of the changes I’m making should probably make their way back to the Flow types, but that’s probably a task for another person/PR.

systemride:
Hello! I was hoping to add some type definitions for remote config although due to the changes in this pull request, I didn't want to create more conflicts. Is there any update on this?

PR #796, opened 2018-01-30
[ios] initializeApp initializes the default app if no appName is given

boogie666:
iOS FIRApp throws when trying to initialize the app default app via the initializeApp(options) method.

Android version seems not to do that. 

This PR brings the two closer together by allowing the default app to be initialized from js with the initializeApp method.

### DIFF
--- a/ios/RNFirebase/RNFirebase.m
+++ b/ios/RNFirebase/RNFirebase.m
@@ -45,7 +45,11 @@ RCT_EXPORT_METHOD(initializeApp:
             firOptions.bundleID = [[NSBundle mainBundle] objectForInfoDictionaryKey:@"CFBundleIdentifier"];
 
             NSString *appName = [RNFirebaseUtil getAppName:appDisplayName];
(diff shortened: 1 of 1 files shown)


Salakar:
@boogie666 thanks for this PR. Am a bit confused though about why this is needed? There should aways be an app name technically as the default app never goes through this logic - it's initialised automatically as per docs.

See: https://rnfirebase.io/docs/v3.2.x/core/initialize-apps

boogie666:
Basically if you want to use messaging you need to use the default app.
But you can't load the default app from a dynamic config.

And in my situation i need both :P. 

plus as i mentioned in my comment, it works fine on android, i think it's a good idea to have the apis as similar as possible.

Salakar:
> But you can't load the default app from a dynamic config.

That's intentional - you have to create this one natively by setting up google service plist file on ios and json file on android. Default should **never** be created dynamically in JS code.

PR #827, opened 2018-02-11
Firestore Transactions

Salakar:
Adds firestore transaction support for both Android & iOS


HofmannZ:
@Salakar, I see you use WIP on most PRs. We do that too but use this nice little integration: [https://github.com/apps/wip](https://github.com/apps/wip). As long as WIP in in the title you cannot merge the PR. Though you might find it useful, cheers!

Salakar:
@HofmannZ Interesting! Thanks for that :)

lib/modules/firestore/Transaction.js
```diff
@@ -0,0 +1,179 @@
...
+   * @param options SetOptions An object to configure the set behavior.
+   *        Pass {merge: true} to only replace the values specified in the data argument.
+   *        Fields omitted will remain untouched.
+   *
+   * @returns {Transaction}
+   */
+  set(
+    documentRef: DocumentReference,
+    data: Object,
+    options?: SetOptions
```
chrisbianca: We need to add SetOptions to the main `DocumentReference.set()` method as well as this was added to the Firestore SDKs recently.

chrisbianca: Scrap that, was thinking of something else

lib/modules/firestore/TransactionHandler.js
```diff
@@ -0,0 +1,231 @@
...
+  type: 'update' | 'error' | 'complete',
+  error: ?{ code: string, message: string },
+};
+
+/**
+ * @class TransactionHandler
+ */
+export default class TransactionHandler {
+  _firestore: Firestore;
+  _transactionListener: Function;
```
chrisbianca: Is there any reason to store this as we never unsubscribe from the listener?

lib/modules/firestore/TransactionHandler.js
```diff
@@ -0,0 +1,231 @@
...
+      } else {
+        pendingResult = await possiblePromise;
+      }
+    } catch (exception) {
+      updateFailed = true; // in case the user rejects with nothing
+      finalError = exception;
+    }
+
+    // reject the final promise and remove from native
+    if (updateFailed) {
```
chrisbianca: At the moment, this only exits if the promise throws an exception, not if the promise isn't actually a promise. Is this correct?

lib/modules/firestore/TransactionHandler.js
Salakar: Hmm true good point, I just based it off the original stuff I did on RTDB transaction, guess we should remove storing on there as well tbh: https://github.com/invertase/react-native-firebase/blob/master/lib/modules/database/transaction.js#L24

lib/modules/firestore/TransactionHandler.js
Salakar: Exits when the update function throws an error (e.g. bad code) or it Promise rejects. Added this a a boolean flag instead of `if (finalError) { } ` as it's possible to `Promise.reject()` with no value/exception which would fall through and not exit.

lib/modules/firestore/TransactionHandler.js
chrisbianca: Yeah, may as well.
When I did the flow types for auth and firestore, I cleaned up a lot of this sort of stuff, but haven't got to database yet.

lib/modules/firestore/TransactionHandler.js
chrisbianca: Ok, was just checking I could only see the boolean being set to true in the try/catch

lib/modules/firestore/TransactionHandler.js
Salakar: Will remove on both, ta

PR #832, opened 2018-02-14
Flow: Use literal values for strings 

skovhus:
I understand that you are working on improving the Flow types and doing some rewriting. But I think I found a long hanging fruit, that I tried to fix.

A few places `react-native-firebase` is not using literal values (`string`) for Flow strings types, but object wrapper (`String`). This seems like a bug to me... And this actually causes a project using  remote config to fail Flow validation:
```
Error: src/main.js:139
139:         .getValue('hasExperimentalFeature')
                       ^^^^^^^^^^^^^^^^^^^^^^^^ string. This type is incompatible with the expected param type of
126:   getValue(key: String) {
                     ^^^^^^ String. See: node_modules/react-native-firebase/dist/modules/config/index.js.flow:126
```

### Background
https://flow.org/en/docs/types/primitives/

https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoAxnAdgZwC5g4CMAXIXgE4CWWA5mALxgDkAngKY7Mbb6EBMZAMqUa9JsyxxmQA

### DIFF
--- a/lib/modules/config/index.js
+++ b/lib/modules/config/index.js
@@ -123,7 +123,7 @@ export default class RemoteConfig extends ModuleBase {
    *    "source" : OneOf<String>(remoteConfigSourceRemote|remoteConfigSourceDefault|remoteConfigSourceStatic)
    *  }
    */
-  getValue(key: String) {
+  getValue(key: string) {
     return getNativeModule(this)
       .getValue(key || '')
       .then(this._nativeValueToJS);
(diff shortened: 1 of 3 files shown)


skovhus:
Note that there is still a bunch of errors after just installing `react-native-firebase`, mainly in `dist/modules/firestore` and `/dist/utils`. Any recommendations here?

chrisbianca:
@skovhus Good spot, thanks for this!  Out of interest, which version of Flow are you running which is giving the errors?

skovhus:
@chrisbianca Thanks. : )

Most versions of Flow should reports this. I'm using `0.63.0`... One reason why you haven't found this might be that no tests is using Flow – so the interface is not validated?

Note that there is still a bunch of errors after just installing react-native-firebase, mainly in `dist/modules/firestore` and `/dist/utils`. Any recommendations here?

chrisbianca:
Can you paste the list of errors?  Have you updated your .flowconfig recently to match the ruleset that RN use for their latest version?

skovhus:
> Can you paste the list of errors? Have you updated your .flowconfig recently to match the ruleset that RN use for their latest version?

I created an issue for this https://github.com/invertase/react-native-firebase/issues/834

Thanks for the quick response.

Issue #834, opened 2018-02-15
Flow errors 


skovhus:
### Issue

With a standard React Native .flowconfig setup using 0.63.0 and React Native 0.53.0, I'm getting 25 errors after installing react-native-firebase:

```
$ flow --show-all-errors --quiet
Error: node_modules/react-native-firebase/dist/modules/auth/PhoneAuthListener.js.flow:113
113:         this[`_${type}Handler`].bind(this)
             ^^^^^^^^^^^^^^^^^^^^^^^ computed property. Indexable signature not found in
113:         this[`_${type}Handler`].bind(this)
             ^^^^ PhoneAuthListener

Error: node_modules/react-native-firebase/dist/modules/core/firebase-app.js.flow:133
133:       this[key] = props[key];
           ^^^^^^^^^ assignment of computed property/element. Indexable signature not found in
133:       this[key] = props[key];
           ^^^^ App

Error: node_modules/react-native-firebase/dist/modules/database/reference.js.flow:317
317:           .then(() => onComplete(null, newRef))
                           ^^^^^^^^^^^^^^^^^^^^^^^^ call of `onComplete`. Function cannot be called on possibly undefined value
317:           .then(() => onComplete(null, newRef))
                           ^^^^^^^^^^ undefined

Error: node_modules/react-native-firebase/dist/modules/database/reference.js.flow:319
319:           .catch(error => onComplete(error, null))
                               ^^^^^^^^^^^^^^^^^^^^^^^ call of `onComplete`. Function cannot be called on possibly undefined value
319:           .catch(error => onComplete(error, null))
                               ^^^^^^^^^^ undefined

Error: node_modules/react-native-firebase/dist/modules/firestore/DocumentReference.js.flow:103
                      v
103:       observer = {
104:         next: optionsOrObserverOrOnNext,
105:         error: observerOrOnNextOrOnError,
106:       };
           ^ object literal. This type is incompatible with
 93:     let observer: Observer;
                       ^^^^^^^^ Observer
  Property `error` is incompatible:
    105:         error: observerOrOnNextOrOnError,
                        ^^^^^^^^^^^^^^^^^^^^^^^^^ Observer. This type is incompatible with
     30:   error?: ObserverOnError,
                   ^^^^^^^^^^^^^^^ ObserverOnError
      Callable property is incompatible:
         30:   error?: ObserverOnError,
                       ^^^^^^^^^^^^^^^ ObserverOnError. Callable signature not found in
[...]

PR #835, opened 2018-02-15
Exclude lib folder from published package

skovhus:
We exclude the `lib` folder to:
- avoid duplicate module errors from Flow*
- to reduce the package size

Example of the duplicated module error:
```
Error: node_modules/react-native-firebase/lib/modules/core/firebase.js:0
Firebase. Duplicate module provider
current provider. See: node_modules/react-native-firebase/dist/modules/core/firebase.js:0
```

See context https://github.com/invertase/react-native-firebase/issues/834

WARNING: If some users are using the `lib` folder directly then it will fail for them.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -62,6 +62,8 @@ ios/RnFirebase.xcodeproj/xcuserdata
 .Trashes
 ehthumbs.db
 Thumbs.dbandroid/gradle
+
+# Project folders
 android/gradlew
 android/build
 android/gradlew.bat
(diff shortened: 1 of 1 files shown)


chrisbianca:
@skovhus Unfortunately, this isn't going to work for us. 

By adding `lib` to `.npmignore`, it also excludes the `lib` directory when installing a specific commit / branch directly from Git.  This is something that's incredibly useful when working on new features as it allows the `dist` folder to be build dynamically.

Instead, I'm going to just remove the `@providesModule` declaration - it looks like this is a hangover from past versions of RN and removing it doesn't have any effect, other than to hopefully remove the error you're seeing: https://github.com/invertase/react-native-firebase/commit/5ca722251795ba36fc6d9cea0d6deab2399002f2

We'll get a release out with this change in over the next few days.

PR #836, opened 2018-02-15
Remove error log from RNFirebaseRemoteConfig

skovhus:
In case the app is offline and the consumer of this library is using Remote Config, it is perfectly fine to get a [`FIRRemoteConfigErrorInternalError`](https://firebase.google.com/docs/reference/ios/firebaseremoteconfig/api/reference/Enums/FIRRemoteConfigError) error, but there is really no reason to show a red error while developing.

screenshot 2018-02-15 19 14 27

Instead, we should leave it up to the consumer of this library to handle any errors.

It seems that only that only `RNFirebaseRemoteConfig.m` should be fixed to not use `RCTLogError` where we also reject.

`RNFirebaseMessaging.m` is also using `RCTLogError` but actual errors where `here is no completion handler with completionHandlerId`.


chrisbianca:
This seems reasonable, thanks for the PR

PR #838, opened 2018-02-19
Move podspec to where "react-native link" expects it to be

mars-lan:


chrisbianca:
@mars-lan Thanks for this, but we have deliberately not put the podspec here because we want `react-native link` to link the `.xcodeproj` file and not the Pod by default.

By using the `.xcodeproj` file, the process can be completed without user intervention.

If we use the podspec, then the user would have to ensure that React is also installed as a Pod, something that can be difficult to get working.

We have kept the podspec to allow people to use Cocoapods if they wish to.

mars-lan:
Unfortunately `react-native link` doesn't seem to work well for podspecs outside of project root and keep adding `.xcodeproj` even after adding `pod RNFirebase` to `Podfile` :(

chrisbianca:
@mars-lan We flagged this with the person that made the change in RN, but unfortunately they have not done anything about it to allow us to configure its usage:

https://github.com/facebook/react-native/pull/15460#issuecomment-346295149

PR #845, opened 2018-02-21
Improving type definitions to better match firebase-js-sdk

ajcrites:
*I will be happy to add tests. Unfortunately, I can't find any documentation for testing / testing types, and I'm not really sure what to add to test types. http://invertase.io/react-native-firebase/#/contributing/testing no longer seems to link to the right place.*

This is an update to the firebase exports to allow importing individual firebase modules. This allows you to do things like `import { database } from 'react-native-firebase';` instead of having to do `import firebase from 'react-native-firebase';` and then `firebase.database()`, etc.

Additionally, separate modules have been created for TypeScript importing to match the original firebase sdk. This allows you to do things like:

```
import { database } from 'react-native-firebase';
import { DataSnapshot } from 'react-native-firebase/database';

...

const ref = database().ref(refPath);
fromEventPattern<DataSnapshot>(handler => ref.on('value', handler));
```

Before, you would have to use `firebase.database` and import `RNFirebase` so you could use `RNFirebase.database.DataSnapshot`, etc.

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -16,46 +16,65 @@ declare module 'react-native-firebase' {
     nativeModuleExists: boolean;
   } & S;
 
+  // type AdmobModule = FirebaseModuleAndStatics<RNFirebase.admob.AdMob>;
+  type AnalyticsModule = FirebaseModuleAndStatics<RNFirebase.Analytics>;
+  type AuthModule = FirebaseModuleAndStatics<
+    RNFirebase.auth.Auth,
+    RNFirebase.auth.AuthStatics
+  >;
+  type ConfigModule = FirebaseModuleAndStatics<RNFirebase.config.Config>;
+  type CrashlyticsModule = FirebaseModuleAndStatics<
+    RNFirebase.crashlytics.Crashlytics
+  >;
+  type DatabaseModule = FirebaseModuleAndStatics<
+    RNFirebase.database.Database,
+    RNFirebase.database.DatabaseStatics
+  >;
+  type FirestoreModule = FirebaseModuleAndStatics<
+    RNFirebase.firestore.Firestore,
+    RNFirebase.firestore.FirestoreStatics
+  >;
+  type FunctionsModule = FirebaseModuleAndStatics<
+    RNFirebase.functions.Functions,
+    RNFirebase.functions.FunctionsStatics
+  >;
+  type IidModule = FirebaseModuleAndStatics<RNFirebase.iid.InstanceId>;
+  // type InvitesModule = FirebaseModuleAndStatics<RNFirebase.invites.Invites>;
+  type LinksModule = FirebaseModuleAndStatics<
+    RNFirebase.links.Links,
+    RNFirebase.links.LinksStatics
+  >;
+  type MessagingModule = FirebaseModuleAndStatics<
+    RNFirebase.messaging.Messaging,
+    RNFirebase.messaging.MessagingStatics
+  >;
+  type NotificationsModule = FirebaseModuleAndStatics<
+    RNFirebase.notifications.Notifications,
+    RNFirebase.notifications.NotificationsStatics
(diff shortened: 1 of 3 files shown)


chrisbianca:
@ajcrites Thanks for this, the TypeScript modules look great, although we're unsure about the changes you've made to the JS exports.  

Is it possible to do the following with the Firebase Web SDK?
```
import { database } from 'firebase';
```

It's not something any of us have come across and doesn't seem to be documented on the Firebase JS reference pages.

ajcrites:
Yes it is possible to do `import { database } from 'firebase';` and `import { DataSnapshot } from 'firebase/database';` which is what gave me the idea for this.

It works a bit differently because the Firebase SDK actually has these as modules in the file system and them namespaces them in the type definitions.

https://github.com/firebase/firebase-js-sdk/blob/master/packages/firebase/index.d.ts#L338
https://github.com/firebase/firebase-js-sdk/blob/master/packages/database/index.ts

However doing it this way would require a much bigger rewrite.

systemride:
Hello! I was hoping to add some type definitions for remote config although due to the changes in this pull request, I didn't want to create more conflicts. Is there any update on this?

ajcrites:
Updated to resolve conflicts with existing typing updates.

ajcrites:
Updated again to resolve conflicts. Anything else that needs to be done to get this merged?

ajcrites:
Updated to resolve conflicts and include typings for new config module.

lib/index.d.ts
```diff
@@ -2254,3 +2287,119 @@ declare module 'react-native-firebase' {
...
+  export type HttpsErrorCode = RNFirebase.functions.HttpsErrorCode;
+  export type FunctionsErrorCode = RNFirebase.functions.FunctionsErrorCode;
+  export type HttpsCallableResult = RNFirebase.functions.HttpsCallableResult;
+  export type Funtions = RNFirebase.functions.Functions;
+  export type HttpsError = RNFirebase.functions.HttpsError;
+}
+
+declare module 'react-native-firebase/firestore' {
+  import { RNFirebase } from 'react-native-firebase';
+  export type FireStore = RNFirebase.firestore.Firestore;
```
Salakar: Can this be `Firestore` lower case `s`? Or is there a conflict

lib/index.d.ts
```diff
@@ -2254,3 +2287,119 @@ declare module 'react-native-firebase' {
...
+  export type UploadTaskSnapshot = RNFirebase.storage.UploadTaskSnapshot;
+  export type FullMetadata = RNFirebase.storage.FullMetadata;
+  export type TaskEvent = RNFirebase.storage.TaskEvent;
+  export type TaskState = RNFirebase.storage.TaskState;
+}
+
+declare module 'react-native-firebase/database' {
+  import { RNFirebase } from 'react-native-firebase';
+  export type Database = RNFirebase.database.Database;
+  export type RnReference = RNFirebase.database.RnReference;
```
Salakar: Not sure why `RnReference` needs separate definitions from `Reference`, should whats in `RnReference` not be a part of `Reference`? 

Appreciate that it was already defined there before this PR so probably out of scope of this PR. Just wanted to get your thoughts

ajcrites:
@Salakar fixed conflicts and updated. I didn't update RnReference since it does still have properties that Reference doesn't have -- they can be kept as separate interfaces for now.

Note that there is an issue on the master branch where `core/firebase.js` is trying to import `../../version.js` which no longer exists. This may need to be addressed separately.

Salakar:
@ajcrites thanks for updating and commenting. 

Re the version file - it  is generated by the [build script](https://github.com/invertase/react-native-firebase/blob/master/package.json#L9) - so this is fine I believe - just needs to have `build` run.

Thanks again for this PR, will merge in prep for v5.

PR #868, opened 2018-03-09
[types] Use literal values for strings

skovhus:
In https://github.com/invertase/react-native-firebase/pull/832 I fixed some incorrect usage of not using the literal values (`string`) for Flow strings types, but object wrapper (`String`) inside the remote config code.

Turns out I clearly missed two cased. : )


chrisbianca:
Thanks for this

PR #881, opened 2018-03-14
Implement multiple database shard support

akshetpandey:


akshetpandey:
@chrisbianca  @Salakar : PR to support url along with app in database

android/src/main/java/io/invertase/firebase/database/RNFirebaseDatabaseReference.java
```diff
@@ -346,8 +348,13 @@ private void handleDatabaseError(ReadableMap registration, DatabaseError error)
    * @return
    */
   private void buildDatabaseQueryAtPathAndModifiers(String path, ReadableArray modifiers) {
-    FirebaseApp firebaseApp = FirebaseApp.getInstance(appName);
-    FirebaseDatabase firebaseDatabase = FirebaseDatabase.getInstance(firebaseApp);
+    FirebaseDatabase firebaseDatabase;
```
chrisbianca: It would make sense to increase the visibility of and make use of the `getDatabaseForApp` method on RNFirebaseDatabase.

lib/utils/apps.js
```diff
@@ -49,9 +49,19 @@ export default {
...
-      if (!APP_MODULES[app]) {
-        APP_MODULES[app] = {};
+    return (serviceUrl: ?string = null): M => {
+      if (serviceUrl && namespace !== 'database') {
+        throw new Error(
+          INTERNALS.STRINGS.ERROR_INIT_SERVICE_URL_UNSUPPORTED(namespace)
+        );
+      }
+
+      let appOrShardName = app.name;
```
chrisbianca: `const appOrShardName = serviceUrl || app.name`

lib/utils/ModuleBase.js
```diff
@@ -42,4 +44,16 @@ export default class ModuleBase {
...
+
+  get serviceUrl(): ?string {
+    return this._serviceUrl;
+  }
+
+  /**
+   * Returns the a key that uniquely identifies the native instance for this module
+   * @return {string}
+   */
+  get nativeModuleKey(): string {
```
chrisbianca: This was in `utils/native.js` to prevent internals from polluting the public API.

Unfortunately Flow doesn't support the concept of `@private` without having to write a separate definitions file.

This is something we might do in the future, if we have time, but for the time being it would be good if we can move this back into `utils/native.js` and also see if there's a way to hide the `serviceUrl` getter as well.

android/src/main/java/io/invertase/firebase/database/RNFirebaseDatabaseReference.java
akshetpandey: Since RNFirebaseDatabaseReference doesn't have an instance of RNFirebaseDatabase and getDatabaseForApp modifies and gets values from its instance variables, I can't make that method static or call it from RNFirebaseDatabaseReference. If its ok I can extract the FirebaseDatabase initialization logic from that method and create a new method that also handles logging and call that from inside RNFirebaseDatabase.

akshetpandey:
@chrisbianca : I have added the requested changes and also rebased the branch on latest master. The RNFirebaseDatabase changes are in its own commit in case you want to to modify it.

chrisbianca:
@akshetpandey thanks for making those changes. I'm going to have a chat with @Salakar to see whether we might make the logging level stuff static on RNFirebaseDatabase.  This means we can consistently ensure that the logging level is set throughout the app and remove the need for the split methods.  If so, I'll update it once we've merged.

Re: @Chenjh1992 's comment about the key is a valid point which I'd overlooked. I think we can probably resolve this in the JS and ensure that there is either never a trailing `/` or always a trailing `/`

chrisbianca:
@akshetpandey Thanks again for this.

I've just merged and made a couple of minor tweaks:
- made logging level static so it can be shared between RNFirebaseDatabase and RNFirebaseDatabaseReference
- enforced that the supplied URL must have a trailing `/`

We'll get this released as part of the next 4.0.0 alpha

PR #882, opened 2018-03-14
Fix flow to support setUserId(null)

leethree:
`null` is accepted for `setUserId`, see [index.d.ts](https://github.com/invertase/react-native-firebase/blob/8060c949a2881cada885849c63c9a4ff93ec5d8c/lib/index.d.ts#L524)


chrisbianca:
@leethree thanks for this, I've included as part of https://github.com/invertase/react-native-firebase/commit/e3ea0d3fc7660a5cea7f8d3e048476468573227b

PR #938, opened 2018-04-04
fixes races bug in react-native > 0.52

taxidermic:
In newer versions of react-native `dispatch_async` in `setCurrentScreen` causes mixed order. It's obvious if you make several calls of `setCurrentScreen` and `logEvent` in a small amount of time.

See: https://github.com/facebook/react-native/blob/0.54-stable/React/Base/RCTBridgeModule.h#L112

### DIFF
--- a/ios/RNFirebase/analytics/RNFirebaseAnalytics.m
+++ b/ios/RNFirebase/analytics/RNFirebaseAnalytics.m
@@ -5,6 +5,7 @@
 #import <FirebaseAnalytics/FIRAnalyticsConfiguration.h>
 
 @implementation RNFirebaseAnalytics
+@synthesize methodQueue = _methodQueue;
 RCT_EXPORT_MODULE();
 
 RCT_EXPORT_METHOD(logEvent:(NSString *)name props:(NSDictionary *)props) {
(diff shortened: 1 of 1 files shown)


ios/RNFirebase/analytics/RNFirebaseAnalytics.m
```diff
@@ -5,6 +5,7 @@
 #import <FirebaseAnalytics/FIRAnalyticsConfiguration.h>
 
 @implementation RNFirebaseAnalytics
+@synthesize methodQueue = _methodQueue;
```
chrisbianca: @taxidermic I'm unsure why we need this queue, as the dispatch method is using `dispatch_get_main_queue()`, not the `methodQueue` you've defined?

taxidermic: Tested, yes, it works without `@synthesize methodQueue = _methodQueue;`!
Will you edit it by yourself, or should I send new pull_request? (I'm not sure how it works here ^^")

Btw, it fixes races **only inside these structures separately**. So, it's still possible, that 
```
firebase.SetCurrentScreen(someProps)
firebase.LogEvent(anotherProps)
```
may mess up.

Though, several calls `SetCurrentScreen` will save their order

taxidermic: I mean:
```
firebase.SetCurrentScreen(A)
firebase.LogEvent(E1)
firebase.SetCurrentScreen(B)
firebase.LogEvent(E2)
```
 E1.screen == A и E2.screen == B, otherwords results will be accurate, though in Firebase Console you may see:
```
E1
screen_view(A)
screen_view(B)
E2
```

chrisbianca: Thanks @taxidermic , this is a good step in the right direction. We'll consider how best to synchronize across all the methods.

PR #970, opened 2018-04-11
Fix for attaching file to image with notification android

judemanutd:
Supplying a string that is a file path for adding images to notifications in android like so `file:///data/user/0/com.example/files/RNFetchBlobTmp_cjyxt7xo568b470ekawvzw.jpg` can be used to recognise that it is a file location that has been submitted but when actually using the file to create a bitmap the `file://` part might need to be removed as it seems to prevent native android from actually finding the file.

Might not be the best way for replacing the `file://` pattern but should be fine.


chrisbianca:
Thanks for this @judemanutd - have you confirmed that it works without the `file://` prefix?

chrisbianca:
Perfect! I have adapted this slightly here: https://github.com/invertase/react-native-firebase/commit/62706f9629c1d61f651e1ba9359e8ad7bee435f6

Thanks for tracking this one down!  It will land as part of v4.0.2

PR #985, opened 2018-04-16
allow using ringtones from RingtoneManager

giladno:
when allowing users to pick a sound from RingtoneManager, RNFirebaseNotification should detect we are passing a Uri instead of trying to load this file from our bundle.


chrisbianca:
Thanks for this - good catch!

PR #1007, opened 2018-04-21
android: allow root project to specify dependency versions

mlc:
React Native projects with multiple native dependencies often end up unintentionally relying on a variety of Android SDK versions, which can lead to unpredictable build-time failures or even run-time crashes. Many libraries (see, e.g., react-community/react-native-maps#2047) have been moving to a system of allowing the solution [recommended by google](https://developer.android.com/studio/build/gradle-tips.html#configure-project-wide-properties) of having the root project set properties with the desired dependency versions. This PR implements that solution for `react-native-firebase`.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -1,5 +1,4 @@
 buildscript {
-  ext.firebaseVersion = '12.0.0'
   repositories {
     jcenter()
     google()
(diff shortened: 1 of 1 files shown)


Salakar:
@mlc thanks for this, makes sense! Will merge and run on ./bridge tests project to confirm - but I don't see anything breaking here as fallback behaviour has been handled 👌

mlc:
thanks, @Salakar!

Salakar:
image

All good, thanks again for this.

PR #1010, opened 2018-04-22
Allow passing null as a sound

giladno:
As per this answer: https://stackoverflow.com/a/46634508/306764


Salakar:
@giladno thanks for this

Issue #1016, opened 2018-04-24
'FirebaseAnalytics/FirebaseAnalytics.h' file not found


andr3a88:
### Issue: Cannot build the project

After updating `RNFirebase´and `Firebase` iOS SDK cannot build the project:

+ `'Firebase.h' file not found with <angled> include; use "quotes" instead` in RNFirebaseUtil.h
+ `'FirebaseAnalytics/FirebaseAnalytics.h' file not found` in Firebase.h
+ `In file included from .../node_modules/react-native-firebase/ios/RNFirebase/admob/RNFirebaseAdMobRewardedVideo.m:2:`

### Environment

iOS 11.3
macOS High Sierra
Xcode 9.3 
ReactNative 0.55.1
RNFirebase Version: 4.0.5

Firebase version: 
Firebase 4.13.0
FirebaseAnalytics
FirebaseCore 4.0.20

jeanregisser:
Do you use CocoaPods for the Firebase and RNFirebase?

AndreiCalazans:
Environment
OS: macOS High Sierra 10.13.4
Node: 9.4.0
Yarn: 1.3.2
npm: 5.6.0
Watchman: 4.7.0
Xcode: Xcode 9.2 Build version 9C40b
Android Studio: 2.3 AI-162.4069837

Packages: (wanted => installed)
react: 16.3.1 => 16.3.1
react-native: 0.55.1 => 0.55.1



I'm having the same issue. I was able to get passed the Firebase.h by adding Firebase.h into `project_root/ios/Firebase/Firebase.h` .

but I'm still getting the second error. 

In `Firebase.h`
``` 
'FirebaseAnalytics/FirebaseAnalytics.h' file not found
```


I installed it manually, I'm not using CocoaPods.

### Steps to reproduce:

1. yarn add react-native-firebase
2. react-native link react-native-firebase
3. Add GoogleService-Info.plist to /ios/appName/
4. Drag Firebase framework files to project root inside xcode as the README of Firebase specified
5. Add the -ObjC flag to "Other Linker Settings":

chrisbianca:
Please take a look at option 2 of https://rnfirebase.io/docs/v4.0.x/installation/ios#1.3.-Install-Firebase-Library

This explains that if you're not using cocoapods, you need to add the frameworks to an `ios/Firebase` folder for them to be picked up by RNFirebase.

AndreiCalazans:
[...]

PR #1017, opened 2018-04-24
Fix podspec so compilation works when RNFirebase is set in the Podfile

jeanregisser:
Not sure when this broke, but this fixes compilation when used in a
project that pulls the dependency using CocoaPods and Xcode 9.3

Otherwise I was getting the following error:

`'Firebase.h' file not found with <angled> include; use "quotes" instead in RNFirebaseUtil.h`

This is the content of the `Podfile`

```ruby
pod 'Firebase/Core'
pod 'Firebase/Auth'
pod 'Firebase/Firestore'
pod 'RNFirebase', :path => '../node_modules/react-native-firebase/ios'
```

### DIFF
--- a/ios/RNFirebase.podspec
+++ b/ios/RNFirebase.podspec
@@ -16,4 +16,5 @@ Pod::Spec.new do |s|
   s.platform            = :ios, "9.0"
   s.source_files        = 'RNFirebase/**/*.{h,m}'
   s.dependency          'React'
+  s.dependency          'Firebase/Core'
 end


chrisbianca:
That's really strange, I'd have thought the fact you had the Firebase/Core dependency already in your podfile would mean that this wasn't necessary.

Out of interest, what version of cocoapods are you running?

jeanregisser:
Thought the same as you. I'm using CocoaPods 1.5.0

What I suspect is CocoaPods only exposes the headers of the direct dependencies declared for each Pod.
Instead of each Pod having access to all the other headers.

jeanregisser:
But I might be wrong, otherwise how would RNFirebase pod be able to find the headers for Firebase/Auth etc.

chrisbianca:
Yeah, exactly what I was wondering! I'll do some digging over the next couple of days to see what I can work out...
I'd prefer not to merge until I understand what's going on.

wbattel4607:
@chrisbianca we've only been able to get the project to build when using this fix. If we don't include the Firebase/Core dependency in the RNFirebase pod spec then we get 

screen shot 2018-04-27 at 7 16 57 pm

chrisbianca:
I'm going to merge this in - I can't see any immediate side effects, and whilst I don't quite understand what would change this behaviour all of a sudden (other than perhaps something in Cocoapods 1.5.0), it seems like it is helping people.

Air-Craft:
I had the same error on v5.6 (after this merge)...  This solved it: https://github.com/invertase/react-native-firebase/issues/1016#issuecomment-549314343

mikehardy:
There are no reported problems anything like that with current (v10.5.x at this point)

PR #1058, opened 2018-05-07
[Android] Add the ability to remove notifications based on the tag

dgruseck:
It would be awesome to be able to remove notification using the tag.


lib/modules/notifications/index.js
```diff
@@ -291,6 +291,21 @@ export default class Notifications extends ModuleBase {
     return getNativeModule(this).removeDeliveredNotification(notificationId);
   }
 
+  /**
+   * Remove a delivered notifications by tag.
+   * @param tag
+   */
+  removeDeliveredNotificationsByTag(tag: string): Promise<void> {
```
chrisbianca: Are you able to move this into `AndroidNotifications.js` as it's Android specific functionality?

You'll see it that file, that the JS does a platform check when a method is called and resolves an empty promise if it's not Android.  If we follow that pattern then we don't need to worry about the iOS implementation that you've added.

android/src/main/java/io/invertase/firebase/notifications/RNFirebaseNotifications.java
```diff
@@ -284,6 +289,7 @@ private WritableMap parseRemoteMessage(RemoteMessage message) {
     }
     if (notification.getTag() != null) {
       androidMap.putString("group", notification.getTag());
```
chrisbianca: Now that we have the `tag` property, I don't think we should also set that as the `group` as they are two separate things.  Are you able to remove the `group`?

jamesisaac:
This would be great, or even just the part which adds the ability to call setTag when creating a local notification (seems like quite an inconsistency between local/remote notifications).

Salakar:
@dgruseck any updates on the requested changes?

android/src/main/java/io/invertase/firebase/notifications/RNFirebaseNotifications.java
dgruseck: Removing the group could be a breaking change for some people, because of that I'm not sure if we should do that.

Salakar:
@dgruseck thanks for updating this, looks fine to me now.

Will merge but could you update the docs to add these on please;

https://rnfirebase.io/docs/master/notifications/reference/AndroidNotification
\/\\----\\/
https://github.com/invertase/react-native-firebase-docs/blob/master/docs/notifications/reference/AndroidNotification.md



---

PR #1073, opened 2018-05-11
Fix ClassCastException while parsing fireDate

dluksza:
ClassCastException was thrown on Android 8.1 while trying to schedule
local notification. Use try-catch approach instead of relaying on the
default parse value.


chrisbianca:
This looks sensible, thanks!

PR #1074, opened 2018-05-11
Implement handling of Android actions in background

dluksza:
There are some cases when local notification action should be handled in
background eg. snoozing the reminder. In case of it launching app UI is
not necessary and would be confusing for the end user.

Therefore there should be a way to handle local notification action in
background.

For this reason new property 'runInBackground' was added to the
AndroidAction class and TypeScript type.

Also new broadcast receiver and service were implemented to handle
properly background actions.

In order to run particular action in background user need to set its
'runInBackground' property to 'true', eg:

```js
  ...
  .android.addAction(new firebase.notifications.Android.Action("snooze",
      "ic_snooze", "Snooze").setRunInBackground(true))
```

Then create a background asynchronous handler:

```js
  const handleAsyncTask = async (notificationOpen: NotifficationOpen) => {
    if (notificationOpen && notificationOpen.notification) {
      const action = notificationOpen.action;
      const notificationId = notificationOpen.notification.notificationId;
      if (action === "snooze") {
        console.log("Reschedule notification for later time", notificationId);
      } else {
        console.log("unsupported action", action);
      }
      // hide the notification
      firebase.notifications().removeDeliveredNotification(notificationId);
    }
  }
```

Next hander should be registered to an headless handler:

```js
  AppRegistry.registerHeadlessTask('RNFirebaseBackgroundNotification', () => handleAsyncTask);
```

Finally AndroidManifest.xml file must be modified, to include receiver
and service definition:

```xml
  <receiver
      android:name="io.invertase.firebase.notifications.RNFirebaseBackgroundNotificationReceiver"
      android:exported="true">
    <intent-filter>
      <action android:name="io.invertase.firebase.notifications.BackgroundAction"/>
    </intent-filter>
  </receiver>
  <service android:name="io.invertase.firebase.notifications.RNFirebaseBackgroundNotificationsService"/>
```

Now when ever 'Snooze' action is pressed it will launch
'handleAsyncTask' function in the background and reschedule the
notification for the later time.

### DIFF
--- a/android/src/main/java/io/invertase/firebase/notifications/RNFirebaseBackgroundNotificationActionReceiver.java
+++ b/android/src/main/java/io/invertase/firebase/notifications/RNFirebaseBackgroundNotificationActionReceiver.java
@@ -1,0 +1,50 @@
+package io.invertase.firebase.notifications;
+
+import android.content.BroadcastReceiver;
+import android.content.Context;
+import android.content.Intent;
+import android.os.Bundle;
+
+import com.facebook.react.HeadlessJsTaskService;
+import com.facebook.react.ReactApplication;
+import com.facebook.react.bridge.Arguments;
+import com.facebook.react.bridge.ReactContext;
+import com.facebook.react.bridge.WritableMap;
+
+import io.invertase.firebase.Utils;
+
+public class RNFirebaseBackgroundNotificationActionReceiver extends BroadcastReceiver {
+  static boolean isBackgroundNotficationIntent(Intent intent) {
+    return intent.getExtras() != null && intent.hasExtra("action") && intent.hasExtra("notification");
+  }
+
+  static WritableMap toNotificationOpenMap(Intent intent) {
+    Bundle extras = intent.getExtras();
+    WritableMap notificationMap = Arguments.makeNativeMap(extras.getBundle("notification"));
+    WritableMap notificationOpenMap = Arguments.createMap();
+    notificationOpenMap.putString("action", extras.getString("action"));
+    notificationOpenMap.putMap("notification", notificationMap);
+    return notificationOpenMap;
+  }
+
+  @Override
+  public void onReceive(Context context, Intent intent) {
+    if (!isBackgroundNotficationIntent(intent)) {
+      return;
+    }
+
+    if (Utils.isAppInForeground(context)) {
+      WritableMap notificationOpenMap = toNotificationOpenMap(intent);
+
+      ReactApplication reactApplication =  (ReactApplication)context.getApplicationContext();
+      ReactContext reactContext = reactApplication.getReactNativeHost().getReactInstanceManager().getCurrentReactContext();
+
+      Utils.sendEvent(reactContext, "notifications_notification_opened", notificationOpenMap);
+    } else {
+      Intent serviceIntent = new Intent(context, RNFirebaseBackgroundNotificationActionsService.class);
+      serviceIntent.putExtras(intent.getExtras());
+      context.startService(serviceIntent);
+      HeadlessJsTaskService.acquireWakeLockNow(context);
+    }
+  }
+}
--- a/android/src/main/java/io/invertase/firebase/notifications/DisplayNotificationTask.java
+++ b/android/src/main/java/io/invertase/firebase/notifications/DisplayNotificationTask.java
@@ -294,9 +294,11 @@ public class DisplayNotificationTask extends AsyncTask<Void, Void, Void> {
   }
 
   private NotificationCompat.Action createAction(Bundle action, Class intentClass, Bundle notification) {
+    boolean runInBackground = action.containsKey("runInBackground") && action.getBoolean("runInBackground");
     String actionKey = action.getString("action");
-    PendingIntent actionIntent = createIntent(intentClass, notification, actionKey);
-
+    PendingIntent actionIntent = runInBackground ?
+      createBroadcastIntent(notification, actionKey) :
+      createIntent(intentClass, notification, actionKey);
     int icon = getIcon(action.getString("icon"));
     String title = action.getString("title");
 
@@ -334,8 +336,19 @@ public class DisplayNotificationTask extends AsyncTask<Void, Void, Void> {
     }
 
     String notificationId = notification.getString("notificationId");
-
     return PendingIntent.getActivity(context, notificationId.hashCode(), intent, PendingIntent.FLAG_UPDATE_CURRENT);
+  }
+
+  private PendingIntent createBroadcastIntent(Bundle notification, String action) {
+    Intent intent = new Intent(context, RNFirebaseBackgroundNotificationActionReceiver.class);
+    intent.addFlags(Intent.FLAG_ACTIVITY_SINGLE_TOP);
+
+    String notificationId = notification.getString("notificationId") + action;
+
+    intent.setAction("io.invertase.firebase.notifications.BackgroundAction");
+    intent.putExtra("action", action);
+    intent.putExtra("notification", notification);
+    return PendingIntent.getBroadcast(context, notificationId.hashCode(), intent, PendingIntent.FLAG_UPDATE_CURRENT);
   }
 
   private RemoteInput createRemoteInput(Bundle remoteInput) {
--- a/android/src/main/java/io/invertase/firebase/notifications/RNFirebaseBackgroundNotificationActionsService.java
+++ b/android/src/main/java/io/invertase/firebase/notifications/RNFirebaseBackgroundNotificationActionsService.java
@@ -1,0 +1,29 @@
+package io.invertase.firebase.notifications;
+
+import android.content.Intent;
+
+import com.facebook.react.HeadlessJsTaskService;
+import com.facebook.react.bridge.WritableMap;
+import com.facebook.react.jstasks.HeadlessJsTaskConfig;
+
+import javax.annotation.Nullable;
+
+import static io.invertase.firebase.notifications.RNFirebaseBackgroundNotificationActionReceiver.isBackgroundNotficationIntent;
+import static io.invertase.firebase.notifications.RNFirebaseBackgroundNotificationActionReceiver.toNotificationOpenMap;
+
+public class RNFirebaseBackgroundNotificationActionsService extends HeadlessJsTaskService {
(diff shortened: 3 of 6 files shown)


android/src/main/java/io/invertase/firebase/notifications/RNFirebaseBackgroundNotificationReceiver.java
```diff
@@ -0,0 +1,47 @@
...
+    Intent serviceIntent = new Intent(context, RNFirebaseBackgroundNotificationsService.class);
+    serviceIntent.putExtras(intent.getExtras());
+    context.startService(serviceIntent);
+
+    if (!isAppOnForeground((context))) {
+      HeadlessJsTaskService.acquireWakeLockNow(context);
+    }
+  }
+
+  private boolean isAppOnForeground(Context context) {
```
chrisbianca: Can you change this to use the method from the `io.invertase.firebase.Utils` class rather than duplicating?

android/src/main/java/io/invertase/firebase/notifications/RNFirebaseBackgroundNotificationReceiver.java
```diff
@@ -0,0 +1,47 @@
...
+import android.content.BroadcastReceiver;
+import android.content.Context;
+import android.content.Intent;
+import android.util.Log;
+
+import com.facebook.react.HeadlessJsTaskService;
+
+import java.util.List;
+
+public class RNFirebaseBackgroundNotificationReceiver extends BroadcastReceiver {
```
chrisbianca: Can we rename this to `RNFirebaseBackgroundNotificationActionReceiver` to make it clear that it handles actions and not notifications themselves

android/src/main/java/io/invertase/firebase/notifications/RNFirebaseBackgroundNotificationReceiver.java
```diff
@@ -0,0 +1,47 @@
...
+
+  @Override
+  public void onReceive(Context context, Intent intent) {
+    Intent serviceIntent = new Intent(context, RNFirebaseBackgroundNotificationsService.class);
+    serviceIntent.putExtras(intent.getExtras());
+    context.startService(serviceIntent);
+
+    if (!isAppOnForeground((context))) {
```
chrisbianca: What happens if the app is in the foreground when the `Snooze` action is hit?
I _think_ it will currently just get lost - instead we need to make sure that the action gets passed through into the standard `onNotificationOpened` handler so that it can be handled normally.

android/src/main/java/io/invertase/firebase/notifications/RNFirebaseBackgroundNotificationsService.java
```diff
@@ -0,0 +1,32 @@
...
+import com.facebook.react.bridge.Arguments;
+import com.facebook.react.bridge.WritableMap;
+import com.facebook.react.jstasks.HeadlessJsTaskConfig;
+import com.google.firebase.messaging.RemoteMessage;
+
+import javax.annotation.Nullable;
+
+import io.invertase.firebase.messaging.MessagingSerializer;
+
+public class RNFirebaseBackgroundNotificationsService extends HeadlessJsTaskService {
```
chrisbianca: As above, can this be renamed to `RNFirebaseBackgroundNotificationActionsService`?

android/src/main/java/io/invertase/firebase/notifications/RNFirebaseBackgroundNotificationsService.java
```diff
@@ -0,0 +1,32 @@
...
+  @Override
+  protected @Nullable HeadlessJsTaskConfig getTaskConfig(Intent intent) {
+    Bundle extras = intent.getExtras();
+    if (extras != null) {
+      WritableMap notificationMap = Arguments.makeNativeMap(intent.getExtras());
+      return new HeadlessJsTaskConfig(
+        "RNFirebaseBackgroundNotification",
```
chrisbianca: Can we change this to `RNFirebaseBackgroundNotificationAction`?

android/src/main/java/io/invertase/firebase/notifications/RNFirebaseBackgroundNotificationsService.java
```diff
@@ -0,0 +1,32 @@
...
+import javax.annotation.Nullable;
+
+import io.invertase.firebase.messaging.MessagingSerializer;
+
+public class RNFirebaseBackgroundNotificationsService extends HeadlessJsTaskService {
+  @Override
+  protected @Nullable HeadlessJsTaskConfig getTaskConfig(Intent intent) {
+    Bundle extras = intent.getExtras();
+    if (extras != null) {
+      WritableMap notificationMap = Arguments.makeNativeMap(intent.getExtras());
```
chrisbianca: We need to make sure that the object being passed to the JS side conforms to the `NotificationOpen` structure.  At the moment it doesn't allow for a `snooze` action with a remote input specified.

Take a look at the `parseIntentForLocalNotification` method in `RNFirebaseNotifications` to see how this is done.

android/src/main/java/io/invertase/firebase/notifications/RNFirebaseBackgroundNotificationReceiver.java
dluksza: Done

android/src/main/java/io/invertase/firebase/notifications/RNFirebaseBackgroundNotificationReceiver.java
dluksza: Done

android/src/main/java/io/invertase/firebase/notifications/RNFirebaseBackgroundNotificationsService.java
dluksza: Done

android/src/main/java/io/invertase/firebase/notifications/RNFirebaseBackgroundNotificationsService.java
dluksza: Done

android/src/main/java/io/invertase/firebase/notifications/RNFirebaseBackgroundNotificationReceiver.java
dluksza: I need to test this. I'm not sure if going through standard `onNotificationOpened` way is the best solution, since this would require code _duplication_ (the same handler function should be called from two places). Would rather handle everything in background.

There is a small drawback for the _always background_ approach. When background action handler is changing the UI state this should be achieved using some kind of event or listener approach.

android/src/main/java/io/invertase/firebase/notifications/RNFirebaseBackgroundNotificationsService.java
dluksza: Done

android/src/main/java/io/invertase/firebase/notifications/RNFirebaseBackgroundNotificationReceiver.java
chrisbianca: I believe it would be better to handle things separately in the foreground and background as it gives more flexibility to the developer if they're like their app to act differently.  You've already touched on UI state which might need to be updated in the foreground.

The functionality that's shared can easily be wrapped inside a method which can be called from both the background and foreground handler to prevent code duplication.

android/src/main/java/io/invertase/firebase/notifications/RNFirebaseBackgroundNotificationReceiver.java
chrisbianca: 👍

android/src/main/java/io/invertase/firebase/notifications/RNFirebaseBackgroundNotificationsService.java
chrisbianca: 👍

dluksza:
Here is the updated version of previous PR. I've addressed all the comments, updated the commit message and rebased it on current master.

android/src/main/java/io/invertase/firebase/notifications/RNFirebaseBackgroundNotificationActionReceiver.java
```diff
@@ -0,0 +1,50 @@
...
+    return notificationOpenMap;
+  }
+
+  @Override
+  public void onReceive(Context context, Intent intent) {
+    if (!isBackgroundNotficationIntent(intent)) {
+      return;
+    }
+
+    if (Utils.isAppInForeground(context)) {
```
chrisbianca: Have you been able to test that this works in the foreground?

Generally the broadcast receiver runs outside the react context, so I'd be surprised if this did work as it is currently.  The way we work around this in other areas is to use a local broadcast receiver which the react application listens to directly.  I'm more than happy to merge and then make these changes myself, but wanted to understand if you'd managed to get it working this way before I do?

dluksza: I've just retested it with [HabitChallenge](http://HabitChallenge.co/) app in three scenarios:
1. app in foreground
1. app in background
1. app not running

In all of them everything worked as expected.

dluksza: BTW. Test were done with `react-native` 0.55.4 and android simulator running Android 8.1.0

chrisbianca: Great! In which case, I will get this merged and we can keep an eye on it.
If we get reports of issues in the foreground then I can easily switch it over to use the local broadcast.

Thank you again for such a comprehensive PR!  I'll look to get some documentation added and we'll include this as part of a release today or tomorrow.

dluksza: Cool! Looking forward for the release! :D

PR #1080, opened 2018-05-13
Add Android API to delete channel and channel group

dluksza:


lib/modules/notifications/AndroidNotifications.js
```diff
@@ -91,4 +91,32 @@ export default class AndroidNotifications {
     }
     return Promise.resolve();
   }
+
+  deleteChannelGroup(groupId: string): Promise<void> {
+    if (Platform.OS === 'android') {
+      if (!(groupId instanceof string)) {
```
Salakar: Shouldn't this be `if (typeof groupId !== 'string') {`, instanceof string would throw `string undefined`?

lib/modules/notifications/AndroidNotifications.js
```diff
@@ -91,4 +91,32 @@ export default class AndroidNotifications {
...
+      return getNativeModule(this._notifications).deleteChannelGroup(
+        groupId
+      );
+    }
+    return Promise.resolve();
+  }
+
+  deleteChannel(channelId: string): Promise<void> {
+    if (Platform.OS === 'android') {
+      if (!(channnelId instanceof string)) {
```
Salakar: Shouldn't this be `if (typeof channnelId !== 'string') {`, instanceof string would throw `string undefined`?

chrisbianca:
@Salakar Other than the changes you've suggested, it looks good to me.  
Thanks @dluksza !!

dluksza:
Updated the PR, applying suggestions from @Salakar.

BTW. what about my other PR https://github.com/invertase/react-native-firebase/pull/1074, will you find some time to look into it?

chrisbianca:
Thanks for making those changes.  I am planning on looking through your other PR later today - it requires a bit more time :)

PR #1085, opened 2018-05-14
Firestore: Allow setting the iOS dispatch queue for callbacks.

bzztbomb:
Use case is preventing processing in the main thread and UI hitches.

For larger documents the time spent converting from the document dictionary to a folly object to be passed into react native can take significant time and cause hitches in UI rendering.  This allows the developer to specify a dispatch queue to handle events in. This seemed like the lowest-impact way to allow this.

I'd argue that by default the firestore dispatch queue should specified to be something other than the main thread.  Since after the work is done it goes to the react native js thread, there's no need to clutter the main thread with this kind of work.  But I'm still pretty new to firebase/firestore, so I may be missing something. 

Thanks!

### DIFF
--- a/ios/RNFirebase/firestore/RNFirebaseFirestore.m
+++ b/ios/RNFirebase/firestore/RNFirebaseFirestore.m
@@ -357,6 +357,9 @@ RCT_EXPORT_METHOD(settings:(NSString *)appDisplayName
         // TODO: Enable when available on Android
         // firestoreSettings.timestampsInSnapshotsEnabled = settings[@"timestampsInSnapshots"];
     }
+    if (settings[@"dispatchQueueName"]) {
+        firestoreSettings.dispatchQueue = dispatch_queue_create([settings[@"dispatchQueueName"] UTF8String], NULL);
+    }
     [firestore setSettings:firestoreSettings];
     resolve(nil);
 }


chrisbianca:
@bzztbomb Thanks for flagging this. You're right, we should be creating a separate dispatch queue for firestore by default rather than needing to specify it as you've done here.  The same is also true of the database module which also does some heavy lifting.

I've updated this here which will come through as part of the next release: https://github.com/invertase/react-native-firebase/commit/fa9bac1a729ef5c424786b939dc284120db34ce2

chrisbianca:
You are very right, sorry, I completely missed that this was a Firestore setting rather than a React Native setting.

I think it would be wise to make this a default configuration for `react-native-firebase` rather than rely on the user to have to set it due to the possible performance implications with the bridge.

I'm just trying to work through this myself - it's a little fiddly as we need this to work across multiple apps.  I'll update you with my progress.

chrisbianca:
Ok, this does what I've described: https://github.com/invertase/react-native-firebase/commit/35474462c61c6a303ec0272193e53dc0406bf0da

It will be included in the next release

bzztbomb:
Chris:

Awesome!  Thank you very much.  I didn't know about running multiple
firebase apps within one process.  heh, always some subtleties you miss
when working with new libraries.

Thanks again for being so responsive!

Brian

chrisbianca:
No problem - thanks for highlighting in the first instance, it's something we'd completely missed.  It should provide much better performance going forward, and we need to look at adding the same to Database and Storage which also have support for a custom callback queue.

PR #1112, opened 2018-05-21
react native firebase with onesignal

adirzoari:


anhtuank7c:
I don't know your's problem but i run **OneSignal** beside **React Native Firebase** without issue.

chrisbianca:
This was highlighted as an error within onesignal itself which should now be resolved.

PR #1113, opened 2018-05-21
Resolve display notification promise

judemanutd:
Once the notification is created the promise was not being resolved so in a situation where the user is waiting on the promise to complete, they would not get a callback and in case a user used async/await this would cause the app to get stuck at that point.


Salakar:
@judemanutd thanks for this, this is now live in the [v4.3.0 release](https://github.com/invertase/react-native-firebase/releases/tag/v4.3.0).

---

PR #1151, opened 2018-05-28
Update build.gradle

alexanderkjeldaas:
Don't set global policy on which gradle versions to support.

`google()` is only supported on newer gradle versions.
An `allprojects{..}` setting shouldn't exist in a component should it?

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -39,14 +39,6 @@ android {
 }
 
 
-allprojects {
-  repositories {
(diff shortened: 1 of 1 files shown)


chrisbianca:
Thanks for flagging this - I incorporated it as part of: https://github.com/invertase/react-native-firebase/commit/dae0645006bc732480731731909547b86448905a

PR #1153, opened 2018-05-29
Fix repository errors

kaladivo:
It seems that jcenter has some issues with providing dependencies to the project. Adding maven repositories should fix it. 

See this: https://stackoverflow.com/questions/50563338/could-not-find-runtime-jar-android-arch-lifecycleruntime1-0-0


chrisbianca:
The line you've added isn't necessary as it is equivalent to `google()`

PR #1164, opened 2018-05-31
[Android] Support for areNotificationsEnabled

geriux:
## Use case

In our app we'd like to know if the users have "blocked" the notifications so we can show them a snackbar with a button to the app's system settings so they can turn them on.

On **iOS** the `hasPermission` method works perfectly, returning true or false depending if they're on or not.

On **Android** it was always returning true, so I made this PR including [areNotificationsEnabled](https://developer.android.com/reference/android/support/v4/app/NotificationManagerCompat#areNotificationsEnabled()) in the same `hasPermission` method which will return true or false depending if the user has blocked them or not.

### DIFF
--- a/android/src/main/java/io/invertase/firebase/messaging/RNFirebaseMessaging.java
+++ b/android/src/main/java/io/invertase/firebase/messaging/RNFirebaseMessaging.java
@@ -5,6 +5,7 @@ import android.content.Context;
 import android.content.Intent;
 import android.content.IntentFilter;
 import android.support.v4.content.LocalBroadcastManager;
+import android.support.v4.app.NotificationManagerCompat;
 import android.util.Log;
 
 import com.facebook.react.bridge.Promise;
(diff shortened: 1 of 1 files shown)


alexsegura:
+1

My face when I discovered `hasPermission` always resolves to true → 😐

There is no way to do it with [PermissionsAndroid](https://facebook.github.io/react-native/docs/permissionsandroid.html)
The only up-to-date library I found out there is [react-native-check-notification-enable](https://github.com/jigaryadav/react-native-check-notification-enable)
This is definitely needed. 
Actually, the behavior is quite strange: I tried to check if this can be implemented using `getToken`, but `getToken` still returns something even after disabling notifications. 

Can we do something for `requestPermission` as well?

chrisbianca:
Thanks for adding this!

Salakar:
@geriux thanks for this, this is now live in the [v4.3.0 release](https://github.com/invertase/react-native-firebase/releases/tag/v4.3.0).

---

PR #1215, opened 2018-06-15
adds support for getToken(string,string) and deleteToken()

ctaintor:
I've added support for `deleteToken()` as well as the other form of `getToken()`. This allows for someone to request or delete FCM tokens for sender IDs other than the default senderid. I also added some tests for instanceId methods - they don't do much but at least will break if someone broke the native implementations.

* [deleteToken](https://firebase.google.com/docs/reference/android/com/google/firebase/iid/FirebaseInstanceId.html#deleteToken(java.lang.String,%20java.lang.String))
* [getToken](https://firebase.google.com/docs/reference/android/com/google/firebase/iid/FirebaseInstanceId.html#getToken(java.lang.String,%20java.lang.String))

### DIFF
--- a/android/src/main/java/io/invertase/firebase/instanceid/RNFirebaseInstanceId.java
+++ b/android/src/main/java/io/invertase/firebase/instanceid/RNFirebaseInstanceId.java
@@ -1,6 +1,8 @@
 package io.invertase.firebase.instanceid;
 
 
+import java.io.IOException;
+
 import android.util.Log;
 
 import com.facebook.react.bridge.Promise;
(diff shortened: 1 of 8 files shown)


ctaintor:
This is what I originally did, but the `getToken()` was in messaging, hence why I added them to messaging and not iid.

chrisbianca:
Yes, there is unfortunately a bit of overlap, but for the majority of use cases, the simple `getToken` is all that developers will need. For these advanced methods, it makes sense to keep them separate in the instanceId module rather than complicate the messaging module further.

ctaintor:
Makes sense. Just to make sure, this will mean that `getToken()` will exist in two places. Cool?

chrisbianca:
Yep, the messaging module will have `getToken()` with no parameters, and instanceId will have the more complicated `getToken()` implementation. Thank you!

ctaintor:
OK - would you prefer the iid version to also support no parameters? Or have it only be the parameter version.

chrisbianca:
Let's go with only the parameter version.

ctaintor:
OK - I'm having trouble getting anything to build right now, but this should work, heh. Cocoapods...

ctaintor:
Have the tests been broken in master? I'm currently unable to install cocoapods:
```
[!] CocoaPods could not find compatible versions for pod "RNFirebase":
  In snapshot (Podfile.lock):
    RNFirebase (= 4.1.0)

  In Podfile:
    RNFirebase (from `../../ios/RNFirebase.podspec`)

It seems like you've changed the constraints of dependency `RNFirebase` inside your development pod `RNFirebase`.
You should run `pod update RNFirebase` to apply changes you've made.
```
(tried with a fresh clone of master and following steps [here](https://rnfirebase.io/docs/master/testing))

ctaintor:
OK - should work now.

chrisbianca:
This looks awesome, thank you for moving it over.

Just one last thing before I can merge it in - are you able to add the Typescript typings to `lib/index.d.ts`?

chrisbianca:
Thank you again for getting this added.

oblador:
@chrisbianca This is great and I need it too, mind creating a release for this?

Salakar:
@bilalsyed001 hmm you're right, can see this PR adds a `FIRMessaging` import + usage for the options param on `getToken`

@ctaintor is this definitely a requirement? If so then we should probably adjust the docs to state it's dependent on messaging.

ctaintor:
You need to pass the apns token. You could change the native signature for getToken to accept the options and inject the apns token from elsewhere. This requires the token to be exposed in js I believe but would mean messaging is only required if you use getToken.

Another option is to move these to messaging.

I won't have time to do this for a while since I'm on vacation until the end of the month.

PR #1265, opened 2018-07-02
added firebase.notifications.Schedule type definitions

shadowmynd:


shadowmynd:
Closing after noticing this was an object - not a formal class.

PR #1291, opened 2018-07-10
Web SDK readme updates

EvanBacon:
Updated the checklist as part of an effort to help confused RN/Expo/Firebase users.

I've noticed that some of the popular google libraries seem to work with Firebase, but for whatever reason (maybe political) they are kept separate. 
If someone can think of a good flow for tying AdMob/Firebase or Analytics/Firebase closer together, let me know! :) 
Notice that I didn't list `Crashlytics` because the web SDK doesn't handle native crashes 😐so it doesn't really count.

## References for my updates

* AdMob: https://stackoverflow.com/a/37322845/4047926
* Analytics: https://support.google.com/analytics/answer/2587086  / https://stackoverflow.com/a/37333205/4047926
* Cloud Firestore: https://firebase.google.com/docs/firestore/
* Dynamic Links: https://firebase.google.com/docs/dynamic-links/
* Offline: https://firebase.google.com/docs/firestore/manage-data/enable-offline
* Remote Config: firebase/firebase-js-sdk#36
* Storage: https://firebase.google.com/docs/storage/web/start

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,28 +35,30 @@ All in all, RNFirebase provides much faster performance (~2x) over the web SDK a
 
 > '**?**' indicates partial support
 
+> '🤔' denotes functionality with separate libraries
+
 | Firebase Features                                                                                                                 | v2.2.x | v3.3.x | v4.2.x | Web SDK |
 | --------------------------------------------------------------------------------------------------------------------------------- | :----: | :----: | :----: | :-----: |
-| **AdMob**                                                                                                                         |   ✅   |   ✅   |   ✅   |   ❌    |
-| **Analytics**                                                                                                                     |   ✅   |   ✅   |   ✅   |   ❌    |
+| **AdMob**                                                                                                                         |   ✅   |   ✅   |   ✅   |   [🤔](https://stackoverflow.com/a/37322845/4047926)    |
+| **Analytics**                                                                                                                     |   ✅   |   ✅   |   ✅   |   [🤔](https://support.google.com/analytics/answer/2587086)    |
 | **App Indexing**                                                                                                                  |   ❌   |   ❌   |   ❌   |   ❌    |
 | **Authentication**                                                                                                                |   ✅   |   ✅   |   ✅   |   ✅    |
 | _-- Phone Auth_                                                                                                                   |   ❌   |   ✅   |   ✅   |   ❌    |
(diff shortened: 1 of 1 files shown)


Salakar:
Ok so I've gone through these and all the original ones on the current readme still stand, notes below:


- AdMob: https://stackoverflow.com/a/37322845/4047926
   - This link is a for linking an AdMob account/property to a Firebase Project, there's no SDK support here - the only way to use this is if the Android or iOS SDK's are accessible. Google Ads JS SDK is 100% reliant on there being a full web environment and is heavily DOM-based, so this wouldn't work either.
- Analytics: https://support.google.com/analytics/answer/2587086  / https://stackoverflow.com/a/37333205/4047926
  - This link is Google Analytics, not Firebase Analytics, you can use Google Analytics through a custom npm package that makes it none DOM reliant, (the Node JS ones) - but this is GA still.
- Cloud Firestore (including Offline above): https://firebase.google.com/docs/firestore/
  - The Firestore JS SDK is DOM based and uses web api's, e.g. last I checked (3 weeks ago it errors with `document is not defined`). It also requires btoa atob API's which RN doesn't have, for it's internal Blob formats.
- Dynamic Links: https://firebase.google.com/docs/dynamic-links/
  - There's no web SDK here? You can create links via REST API calls, yes - but you cannot retrieve inbound links that triggered a web app to load - there's no concept of that in the Web, only on Android & iOS devices.
- Offline: https://firebase.google.com/docs/firestore/manage-data/enable-offline
  - Uses window.localStorage and DB afaik (don't exist in RN), but you'd still need to get the Firestore SDK to work in the first place as above.
- Remote Config: firebase/firebase-js-sdk#36
  - This issue is still open with no support indicated? There's a REST API but this can only be used with Admin service account credential - which should never be used in an RN app as these credentials give full control over the Firebase and GCP proect 
  - these credentials are server side only.
- Storage: https://firebase.google.com/docs/storage/web/start
  - `?` Readme indicates partial support. You can upload a file with the Web SDK yes, but it has to go over the RN bridge in some form at least twice as the Web SDK has no concept of device file storage so you'd need to do it via `putString()`. Doing this is bad and will lock the bridge up, especially for large files. Twice over the bridge is also a conservitive estimate as well, if you need to modify the file, resizing, cropping, effects etc this will go over the bridge several times.
   - RNFB has a `putFile(pathToFile)` to allow all File based actions to stay natively like most RN modules do.

PR #1294, opened 2018-07-11
Honor repeatInterval when rescheduling outdated notication

dluksza:


android/src/main/java/io/invertase/firebase/notifications/RNFirebaseNotificationManager.java
```diff
@@ -365,11 +365,26 @@ private void scheduleNotification(Bundle notification, @Nullable Promise promise
...
         newFireDate.set(Calendar.HOUR_OF_DAY, currentFireDate.get(Calendar.HOUR_OF_DAY));
         newFireDate.set(Calendar.MINUTE, currentFireDate.get(Calendar.MINUTE));
         newFireDate.set(Calendar.SECOND, currentFireDate.get(Calendar.SECOND));
 
+        switch (schedule.getString("repeatInterval")) {
+          case "minute":
+            newFireDate.add(Calendar.MINUTE, 1);
+            break;
+          case "hour":
+            newFireDate.add(Calendar.HOUR, 1);
```
Salakar: Would this not still have the same issue as before;

E.g.: I set a `fireDate` of 3 hours ago, it comes through this logic with a `repeatInterval` of an hour so this adds an hour but it would still b 2 hours behind - so fires immediately?

In the first instance setting a `fireDate` in the past shouldn't be allowed on the validations in the JS code - that stops one thing. But; when the`RNFirebaseNotificationsRebootReceiver` triggers `rescheduleNotifications()` after a device reboot there's a good chance `fireDates` can be in the past, so we'd still need to compensate for this anyway 🤔

dluksza: You are right, it is not as easy as I initially expected. :|

dluksza:
@Salakar would you find time to review this again?

ifsnow:
@Salakar It looks like there was a mistake in my previous PR. I think @dluksza's implementation is good. I'm really sorry to make you work twice.

Issue #1267, opened 2018-07-02
Unable to load data when app is backgrounded (triggered by notification)


tclarke-scottlogic:
If I get a push notification on iOS, I want to pull some related data from the Firebase database in the background, so my user can read their articles on the train.

I'm getting the appropriate onDisplayNotification event, and it's reliably triggering, but when I query the database, it doesn't load the relevant data until the app is foregrounded, by which time the user may be offline.

So how can I retrieve entries from firebase.database() when my app is backgrounded?

chrisbianca:
@tclarke-scottlogic This is related to: https://github.com/invertase/react-native-firebase/issues/971

In trying to simplify the messaging and notification process as part of v4, we had hidden some of the control that iOS offers when handling background messages and notifications.  We need to reinstate this functionality in a simple to understand manner.  I'm going to close this issue to remove the duplicate, but please track #971 for progress.

pragneshAliansoftware:
Any fix for this issue? I am also facing this problem in v4.0.0

Issue #971, opened 2018-04-11
iOS remote notification handler and background data support


dougkeen:
### Issue

In the 3.x versions of react-native-firebase, a `finish` method was available in the messaging module that allowed the developer to invoke the `fetchCompletionHandler` provided in the iOS app delegate method `- (void)didReceiveRemoteNotification:(NSDictionary *)userInfo fetchCompletionHandler:(void (^)(UIBackgroundFetchResult))completionHandler`. This mirrored the implementation provided in react native's own [PushNotificationIOS implementation](https://facebook.github.io/react-native/docs/pushnotificationios.html#content). According to the [Apple documentation](https://developer.apple.com/documentation/uikit/uiapplicationdelegate/1623013-application?language=objc#parameters), that `fetchCompletionHandler` *must* be called, and should be called "as soon as possible".

This capability was removed in the 4.x overhaul, and the app-delegate-provided `fetchCompletionHandler` is now ignored.

Without a way to call that callback, iOS will consider background fetches executed through react-native-firebase as having the worst possible performance. In short order, silent pushes will be throttled or disabled for the app ([this article](https://topologyeyewear.github.io/engineering-blog/2017/11/20/background_transfer/) has some good information on how the OS "punishes" apps with seemingly long running background tasks).

### Environment

1. Application Target Platform: iOS

2. Development Operating System: N/A

3. Build Tools: N/A

4. React Native version: N/A

5. RNFirebase Version: 4.0.1

6. Firebase Module: notifications, messaging

chrisbianca:
It turns out that there were a couple of places where completionHandler should have been being called, but wasn't. This has been updated pending release in v4.0.3.

I'd be interested in hearing your views on whether we should be giving more control over how this method should be called?

dougkeen:
Thanks for the clarification. Things aren't quite as dire as I thought on first glance :)

I still think there's a need to give the developer control over when that callback is invoked. If my understanding is correct, the javascript that's executed in response to the event that's sent over the bridge happens asynchronously with respect to the native handler. That could be problematic if you want to do some work in the `onMessage` handler.

As an example, let's say we want to download some data from the server, do some data transformation, then store the result somewhere when we receive a silent push. Assuming the `onMessage` handler is executed asynchronously with respect to the native handler, consider this situation:
```
Native thread                 JS thread
--------------------------    ---------------------
+ Receive remote message 
+ Send event over bridge      + onMessage called
+ Call completionHandler      + perform http fetches
[...]

PR #1393, opened 2018-08-13
Background notification fix for ios

ryanggrey:
Fixes:
- https://github.com/invertase/react-native-firebase/issues/1267
- https://github.com/invertase/react-native-firebase/issues/971

Note that I haven't implemented a fix for foreground notifications or messages yet. Thought it would be best to get feedback on the approach before extending to include foreground notifications and messages.

One of the key considerations was to avoid introducing memory leaks in objc code. 

If the (library) user is *not* listening for `onNotificationDisplayed` then we have to call the completion handler (auto-complete). We have no user listener to chain to the end of so we need to do this from within the internal listener.

If the (library) user *is* listening for `onNotificationDisplayed` then we need to try and auto-complete after the user has finished _all_ work in their listener. This is because the user must have the chance to call the completion handler themselves with the correct parameter (instead of our assumed `UIBackgroundFetchResultNoData`) _before_ we attempt auto-completion. This has resulted in a change to the `OnNotification` type - a `Promise` must be returned so we can chain to the end of it. It would be simpler for us to not attempt auto-completion in this case (so that the `OnNotification` type doesn't need to change) however I imagine it will be easy for upgrading users to miss the need to call the new `done` param. In summary - the chance of upgrading users suffering memory leaks is high so we should guard against it at the cost of adding complexity and changing an existing function type. Any opinions on this would be appreciated.


ryanggrey:
It's also occurred to me that docs will need to be updated if this PR is accepted. I found a few references to the changed API/registration function here - https://github.com/invertase/react-native-firebase-docs/tree/v4.3.x/docs/notifications. Anywhere else I should check for references? I'd be happy to make a docs PR once the approach for this fix has been approved.

Salakar:
@ryanggrey thanks a ton for starting this off.

If we release under a major release then upgrading issues can be mitigated as part of upgrade guide for v4 to v5. The next major release is v5 which in progress right now - would be good therefore to land this as part of that.

Could we not move the JS `completionHandler` logic to the `Notification` class - it'd be much simpler, specifically the [`IOSNotification`](https://github.com/invertase/react-native-firebase/blob/master/src/modules/notifications/IOSNotification.js) class - the end-user gets sent an instance of this already in their listeners. Something like `await notification.ios.complete(result)`. The handlerKey can be sent natively as part of the notification payload on iOS e.g. attach it [somewhere here](https://github.com/invertase/react-native-firebase/pull/1393/files#diff-27cb0e541ec7db4b073768831ce8e9dfR121) - then when the [`IOSNotification`](https://github.com/invertase/react-native-firebase/blob/master/src/modules/notifications/IOSNotification.js) class constructs itself after receiving the `NativeNotification` payload it can do whatever is needed internally for the auto completionHandler logic & manual logic - all self contained then.

A potential `autoComplete` solution  then based on the above could be a short timeout (say 5 seconds?) on the `IOSNotification` class when constructed from `NativeNotification` and only when handlerKey is present - user land code can synchronously call `notification.ios.cancelAutoComplete();` immediately on receiving the notification in their listener to cancel it;  then it's up to the user to call `await notification.ios.complete(result)` when they're ready - otherwise the time out continues and calls `notification.ios.complete(result)` with `UIBackgroundFetchResultNoData` automatically. Or a global setting on `notifications()` e.g. `notifications().disableIOSAutoComplete()` instead so we don't have to deal with timeouts - its then becomes handle custom completion or not - no mix and match - enabled by default - which causes JS to call `notification.ios.complete(result)` automatically on construction of a `IOSNotification` from NativeNotification payload with a handlerKey.

In terms of docs - could you make any PRs to the v5 branch, https://github.com/invertase/react-native-firebase-docs/tree/v5.0.x - you can also use GitHub search at the top to search for occurrences on master (which is in sync with v5 currently)

My main feedback code wise is that everything internal or public API wise should be prefixed with `iOS` as it's an iOS only feature - makes it easier for end users and contributors to distinguish platform specific logic; e.g. all variations of `backgroundFetchResult` become `iOSBackgroundFetchResult` - if it's all moved to the IOSNotification class then that's already all nicely platform specified. Will add some misc review notes on code as well.

ios/RNFirebase/notifications/RNFirebaseNotifications.m
```diff
@@ -106,6 +118,9 @@ - (void)didReceiveRemoteNotification:(NSDictionary *)userInfo
         return;
     }
 
+    NSDictionary *notification = [self parseUserInfo:userInfo];
```
Salakar: Could attach the handlerKey onto the notification payload here

ios/RNFirebase/notifications/RNFirebaseNotifications.m
```diff
@@ -95,6 +98,15 @@ - (void)didReceiveLocalNotification:(nonnull UILocalNotification *)localNotifica
     }
 }
 
+RCT_EXPORT_METHOD(complete:(NSString*)handlerKey fetchResult:(UIBackgroundFetchResult)fetchResult) {
+    void (^completionHandler)(UIBackgroundFetchResult) = completionHandlers[handlerKey];
+    completionHandlers[handlerKey] = nil;
+
+    if(completionHandler != nil) {
+        completionHandler(fetchResult);
```
Salakar: `completionHandler` should be removed from the `completionHandlers` on consumption?

ryanggrey: This happens on line 103. We do it before calling the `completionHandler` as the OS may suspend execution when `completionHandler` is called.

ryanggrey:
@Salakar Thanks for the feedback - much appreciated! I've opted for your "all in" suggestion where the user opts to go all in and handle completion themselves by setting a global `firebase.notifications().ios.shouldAutoComplete` and we don't use timers.

All iOS specific API is now on `notification.ios` or `notifications.ios` and completion API has been moved onto `IOSNotification`. 

Again I have yet to implement for foreground notifications or messages. I'll wait until approval of the general approach has happened before doing that. Would be great to get your feedback on the update.

src/index.d.ts
```diff
@@ -1165,7 +1168,7 @@ declare module 'react-native-firebase' {
         ): () => any;
 
         onNotificationDisplayed(
-          listener: (notification: Notification) => any
+          listener: (notification: Notification, done: CompletionHandler) => any
```
ryanggrey: Need to remove `done` param as it doesn't exist anymore.

src/modules/notifications/index.js
```diff
@@ -33,7 +36,9 @@ import type {
   Schedule,
 } from './types';
 
-type OnNotification = Notification => any;
+type CompletionHandler = BackgroundFetchResultValue => void;
+
+type OnNotification = (Notification, CompletionHandler) => any;
```
ryanggrey: Remove `CompletetionHandler` from `OnNotification`. This means we don't need `CompletionHandler` at all in this file.

ios/RNFirebase/notifications/RNFirebaseNotifications.m
Salakar: Oh, I meant `[completionHandlers removeObjectForKey:handlerKey];` specifically, setting to `nil` still keeps the key + a nil object in the dictionary afaik?

ryanggrey: When I was debugging this I checked and `completionHandlers[handlerKey] = nil` definitely resulted in the key also being removed. Happy to change to be sure though if you'd prefer `[completionHandlers removeObjectForKey:handlerKey]`?

Salakar: If that's the case then as-is is fine - my obj-c some days is 💩

Salakar:
@ryanggrey thanks for making the changes. The "all-in" approach and API you've done looks good to me 👍

ios/RNFirebase/notifications/RNFirebaseNotifications.m
Salakar: Using the `notificationId` is a much better approach, stick with that. 👍

ryanggrey:
@Salakar Thanks for reviewing. I can make additional changes to introduce the completion handler API for foreground notifications and messages in this PR or in subsequent PRs. Which would you prefer?

Salakar:
@ryanggrey I'm fine with either - both fall under adding the completion handler logic/features; so in one PR is fine, up to you - no preference on this one.

ryanggrey:
@Salakar I've had a quick look and implementing for foreground notifications is going to be slightly more work due to the fact that `didReceiveRemoteNotification:fetchCompletionHandler` does not send a JS event but instead defers that to `userNotificationCenter:willPresentNotification:withCompletionHandler:` in order to avoid duplicate JS notifications/events. The problem arises because each of those methods parses the "notification" (`NSDictionary` in the case of the prior and `UNNotification` in the case of the latter) differently resulting in different notification Ids. Couple this with the fact that its the completion handler in `didReceiveRemoteNotification:fetchCompletionHandler` that we need to store and the Id sent to JS in `userNotificationCenter:willPresentNotification:withCompletionHandler:` that we need to use when calling the native `complete` method and we have a problem that needs to be solved.

Given that this PR as it stands solves #1267 & #971 it might be nice to get this merged so people can benefit from the fix and raise another issue for foreground notification completion handlers and message completion handlers to be dealt with separately. Happy to help with raising issues if needed. I'll get on with a PR for docs next.

Salakar:
@ryanggrey ok, I agree, we can tackle foreground in a later release / another issue. Are you happy for me to merge this as is or was there any work you had left to on this PR? We can get this in for v5.0.0

If you could raise another issue in regards to the above we can continue the discussion there.

Thanks for all your help and your contributions on this, much appreciated! 👍

ryanggrey:
Added docs PR - https://github.com/invertase/react-native-firebase-docs/pull/102

ryanggrey:
Raised foreground notification and message issue - https://github.com/invertase/react-native-firebase/issues/1408

Salakar:
@ryanggrey looks like theres a condition on this change where `notifications` is undefined and causes a crash, 

From Discord:

image

cc @ovaris

Salakar:
Hmm actually i think this is user code creating a new notification, e.g.:

```js
const notification = new Notification();
```

Which means the second arg is not passed, could move the second arg internally and just get notifications ourselves by importing firebase?

ryanggrey:
@salakar Sorry if this has caused any problems and for the lateness of my reply. I’ve been quite busy lately and not sure I’ll be able to look into this for a while. I’ll check in here again once I have some more time available. 

I’m guessing this means the change will be reverted and it won’t make it into v5v

Salakar:
@ryanggrey I'll look at fixing it for the next RC if you don't beat me to it, no problem 👍

PR #1434, opened 2018-08-28
enabled automatic reference count for ios

Aigh-test:
Disabled ARC was marked as vulnerability by security scanner. This PR enables ARC


Salakar:
@Aigh-test thanks for this, can you re-sync up with master so the CI will trigger and test this. Thanks

Salakar:
Looks like you're using your fork for other purposes now, will close this as the scope of the original PR has now changed. Feel free to open another with just this change if you're still interested. Thanks

PR #1447, opened 2018-08-31
Fixing transitive dependencies vendored as static frameworks

Ricowere:
Hello React native team!

First and most important, incredible work you're doing with this library.

I'd like to expose the issues we faced integrating this library into our project and how we tackle them. We thought it might help for the community to have this amend in the official repository.

**The Context**
Currently we're runnning a _swift_ based project with a _cocoapods_ setup for managing the dependencies.

**The Problem**
When we add this new dependency as another pod in our setup via standard procedure (pod install ..) we found this error:

```
The 'Our App' target has transitive dependencies that include static binaries: (firebase deps...)
```

**The Solution**
The solution we found is to let cocoapods (base on the recent version they deployed) to bundle react-native-firebase podspec with the flag `s.static_framework = true` to allow to be bundled as static dependency if needed.
Also for make this work we had to modify some import headers from `#import <RNFirebaseUtil.h>` to `#import "RNFirebaseUtil.h"`

And then problem solve. All happy here with RNFirebase integrated successfully in our iOS app.

Let us know what do you think guys.

Regards!
David.

### DIFF
--- a/ios/RNFirebase.podspec
+++ b/ios/RNFirebase.podspec
@@ -15,6 +15,7 @@ Pod::Spec.new do |s|
   s.social_media_url    = 'http://twitter.com/RNFirebase'
   s.platform            = :ios, "9.0"
   s.source_files        = 'RNFirebase/**/*.{h,m}'
+  s.static_frameworks   = true
   s.dependency          'React'
   s.dependency          'Firebase/Core'
 end
--- a/ios/RNFirebase.podspec
+++ b/ios/RNFirebase.podspec
@@ -15,7 +15,7 @@ Pod::Spec.new do |s|
   s.social_media_url    = 'http://twitter.com/RNFirebase'
   s.platform            = :ios, "9.0"
   s.source_files        = 'RNFirebase/**/*.{h,m}'
-  s.static_frameworks   = true
+  s.static_framework    = true
   s.dependency          'React'
   s.dependency          'Firebase/Core'
 end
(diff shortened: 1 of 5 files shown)


Salakar:
@Ricowere welcome 🎉  

Thanks for submitting this PR with some helpful detail 👌 Would import change interfere with non Pod users? Quite a lot of people choose to link manually in their project rather than using Pods.

If it does then an alternative perhaps could be removing the `s.dependency          'Firebase/Core'` line from the pod  - might fix your issue without that being specified - would need to be added to your own pod spec instead?

Ricowere:
Sorry guys for the tremendous delay. I was thinking to make sure everything works in different set ups. We could add a double check for that import. Something like:

```
#if __has_include(<Firebase.h>)
#import <Firebase.h>
#elif __has_include("Firebase.h")
#import "Firebase.h"
#else
#import "Firebase.h"
#endif
```

Let me try during the day either ways, our setup and a standard setup.

Salakar:
Hey @Ricowere; apologies for the delay on this; as part of v6 all React Native Firebase Pods are now vendored as static frameworks (so this will no longer be an issue). If you're interested; I implemented this in https://github.com/invertase/react-native-firebase/pull/1958.

[[The full changelog can be viewed here](https://github.com/invertase/react-native-firebase/blob/master/CHANGELOG.md#crashlytics)]

Thanks for submitting this and for your patience; also apologies that I couldn't merge this - I was unsure at the time of the implications of changing the imports. The restructure of v6 made this a lot easier to deal with.

Will close this now.

---

PR #1476, opened 2018-09-10
[auth] Support updatePhoneNumber

oblador:
Regarding the `updatePhoneNumber` method, the documentation states 
>The following methods are not supported in RNFirebase as they cannot work in the React Native environment or have a different implementation.

I didn't see any reason why we couldn't support it, so went ahead and made an implementation. 

I've tried it manually on iOS and Android but didn't include any E2E tests as it requires some changed settings in the firebase account and could see phone auth is absent from current E2E suite. Please advice on how to proceed or if they should stay omitted for now.

### DIFF
--- a/src/modules/auth/User.js
+++ b/src/modules/auth/User.js
@@ -251,6 +251,24 @@ export default class User {
   }
 
   /**
+   * Update the current user's phone number
+   *
+   * @param  {AuthCredential} credential Auth credential with the _new_ phone number
+   * @return {Promise}
+   */
+  updatePhoneNumber(credential: AuthCredential): Promise<void> {
+    return getNativeModule(this._auth)
+      .updatePhoneNumber(
+        credential.providerId,
+        credential.token,
+        credential.secret
+      )
+      .then(user => {
+        this._auth._setUser(user);
+      });
+  }
+
+  /**
    * Update the current user's profile
    * @param  {Object} updates An object containing the keys listed [here](https://firebase.google.com/docs/auth/ios/manage-users#update_a_users_profile)
    * @return {Promise}
@@ -314,15 +332,6 @@ export default class User {
       INTERNALS.STRINGS.ERROR_UNSUPPORTED_CLASS_METHOD(
         'User',
         'reauthenticateWithRedirect'
-      )
-    );
-  }
-
-  updatePhoneNumber() {
-    throw new Error(
-      INTERNALS.STRINGS.ERROR_UNSUPPORTED_CLASS_METHOD(
-        'User',
-        'updatePhoneNumber'
       )
     );
   }
--- a/android/src/main/java/io/invertase/firebase/auth/RNFirebaseAuth.java
+++ b/android/src/main/java/io/invertase/firebase/auth/RNFirebaseAuth.java
@@ -712,6 +712,63 @@ class RNFirebaseAuth extends ReactContextBaseJavaModule {
   }
 
   /**
+   * updatePhoneNumber
+   *
+   * @param provider
+   * @param authToken
+   * @param authSecret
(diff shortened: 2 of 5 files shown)


Salakar:
@oblador woo thanks for this. 🎉 The initial reason it was not supported was because it was missing on one of the platforms - but as you've pointed out they're both there now (we did have an internal todo for this). 

In terms of the phone auth e2e testing - I've only recently migrated some of the tests from our internal tests to be public and on CI - I will get a dummy testing phone number and code added to the console and report back here - if you're willing to add the tests for it as well (would be great)?

Would also be best to add e2e tests after I land https://github.com/invertase/react-native-firebase/pull/1472 (next day or 2) - CI will be back to normal then for iOS at least and locally running tests for both platforms.

Could you also update the typescript def - around here: https://github.com/invertase/react-native-firebase/blob/master/src/index.d.ts#L778

Thanks

---

oblador:
Cool. Have updated type definitions, LMK if you want me to wait until the test setup is fixed or if we should merge and then you can add the tests while you're at it (the 0.57 RC's has been a quite troublesome for me).

Salakar:
Leave the rest with me, this is a great start. Will ping back here I'd there's any issues.

If you need an assist with anything 57 let me know, have got it working a couple of times now 😅 rc4 is fairly stable

src/modules/auth/User.js
```diff
@@ -250,6 +250,24 @@ export default class User {
       });
   }
 
+  /**
+   * Update the current user's phone number
+   *
+   * @param  {AuthCredential} credential Auth credential with the _new_ phone number
+   * @return {Promise}
+   */
+  updatePhoneNumber(credential: AuthCredential): Promise<void> {
```
Salakar: Promise returning a User instance, instead of void here

src/modules/auth/User.js
```diff
@@ -250,6 +250,24 @@ export default class User {
...
+    return getNativeModule(this._auth)
+      .updatePhoneNumber(
+        credential.providerId,
+        credential.token,
+        credential.secret
+      )
+      .then(user => {
+        this._auth._setUser(user);
```
Salakar: Add a return here to return the user, setUser updates and returns the user.

oblador:
RC4 still doesn't work very well with Jest for me and it seems to break Storybook (although that might just be a Storybook issue). 

Anyway, did requested changes although they were inconsistent with the other `update*` methods.

Salakar:
I based by requested changes of the proposed types: `updatePhoneNumber(credential: AuthCredential): Promise<User>;` - I agree they should be consistent however. Just took a look at the JS SDK reference and they all return void so you were right, just the types were wrong, my bad - https://firebase.google.com/docs/reference/js/firebase.User#updatePhoneNumber

I'm happy to change back if you want

oblador:
Cool, reverted and updated TS definition instead :-)

Salakar:
@oblador thanks for sorting.

RN 0.57 release is out now so i'll get the testing framework sorted and get this merged. I've also added the following testing phone numbers and codes to the testing project:

```
+44 7445 123456
123456

+44 7445 123457
654321

+44 7445 123458
112233
```

oblador:
Amazing! Any chance of a RC 2 release?

Salakar:
Of course, aiming to publish rc2 late afternoon tomorrow - still need to add the tests for these specifically. 

I've had to add the `AuthSettings` API to enable phone Auth testing support, so took some time - some phone Auth tests have been added here and working: https://github.com/invertase/react-native-firebase/blob/master/tests/e2e/auth/phone.e2e.js, just need to test these ones now specifically unless you beat me to it 🙃

PR #1491, opened 2018-09-14
[android][utils] `isAppInForeground` not accounting for React Native LifecycleState

Salakar:
Prior to this change, this utility in some rare cases would return true that Activity was in the foreground, however, React Native was still in the process of resuming, this, for example, led to crashes in HeadlessJS Notification tasks: "Tried to start task RNFirebaseBackgroundMessage while in foreground, but this is not allowed."

---

### DIFF
--- a/android/src/main/java/io/invertase/firebase/Utils.java
+++ b/android/src/main/java/io/invertase/firebase/Utils.java
@@ -9,6 +9,7 @@ import com.facebook.react.bridge.ReactContext;
 import com.facebook.react.bridge.ReadableArray;
 import com.facebook.react.bridge.ReadableMap;
 import com.facebook.react.bridge.WritableMap;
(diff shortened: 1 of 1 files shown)


vrozaev:
Hello @Salakar!

I need you help and I ready to provide any necessary information. It looks that this issue still can be reproduced (react-native-firebase@5.3.1, react-native@0.59.10). 

We still have such exception:

> java.lang.IllegalStateException: Tried to start task RNFirebaseBackgroundMessage while in foreground, but this is not allowed.

The exception throws in this line: https://github.com/facebook/react-native/blob/master/ReactAndroid/src/main/java/com/facebook/react/jstasks/HeadlessJsTaskContext.java#L98

How you originally find out that isAppInForeground is incorrect? Can you somehow check - is that issue reproduce on your projects or not?

PR #1513, opened 2018-09-21
[ts] fix Messaging requestPermission return type

squirly:
See: https://github.com/invertase/react-native-firebase/blob/619cd17516a88dff8b89b749888461fe91480737/src/modules/messaging/index.js#L129


Salakar:
I think it's best if this function is changed to correctly resolve true or false rather than erroring if false.

Salakar:
I'll accept this for now and look at changing for v6 so the change is behind a major release.

Thanks again for this.

---

PR #1533, opened 2018-09-27
Add functions type

zhigang1992:
Looks like `functions` is missing from the exports.

Need help with the rest of the steps, should I do `yarn build` or something?


Salakar:
@zhigang1992 thanks for this. The CI automatically validates TS declarations and it has passed so this seems fine to me. 👍

PR #1580, opened 2018-10-09
Fix typings to for `createShortDynamicLink`

evelynccordner:
'createShortDynamicLink' should accept a link as well as a type parameter, similar to 'createDynamicLink'. 
See documentation here: https://rnfirebase.io/docs/v4.3.x/links/reference/links#createShortDynamicLink


Salakar:
Thank you 👍

Issue #1541, opened 2018-09-30
Reference.toJSON is not implemented


gunslingerfry:
### Issue

[Documentation](https://rnfirebase.io/docs/v5.x.x/database/reference/Reference) lists a Reference.toJSON function, this function does not exist.

### Environment

1. Application Target Platform:
Both

2. Development Operating System:
Ubuntu 18.04

3. Build Tools:

4. `React Native` version:
0.56.0

5. `React Native Firebase` Version:
5.0.0

6. `Firebase` Module:
database

7. Are you using `typescript`?
no

---

eragon512:
is this issue available/open? I'd like to take it up, if so

Salakar:
@eragon512 it is, straightforward one also, it just needs to return the value of Reference.toString() to match the WEB SDK - see: https://github.com/firebase/firebase-js-sdk/blob/master/packages/database/src/api/Query.ts#L577

PR #1595, opened 2018-10-14
[WIP] attempting implementation of Reference.toJSON()

eragon512:
* attempting to implement Reference.toJSON() as per the docs

Referencing #1541 

I am unsure if I am on the right track to solve this issue. 
Kindly review the PR and let me know if I am on the right track and what are the next steps to be taken. 
Thanks!!


Salakar:
@eragon512 thanks for starting this. The simplest way to match the Web SDK behaviour would be to add the following after this line: https://github.com/invertase/react-native-firebase/blob/master/src/modules/database/Reference.js#L504

```js
  toJSON(): string {
    return this.toString();
  }
```

As you can see on the Firebase Web JS SDK here: https://github.com/firebase/firebase-js-sdk/blob/master/packages/database/src/api/Query.ts#L577 this is purely for serialization purposes only and it only returns the same as `toString()`

eragon512:
Implemented the function as suggested.
Please review and let me know if any further changes are needed

PR #1624, opened 2018-10-22
[Android][IOS][Firestore] add arrayRemove and arrayUnion to FieldValue 

### LINKED ISSUE #1389: FieldValue.arrayUnion() and FieldValue.arrayRemove() APIs.
These two features are not working.

please update firestore...

andrevus:
### Summary

Add methods arrayRemove and arrayUnion to work with arrays in firestore.

Fixes #1389

### Checklist
* [x]  Supports `Android`
* [x]  Supports `iOS`
* [x]  `e2e` tests added or updated in [/tests/e2e/*](/tests/e2e)
* [x]  Updated the documentation in the [docs repo](https://github.com/invertase/react-native-firebase-docs)

https://github.com/invertase/react-native-firebase-docs/pull/134

* [x]  Flow types updated
* [x]  Typescript types updated

### DIFF
--- a/src/modules/firestore/FieldValue.js
+++ b/src/modules/firestore/FieldValue.js
@@ -4,6 +4,14 @@
  */
 
 export default class FieldValue {
+  get elements() {
+    return this._elemenets;
+  }
+
+  set elements(elements) {
+    this._elemenets = elements;
+  }
+
   static delete(): FieldValue {
     return DELETE_FIELD_VALUE;
   }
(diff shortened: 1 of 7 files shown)


src/modules/firestore/FieldValue.js
```diff
@@ -2,16 +2,39 @@
...
   static delete(): FieldValue {
     return DELETE_FIELD_VALUE;
   }
 
   static serverTimestamp(): FieldValue {
     return SERVER_TIMESTAMP_FIELD_VALUE;
   }
+
+  static arrayUnion(...elements: AnyJs[]) {
+    ARRAY_UNION_FIELD_VALUE.elements = elements;
```
Salakar: A small bug here; this should return a new instance of FieldValue and not be a singleton like the others.

If you add a test with 2x `FieldValue`.`arrayUnion` or 2x `arrayRemove` (with different values) then it'll fail as only the 2nd `FieldValue` will overwrite the elements of the first `FieldValue`, if that makes sense?

Changing this will the `if (value === ARRAY_UNION_FIELD_VALUE) {` checks - so they'd need changing also.

src/modules/firestore/FieldValue.js
```diff
@@ -2,16 +2,39 @@
...
+    return ARRAY_UNION_FIELD_VALUE;
+  }
+
+  static arrayRemove(...elements: AnyJs[]) {
+    ARRAY_REMOVE_FIELD_VALUE.elements = elements;
```
Salakar: Same bug as above.

src/modules/firestore/FieldValue.js
```diff
@@ -2,16 +2,39 @@
...
+    return ARRAY_REMOVE_FIELD_VALUE;
+  }
 }
 
 export const DELETE_FIELD_VALUE = new FieldValue();
 export const SERVER_TIMESTAMP_FIELD_VALUE = new FieldValue();
+export const ARRAY_UNION_FIELD_VALUE = new FieldValue();
```
Salakar: Same bug as above.

src/modules/firestore/FieldValue.js
```diff
@@ -2,16 +2,39 @@
...
+export const ARRAY_REMOVE_FIELD_VALUE = new FieldValue();
```
Salakar: Same bug as above.

PR #1832, opened 2019-01-14
fixed UploadTaskSnapshot ref type

abdullah-sr:
I noticed that `ref` in `UploadTaskSnapshot` was of type string and not actually `Reference`


Salakar:
@abdullah-sr thanks for reporting this, the JS code though should be returning an instance of the reference here though - so I'll get this fixed in v6.0.0. Thanks for this PR though.

PR #1853, opened 2019-01-23
adds option to specify shape of callable function params and result

unao:
It improves `httpsCallable` interface by allowing to specify shape of input output with generics.
If none specified it will default to `any` (current implementation).


Salakar:
@unao thanks for this, makes sense! Will merge for next patch release.

PR #1860, opened 2019-01-25
Fix timestampToUTC timezone handling

Strate:
TimeZone.getOffset() is in milliseconds, so it should be added to milliseconds, not to seconds.

`Date(long date)` this constructor accepts milliseconds from unix epoch, which considered to be in UTC time. So, there should not be any extra timezone handle.

### Summary

I found this bug in GMT+3, all dates were shifted to coulple of months to future. With this fix all dates becomes correct.

Please note, that I am originally is not java/android developer, I just used the debugger to find this. So, plz, check carefully.

### DIFF
--- a/android/src/main/java/io/invertase/firebase/Utils.java
+++ b/android/src/main/java/io/invertase/firebase/Utils.java
@@ -28,8 +28,7 @@ public class Utils {
   private static final String TAG = "Utils";
 
   public static String timestampToUTC(long timestamp) {
-    Calendar calendar = Calendar.getInstance();
-    Date date = new Date((timestamp + calendar.getTimeZone().getOffset(timestamp)) * 1000);
+    Date date = new Date(timestamp * 1000);
     SimpleDateFormat format = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss'Z'", Locale.US);
     format.setTimeZone(TimeZone.getTimeZone("UTC"));
     return format.format(date);


Salakar:
Thanks for this @Strate - I'll test this in v6 and if all ok I'll accept this PR for merging into v5

mikehardy:
@Salakar - did you see anything in your v6 testing?

Salakar:
@mikehardy this change isn't mergeable as is - the offset looks to be completely removed here, it's still required, just needs to be like this though I think:

```java
Date date = new Date(((timestamp * 1000) + calendar.getTimeZone().getOffset(timestamp * 1000)));
```

Strate:
@salakar as I stated in PR, looks like there is no need to alter time stamp with timezone offset at all, because it comes from FB in UTC, and Date constructor expects it in UTC too.

Strate:
@salakar what can I do to make this fix available for v5? I use for a long time patched version of library without any issue. You say that timezone offset needed, but I can not get a reason. Date constructor accepts timestamp in UTC, and firebase’s timestamp comes with UTC.

mikehardy:
For a v5 release @Salakar and I need to either 1- make the android emulator work in the existing test project so we can do releases, or 2- finish the react-native 0.60 test project conversion in v5.x.x - both with the end result that android e2e tests run and unblock releases. From my perspective that's the only thing blocking integration of fixes like this again but I haven't had time to finish item 2 and I'm not in control of item 1 in the CI environment so it's been parked

I'm using a patch myself for #2476 on the v5.x.x branch which is a little silly since I have maintainer powers but with CI blocked I'm not comfortable releasing

mikehardy:
@Salakar I only have a couple PRs that still apply to v5 and are unmerged.

There is a fundamental question here that is unresolved.

**Do we need to offset by timezone or not**

My reading of Date constructor documentation indicates that @Strate is correct. The Date constructor expects a UTC date. So no translation should be needed.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date

So I'd merge this as is to v5 unless I hear differently

russellwheatley:
Hey @Strate, thank you for your pull request. We’re no longer actively supporting `<= v5` development so I’m going to go ahead and close this. If this PR change still applies to the latest version of React Native Firebase then please feel free to raise a new PR targeting the latest version with your changes.

PR #1864, opened 2019-01-26
Messaging from secondary firebase app

erezacohen:
### Summary

Support Messaging with secondary Firebase App!

Expanded React-Native to support native SDK functionality:
- Support getToken by senderID (which is different from main Firebase App).
`firebase.messaging().getToken(senderId?: string)`

- Support deleteToken by senderID (which is different from main Firebase App).
`firebase.messaging().deleteToken(senderId?: string)`

### Checklist

- [x] Supports `Android`
- [x] Supports `iOS`
  - **LINK TO DOCS PR HERE**
- [x] Flow types updated

### Release Plan

---


mikehardy:
This looks neat, but is a breaking change. Since the v5 / v6 transition means there is not really semver number space for breaking changes, but messaging isn't in v6 yet, the only way to get this in v5 would be to have a backwards-compatible API available (maybe deprecated if it's just a stopgap during v5->v6) 

The API in question here appears to be a "synthetic" one (local to this module) as opposed to something direct in the SDK APIs so there shouldn't be any problem supporting both styles with different local method calls, at which point this wouldn't be a breaking change I think, and might have a chance for merge quickly?

erezacohen:
Hi Mike, thanks for reaching out. 

When will Messaging be available on v6? I would like to make sure this feature is supported when available on v6.

Thanks,
Erez

mikehardy:
Due dates in open source don't mean anything, but the tracker issue for v6 is pinned in this repo: https://github.com/invertase/react-native-firebase/issues/2025

Best current guess (subject to change) is July-ish I believe

Salakar:
This will definitely be supported in v6.

erezacohen:
Awesome, happy to hear that! 
I’ll try to make the backward compatible changes, hopefully I’ll have time to do it soon.

Thanks,
Erez

Salakar:
Also; work on v6 messaging & notifications starts in ~4 weeks

mikehardy:
We've just finished shaping v5 up for maintenance mode, and as long as this doesn't skew from v6 (it should not) and is backwards compatible (dependent on @erezacohen ) - I can get this into v5 I believe

Salakar:
I'm fine with this for v5.x.x, my only minor nitpick is that the senderId default value can probably be done in js instead of on each native platform, via `this.app.options.messagingSenderId`

But I'm fine either way 👍 will leave it up to @mikehardy's discretion

src/index.d.ts
```diff
@@ -1230,9 +1230,9 @@ declare module 'react-native-firebase' {
         /**
          * Returns the devices FCM token.
          */
-        getToken(): Promise<string>;
+        getToken(senderId?: string): Promise<string>;
 
-        deleteToken(authorizedEntity?: string, scope?: string): Promise<void>;
```
mikehardy: if this part didn't change the typescript method signature, this wouldn't be breaking. But I'm not sure how this can actually be valid as a type change, it seems like we're losing an API completely (that I would assume is necessary?) in order to get a new one that can take a senderId? I am probably missing something, but I need to understand

src/modules/messaging/index.js
```diff
@@ -79,12 +79,12 @@ export default class Messaging extends ModuleBase {
     return this._ios;
   }
 
-  getToken(): Promise<string> {
-    return getNativeModule(this).getToken();
```
mikehardy: I agree with the other remove comment from Salakar - this could be the place where you check for senderID null and use the javascript level token, then we have one set of branching logic in javascript vs two sets, one in each native area. It should serve to simply the native code a bit

src/index.d.ts
Salakar: We can make this non-breaking by keeping matching the JS code in each method to how iid() does it, with default arg values to be the same as these: 

https://github.com/invertase/react-native-firebase/blob/v5.x.x/src/modules/iid/index.js#L51-L52

https://github.com/invertase/react-native-firebase/blob/v5.x.x/src/modules/iid/index.js#L65-66

That way the old method signature with no args still continues to work as it did before.

mikehardy:
@erezacohen if you make the changes from review above, I can merge them into a release, I'm shaping up 5.5.0 right now, but maybe right after?

erezacohen:
Hi @mikehardy - Sorry for the delay. I'm handling my startup's fundraising so hectic times :) Let me know when you're done shaping up 5.5.0.

mikehardy:
v5.5.x is already up to v5.5.4 now :-), I could merge something in but there is review to handle above on your part. Good luck fundraising! That is hectic times

erezacohen:
Lol, ok I'll do my best to handle this ASAP. Thanks :)

mikehardy:
This is still building cleanly and testing out on v5.x.x, if the API had default args following the linked pattern and branched in JS instead of native, I'd merge and release

mancas:
Do we have any update on this?

Salakar:
I'm going to close this PR for now as stale, this functionality exists on v6 also.

sergeymorkovkin:
Is there any progress on this one year later? There is no clarity on whether messaging and crashlytics support secondary apps in 2021. A lot of contradicting information.

sergeymorkovkin:
Answered [here](https://stackoverflow.com/questions/62253691/firebasemessaging-getinstancefirebaseapp-for-secondary-app-supposed-to-be-publ).

Avishayy:
@mikehardy hi again

Secondary apps are supported but not for messaging and crashalytics [due to Firebase's team decision](https://stackoverflow.com/a/62710006/2135320), which is a bummer. Unfortunately I have to use a third party library that also registers a messaging app and both mine and the third app use the messaging module for various reasons (and cannot be merged). Is `react-native-firebase` going to be accepting of a second iteration implementation of this PR (for 0.6, of course)?

mikehardy:
With sincere apologies, we have enough work simply supporting the API surface area that firebase-ios-sdk/firebase-android-sdk present, we don't really have the capacity to go beyond that, even if philosophically it was something we would do - and it's worth mentioning that philosophically, we don't do that in general (or in specific...I can't think of examples).

I can say this: our native code moves very slowly. We are not in the habit of altering it for no reason, and even when altered it's typically a surgical change. There are zero plans to rewrite native stuff in any way at the moment. We do have plans to adopt the v9 firebase-js-sdk APIs and adopt typescript at the JS layer but even though those changes are dramatic code-wise they are probably not too hard to deal with. I mention all this to say that while in general as a developer using react-native-firebase in my own apps I don't like the idea of maintaining a fork or using patch-package, in my opinion, it's a contender as a solution for you.

Avishayy:
I see Mike, thank you.

I'm putting our fork here in case anyone needs to support this in their own app:
https://github.com/knockapp/react-native-firebase

Some notes for whoever sees this:
1. Before being able to use the secondary app, you need to initialize it through the `app` package (`app.initializeApp()`)
2. In regards to the fork, in addition to the changes to the native code itself – there's also a small hack needed to create `lib/version.js` since lerna publishes the version files to the registry but these are not git tracked ([in fact, they are ignored](https://github.com/invertase/react-native-firebase/blob/main/.gitignore#L543)] and as we are using a git repo and not a registry, this is needed
3. [yarn / npm will break if you use a lerna repo as the git url](https://github.com/lerna/lerna/issues/2074), i.e. this will not work:
```
"@react-native-firebase/messaging": "git+ssh://git@github.com:knockapp/react-native-firebase.git#main",
```
the git repo needs to point to the actual package (`packages/messaging` in our case) but maintaining a fork that holds only a subfolder of the whole original repo is much more of a hassle. A solution we found is using [gitpkg](https://gitpkg.vercel.app/) which returns a url for a subfolder of the original repo, so it will look like this now:
```
"@react-native-firebase/messaging": "https://gitpkg.now.sh/knockapp/react-native-firebase/packages/messaging?main",
```

Hope this helps for anyone!

PR #1872, opened 2019-01-31
Cross-env to fix Windows build

SebastienGllmt:
### Summary

Simple fix for env-variable to allow building on Windows. This is a common solution and you can see this in many other repos. [Example here](https://github.com/prescottprue/react-redux-firebase/blob/master/package.json#L19)

### Release Plan

[JS][BUGFIX] [package.json] - Fix build on Windows (BABEL_ENV error)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2487,6 +2487,16 @@
         "object-assign": "^4.1.1"
       }
     },
+    "cross-env": {
+      "version": "5.2.0",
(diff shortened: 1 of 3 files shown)


Salakar:
Hey @SebastienGllmt, thanks for this.

Any chance you can send this over to the v5.x.x branch instead, master is now the WIP v6.x.x release sorry 🙈

SebastienGllmt:
Okay. Fixed this in #1879

PR #1974, opened 2019-03-14
Remove Expo warning

brentvatne:
I submitted this to `master` rather than `5.x.x` as suggested by the PR template because it looks like 6.x is in development on master and this would apply to the latest version, if this is wrong I will adjust base, let me know. Alternatively it could be merged here and cherrypicked onto `5.x.x` if it even makes sense for it to be there.

### Summary

ExpoKit is being deprecated in favour of using Expo APIs like other React Native packages (as of [this change](https://blog.expo.io/you-can-now-use-expo-apis-in-any-react-native-app-7c3a93041331)). The current warning message would be visible in apps that use `@unimodules/core`. Unlike ExpoKit, using Unimodules is pretty lightweight and should not impact your work as react-native-firebase maintainers.

### Checklist

- [x] Supports `Android`
- [x] Supports `iOS`
- Other stuff doesn't apply

### Test Plan

Run an Android app.

### Release Plan

[ANDROID][ENHANCEMENT] - Remove no longer valid warning about Expo


brentvatne:
CI failures seem unrelated

Salakar:
@brentvatne thanks for flagging this, you're right about this logic breaking in `unimodules` - it didn't exist at the time 🙈 happy to remove the build gradle warning - this warning is only on master (unreleased) and not on v5.

The warning was aimed at a stop-gap measure until I could get the time to go through usage in ExpoKit (there were some reports of React Native Firebase not finding native modules e.g. `NativeModules.X` not existing and I've not had the time to triage). 

I've reverted the changes in `SharedUtils.java` as these are unrelated to the build warning and were also failing CI. They are leftovers from early attempts/experiments ([source](https://github.com/Salakar/react-native-firebase-private/tree/mono-refactor/packages/library/core/android/src/main/java/io/invertase/firebase/common)) to make React Native Firebase compatible with Expo's ModuleRegistry/Expo itself; I'd like to revisit that in future so would like to keep those there (though the class may need changing); it can be used to detect the Platform to disable/enable certain features e.g. disable Firebase features that break sandbox on Expo.

Salakar:
As per the Contributing guide [ownership section](https://github.com/invertase/react-native-firebase/blob/master/CONTRIBUTING.md#ownership) you've now been added as a collaborator - I'm sure you're aware but; you're not obligated to accept or contribute more than you normally would. 👍

Issue #2071, opened 2019-04-21
Analytics docs don't mention how to enable automatic audience metrics on iOS


mikehardy:
## Documentation Feedback

If I am not mistaken, on iOS you need to link in a separate Pod on order to get the benefit of automatic collection of audience metrics.

https://stackoverflow.com/questions/54437891/firebase-linking-to-libadidaccess-a

https://support.google.com/firebase/answer/6318039

```
For iOS apps, the app must collect IDFA in order to automatically derive the Age,
Gender, and Interests properties.
```

I believe the documentation update should be something along the lines of 

```
If you would like to enable Firebase Analytics to generate automatic audience 
metrics for iOS, you must also link in the IDFA support libraries, available via 
the CocoaPod GoogleIDFASupport
```

Seems like it should be here with some links out to supporting docs:
https://rnfirebase.io/docs/v5.x.x/analytics/ios

Happy to PR this if it seems sensible.

---

Think `react-native-firebase` is great? Please consider supporting the project with any of the below:

- 👉 Donate via [Open Collective](https://opencollective.com/react-native-firebase/donate)
- 👉 Follow [`React Native Firebase`](https://twitter.com/rnfirebase) and [`Invertase`](https://twitter.com/invertaseio) on Twitter
- 👉 Star this repo on GitHub ⭐️
- 👉 Contribute; see our [contributing guide](./../../CONTRIBUTING.md)

Ehesp:
@Salakar how does this apply to v6?

mikehardy:
for me - I'd take whatever Podfile you publish as an example and add a commented out line for GoogleIDFASupport (currently version 3.14.0) with a note it is optional for iOS audience measurement, along with a note in the analytics docs since analytics otherwise has no setup. But I'm just thinking out loud on what I actually did, and what would have prompted me to do it

Salakar:
@Ehesp for v6 it'd be as @mikehardy mentioned, we can put this on `/oss/react-native-firebase/v6/analytics/ios` as an optional step?

@mikehardy if you're willing to send a docs PR up for this; this is the place for it to be added to for the new site: https://github.com/invertase/react-native-firebase/blob/master/docs/analytics/ios.md

mikehardy:
that pointer to the spot for the new-new is helpful - I'm on it. thanks

mikehardy:
Opened #2082 for v6 and https://github.com/invertase/react-native-firebase-docs/pull/180 for v5.x

[...]

PR #2082, opened 2019-04-26
docs: noted iOS pod needed for iOS audience stats

mikehardy:
Per discussion on #2071 - added a note about the cocoapod required for automatic audience metric collection on iOS

I will send a separate one to v5

I am skipping the template, this is a docs change as discussed already, please don't shoot me :pray:


Salakar:
The template is for v5, needs a full rework for v6 ha 😭 so no worries there!

LGTM, will merge - thanks for this 🙏

PR #2123, opened 2019-05-08
[v6][DOCS] Fix quick-start.md for ml-kit

gianpaj:
https://invertase.io/oss/react-native-firebase/v6/ml-kit/quick-start


Salakar:
Good spot, thanks so much for sending a PR, appreciate it 🎉

Issue #2126, opened 2019-05-09
[v6] 🔥Unable to install as per quick start guide


matthieunelmes:
## Issue

Describe your issue here

As per docs tried running:
`npx @react-native-community/cli@next init --template=@react-native-firebase/template@alpha <myProjectName>`

Install does not complete, error:
`error Error: Cannot find module '/var/folders/s7/v2pwtdc96fzbv2s8j10mf3340000gn/T/rncli-init-template-k19y8J/node_modules/@react-native-firebase/template@alpha/template.config'`

---

---

Think `react-native-firebase` is great? Please consider supporting all of the project maintainers and contributors by donating via our [Open Collective](https://opencollective.com/react-native-firebase/donate) where all contributors can submit expenses. [[Learn More]](https://invertase.io/oss/react-native-firebase/contributing/donations-expenses)

- 👉 Check out [`React Native Firebase`](https://twitter.com/rnfirebase) and [`Invertase`](https://twitter.com/invertaseio) on Twitter for updates on the library.

Salakar:
cc @thymikee - think this is a bug on the CLI with templates, the path to the template.config file being loaded has the NPM tag `@alpha` tag in the path - should be stripped out I'm guessing. Must be a recent bug - was working a couple of days ago.

NdauwaRafael:
I have the same issue

thymikee:
@Salakar we didn't have a release in ~20 days. This is a bug indeed, fixed in https://github.com/react-native-community/cli/pull/364. We'll have a release shortly, right after https://github.com/react-native-community/cli/pull/372 is merged (I'm waiting for someone to review it once again, help appreciated 😃)

thymikee:
Published `@react-native-community/cli@2.0.0-alpha.18` with the fix

matthieunelmes:
Thanks for jumping on this, look forward to trying it out

matthieunelmes:
Sorry guys, no dice. 

Can successfully install via the template, trying to follow docs and have installed .plist into the Xcode project but it's failing to build. Have tried auto-linking, manual linking, installing via pods, building via workspace but nothing is working. Keep getting:

`'Firebase/Firebase.h' file not found`

Not really the quick start I had in mind 😕and the docs fell very sporadic. I would love to contribute to the docs if only I knew the correct process to get a working build.

mikehardy:
@matthieunelmes please post your own issue, we have no idea what your setup is, and your error seems to be unrelated

[...]

PR #2140, opened 2019-05-17
docs: note the v6 template init java package / ios bundle names

mikehardy:
This is the smallest PR ever, so I am rudely skipping your well-constructed PR template :-)

That said, you may still think the proposed wording chunk I added is terrible and reject, fine by me. Might be interesting to note what the ios bundle is called as well but I didn't check (sorry)

### DIFF
--- a/pages/quick-start/new-project.md
+++ b/pages/quick-start/new-project.md
@@ -21,7 +21,7 @@ npx @react-native-community/cli@next init --template=@react-native-firebase/temp
 
 > Substitute `<myProjectName>` for the name of your new project.
 
-This will create a new React Native project in a folder called `<myProjectName>` with the `@react-react-native/app` package pre installed.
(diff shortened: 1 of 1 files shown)


Salakar:
Thanks for this @mikehardy - this is a good change so don't worry - best to be clear in the docs that the package name is based on the specified `<projectName>`

iOS is the same, so if you'd like to update this PR to mention that also that would be cool :) 

PR has 2 approvals anyway if you just want to go ahead and squash merge yourself, thanks.

mikehardy:
Great - I'll wait till the #2126 thing is sorted out so I know I can actually do the thing and look at iOS, then I'll add the ios bit and hit the merge button

mikehardy:
In XCode, it comes out like so: `org.reactjs.native.example.firebaseanalyzertests` FWIW

mikehardy:
Okay, I triple-checked that my edits look good afterwards, going to squash and merge. Blame me later if it's terrible :-)

PR #2184, opened 2019-06-01
docs(crashlytics): iOS manual integration steps

jojonarte:
Update ios.md for steps in integrating crashlytics to iOS application

### Summary

I was working on integrating Crashlytics manually on react-native specifically iOS, I've found that there is no documentation yet, so I made this PR in response.

### Checklist
- [x] Supports `iOS`


docs/crashlytics/ios.md
```diff
@@ -4,8 +4,34 @@ description: Manually integrate Crashlytics into your iOS application.
...
 
 # iOS Setup
-
 > The following steps are only required if your environment does not have access to React Native
 auto-linking. 
 
-## TODO
+## Update Podfile
+
+```
```
Salakar:
```suggestion
```ruby{9-10}
```

This will highlight lines 9 & 10 in the code block when on the docs website

docs/crashlytics/ios.md
```diff
@@ -4,8 +4,34 @@ description: Manually integrate Crashlytics into your iOS application.
...
+# Uncomment the next line to define a global platform for your project
+# platform :ios, '9.0'
+
+target 'MyProject' do
+# Comment the next line if you're not using Swift and don't want to use dynamic frameworks
+use_frameworks!
+
+# Pods for PodTest
+pod 'Fabric', '~> 1.10.1'
```
Salakar: cc @ehesp - these should be using the versions config variables; not sure how they work though, can you help out here. Thanks

docs/crashlytics/ios.md
```diff
@@ -4,8 +4,34 @@ description: Manually integrate Crashlytics into your iOS application.
...
+pod 'Crashlytics', '~> 3.13.1'
```
Salakar: cc @ehesp - these should be using the versions config variables; not sure how they work though, can you help out here. Thanks

docs/crashlytics/ios.md
mikehardy:
```suggestion
pod 'Fabric', '~> {{ ios.fabric.tools }}'
```

Is how it would go in v5.x.x but I think it will need a corresponding change in https://github.com/invertase/react-native-firebase/blob/master/docs/variables.yaml for it to work? It doesn't seem to have the majority of variables in there and I don't see corresponding files per package...

docs/crashlytics/ios.md
mikehardy:
```suggestion
pod 'Crashlytics', '~> {{ ios.firebase.crashlytics }}'
```

Is how it would go in v5.x.x but I think it will need a corresponding change in https://github.com/invertase/react-native-firebase/blob/master/docs/variables.yaml for it to work? It doesn't seem to have the majority of variables in there and I don't see corresponding files per package...

jojonarte:
I updated the branch with requested changes.

PR #2200, opened 2019-06-04
[tests] add support for new ci

Salakar:
-


tests/package.json
```diff
@@ -25,7 +25,7 @@
     "detox": "^9.0.4",
     "fbjs": "^0.8.16",
     "firebase-admin": "^5.12.0",
-    "jet": "https://github.com/mikehardy/jet.git#v0.3.0-dev",
+    "jet": "^0.4.2",
```
mikehardy: I'm not sure about this one - I think with jet 0.4.x you have androidx, so I worked on the 0.3.x line on purpose, it was the freshest non-androidx version. I was thinking an 0.3.1 for react-native-firebase v5, at least / until all the AndroidX stuff resolves (which it might actually, shortly)

PR #1940, opened 2019-03-02
[Crashlytics][ENHANCEMENT] v5.x.x

bthurlow:
### Summary

Wanted to help improve crashlytics js reporting so I implemented a variation of react-native-fabric's custom exception.  I also added two missing methods that are in the crashlytics api but were not included in RNFirebaseCrashlytics.

Also, I addressed #1643 in the documentation to hopefully make it more clear how to use this module.

### Checklist

- [ ✔️ ] Supports `Android`
- [ ✔️ ] Supports `iOS`
- [ ✔️ ] `e2e` tests added or updated in [/tests/e2e/\*](/tests/e2e)
- [ ✔️ ] Updated the documentation in the [docs repo](https://github.com/invertase/react-native-firebase-docs)
https://github.com/invertase/react-native-firebase-docs/pull/168
- [ ? ] Flow types updated
Pretty sure I updated those as I searched the entire project for any files related to crashlytics.
- [ ✔️ ] Typescript types updated

### Test Plan

Implemented and ran e2e tests on android.  Need help on iOS.
Android e2e results:
image

Also ran android code through a debug version of a production app successfully.
Crashlytics Examples:
image
image

### Release Plan

[GENERAL][ENHANCEMENT][CRASHLYTICS]
- Added recordCustomError method which allows you to specify your own filename, linenumber, class, and fuction as well as any additional data you which to show in the stack trace.
- Added setUserName method to allow you to set the user name to be attached to any following errors.
- Added setUserEmail method to allow you to set the user email to be attached to any following errors.

🔥 
---

Think `react-native-firebase` is great? Please consider supporting the project with any of the below:

- 👉 Donate via [Open Collective](https://opencollective.com/react-native-firebase/donate)
- 👉 Follow [`React Native Firebase`](https://twitter.com/rnfirebase) and [`Invertase`](https://twitter.com/invertaseio) on Twitter
- 👉 Star this repo on GitHub ⭐️
- 👉 Contribute; see our [contributing guide](/CONTRIBUTING.md)

PR #2237, opened 2019-06-14
[Crashlytics][ENHANCEMENT] add recordCustomError/setUserName/setUserEmail update to #1940

bthurlow:
Update to PR #1940.

Additional testing performed.
New test results.
image

### DIFF
--- a/tests/e2e/crashlytics/crashlytics.e2e.js
+++ b/tests/e2e/crashlytics/crashlytics.e2e.js
@@ -1,4 +1,4 @@
-describe.only('crashlytics()', () => {
+describe('crashlytics()', () => {
   // todo test is flakey due to a detox error occurring sometimes;
   // Error: the string "Error when sending event: websocketFailed with body:
   // {\n    id = 0;\n    message = \"The operation couldn\\U2019t be completed.
--- a/tests/e2e/crashlytics/crashlytics.e2e.js
+++ b/tests/e2e/crashlytics/crashlytics.e2e.js
@@ -1,4 +1,4 @@
-describe('crashlytics()', () => {
+describe.only('crashlytics()', () => {
   // todo test is flakey due to a detox error occurring sometimes;
   // Error: the string "Error when sending event: websocketFailed with body:
   // {\n    id = 0;\n    message = \"The operation couldn\\U2019t be completed.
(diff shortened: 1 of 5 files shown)


mikehardy:
@bthurlow it looks like your commit username is somehow not your github username so the CLA bot isn't picking up that "the contributor" (who it think is different) accepted. Not sure if you can try something to fix or if @Salakar needs to override. 

Either way, I'll pull this and see what I see, I liked the feature the first time :-), so hopefully we can get it in for real this time.

bthurlow:
@mikehardy Yeah, I messed that up on the first PR, I think my unmapped name needs to be removed from the CLAssistant. Hope it works well this time around.

tests/e2e/crashlytics/crashlytics.e2e.js
```diff
@@ -27,9 +27,14 @@ describe('crashlytics()', () => {
...
     });
 
-    xit('should error on a non a string value', async () => {
-      // TODO lib needs validations adding
-      await firebase.crashlytics().log(123456);
+    it('should error on a non string value', async () => {
+      if (device.getPlatform() !== 'ios') {
+        try {
+          await firebase.crashlytics().log(123456);
+        } catch (e) {
```
mikehardy: out of curiousity - shouldn't all of these contain some sort of 'fail' assertion right after the await in the try block, before the catch, to assert that it really did error since it should never get past the await? I've done lots of java but not much javascript testing so I might be off base, but in java testing I'd put a fail after the line that should throw, to make sure I wasn't just silently passing through everything...

bthurlow: If my memory serves correct, in JS if you don't catch the throw it will automatically fail out as an uncaught exception.

mikehardy: Nothing else jumped out at me - I'll read again and check locally for android e2e, I asked Salakar about the CLA thing (he has the admin oomph to fix that)

mikehardy: > If my memory serves correct, in JS if you don't catch the throw it will automatically fail out as an uncaught exception.

Hmm - but what happens in these structures as they are (with proper try/catch) if the method call doesn't throw? wouldn't it silently pass even though we expect an exception? (at least for the ones where we are purposely sending in junk, expecting it to throw)

bthurlow: I see where you're going with that, that's why it took me so long to rewrite the tests, I had to make sure on the forced errors that I was actually passing the test on the error call and not on the junk args.

I had a ton of console.log calls to verify everything going in and coming out when I was reworking the test suite.

mikehardy: Okay - that makes sense but does seem like a Java / Javascript difference I'll have to get over personally ;-). Ok, I'll check this via android / e2e here and re-read etc then ideally merge it in. 💪

mikehardy:
These passed locally for me :muscle: I'm willing to try again

mikehardy:
possibly worth mentioning, we have Android enabled in CI now so even though I successfully tested this locally just in case, CI will no longer let Android things slip through

PR #2247, opened 2019-06-18
[BREAKING CHANGE][TYPES][Firestore] Document id can't be null

iwikal:
### Summary

I looked in all the official firebase SDKs for different platforms and they all say that ID is guarranteed to be as string. Also, it seems that when the typescript definitions were added, they followed the flow types which also said id could be null, but the flow types have since changed to correctly reflect the official SDKs, while the TS types remained as they were.

Links to SDKs:
https://firebase.google.com/docs/reference/js/firebase.firestore.DocumentReference.html#id
https://firebase.google.com/docs/reference/js/firebase.firestore.DocumentSnapshot.html#id
https://firebase.google.com/docs/reference/swift/firebasefirestore/api/reference/Classes/DocumentReference#/c:objc(cs)FIRDocumentReference(py)documentID
https://firebase.google.com/docs/reference/swift/firebasefirestore/api/reference/Classes/DocumentSnapshot#/c:objc(cs)FIRDocumentSnapshot(py)documentID
https://firebase.google.com/docs/reference/android/com/google/firebase/firestore/DocumentReference.html#getId()
https://firebase.google.com/docs/reference/android/com/google/firebase/firestore/DocumentSnapshot.html#getId()

### Checklist

- [x] Supports `Android`
- [x] Supports `iOS`
  - **LINK TO DOCS PR HERE**
- [ ] Flow types updated

[TYPES][BUGFIX][Firestore] - Document id is no longer possibly null in typescript

:fire:

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -2437,7 +2437,7 @@ declare module 'react-native-firebase' {
 
       interface DocumentReference {
         readonly firestore: Firestore;
-        readonly id: string | null;
+        readonly id: string;
         readonly parent: CollectionReference;
         readonly path: string;
 
@@ -2526,7 +2526,7 @@ declare module 'react-native-firebase' {
 
       interface DocumentSnapshot {
         readonly exists: boolean;
-        readonly id: string | null;
+        readonly id: string;
         readonly metadata: Types.SnapshotMetadata;
         readonly ref: DocumentReference;
 


mikehardy:
Thank you!

Please provide a link to the firebase SDK doc so I can use that to tell others, but otherwise this seems fine. It's a breaking change to remove null from a typing but we just created "semver space" to do that for v5 maintenance, so this can go into 5.5.0

mikehardy:
Oh - and could you check master (v6)? I believe in packages/firestore/ this may have already been migrated and I'm not sure if the type there is correct

iwikal:
Pull request description updated!
I looked and firestore v6 seemed to lack typings, I grepped through the entire repo and didn't find any matches for `interface DocumentReference`, and https://github.com/invertase/react-native-firebase/blob/master/packages/firestore/lib/index.d.ts seems to mostly contain boilerplate that I don't really understand. Also, the firestore package is missing from https://github.com/invertase/react-native-firebase/blob/master/CHANGELOG.md so it seems that it's not yet migrated.

mikehardy:
Thanks again! Ah, might be so - regarding v6, thanks for checking. I'll check with the other guys about it. Thanks for the document links, I just wanted to make very very sure and needed to see it :-). v5.5 is already mostly done, I should be able to release this week. In the meantime feel free to use this change via patch-package as it will come through in the same shape on release

PR #2256, opened 2019-06-19
[IOS][BUGFIX][CRASHLYTICS] - fix crashlytics pod integration and compile issue

maxkomarychev:
- include `Fabric` and `Crashlytics` as dependencies of a subspec
`Crahslytics` so that includes of <Crashlytics/Crashlytics.h> and
<Fabric/Fabric.h> can properly be resolved by native module
- fix syntax errors in `RNFirebaseCrashlytics.m`

### Summary

In current setup with cocoapods checks for include in crashlytics-related files will not work and attempt to call `firebase.crahslytics()` in the code will lead to error "You attempted to use a firebase module that's not installed natively on your iOS project....".

### Checklist

- [ ] Supports `Android`
- [x] Supports `iOS`
- [ ] `e2e` tests added or updated in [/tests/e2e/\*](/tests/e2e)
- [ ] Updated the documentation in the [docs repo](https://github.com/invertase/react-native-firebase-docs)
  - **LINK TO DOCS PR HERE**
- [ ] Flow types updated
- [ ] Typescript types updated

### Release Plan

[CATEGORY][type] [LOCATION] - Message

[IOS][BUGFIX][CRASHLYTICS] - fix crashlytics integration via cocoapods

---

Think `react-native-firebase` is great? Please consider supporting the project with any of the below:

- 👉 Donate via [Open Collective](https://opencollective.com/react-native-firebase/donate)
- 👉 Follow [`React Native Firebase`](https://twitter.com/rnfirebase) and [`Invertase`](https://twitter.com/invertaseio) on Twitter
- 👉 Star this repo on GitHub ⭐️
- 👉 Contribute; see our [contributing guide](/CONTRIBUTING.md)

PR #2051, opened 2019-04-11
[links] Add ability to specify domainUriPrefix instead of dynamicLinkDomain

JYeop:
Made a change for the custom links.
If you make a link with customization(eg>abc.com/link), package throws an error.

> Error: Use setDomainUriPrefix() instead, setDynamicLinkDomain() is only applicable for *.page.link and *.app.goo.gl domains.

So, I changed the v5.3.1 code just one line.
One thing to consider, setDomainUriPrefix must provided with 'https://' or 'http://'.
```js
const link = new firebase.links.DynamicLink(
          `https://abc.com/link/?invitedby=${String(Name)},
           "abc.com/link" // Wrong
          "https://abc.com/link" // Right
```
 

### Checklist

- [x] Supports `Android`
- [ ] Supports `iOS`
- [ ] `e2e` tests added or updated in [/tests/e2e/\*](/tests/e2e)
- [ ] Updated the documentation in the [docs repo](https://github.com/invertase/react-native-firebase-docs)
  - **LINK TO DOCS PR HERE**
- [ ] Flow types updated
- [ ] Typescript types updated

### Release Plan

[CATEGORY][type] [LOCATION] - Message

<!--
  **INTERNAL tagged notes will not be included in the next version's release notes.**

    CATEGORY
  [----------]      TYPE
  [ TYPES    ] [-------------]       LOCATION
  [ JS       ] [ BREAKING    ] [------------------]
  [ GENERAL  ] [ BUGFIX      ] [ {FirebaseModule} ]
  [ INTERNAL ] [ ENHANCEMENT ] [ {Filename}       ]
  [ IOS      ] [ FEATURE     ] [ {Directory}      ]   |-----------|
  [ ANDROID  ] [ MINOR       ] [ {Framework}      ] - | {Message} |
  [----------] [-------------] [------------------]   |-----------|

 EXAMPLES:

[ANDROID][ENHANCEMENT][android/src/main/java/io/invertase/firebase/links/RNFirebaseLinks.java // line 222] - Modified code for enabling custom domain.
---

PR #2258, opened 2019-06-19
[BUGFIX] - patch releases to 5.5.3

mikehardy:
The 5.5.0 release had a crashlytics compile bug that slipped through because of a combination of problems allowing that chunk of code to not even attempt compilation in CI

The 5.5.1 release includes a PR #2256 from @maxkomarychev that fixed the compile issue *and* (important for the future) fixed the podfile issue that allowed it to slip through and re-enabled v5 crashlytics for iOS in CI

The 5.5.0 and 5.5.1 release still had a compile error for Analytics #2259 - I'm still combing through release docs on the pods but analytics isn't in CI (:scream:) which is the same way the crashlytics error slipped through. I'll work on that later

And now 5.5.3 fixes issue #2262 with the new domainURIPrefix links parameter - I didn't plumb that breaking change correctly, this one isn't in my app yet so I couldn't see it via integration tests, and there's no v5 links e2e support so it slipped past me as I applied the change from #2051 during the SDK update 

build errors are sort of "stop work" important so I released 5.5.1 and 5.5.2 and this is the versioning change that goes with it via [v5 the release process](https://github.com/invertase/react-native-firebase/blob/v5.x.x/RELEASE_PROCESS.md) we're currently using

When this is approved I need someone with git-power to create the v5.5.1,2,3 tags. Release notes will follow after v5.5.x is proven successfully integrated - I'm prioritizing support at the moment

Action items that are doable - check if I can get Analytics and Links at least in the compilation for e2e. Doing more doesn't make sense with v6 covering them fully but it'll put some more guardrails around the process

---

### DIFF
--- a/RELEASE_PROCESS.md
+++ b/RELEASE_PROCESS.md
@@ -14,7 +14,7 @@ This document is an attempt to write down all the steps in the release process f
 
 ## SDK Version numbers
 
-1. Update the firebase registration version numbers to match the next version number
+1. Update and commit the firebase registration version numbers to match the next version number
    - iOS here: <https://github.com/invertase/react-native-firebase/blob/v5.x.x/ios/RNFirebase/RNFirebase.m#L20>
    - Android here: <https://github.com/invertase/react-native-firebase/blob/v5.x.x/android/src/main/java/io/invertase/firebase/ReactNativeFirebaseAppRegistrar.java#L36>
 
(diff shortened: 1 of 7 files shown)


mikehardy:
Unfortunately something is wrong with CI right now - watchman is having permissions errors and this is usually file handles running out and CI needing a reboot or similar if I recall? 5.5.2 successfully integrates for me though I have 5.5.x with pods 6.current working successfully in my full app now

mikehardy:
I've finished reviewing the iOS release notes and I think these 2 were the only breaking changes from pods 5.x.x to pods 6.current where the v5 codebase had not already moved away from deprecated things that were then removed

I have not attempted to handle any of the new deprecation added though, so v5 has technical debt now. That should be fine with v6 coming

Still the CI errors though - I can't fix that sorry

PR #2235, opened 2019-06-14
Allows react-native-firebase to be used via Podfile with use_frameworks!

MoOx:
Ref https://github.com/invertase/react-native-firebase/issues/252#issuecomment-501692286

### Checklist

- [ ] Supports `Android`
- [x] Supports `iOS`
- [ ] `e2e` tests added or updated in [/tests/e2e/\*](/tests/e2e)
- [ ] Updated the documentation in the [docs repo](https://github.com/invertase/react-native-firebase-docs)
- [ ] Flow types updated
- [ ] Typescript types updated

### Test Plan

I tested this  like that:

- installed rnfirebase v5 via yarn
- modified podspec in `node_modules/react-native-firebase/ios` after install (as in the pr)
- run `pod install` (cocoapods read the podspec)
- I can build my app without the 3 tricks mentionned in https://github.com/invertase/react-native-firebase/issues/252#issuecomment-501692286

### Release Plan

[IOS][BUGFIX] [ios/RNFirebase.podspec] - Allows react-native-firebase to be used via Podfile with use_frameworks!

---

Think `react-native-firebase` is great? Please consider supporting the project with any of the below:

- 👉 Donate via [Open Collective](https://opencollective.com/react-native-firebase/donate)
- 👉 Follow [`React Native Firebase`](https://twitter.com/rnfirebase) and [`Invertase`](https://twitter.com/invertaseio) on Twitter
- 👉 Star this repo on GitHub ⭐️
- 👉 Contribute; see our [contributing guide](/CONTRIBUTING.md)

Issue #252, opened 2017-07-14
RNFirebase doesn't compile with use_frameworks!


paulftw:
I really love using your library but unfortunately iOS compilation breaks when I add `use_frameworks!` to the Podfile.

I was able to combine RNFirebase and [FirebaseUI-iOS](https://github.com/firebase/FirebaseUI-iOS). Phone Authentication in FirebaseUI triggered `firebase.auth().onAuthStateChanged` with correct user Id, which felt great.

However, in order to add "Login with Facebook" I have to enable `use_frameworks!`. Otherwise FirebaseUI won't build - it's a [known issue with FirebaseUI/FBSDKLoginKit combo](https://github.com/firebase/FirebaseUI-iOS/issues/194).
Enabling use_frameworks breaks RNFirebase.

***Quickest way to reproduce*** the issue is to run this script, it will create and build a RN project:
```
cd $(mktemp -d)
pwd > YOU_ARE_HERE.txt
echo FooBar in created on $(date). Working dir >> ~/my_foobars.txt
echo $(pwd) >> ~/my_foobars.txt

open .
react-native init FooBar
cd FooBar
react-native install react-native-firebase
cat >ios/Podfile <<EOF
platform :ios, '9.0'

# comment out the next line to make RNFirebase great again.
use_frameworks!

react_native_path = '../node_modules/react-native'
target 'FooBar' do
  pod 'Yoga', :path => react_native_path + '/ReactCommon/yoga'
  pod 'React', :path => react_native_path, :subspecs => [
    'BatchedBridge',
    'Core',
  ]

  pod 'FirebaseAnalytics'
  pod 'Firebase/Core'
  pod 'Firebase/AdMob'
  pod 'Firebase/Auth'
  pod 'Firebase/Analytics'
  pod 'Firebase/Crash'
  pod 'Firebase/Database'
  pod 'Firebase/DynamicLinks'
  pod 'Firebase/Messaging'
  pod 'Firebase/RemoteConfig'
  pod 'Firebase/Storage'
  pod 'FirebaseUI/Auth'
  pod 'FirebaseUI/Phone'

  pod 'RNFirebase', :path => '../node_modules/react-native-firebase'
end
EOF

cd ios && pod install && cd .. && react-native run-ios
```

### Error:
```
[...]

PR #2328, opened 2019-07-06
[v5] Remove recursive header search from ios podspec

IljaDaderko:
### Summary
I believe following commit introduces an issue for cocoapods users (of whom there will be many now as react-native uses cocoapods since v0.60) https://github.com/invertase/react-native-firebase/commit/b85153de6bd92e8112d75d2cb22ee9ccb9fa294e#diff-1b2b06e7c58d8929c91f917ce51a7454R28

The gist of it is that that it introduces back this issue https://github.com/firebase/firebase-ios-sdk/issues/1916
It is due to some header include in gRPC-C++ library that recursive header search will try to include, but won't find, hence causing build error.

cc @MoOx as he authored original PR, I am not sure if removing this will cause any issues, so wanted to discuss it with you.

### Checklist

- [ ] Supports `Android`
- [x] Supports `iOS`
  - **LINK TO DOCS PR HERE**
- [ ] Flow types updated

### Test Plan

I tested 5.5.4 and it failed my iOS build, reverting back to 5.5.3 solved that issue. After looking through diff and various GitHub comments only related change I found was introduction recursive header search.

There shouldn't be any issues with removing it from what I see, I am also not sure why it was added, so I cc'd in author of that PR, maybe he'll be able to explain this better.

### Release Plan

[iOS][BUGFIX] [podspec] - remove recursive header path search

---

### DIFF
--- a/ios/RNFirebase.podspec
+++ b/ios/RNFirebase.podspec
@@ -23,8 +23,4 @@ Pod::Spec.new do |s|
   end
   # allow this package to be used with use_frameworks!
   s.static_framework = true
-  # fix recursive header flag being skipped by cocoapods when using this as a framework
-  s.xcconfig = {
-    'HEADER_SEARCH_PATHS' => '${PODS_ROOT}/Headers/Public/**'
-  }
 end


mikehardy:
I in turn am curious what about your build broke that was not detected in CI? Can you determine what is different between your project and the tests ios project in the v5.x.x branch?

I'm inclined to merge this if it fixes a build break even if this in turn breaks use_frameworks as I believe frameworks broke in React-native 0.60 anyway with a patch not planned till. 61

But I want to understand and guard against it in the future

IljaDaderko:
I'll need to clone the repo and open up ios project to inspect it, first two things that come to mind however

1. Test project is not based on cocoapods (this is not the case, so we can cross this out)
2. Test project doesn't use new XCode 10 build system?
3. There is some other extra configuration for things like header search paths or something else (my project uses xcode defaults for basically all these areas)

I'll investigate this further once I can clone ios project and open it in xcode 👍

mikehardy:
I think ítem 2 at least, thanks for the investigation! Ios build is a consistent trouble area for new (and existing!) users

IljaDaderko:
In theory we can specify a more specific header search path (I'm just not sure what we need it for, is it just rnfb ones)? This way it won't look up gRPC-C++ ones or any other libs that might cause similar issues.

MoOx:
> cc @MoOx as he authored original PR, I am not sure if removing this will cause any issues, so wanted to discuss it with you.

This should make this build impossible when using cocoapods

I am curious why you did get issues... I guess you are using cocoapods? Are you using the latest pod version of firebase?

I am currently having this in my podfile

```ruby
source 'https://github.com/CocoaPods/Specs.git'

platform :ios, '9.0'
swift_version = '4.2'

target 'ADM' do
  # required for another (not shown) private dep
  use_frameworks!

  # React-Native
  pod 'React', :path => '../node_modules/react-native', :subspecs => [
    'Core',
    'CxxBridge',
    'DevSupport',
    'RCTActionSheet',
    'RCTAnimation',
    'RCTBlob',
    'RCTGeolocation',
    'RCTImage',
    'RCTLinkingIOS',
    'RCTNetwork',
    'RCTSettings',
    'RCTText',
    'RCTVibration',
    'RCTWebSocket',
  ]
  # React-Native dependencies
  pod 'yoga', :path => '../node_modules/react-native/ReactCommon/yoga'

  # Firebase
  pod 'Firebase', '~> 5.20.0'
  pod 'Firebase/Core', '~> 5.20.0'
  pod 'Firebase/Analytics', '~> 5.20.0'
  pod 'Firebase/DynamicLinks', '~> 5.20.0'
  pod 'Firebase/Messaging', '~> 5.20.0'

  # React-Native libraries
  pod 'RNFirebase', :path => '../node_modules/react-native-firebase/iOS'
  #...
end
```

IljaDaderko:
I think there are 2 differences we have

1. I'm on version `6.2.0` for firebase pods
2. I believe your project uses Swift and has to specify `use_frameworks` flag in Podfile?

Basically, the issue I'm getting is due to one of firebase pods relying on `gRPC-C++` library. This library includes some header that doesn't exist within it. Internally `gRPC-C++` disables search for this header include, so it works fine for them, but as soon as you have recursive header search path in parent project (i.e. my project and yours) there is a build error as now parent project is trying to include that non existent header from `gRPC-C++`.

We can probably keep recursive header search paths, but is there a way to make them search only header files within react native firebase code?

MoOx:
Yes, you can dynamically readjust this like in my previous hack to make react-native-firebase works with podfile.

See https://github.com/invertase/react-native-firebase/issues/252#issuecomment-501692286

You should be able to do something like

```
post_install do |installer|
  rnfirebase = installer.pods_project.targets.find { |target| target.name == 'RNFirebase' }
  rnfirebase.build_configurations.each do |config|
    config.build_settings['HEADER_SEARCH_PATHS'] = '$(inherited) ${PODS_ROOT}/Headers/Public/*'
  end
end
```

IljaDaderko:
@MoOx I assume this workaround needs to be included in my Podfile? If so it could be an issue since rn 0.60 will require everyone to include this in their pod files in order to make rnfb work.

Can you think of any other way to make `use_frameworks!` work, but without adding recursive header path search to user projects / requiring them to include workaround into Podfile?

I think those not familiar with iOS / CocoaPods will struggle with both ^

MoOx:
> I assume this workaround needs to be included in my Podfile?

Yes

>  If so it could be an issue since rn 0.60 will require everyone to include this in their pod files in order to make rnfb work.

Why? I am using podfile for all my deps before 0.60 & I am not having any issue, I don't see why this would cause issue.

And I am not sure this has to do something with use_framework. 

You could ask: so why use_framework is involved here?

It has to do with me wanting to move all my RN deps in my podfile (to avoid having a huge xcodeproj file) & before my trick/my change to RNFirebase, it was just impossible.

IljaDaderko:
> I am using podfile for all my deps before 0.60 & I am not having any issue

Oh no, it doesn't have to be 0.60 specifically, I only mentioned this version since majority of users will start using cocoapods starting with it.

There are no issues with v5.5.3 of react-native-firebase and nothing specifically related to `use_frameworks!`, it's the addition of recursive header search paths to .podspec (that this PR removes) that causes issues in v5.5.4.

The only differences I can think of between our project that break it for me, but make it work for you are

1. Do you use legacy build system in XCode, or new one introduced in XCode 10?
2. Did you add any extra tweaks to your project config, particularly around header search paths?
3. Perhaps this issue only pops up once v6 firebase pods are used, any chance you can test that?
4. Could be build cache related, can you clean your build folder / Podfile.lock and install all of these fresh with react-native-firebase v5.5.4

MoOx:
> There are no issues with v5.5.3 of react-native-firebase and nothing specifically related to use_frameworks!, it's the addition of recursive header search paths to .podspec (that this PR removes) that causes issues in v5.5.4.

Did you where using RNFirebase from podfile with 5.5.3 ?

My previous PR incorporate 2 distinct fixes:
- one for use_framework compatibility: https://github.com/invertase/react-native-firebase/pull/2235/files#diff-1b2b06e7c58d8929c91f917ce51a7454R25
- one to avoid header search path issue https://github.com/invertase/react-native-firebase/pull/2235/files#diff-1b2b06e7c58d8929c91f917ce51a7454R27 (and it seems it was a [know issue](https://github.com/invertase/react-native-firebase/issues/252#issuecomment-501695233)

To answer your question

1. No

image

2. No. I have currently just one thing that disable bitcode for some others (private) deps that have nothing to do with firebase

3. I use Firebase 5.20 & I currently don't have time to upgrade to 6.x as I have already spent 3 weeks for my clients to upgrade a codebase to RN 0.52 to 0.59 with all deps (including this one)

4. I always clean my cache, & while I made my trick/fix I probably cleaned/reinstalled pods like 256 times :)

IljaDaderko:
Ty for the info, I will make small repo with project that reproduces this error to look at, since I don't see anything else besides maybe firebase pod version that could be causing this header issue. Maybe it'll help to debug it further

IljaDaderko:
Alright here is the repro, this is basically fresh init of rn 0.60 and I added my firebase pods as well as react native firebase one.

Once you install all js / ios dependencies and run this (I'm on XCode 10.2.1, normal build system) you should see these errors (closer to the end of the build)

Screenshot 2019-07-08 at 12 26 28

MoOx:
Did you check with this repo that you don't have the issue with 5.5.3?

Could you check if the issue is still here with Firebase 5.20 for example?

IljaDaderko:
Made a separate branch fro 5.5.3 https://github.com/IljaDaderko/RNFBTest/tree/v5.5.3 this one builds fine. 

I haven't tried it with `5.20` pods yet (running out of time).

IljaDaderko:
```
s.xcconfig = {	
  'HEADER_SEARCH_PATHS' => '${PODS_ROOT}/Headers/Public'	
}
```

Does it need to be recursive @MoOx ? Config above works as non-recursive one.

mikehardy:
I'm just chiming in again to offer encouragement - thank you guys for really inspecting this. The iOS build has been a massive source of issues the entire time I've been watching the project and I would love to merge something that was well understood and seemed to work for everyone...

MoOx:
> Does it need to be recursive @MoOx ? Config above works as non-recursive one.

Yeah that's one of the fix as explained above. By default cocoapods add the same header search path but as non recursive.

IljaDaderko:
Dang 😕so this is a situation

1. To support Swift / `use_frameworks` projects we need recursive header search paths
2. But this breaks normal objc builds

I googled around and found this answer https://stackoverflow.com/a/3428303/911930 will adding `${SRCROOT}` solve this?

@mikehardy to address issues with iOS build that happened prior, I think majority of those will be solved by users including this lib through cocoapods as opposed to static lib i.e. it should solve errors related to XCode 10 etc... and hopefully we can figure out how to allow `use_frameworks` without breaking other builds here 👍

paolospag:
I am facing the same issue with Xcode 10.2.1, `react-native@0.60.0` and `react-native-firebase@5.5.4`.

I followed the whole discussion, but I didn't understand what the solution to adopt as a workaround.

Can someone help me understand how to overcome the problem?

IljaDaderko:
@paolospag I think easiest thing to do atm is to downgrade back to v5.5.3, I'd give more time for discussion on this PR. If we can't find a solution that works for both standard and `use_frameworks` approach, I'd suggest to remove recursive header search from podspec and add appropriate documentation around `use_frameworks` workaround

mikehardy:
Yeah, sorry if 554 isn't working well for all but I have been trying to do tiny releases with full release notes, so downgrades (or upgrades I guess) are tiny and you can wait most likel without missing anything

But hopefully an amazing solution for everyone is where it ends up 😁

Salakar:
I came across this today myself on a fresh project, I think it's best that we do revert this for now given that this is breaking the default flow for using the library.

So I'm gonna go ahead and merge this but I would still like to have this in a future release if we can figure out the quirks, so hopefully, we can keep the discussion going on this.

Salakar:
👋 this is now live in v5.5.5 - [[changelog]](https://rnfirebase.io/docs/v5.x.x/releases/v5.5.x#Changelog)

MoOx:
As this has been released, can we document how to use use_framework?

This should be enough:


---
If you have `use_framework!` in your Podfile, you might need to add in it the following code to adjust the header paths to avoid the error `'Firebase.h' file not found with <angled> include; use "quotes" instead`

```ruby
post_install do |installer|
  rnfirebase = installer.pods_project.targets.find { |target| target.name == 'RNFirebase' }
  rnfirebase.build_configurations.each do |config|
    config.build_settings['HEADER_SEARCH_PATHS'] = '$(inherited) ${PODS_ROOT}/Headers/Public/*'
  end
end
```

---

2 questions @mikehardy:

- Where to add this doc?
- I saw an iOS coworker that asked my why angled import where used. It seems it should be quotes import as angled are reserved to framework (if I understood correctly). Maybe using quotes would avoid the need for this workaround & would not create issue for non use_framework user?

mikehardy:
Each docs page has an edit button, the ios install page seems like the right spot? For the second question I wish my ios skills were strong enough to say but I am not authoritative about ios anything sorry

PR #2434, opened 2019-08-01
[natural-language] Update `suggestReplies` to accept an array of objects

Ehesp:


packages/ml-natural-language/e2e/smartReply.e2e.js
```diff
@@ -20,175 +20,187 @@ const timestamp = Date.now();
...
-      const conversation = firebase.mlKitLanguage().newSmartReplyConversation(5);
-      for (let i = 0; i < 10; i++) {
-        conversation.addLocalUserMessage(`${i}`, timestamp);
-        conversation.messages.length.should.equal(i >= 5 ? 5 : i + 1);
-        if (i >= 5) {
-          // checking of older first gets sliced off
-          conversation.messages[0].text.should.equal(`${i + 1 - 5}`);
-        }
-      }
+    it('resolves an empty if empty array if provided', async () => {
```
Salakar:
```suggestion
    it('resolves an empty array if empty array if provided', async () => {
```

packages/ml-natural-language/lib/index.js.flow
```diff
@@ -203,17 +240,22 @@ export interface Module extends ReactNativeFirebaseModule {
...
   /**
-   * Returns a new instance of SmartReplyConversation.
+   * Returns suggested replies for a conversation.
    *
    * #### Example
    *
    * ```js
-   * const conversation = firebase.mlKitLanguage().newSmartReplyConversation();
+   * const replies = await firebase.mlKitLanguage().suggestReplies([
+   *   { text: 'We should catchup some time!', },
```
Salakar:
```suggestion
   *   { text: 'Hey, it's been a while!', },
```

Or w/e - you have 2 duplicate lines, doesn't make sense.

packages/ml-natural-language/lib/index.d.ts
```diff
@@ -241,17 +164,55 @@ export namespace MLKitLanguage {
...
     /**
-     * Returns a new instance of SmartReplyConversation.
+     * Returns suggested replies for a conversation.
      *
      * #### Example
      *
      * ```js
-     * const conversation = firebase.mlKitLanguage().newSmartReplyConversation();
+     * const replies = await firebase.mlKitLanguage().suggestReplies([
+     *   { text: 'We should catchup some time!', },
```
Salakar:
```suggestion
     *   { text: "Hey, long time no speak!", },
```

Or w/e - you have 2 duplicate lines, doesn't make sense.

packages/ml-natural-language/lib/validateTextMessage.js
```diff
@@ -0,0 +1,97 @@
...
+      "'textMessage' expected an object value",
+    );
+  }
+
+  const out = {
+    timestamp: Date.now(),
+    isLocalUser: true,
+  };
+
+  if (!hasOwnProperty(textMessage, 'text')) {
```
Salakar: This isn't needed, the `isString` check below will return false if the property doesn't exist.

packages/ml-natural-language/lib/validateTextMessage.js
```diff
@@ -0,0 +1,97 @@
...
+    if (textMessage.text.length === 0) {
+      throw new Error(
+        "'textMessage.text' expected string value to not be empty",
+      );
+    }
+
+    out.text = textMessage.text;
+  }
+
+  if (hasOwnProperty(textMessage, 'timestamp')) {
```
Salakar: This isn't needed, the `isNumber` check below will return false if the property doesn't exist as it will be `undefined`.

packages/ml-natural-language/lib/validateTextMessage.js
```diff
@@ -0,0 +1,97 @@
...
+    if (!isNumber(textMessage.timestamp)) {
+      throw new Error(
+        "'textMessage.timestamp' expected number value (milliseconds)",
+      );
+    }
+
+    out.timestamp = textMessage.timestamp;
+  }
+
+  if (hasOwnProperty(textMessage, 'isLocalUser')) {
```
Salakar: This isn't needed, the `isBoolean` check below will return false if the property doesn't exist as it will be `undefined`.

packages/ml-natural-language/lib/validateTextMessage.js
```diff
@@ -0,0 +1,97 @@
...
+        "'textMessage.isLocalUser' expected boolean value",
+      );
+    }
+
+    out.isLocalUser = textMessage.isLocalUser;
+  }
+
+  if (out.isLocalUser && hasOwnProperty(textMessage, 'userId')) {
+    throw new Error(
+      "'textMessage.userId' expected 'textMessage.isLocalUser' to be false when setting a user ID.",
```
Salakar: Do we need to validate this one, the userId would get ignored anyway?

packages/ml-natural-language/lib/validateTextMessage.js
```diff
@@ -0,0 +1,97 @@
...
+    );
+  } else if (!out.isLocalUser && !hasOwnProperty(textMessage, 'userId')) {
+    throw new Error(
+      "'textMessage.userId' expected a string value",
+    );
+  } else if (!out.isLocalUser && hasOwnProperty(textMessage, 'userId')) {
```
Salakar: These 2 if statements are conflicting (76 & 80) - it will always throw?

packages/ml-natural-language/lib/index.js
```diff
@@ -89,19 +90,30 @@ class FirebaseMlKitLanguageModule extends FirebaseModule {
...
-    return new SmartReplyConversation(this.native, messageHistoryLimit);
+    if (messages.length === 0) {
+      return Promise.resolve([]);
+    }
+
+    const validated = [];
+
+    for (let i = 0; i < messages.length; i++) {
+      try {
+        validated.push(validateTextMessage(messages[i]));
```
Salakar: Can we not just loop over the original array and validate, without pushing to another array - it'll throw anyway? If you need the index you can pass it to the validate fn.

packages/ml-natural-language/lib/index.d.ts
```diff
@@ -241,17 +164,55 @@ export namespace MLKitLanguage {
...
+     *   { text: 'I know right, it has been a while..', userId: 'xxxx', isLocalUser: false },
+     *   { text: 'Hey, long time no speak!' },
```
Salakar:
```suggestion
     *   { text: 'We should catchup some time!', },
```

packages/ml-natural-language/e2e/smartReply.e2e.js
```diff
@@ -20,175 +20,181 @@ const timestamp = Date.now();
...
-        const conversation = firebase.mlKitLanguage().newSmartReplyConversation();
-        conversation.addRemoteUserMessage(text, timestamp, remoteUserId);
-
-        conversation.messages.length.should.equal(1);
-        conversation.messages[0].text.should.equal(text);
-        conversation.messages[0].timestamp.should.equal(timestamp);
-        conversation.messages[0].remoteUserId.should.equal(remoteUserId);
+    it('returns suggested replies', async () => {
+      const replies = await firebase.mlKitLanguage().suggestReplies([
+        { text: 'We should catchup some time!' },
```
Salakar: Duplicate line

packages/ml-natural-language/lib/validateTextMessage.js
Ehesp: It's optional, so it's checking whether it exists before validating.

packages/ml-natural-language/lib/validateTextMessage.js
Ehesp: Removing that will make it required field?

packages/ml-natural-language/lib/validateTextMessage.js
Ehesp: - 1: local user & user ID = error
- 2: Remote user & no user ID = error
- 3: Remote user & user ID (which it will have to go to = validate user ID. Can remove the `&& hasOwnProperty(textMessage, 'userId')` check.

Think its fine, tests pass?

packages/ml-natural-language/lib/validateTextMessage.js
Ehesp: Not sure - could make the user think they need to/gives better results if they add it.

packages/ml-natural-language/lib/index.js
Ehesp: The validate returns a new object with defaults - I can wrap the try catch around the for?

PR #2453, opened 2019-08-07
Move Firestore settings to share preferences

Ehesp:
Currently calling settings multiple times or after usage throws an error/warning. This moves it off to shared prefs and sets the latest settings the first time the app is requested.


packages/firestore/android/src/main/java/io/invertase/firebase/firestore/UniversalFirebaseFirestoreCommon.java
```diff
@@ -20,13 +20,65 @@
...
+
+    firestoreSettings.setCacheSizeBytes(cacheSizeBytes);
+    firestoreSettings.setHost(host);
+    firestoreSettings.setPersistenceEnabled(persistence);
+    firestoreSettings.setSslEnabled(ssl);
+
+    try {
+      firebaseFirestore.setFirestoreSettings(firestoreSettings.build());
+    } catch (Exception exception) {
+      throw exception;
```
Ehesp: @Salakar what to do here? On DB we check a string contains. Is this needed?

PR #2526, opened 2019-08-28
[v6] utils Logging

IljaDaderko:
Work in progress:

- [x] Adjust colours
- [x] Add logging to native events
- [x] Disable logging by default (from test app.js)
- [x] Better params formmating in info logger

I tried to make it nice ¯\_(ツ)_/¯

Screenshot 2019-09-04 at 17 13 13


packages/app/lib/index.d.ts
```diff
@@ -240,6 +240,7 @@ export namespace ReactNativeFirebase {
  */
 export namespace Utils {
   import FirebaseModule = ReactNativeFirebase.FirebaseModule;
+  type LoggerConfig = { enableMethodLogging: boolean; enableEventLogging: boolean };
```
Ehesp: This will get added to the docs so worth having a quick description above it.

mikehardy:
This is potentially interesting currently by grafting the API work and docs work onto the current RNFBDebug global toggle

PR #2531, opened 2019-08-30
Remove executable bit from non-executable files

friederbluemle:
A bunch of non-executable files in the repo erroneously had the executable bit set. This fixes is by changing the file mode from `755` to `644`.


Salakar:
Interesting find, will get this merged - how did you come across this if I may ask?

friederbluemle:
@Salakar - Thanks for merging. I noticed this because my terminal uses a different color for files that have the executable bit set. ;)

PR #2476, opened 2019-08-11
[BUGFIX][Compile][Android] - backwards-compatible AndroidX dependency

mikehardy:
I needed this in my project to successfully use this with modern (divergent) numbers for the support libraries - core is at 1.0.2 but support-v4 is at 1.0.0 and media is at 1.0.1 etc.

use 'supportV4Version' in your ext{} block in build.gradle to use it

Assuming this passes CI and is good to go, I'll update the docs repo for changelogs for a 5.5.7 release etc (and I'm traveling back from vacation right now, I could work to shape up a 5.5.7 release and release it if that helps)

This was supposed to be in #2445 but I didn't make a feature branch and clobbered the change 🤦‍♂ I am using this locally in production though

### Checklist

- [x] Supports `Android`
- [ ] Supports `iOS`
- [ ] `e2e` tests added or updated in [/tests/e2e/\*](/tests/e2e)
- [ ] Updated the documentation in the [docs repo](https://github.com/invertase/react-native-firebase-docs)
  - **LINK TO DOCS PR HERE**
- [ ] Flow types updated
- [ ] Typescript types updated

### Release Plan

[CATEGORY][type] [LOCATION] - Message

---

Think `react-native-firebase` is great? Please consider supporting the project with any of the below:

- 👉 Donate via [Open Collective](https://opencollective.com/react-native-firebase/donate)
- 👉 Follow [`React Native Firebase`](https://twitter.com/rnfirebase) and [`Invertase`](https://twitter.com/invertaseio) on Twitter
- 👉 Star this repo on GitHub ⭐️
- 👉 Contribute; see our [contributing guide](/CONTRIBUTING.md)

PR #2596, opened 2019-09-18
[v5] Android: Notifications: add support for channelId & image fields

mcuelenaere:
### Checklist

- [X] Supports `Android`
- [ ] Supports `iOS`
- [ ] Flow types updated

### Test Plan

Send a test notification to the device with the app in foreground and `console.log()` the message.

### Release Plan

[ANDROID][ENHANCEMENT] [NOTIFICATIONS] - channelId is parsed in remote notifications
[ANDROID][ENHANCEMENT] [NOTIFICATIONS] - imageUrl is parsed in remote notifications

### DIFF
--- a/android/src/main/java/io/invertase/firebase/notifications/DisplayNotificationTask.java
+++ b/android/src/main/java/io/invertase/firebase/notifications/DisplayNotificationTask.java
@@ -135,9 +135,14 @@ public class DisplayNotificationTask extends AsyncTask<Void, Void, Void> {
         }
 
         if (bigPicture.containsKey("largeIcon")) {
-          Bitmap largeIcon = getBitmap(bigPicture.getString("largeIcon"));
-          if (largeIcon != null) {
-            bp = bp.bigLargeIcon(largeIcon);
+          String largeIconStr = bigPicture.getString("largeIcon");
+          if (largeIconStr == null) {
+            bp = bp.bigLargeIcon(null);
+          } else {
+            Bitmap largeIconBitmap = getBitmap(largeIconStr);
+            if (largeIconBitmap != null) {
+              bp = bp.bigLargeIcon(largeIconBitmap);
+            }
           }
         }
 
--- a/android/src/main/java/io/invertase/firebase/notifications/RNFirebaseNotifications.java
+++ b/android/src/main/java/io/invertase/firebase/notifications/RNFirebaseNotifications.java
@@ -397,6 +397,9 @@ public class RNFirebaseNotifications extends ReactContextBaseJavaModule implemen
       androidMap.putString("group", notification.getTag());
       androidMap.putString("tag", notification.getTag());
     }
+    if (notification.getChannelId() != null) {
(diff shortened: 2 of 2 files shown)


mikehardy:
I don't believe this is necessary? 

https://github.com/invertase/react-native-firebase/blob/v5.x.x/android/src/main/java/io/invertase/firebase/notifications/DisplayNotificationTask.java#L83


https://rnfirebase.io/docs/v5.x.x/notifications/reference/AndroidNotification#channelId


Please feel free to re-open if I am wrong, but please start by opening an issue with a clear description of the problem being solved - there is no context here

mcuelenaere:
The issue here is that the following code:
```javascript
firebase.notifications().onNotification(notification => {
  console.log(notification);
});
```
will never include the Android notification channelId (`notification.android.channelId`), since the code required to populate it was absent.

As a result, if a remote notification comes in (for a specific notification channel) while the app is in the foreground, we can never show it with the correct notification channel (since the JS side is unable to obtain the proper `channelId`).

I cannot re-open this issue (lack of permissions).

mikehardy:
Ah okay, I see what you mean - and I reopened it before I just read some more. This begs the question then, in general should those two areas of code - android-specific remote-message-parsing and local-notification-construction be in sync? It looks to me like they are out of sync at the moment? @Salakar / @Ehesp - ?

mcuelenaere:
The same is true for `RemoteMessage.getImageUrl()` and `largeIcon`, although that's not a 1-on-1 mapping; it should probably be put in the `bigPicture` field.

Do you want me to do this in a separate PR?

mikehardy:
One PR is fine by me - thanks for checking on other attributes - note that we specifically have an Android CI issue right now that is backing up v5 releases so even if this were merged (I'm +1 on it, for what it's worth) you'll need to either use patch-package to pull the patch in or depend on git pinned to the v5.x.x branch or the post-merge commit hash until we get releases moving again. Not ideal, sorry

mikehardy:
I have approved, I requested a review from @Ehesp and/or @Salakar  just in case, and CI will fail on ci.android.test no matter what at the moment, whether this change would pass or not.

mikehardy:
Okay, I'm going to merge this in now that there has been a second set of eyes - sorry I misunderstood at first but thank  you for being persistent! :-)

You can depend on the commit hash from git in your package.json until we get this released and you'll be guaranteed to get reproducible builds that way. We'll hopefully get v5 android e2e tests sorted out in CI soon so we can release a 5.5.7 with this in it

mikehardy:
Or actually - I can't override the required status but @Salakar maybe this and the compile fix https://github.com/invertase/react-native-firebase/pull/2476 could be merged by you regardless? I'm using #2476 daily and this one seems fine also - I'll handle fallout if not

Salakar:
Done 💥  

😄

mikehardy:
Thanks man - I'll try to finish up v5 RN60 upgrade but my work project has some nifty iOS issue right now that's fully consuming.

PR #2604, opened 2019-09-20
feat(firestore): add QueryDocumentSnapshot interface

iwikal:
This library is amazing, kudos for making it! I had some thoughts on the firestore typescript declarations, and hope there's room for this change before v6 is released.

### Summary

Most of the time when I get data from firestore, I get a whole query with many documents, all guaranteed to exist. Since I know they should exist, I assert data to be an object, like so:
```ts
const collectionSnap = await firestore().collection('coll').get()
for (const documentSnap of collectionSnap) {
  const data = documentSnap.data() as { foo: string }
  console.log(data.foo)
}
```
But this trips me up when I get a single document! I forget to check that it exists, and erroneously assert that data isn't undefined when it might be.
```ts
const documentSnap = await firestore().doc('coll/doc').get()
const data = documentSnap.data() as { foo: string }
console.log(data.foo) // crash: undefined is not an object
```

If DocumentSnapshot is represented by a union type, TypeScript can draw
the connection between documentSnap.exists being true and data()
returning an object.
```ts
const documentSnap = await firestore().doc('coll/doc').get()
if (documentSnap.exists) {
  // TypeScript is now aware that documentSnap.data cannot return
  // undefined, and will not complain when I try to access myValue
  console.log(documentSnap.data().myValue)
}
```

This also lets us express the guarantee that a CollectionSnapshot only
contains documents that exist.
```ts
const collectionSnap = await firestore().collection('coll').get()
for (const documentSnap of collectionSnap.docs) {
  // TypeScript is now aware that documentSnap.data cannot return
  // undefined, and will not complain when I try to access myValue
  console.log(documentSnap.data().myValue)
})
```

### Checklist

- [ ] Supports `Android`
- [ ] Supports `iOS`
  - **LINK TO DOCS PR HERE**
- [ ] Flow types updated

### Test Plan

I added some code in `type-test.ts` that assumes the guarantees that this change adds. I couldn't figure out if you had any tools aimed at checking typescript correctness across the whole repo, so I guess we'll see if the CI tests pass.

### Release Plan

[TYPES][BREAKING] [Firestore] - Improved TS types with more precise function types, and union types for DocumentChanges and DocumentSnapshots.

---


packages/firestore/lib/index.d.ts
```diff
@@ -430,14 +486,26 @@ export namespace Firestore {
...
+
   /**
    * A DocumentSnapshot contains data read from a document in your Firestore database. The data can be extracted with
    * .`data()` or `.get(<field>)` to get a specific field.
    *
    * For a DocumentSnapshot that points to a non-existing document, any data access will return 'undefined'.
    * You can use the `exists` property to explicitly verify a document's existence.
    */
-  export interface DocumentSnapshot {
+  interface BaseDocumentSnapshot {
```
Ehesp: Should this be exported?

Ehesp:
Wow thanks for this. 

My only reservation so far, is that we draw on TypeScript for our documentation website. There's a few things in here which I'm not sure we handle, such as the nested `namespace Query` so will have to look into adding support for that.

It will also add pages for `NonexistingDocumentSnapshot`, which isn't something found in the usual Firebase docs site, so whether this will confuse users I'm not sure... But good TS support is equally as important.

packages/firestore/lib/index.d.ts
iwikal: I couldn't think of any reason why the consumer would want to interact with it directly. It feels like an implementation detail, best left private if we want to change it later.

Ehesp: The only thing is the docs would mention methods are extended from it - but it should handle that, just the user won't be able to see the interface.

iwikal:
If this PR takes things too far from the official SDKs, and since v6 is released, I could make a new one with just the QueryDocumentSnapshot part. That's a nice feature in its own right. Also, it shouldn't be majorly breaking. Worst case, people might get linter warnings about unnecessary checks and constant conditionals.

iwikal:
Turns out I had misinterpreted the docChanges api: `DocumentChange.doc` is always a document that exists, so the union thing is completely unnecessary. It could just be a QueryDocumentSnapshot.

Salakar:
@iwikal thanks for this PR, whats the current status of it, is it good to go? If so could you fix any merge conflicts with master if thats ok. Thank you

iwikal:
I think it's ready!

PR #2723, opened 2019-10-14
Fix module links

karhatsu:
### Summary

The module links in the v6 migration page did not work. Clicking them returned Page not found. Adding `v6/` to them should fix the issue.

### Release Plan

[GENERAL][BUGFIX] - Fix module links in the v6 migration page


Salakar:
@karhatsu thank you for sorting this!

PR #2726, opened 2019-10-14
[v5] [iOS] Register for remote notifications fix

mcuelenaere:
### Summary

We've observed crashes on iOS 13 when calling `RNFirebaseMessaging registerForRemoteNotifications:rejecter:` in the wild.

When checking the code, `registerForRemoteNotifications` is not being called from the main thread (while 5 lines above it, in `requestPermission`, it is being done so).

This also corresponds with the crash stack traces we've seen, none of them are on the main thread.

Bottom line: calling this method from the main thread has no negative impact and should hopefully fix these crashes.

### Checklist

- [ ] Supports `Android`
- [X] Supports `iOS`
- [ ] `e2e` tests added or updated in packages/**/e2e
- [ ] Flow types updated
- [ ] Typescript types updated

### Test Plan

Call `registerForRemoteNotifications` and notice app not crashing.

### Release Plan

[IOS][BUGFIX] [MESSAGING] - Register for remote notifications should run on main thread

---

Think `react-native-firebase` is great? Please consider supporting the project with any of the below:

- 👉 Donate via [Open Collective](https://opencollective.com/react-native-firebase/donate)
- 👉 Follow [`React Native Firebase`](https://twitter.com/rnfirebase) and [`Invertase`](https://twitter.com/invertaseio) on Twitter
- 👉 Star this repo on GitHub ⭐️
- 👉 Contribute; see our [contributing guide](/CONTRIBUTING.md)

PR #2738, opened 2019-10-16
[v5] Check paths in Query.isEqual

iwikal:
### Summary

`Query.isEqual` should return false when you compare two queries on different collections.

### Checklist

- [x] Supports `Android`
- [x] Supports `iOS`
- [x] `e2e` tests added or updated in packages/**/e2e
- [ ] Flow types updated

### Test Plan

The plan is to wait and see if CI accepts the PR.

### Release Plan

[JS][BUGFIX] [firestore] - Query.isEqual now takes the collection path into consideration

---

### DIFF
--- a/src/modules/firestore/Query.js
+++ b/src/modules/firestore/Query.js
@@ -199,6 +199,13 @@ export default class Query {
       return false;
     }
 
+    if (
+      this._referencePath.relativeName !==
+      otherQuery._referencePath.relativeName
+    ) {
+      return false;
+    }
+
     if (this._fieldFilters.length !== otherQuery._fieldFilters.length) {
       return false;
     }
--- a/tests/e2e/firestore/query.e2e.js
+++ b/tests/e2e/firestore/query.e2e.js
@@ -1,7 +1,7 @@
 describe('firestore()', () => {
   describe('Query', () => {
     describe('isEqual()', () => {
-      it(`returns true if two Queries have the same .where and .orderBy calls`, () => {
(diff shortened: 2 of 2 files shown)


iwikal:
I assume you want the same functionality as in the web sdk, where path does matter.

iwikal:
I can't really make sense of the build errors; something about a timeout in beforeAll? I couldn't build an android apk to test locally on my linux machine either.

mikehardy:
The ci.android.test build right now is non-functional for unrelated reasons - whether your change is correct or not

ci.ios.test should be passing though, there's a separate PR (#2726) that is also experiencing failures on CI there while it works local, and we're not sure why. We are investigating

mikehardy:
@Salakar @Ehesp I think this is also an issue on v6 but the line has moved it is here now https://github.com/invertase/react-native-firebase/blob/master/packages/firestore/lib/FirestoreQuery.js#L176 (although it's missing, so it's not there - there is no path check)

The underlying assumption that we should mirror web and check for path also seems valid, so I think this should be merged? And v6 should get the same treatment?

iwikal:
It seems that the web sdk implements CollectionReference as a subclass of Query. It follows that if the Query.isEqual method compares paths, CollectionReference.isEqual can be straight up inherited from that. Maybe it's something we don't want to do for performance reasons, but it would be correct.

russellwheatley:
This was merged into v6: https://github.com/invertase/react-native-firebase/pull/3738, but will not be getting merged into v5 as it is no longer supported. Thank you for bringing this to our attention @iwikal, much appreciated.

PR #2749, opened 2019-10-18
fix(auth): Fix iOS event subscriptions not correctly removing on reload

Ehesp:
Correctly remove iOS auth handlers when app is reloaded


mikehardy:
nice!

PR #2792, opened 2019-10-29
fix(crashlytics): setCrashlyticsCollectionEnabled return promise

justinwh:
### Summary

The JS return type of crashlytics.setCrashlyticsCollectionEnabled() is listed as Promise<null>, but the native code is actually returning void.  This change adapts the native code to match the JS declaration.

### Checklist

- [x] Supports `Android`
- [x] Supports `iOS`
- [ ] Flow types updated

### Release Plan

[TYPES][BUGFIX] [CRASHLYTICS] - Fix setCrashlyticsCollectionEnabled() to return Promise (as declared)

---


Ehesp:
Hmm, this is still a promise though as we call native.. @Salakar

Ehesp:
Yeah, however anything going over the bridge is a promise. I think the correct return type should be `Promise<void>`

justinwh:
Yeah, there's another way to fix this, which is to make the native `setCrashlyticsCollectionEnabled` return `Promise<void>` like most of the other methods in the module do.

This method isn't _currently_ doing any async work, but it _could_ be made to still return a promise to future-proof the API (e.g. if in a later revision the method needs to do some async work, the API wouldn't need to change).

If you prefer that direction, I can submit a new PR along those lines.

Salakar:
> Yeah, there's another way to fix this, which is to make the native `setCrashlyticsCollectionEnabled` return `Promise<void>` like most of the other methods in the module do.
> 
> This method isn't _currently_ doing any async work, but it _could_ be made to still return a promise to future-proof the API (e.g. if in a later revision the method needs to do some async work, the API wouldn't need to change).
> 
> If you prefer that direction, I can submit a new PR along those lines.

Hey, thanks for sending this PR, for consistency could you change this to be `Promise<void>` for now - when/if we migrate the modules to TurboModules then we can make these synchronous but this would happen through semver regardless.

justinwh:
I updated this PR to implement the requested approach—making the native return type match what's declared in the JS.

Issue #2412, opened 2019-07-26
registerHeadlessTask('RNFirebaseBackgroundMessage' does not triggered function


c0rrectanswer:
## Issue

index.js
```
const NotificationHandler = async (message) => {
    console.warn('RNFirebaseBackgroundMessage: ', message);
    return Promise.resolve();
};

AppRegistry.registerHeadlessTask('RNFirebaseBackgroundMessage', () => NotificationHandler);
```

**NotificationHandler** does not get triggered although I have made the following changes.

---

### iOS

<details><summary>Click To Expand</summary>
<p>

#### `ios/Podfile`:

- [ ] I'm not using Pods
- [ ] I'm using Pods and my Podfile looks like:

```ruby
# N/A
```

#### `AppDelegate.m`:

```objc
// N/A
```

</p>
</details>

---

### Android

<details><summary>Click To Expand</summary>
<p>

#### `android/build.gradle`:

```groovy
// N/A
```

#### `android/app/build.gradle`:

```groovy
implementation 'me.leolin:ShortcutBadger:1.1.21@aar'
```

#### `android/settings.gradle`:

```groovy

// N/A
```

#### `MainApplication.java`:

```java
new RNAppAuthPackage(),
new RNFirebasePackage(),
new RNFirebaseCrashlyticsPackage(),
new RNFirebaseMessagingPackage(),
new RNFirebaseNotificationsPackage()
```

#### `AndroidManifest.xml`:

```xml
<service android:name="io.invertase.firebase.messaging.RNFirebaseMessagingService">
  <intent-filter>
    <action android:name="com.google.firebase.MESSAGING_EVENT" />
  </intent-filter>
</service>
<service android:name="io.invertase.firebase.messaging.RNFirebaseBackgroundMessagingService" />

```

</p>
</details>

---

## Environment

<details><summary>Click To Expand</summary>
<p>

**`react-native info` output:**

```
 OUTPUT GOES HERE
```

- **Platform that you're experiencing the issue on**:
  - [ ] iOS
  - [x] Android
  - [ ] **iOS** but have not tested behavior on Android
  - [ ] **Android** but have not tested behavior on iOS
  - [ ] Both
- **`Firebase` module(s) you're using that has the issue:**
    "react-native": "0.59.9",
    "react-native-firebase": "5.2.2",
    "react-native-image-picker": "^1.0.1",
    "react-native-open-notification": "^0.1.3",
    "react-native-shortcutbadger": "^0.2.0",
[...]

PR #2806, opened 2019-11-01
docs(messaging): note headless task key change in release notes

adamfellon:
When converting to v6 the key used when registering a headless task needs to be changes to in order to register the correct task.

### Summary

Want to represent this comment for users who encounter headless task issues: https://github.com/invertase/react-native-firebase/issues/2412#issuecomment-548934740

### Checklist

- [x] Supports `Android`

### DIFF
--- a/pages/migrating-to-v6.md
+++ b/pages/migrating-to-v6.md
@@ -397,12 +397,14 @@ No breaking changes.
 `@react-native-firebase/messaging`
 
 - [android] The manually added `RNFirebaseMessagingService` service in your `AndroidManifest.xml` file is no longer required - you can safely remove it.
+- [android] For background messages, the registered background headless task name has been changed.  You will need to change `RNFirebaseBackgroundMessage` to `ReactNativeFirebaseMessagingHeadlessTask` when you call `AppRegistry.registerHeadlessTask(` in your `index.js`
 - [ios] The manually added `RNFirebaseMessaging` usages in your `AppDelegate` files are no longer required - you can safely remove them.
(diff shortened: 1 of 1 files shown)


mikehardy:
CI seems to have had a major, unrelated failure @Salakar  ?

Salakar:
> CI seems to have had a major, unrelated failure @Salakar ?

Its just incorrectly not ignoring the docs branch, have updated to branch rules so shouldn't do it again.


----


Generally ok with this change but really `AppRegistry.registerHeadlessTask` shouldn't be called manually on v6. This is automatically done for you if you use [`messaging().setBackgroundMessageHandler(fn)`](https://invertase.io/oss/react-native-firebase/v6/messaging/reference/module#setBackgroundMessageHandler) - right now this is all it does internally - but moving it into the API gives us the flexibility in future to change it or further augment the API with minimal effort for users + it also matches the Firebase Web SDK API.

Would you be able to update this to mention calling that method instead? As with the headless task API; it should be called early on in your apps code.


----

image

adamfellon:
Ah, yeah I think I should even be calling that function in my code as well as it'll avoid any issues with renaming in the future.  

I don't use Typescript so I think I missed the example from the interface.

I don't know how constants/statics from the `NativeModules` work but is it worth it to go as far as exporting `private static final String TASK_KEY = "ReactNativeFirebaseMessagingHeadlessTask";` out to be used in `lib/index.js` when it registers the headless task?

Salakar:
Superseded by #3309 and friends - thanks!

Salakar:
Closing as #3339 which supersedes this has now been merged and is available on an RC release (6.4.0-rc4).

Thank you for your contribution

PR #2921, opened 2019-11-24
docs(crashlytics): add note about required module import

ScreamZ:
Improve documentation for crashlytics


docs/crashlytics/quick-start.md
```diff
@@ -53,6 +53,14 @@ Once installed, import the Crashlytics package into your project:
 import crashlytics from '@react-native-firebase/crashlytics';
 ```
 
+**You need to import the library even if you're not using any functions of the Crashlytics SDK. By doing this, you enable the listener that get the JS stacktrace into the Crash Reporting console in Firebase**
```
Ehesp: Would you be able to remove the `**`'s? Aside from that's it's good.

ScreamZ: Done

Issue #2943, opened 2019-11-28
UIWebView : [Admob] (Deprecated Api Usage)


yannou788:
## Issue

"@react-native-firebase/admob": "6.1.0",
TMS-90809: Deprecated API Usage - Apple will stop accepting submissions of apps that use UIWebView.

After some research, I found that in **RNFBAdMob.podspec** there was an dependency with **PersonalizedAdConsent** (https://github.com/googleads/googleads-consent-sdk-ios/tree/master/PersonalizedAdConsent/PersonalizedAdConsent) which contains call to "UIWebView".

---

### iOS

<details><summary>Click To Expand</summary>
<p>

#### `ios/Podfile`:

- [ ] I'm not using Pods
- [x] I'm using Pods and my Podfile looks like:

```ruby
# N/A
```

#### `AppDelegate.m`:

```objc
// N/A
```

</p>
</details>

---

### Android

<details><summary>Click To Expand</summary>
<p>

#### Have you converted to AndroidX?

- [ ] my application is an AndroidX application?
- [ ] I am using `android/gradle.settings` `jetifier=true` for Android compatibility?
- [ ] I am using the NPM package `jetifier` for react-native compatibility?

#### `android/build.gradle`:

```groovy
// N/A
```

#### `android/app/build.gradle`:

```groovy
// N/A
```

#### `android/settings.gradle`:

```groovy
// N/A
```

#### `MainApplication.java`:

```java
// N/A
```

#### `AndroidManifest.xml`:

```xml
<!-- N/A -->
```

</p>
</details>

---

## Environment

<details><summary>Click To Expand</summary>
<p>

**`react-native info` output:**

```
 OUTPUT GOES HERE
```

- **Platform that you're experiencing the issue on**:
  - [x ] iOS
  - [ ] Android
  - [ ] **iOS** but have not tested behavior on Android
  - [ ] **Android** but have not tested behavior on iOS
  - [ ] Both
- **`react-native-firebase` version you're using that has this issue:**
  - `6.1.0`
- **`Firebase` module(s) you're using that has the issue:**
  - `Admob`
- **Are you using `TypeScript`?**
  - `N`
  
</p>
</details>

---

Think `react-native-firebase` is great? Please consider supporting all of the project maintainers and contributors by donating via our [Open Collective](https://opencollective.com/react-native-firebase/donate) where all contributors can submit expenses. [[Learn More]](https://invertase.io/oss/react-native-firebase/contributing/donations-expenses)

[...]

PR #2944, opened 2019-11-28
Add JS in bug report for package.json/firebase.json

mikehardy:
For v6 I really need to see firebase.json and package.json  or I don't know what versions people are running anymore as they aren't in the other files


mikehardy:
As noticed in #2943

PR #2969, opened 2019-12-04
docs(auth): fix incorrect auth credential example

Phaen:
This one had me confused for quite a bit myself.


mikehardy:
But the AuthCredential can be used for both and is a valid argument for both the example method you removed as well as the one you added. The old example was not incorrect I don't think?

Phaen:
@mikehardy `linkWithCredential()` has to be called on an user, it does not automatically select the current user and link that one.

mikehardy:
@Phaen of course! Thanks, and great catch

PR #2975, opened 2019-12-05
Fixes default re-export syntax

joelgallant:
### Summary
This export is not valid when being parsed by typescript or webpack. The updated version is equivalent.


Salakar:
@joelgallant thank you for this, much appreciated 🎉

PR #2981, opened 2019-12-09
Load Android Gradle Plugin conditionally

SaeedZhiany:
### Summary

This wraps the Android Gradle plugin dependency in the buildscripts section of android/build.gradle in a conditional:

```
if (project == rootProject) {
    // ... (dependency here)
}
```

The Android Gradle plugin is only required when opening the project stand-alone, not when it is included as a dependency. By doing this, the project opens correctly in Android Studio, and it can also be consumed as a native module dependency from an application project without affecting the app project (avoiding unnecessary downloads/conflicts/etc).

for more info, you can refer to [this thread](https://github.com/facebook/react-native/pull/25569) and especially [this comment.](https://github.com/facebook/react-native/pull/25569#issuecomment-532504277)

### Checklist

- [x] Supports `Android`
- [ ] Supports `iOS`
- [ ] Flow types updated

### Release Plan

[INTERNAL][ENHANCEMENT] [AndroidGradleFile] - Load Android Gradle Plugin only when this repository opened as a stand-alone android project in Android Studio.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -3,15 +3,21 @@ def safeExtGet(prop, fallback) {
 }
 
 buildscript {
-  repositories {
-    google()
-    jcenter()
-    maven {
-      url 'https://maven.fabric.io/public'
+  // The Android Gradle plugin is only required when opening the android folder stand-alone.
+  // This avoids unnecessary downloads and potential conflicts when the library is included as a
+  // module dependency in an application project.
+  if (project == rootProject) {
+    repositories {
+      google()
+      jcenter()
+      maven {
+        url 'https://maven.fabric.io/public'
+      }
     }
-  }
-  dependencies {
-    classpath 'com.android.tools.build:gradle:3.4.1'
+
+    dependencies {
+      classpath("com.android.tools.build:gradle:3.5.2")
+    }
   }
 }
 


mikehardy:
Oh interesting - I thought we already did that but @SaeedZhiany I think rather it was on v6? And this is v5. I do see the v5 branch living on (as a life support thing) for quite a while so I'm fine with this change.

We are having CI issues on v5 though, so testing is very important. Can you explain the exact steps you took to test the change?

(and for reference I know this change is easy however I attempted to apply it to react-native-webview for instance and it failed because of the need for the kotlin plugin on the classpath - so it is possible for this change to go wrong...)

SaeedZhiany:
Yeah, it was for v6 and this is for v5.

> We are having CI issues on v5 though, so testing is very important. Can you explain the exact steps you took to test the change?

to be honest, nothing. because it's just like the other gradle files and I didn't expect to see build failure.
the only difference is the maven repository usage.

```
maven {
      url 'https://maven.fabric.io/public'
}
```

but I could find the full stack trace of CI to see what causes the failure.
also, the `ci.analyse` has been failed not `ci.android.build`, the second one is not started yet and I'm wondering why? does `ci.android.build` start after successful passing of the `ci.analyse` phase?

SaeedZhiany:
@mikehardy 
Yeah, it was for v6 and this is for v5.

> We are having CI issues on v5 though, so testing is very important. Can you explain the exact steps you took to test the change?

to be honest, nothing. because it's just like the other gradle files and I didn't expect to see build failure.
the only difference is the maven repository usage.

```
maven {
      url 'https://maven.fabric.io/public'
}
```

but I could not find the full stack trace of CI to see what causes the failure.
also, the `ci.analyse` has been failed not `ci.android.build`, the second one is not started yet and I'm wondering why? does `ci.android.build` start after the successful passing of the `ci.analyse` phase?

mikehardy:
@SaeedZhiany I should been more clear about the CI failures :-). They are unrelated to your work, and nothing you can do here will make CI work again for v5, it's a real mess at the moment 😬 

So what I'm asking for is local testing. What have you yourself done to test this? You can for instance (on v5, this does not work on v6) simply refer to your react-native-firebase branch via a git reference in your package.json and after installing it like that verify the build.gradle has the change you expect, and still works. That would be sufficient proof and is how I am verifying my v5.x.x work temporarily while CI is non-functional

SaeedZhiany:
No problem @mikehardy, It's easy to test. I'll inform you in a few minutes later

SaeedZhiany:
@mikehardy 
I tested it by applying the same change to my current RNF module's gradle file in my project.
the project successfully built in both debug and release mode. the app has been launched successfully but my server is not currently available to check possible runtime failure, however, I don't think it happens because of this change.

do more tests needed?

mikehardy:
Nope, I agree with this change that is totally adequate testing, thanks for doing it. I'm on release engineer mode this morning with RNFBv5 so I think I can even get this out today. Thanks!

mikehardy:
I have manually pulled this commit from your fork, so this in my patch queue and it will release in 5.5.7, but not via github merge. I won't update this PR (except to close it if it does not auto-close) but 5.5.7 will have it, thanks @SaeedZhiany

PR #3034, opened 2019-12-22
Improved error documentation of Clouds module - Fixes #3033

### LINKED ISSUE #3033: Can we talk about error documentation please?
## Documentation Feedback
Hello guys!
First of all, great package. It's a huge help to developers everywhere!
Now back to the topic at hand:
I find it pretty irking how little documentation there is about there error codes and messages that this package will return for the different Firebase modules. And if it's the case that this package returns errors identical to Firebase Web, I think that should be explicitly stated somewhere, even if it may seem like an obvious assumption to make.
I understand that you guys only recently upgraded to v6, so there's a lot of work going on right now, but I'd love if this were improved because I find it super important.

andersonaddo:
This is a suggested way to improve the error documentation of this package. I've talked more about my feelings about the current state of the error documentation in #3033

I've also suggested some enhancements to the [NaticeFirebaseError doc page](https://invertase.io/oss/react-native-firebase/v6/app/reference/nativefirebaseerror) that will hopefully come through sooner or later.

### DIFF
--- a/docs/storage/index.md
+++ b/docs/storage/index.md
@@ -40,6 +40,14 @@ directly from your device and from your Firebase Cloud Storage bucket.
 	</Block>
 </Grid>
 
+## Errors
+The functions in this module can throw two different types of errors. If ill-formed urls are passed into them, they will throw
+[generic Errors](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error) with messages 
+explaining the problem that caused them.
+
+If error was thrown by Firebase itself, then the function will throw a [NativeFirebaseError](https://invertase.io/oss/react-native-firebase/v6/app/reference/nativefirebaseerror). The codes for these errors are the same as the codes for [Firebase Cloud Storage Web errors](https://firebase.google.com/docs/storage/web/handle-errors).
+
+
 ## Learn more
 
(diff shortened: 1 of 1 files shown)


docs/storage/index.md
```diff
@@ -40,6 +40,14 @@ directly from your device and from your Firebase Cloud Storage bucket.
 	</Block>
 </Grid>
 
+## Errors
+The functions in this module can throw two different types of errors. If ill-formed urls are passed into them, they will throw
+[generic Errors](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error) with messages 
+explaining the problem that caused them.
+
+If error was thrown by Firebase itself, then the function will throw a [NativeFirebaseError](https://invertase.io/oss/react-native-firebase/v6/app/reference/nativefirebaseerror). The codes for these errors are the same as the codes for [Firebase Cloud Storage Web errors](https://firebase.google.com/docs/storage/web/handle-errors).
```
Salakar:
```suggestion
If the error was thrown by Firebase itself, then the function will throw a [NativeFirebaseError](https://invertase.io/oss/react-native-firebase/v6/app/reference/nativefirebaseerror). The codes for these errors are the same as the codes for [Firebase Cloud Storage Web errors](https://firebase.google.com/docs/storage/web/handle-errors).
```

PR #3047, opened 2019-12-27
fix(app): typo

### LINKED ISSUE #3046: Typo in /packages/app/lib/internal/registry/app.js
There's a typo here. It should be `initialized`. Missing the "i":
https://github.com/invertase/react-native-firebase/blob/b40d44c57e81a4110cbb4e6204c400cb8f78b00a/packages/app/lib/internal/registry/app.js#L176

Just as a side note, this kind of typo (there may be a bunch of other ones, who knows) wouldn't exist if the code was written in TypeScript.

gustavopch:
Fixes #3046 

[JS][BUGFIX] [APP] - Fix a typo

### DIFF
--- a/packages/app/lib/internal/registry/app.js
+++ b/packages/app/lib/internal/registry/app.js
@@ -173,7 +173,7 @@ export function initializeApp(options = {}, configOrName) {
   return getAppModule()
     .initializeApp(options, { name })
     .then(() => {
-      app._intialized = true;
+      app._initialized = true;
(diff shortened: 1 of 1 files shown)


mikehardy:
Excellent - thanks @gustavopch - I grepped through the whole code base and this seems to be the only instance of 'intialized', whereas there are a few references to what appear to be the same symbol as 'initialized' so I think this is a good change

PR #3087, opened 2020-01-12
Added Firestore type parameters

Phaen:
### Summary

Whenever `DocumentSnapshot.get` was used to retrieve a document field, the user would have to deal with the type assertion themselves, leading to ugly `as` statements throughout their code. Additionally, the typing might not be clear to the user, as firestore accepts `Date` as a field type when storing data, but would return it as a `FirebaseFirestoreTypes.Timestamp` upon retrieval. By allowing the user to pass the type and constricting the type parameter, it should help alleviate these issues.

### Checklist

- [x] Supports `Android`
- [x] Supports `iOS`
- [ ] Flow types updated

### Release Plan

[TYPES] [BREAKING] [FIRESTORE] - Added type parameters to `DocumentSnapshots.get`

---


packages/firestore/lib/index.d.ts
```diff
@@ -494,12 +512,12 @@ export namespace FirebaseFirestoreTypes {
      * ```js
      * const user = await firebase.firestore().doc('users/alovelace').get();
      *
-     * console.log('Address ZIP Code', user.get('address.zip'));
+     * console.log('Address ZIP Code', user.get<string>('address.zip'));
```
russellwheatley: you wouldn't put `<string>` in the actual code.

russellwheatley: To clarify, it presumes the developer is using Typescript.

packages/firestore/lib/index.d.ts
```diff
@@ -175,6 +175,24 @@ export namespace FirebaseFirestoreTypes {
    */
   export type DocumentChangeType = 'added' | 'removed' | 'modified';
 
+  /**
+   * The types for a DocumentSnapshot field that are supported by Firestore.
+   */
+  export type DocumentFieldType =
+    | string
```
russellwheatley: Apologies. A couple of other things that I missed. the `|` operator shouldn't be on line 182.

packages/firestore/lib/index.d.ts
```diff
@@ -499,7 +517,7 @@ export namespace FirebaseFirestoreTypes {
      *
      * @param fieldPath The path (e.g. 'foo' or 'foo.bar') to a specific field.
      */
-    get(fieldPath: string | FieldPath): any;
+    get<fieldType extends DocumentFieldType>(fieldPath: string | FieldPath): fieldType;
```
russellwheatley: Shouldn't this just be `get(fieldPath: string | FieldPath): DocumentFieldType;`? `fieldType` doesn't actually extend `DocumentFieldType` with any further properties

Phaen: It defines a new named type to be specifically repeated for the returning value. Is there a different way that you're supposed to do that?

packages/firestore/lib/index.d.ts
Phaen: This operator is indeed redundant, but harmless. It's automatically generated by Prettier for consistent styling.

packages/firestore/lib/index.d.ts
Phaen: I forgot that we're not using Typescript in any of the examples, good one!

Issue #3084, opened 2020-01-09
[v6] Javascript source maps not available in crashlytics


kbanashek:
## Issue

I'm not seeing javascript stack traces being logged by default after implementing v6.2 running on iOS simulator & RN 61.5

![image](https://user-images.githubusercontent.com/2280402/72094188-d6be9280-32e3-11ea-9390-f5c2decf2e14.png)

![image](https://user-images.githubusercontent.com/2280402/72094153-ca3a3a00-32e3-11ea-807c-c242b77c518d.png)

Salakar:
.crash() is a native method so causes a native error. Use `recordError(new Error('hello world'))` if you want to record a JS error

kbanashek:
@Salakar Does `recordError` fire on both android and iOS? I was not seeing it fire on iOS

PR #3092, opened 2020-01-13
docs(crashlytics): notes on what to expect for crashes

mikehardy:
### Summary

Related to #3084 and other user requests about the crash APIs

This is documentation only

Possibly worth mentioning that I frequently advise people "there is an edit button at the top of each docs page" but that's wrong for v6! I wasn't aware that the docs are auto-generated from the types defs but they are I think? Which is why I'm doing a PR against this file

Is there any way to hook up an edit button to go to the types files so my assertion about edit buttons is true again? Otherwise I can give more nuanced advice, but that edit button was a neat trick...

### Release Plan

[TYPES][ENHANCEMENT] [Crashlytics] - Add docs noting what traces will look like

---

### DIFF
--- a/packages/crashlytics/lib/index.d.ts
+++ b/packages/crashlytics/lib/index.d.ts
@@ -79,7 +79,7 @@ export namespace FirebaseCrashlyticsTypes {
     isCrashlyticsCollectionEnabled: true;
 
     /**
-     * Cause your app to crash for testing purposes.
+     * Cause your app to crash for testing purposes. This is a native crash and will not contain a javascript stack trace.
      *
      * #### Example
      *
@@ -107,7 +107,9 @@ export namespace FirebaseCrashlyticsTypes {
     /**
      * Record a JavaScript Error.
      *
-     * The JavaScript stack trace is converted into a mock native iOS or Android exception before submission.
+     * The JavaScript stack trace is converted into a mock native iOS or Android exception before submission. 
+     * The line numbers in the stack trace (if available) will be relative to the javascript bundle built by your packager,
+     * after whatever transpilation or minimization steps happen. You will need to maintain sourcemaps to decode them if desired.
      *
      * #### Example
      *


Salakar:
> Possibly worth mentioning that I frequently advise people "there is an edit button at the top of each docs page" but that's wrong for v6! I wasn't aware that the docs are auto-generated from the types defs but they are I think? Which is why I'm doing a PR against this file

Edit button is correct still, it's only the reference docs that don't have an edit button, instead each example has a `< >` button top right of the code block that will take you to the specific line in the relevant `.d.ts` file that can be edited;

image

Though I've just noticed a bug with that, the url it goes to has a double `/packages/` in it for some reason, perhaps @Phaen / @Ehesp could take a look, was working a while back, looks like it may be a underlying graphql change as it purely renders the `sourceUrl`, e.g. https://github.com/invertase/invertase/blob/master/packages/documentation/src/reference/kinds/Property.js#L41

mikehardy:
Ah! I had not noticed that code block icon. Does that work if there are not examples? Is it possible to have an edit button at the top for the whole types file also? I have a pretty high hit-ratio on users making docs PRs after I beseech them so the easier it is, the more free docs PRs we'll get I'm thinking

Salakar:
> Ah! I had not noticed that code block icon. Does that work if there are not examples? Is it possible to have an edit button at the top for the whole types file also? I have a pretty high hit-ratio on users making docs PRs after I beseech them so the easier it is, the more free docs PRs we'll get I'm thinking

@Phaen has done a PR internally to fix the broken links, Pablo could you take a look at adding the edit button onto reference pages also when you have some time, can copy the button from the normal pages and just change the link to point to the `sourceUrl` generated from the gatsby plugin iirc - thanks!

Salakar:
Also PR LGTM, so will merge, thanks Mike!

PR #3107, opened 2020-01-17
chore(docs): select the right xcode project/workspace to open

nvthai:
### Summary
update `ios-firebase-credentials` docs: should open `.xcworkspace` if using Pods and in reverse, open `.xcodeproj`


Salakar:
Good spot, thanks!

Issue #2886, opened 2019-11-17
🔥 Attempt to invoke virtual method 'void com.google.firebase.perf.metrics.HttpMetric.setHttpResponseCode(int)' on a null object reference while using perf library.


shardick:
## Issue

Having [perf/unknown] Attempt to invoke virtual method 'void com.google.firebase.perf.metrics.HttpMetric.setHttpResponseCode(int)' on a null object reference while using perf library.

```javascript
var baseUrl = url.split('?')[0];

            const metric = await perf().newHttpMetric(baseUrl, method);

            var myHeaders = new Headers();
            myHeaders.set('user-agent', 'Trucky ' + Platform.OS + ' ' + DeviceInfo.getReadableVersion());
            myHeaders.set("x-platform", Platform.OS);

            var myInit = {
                method: method,
                headers: myHeaders,
                mode: 'cors',
                cache: 'no-cache'
            };

            if (myInit.method == "POST") {
                //console.warn(payload);

                myInit
                    .headers
                    .set('Accept', 'application/json, text/plain, */*');
                myInit
                    .headers
                    .set('Content-Type', 'application/json');

                if (auth)
                    myInit.headers.set('x-access-token', this.token);

                myInit.body = JSON.stringify(payload);
            }

            //console.log(this.config.serviceUrl + url);

            var response = await fetch(this.config.serviceUrl + url, myInit);

            metric.setHttpResponseCode(response.status);
            metric.setResponseContentType(response.headers.get('Content-Type'));
            //metric.setResponsePayloadSize(response.headers.get('Content-Length'));

            await metric.stop();

            var json = await response.json();
```
---

### iOS

<details><summary>Click To Expand</summary>
<p>

#### `ios/Podfile`:

- [ ] I'm not using Pods
- [x] I'm using Pods and my Podfile looks like:

```ruby
# N/A
```

#### `AppDelegate.m`:

```objc
// N/A
```

</p>
</details>

---

### Android

<details><summary>Click To Expand</summary>
<p>

#### Have you converted to AndroidX?

- [x] my application is an AndroidX application
- [x] I am using `android/gradle.settings` `jetifier=true` for Android compatibility?
- [] I am using the NPM package `jetifier` for react-native compatibility?

#### `android/build.gradle`:

```groovy
// Top-level build file where you can add configuration options common to all sub-projects/modules.

buildscript {
    ext {
        buildToolsVersion = "28.0.3"
        minSdkVersion = 21
[...]

PR #3110, opened 2020-01-18
docs(perf): Add HTTP metric.start command to example

jack828:
### Summary

#2886 and other users (including me!) copied and pasted the quick start example and started modifying away into my project, and then all my HTTP requests started throwing rejections. Turns out, I didn't read the module docs correctly - a metric must be started before you can use it!
This docfix should hopefully help prevent some frustrated googling.

Thanks for an amazing library :blue_heart:

### Checklist

N/A

### Test Plan

N/A

### Release Plan

[INTERNAL][minor] [perf] - Quick start documentation update for HTTP metrics

(I don't know if this is needed for doc updates!)
---

:fire:

### DIFF
--- a/docs/perf/quick-start.md
+++ b/docs/perf/quick-start.md
@@ -80,13 +80,16 @@ async function getRequest(url) {
   // Define meta details
   metric.putAttribute('user', 'abcd');
 
+  // Start the metric
+  await metric.start();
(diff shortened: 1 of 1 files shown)


Ehesp:
Good catch. Thanks!

PR #3120, opened 2020-01-21
fix(messaging) fix typescript signature for setBackgroundMessageHandler

andersonaddo:
It wasn't specified that it should actually return a promise.
If it doesn't return a promise, you start to get `cannot resolve .then of undefined` errors.


Ehesp:
Good catch! Yep it's a Headless task and requires a promise to be returned.

PR #3122, opened 2020-01-22
Shared keychain ios

### LINKED ISSUE #3074: [Feature Request] store auth state in iOS shared keychain via `useUserAccessGroup` method
Hi there. First off, upon choosing to open a new feature req issue, I was directed to your [Feature Request Board](https://invertase.io/oss/react-native-firebase/feedback/), although there wasn't anything there apart from an email form. Please let me know if this is not appropriate to discuss here.

## Use-case background

I am currently developing an iOS app, where one of the main features is interacted with via a share extension. Both the main app and the share extension need access to the same Firebase auth state.

Currently the solution we have to share these states is set up a shared keychain (via `react-native-keychain` + an iOS access group) then, upon logging into Firebase on the main app and receiving a token, storing that token in the shared keychain. When the share ext starts, it checks the keychain for an existing token and uses that to log itself into Firebase.

This seems to work, however it is flawed in the way that if the user does not start the share extension before the token expires, they are unable to log into Firebase any more on the share ext.

## Idea

This page describes use of a `useUserAccessGroup` method on the Firebase Auth module to tell Firebase which access group to use and it seems to automatically manage the shared keychain:
https://firebase.google.com/docs/auth/ios/single-sign-on

This looks perfect for my use-case. Unfortunately this method is not provided in the React Native Firebase Auth module. 

**My feature request would be to add support for this method somehow.**

### What I've played with

I thought perhaps I can just add that method call into both my `AppDelegate.m` on the main app and the share ext's equivalent:
```[FIRAuth.auth useUserAccessGroup:@"MY_ACCESS_GROUP" error:nil];```

Doing this however does not give the expected results. After logging in on the main app, calling `firebase.auth().currentUser` (React Native Firebase API) on the share ext does not return a user.

<br />
Any ideas, questions, about how to do this? I personally have no experience with native ObjC or Swift iOS programming and am unsure of the difficulty of this request

### PR DESCRIPTION
**Pull request by russellwheatley**
### Summary

fixes #3074

### Checklist

- [ ] Supports `Android`
- [ ] Supports `iOS`
- [ ] Flow types updated

### Release Plan

---

### DIFF
--- a/packages/auth/ios/RNFBAuth/RNFBAuthModule.m
+++ b/packages/auth/ios/RNFBAuth/RNFBAuthModule.m
@@ -165,9 +165,15 @@ RCT_EXPORT_METHOD(useUserAccessGroup:
     :(RCTPromiseResolveBlock) resolve
     :(RCTPromiseRejectBlock) reject
 ) {
+  NSError *error;
+  [[FIRAuth authWithApp:firebaseApp] useUserAccessGroup:userAccessGroup error:&error];
 
-  [[FIRAuth authWithApp:firebaseApp] useUserAccessGroup:userAccessGroup error:nil];
+  if(!error){
     [self promiseNoUser:resolve rejecter:reject  isError:NO];
+  } else {
+    [self promiseRejectAuthException:reject error:error];
+  }
+  return;
 }
 
 RCT_EXPORT_METHOD(signOut:
--- a/packages/auth/ios/RNFBAuth/RNFBAuthModule.m
+++ b/packages/auth/ios/RNFBAuth/RNFBAuthModule.m
@@ -169,7 +169,7 @@ RCT_EXPORT_METHOD(useUserAccessGroup:
   [[FIRAuth authWithApp:firebaseApp] useUserAccessGroup:userAccessGroup error:&error];
 
   if(!error){
-    [self promiseNoUser:resolve rejecter:reject  isError:NO];
+    [self promiseNoUser:resolve rejecter:reject isError:NO];
   } else {
     [self promiseRejectAuthException:reject error:error];
   }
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: packages/auth/ios/RNFBAuth/RNFBAuthModule.m [hunk 1]
```diff
@@ -159,6 +159,23 @@ - (void)invalidate {
...
+  (FIRApp *) firebaseApp
+    :(NSString *) userAccessGroup
+    :(RCTPromiseResolveBlock) resolve
+    :(RCTPromiseRejectBlock) reject
+) {
+  NSError *error;
+  [[FIRAuth authWithApp:firebaseApp] useUserAccessGroup:userAccessGroup error:&error];
+
+  if(!error){
+    [self promiseNoUser:resolve rejecter:reject  isError:NO];
```
Salakar:
```suggestion
    [self promiseNoUser:resolve rejecter:reject isError:NO];
```

packages/auth/lib/index.d.ts
```diff
@@ -1594,6 +1594,19 @@ export namespace FirebaseAuthTypes {
      * @param code A password reset code.
      */
     verifyPasswordResetCode(code: string): Promise<void>;
+    /**
+     * Switch userAccessGroup and current user to the given accessGroup and the user stored in it.
+     * Sign in a user with any sign in method, and the same current user is available in all
+     * apps in the access group.
+     *
+     * Set the access group to nil to remove user's auth state, and the user state will not be
```
Salakar:
```suggestion
     * Set the `useAccessGroup` argument to `null` to stop sharing the auth state (default behaviour), the user state will no longer be
```

mikehardy:
For unrelated reasons I am looking into transferring ios apps and what the side effects are.

If you transfer an app you will lose access to whatever keychain data is stored. This is not a common case but still might be worth a docs mention. Or maybe just putting it here is adequate https://developer.apple.com/library/archive/technotes/tn2311/_index.html#//apple_ref/doc/uid/DTS40014135-CH1-A_ONE_TIME_LOSS_IN_KEYCHAIN_DATA_WILL_OCCUR_IF_YOU_SWITCH_YOUR_APP_ID_PREFIX

PR #3153, opened 2020-01-30
fix(types): remove types in ConfirmationResult.js

FDiskas:
'types' can only be used in a .ts file.


Salakar:
Them pesky types getting everywhere 👀 

Thanks for the PR 😸

FDiskas:
Migrate to typesript? :)

PR #3202, opened 2020-02-13
docs(database): added missing use of .push() type docs

andersonaddo:
There was formally no actual use of the .push method in it's documentation example snippet.


Salakar:
Woops, good spot! Thanks for the PR 🎉

PR #3218, opened 2020-02-17
docs(remote-config): remove unnecessary await in example

andersonaddo:
In the quick start section of the Remote Config docs, getValue is called as if it's asynchronous, which I don't believe it is.


Salakar:
Good spot, thanks :)

PR #3272, opened 2020-03-04
docs: spelling mistake corrected.

geekrepos:
Spelling mistake corrected in ReadMe file in version 5.x.x


mikehardy:
```
   Error: Can't find a simulator to match with "iPhone 8, iOS 12.1", run 'xcrun simctl list' to list your supported devices.
        It is advised to only state a device type, and not to state iOS version, e.g. "iPhone 7"
      at AppleSimUtils.findDevicesUDID (/Users/ciuser/Library/buildAgent/work/41c4e0b4ce6a9b09/tests/node_modules/detox/src/devices/ios/AppleSimUtils.js:46:13)
    1) "before all" hook in "{root}"
```

:thinking:

PR #3130, opened 2020-01-24
Fix `getToken()` iOS function for messaging

anfriis:
This function often throws an exception:
`Cloud Messaging "[messaging/unknown] The operation couldn’t be completed"`
This fixes it and returns the FCM token successfully.

Fixes #2657 

### Checklist

- [x] Supports `Android`
- [x] Supports `iOS`
- [x] `e2e` tests added or updated in packages/**/e2e
- [x] Flow types updated
- [x] Typescript types updated

### Test Plan

I have tested on two different iPhone's, trying to get FCM token several times using this function, and it works now whereas it didn't before because it threw the exception above.

### Release Plan
[IOS][BUGFIX] [messaging] - Fix `getToken()` iOS function for messaging
 

[CATEGORY][type] [LOCATION] - Message

---

Think `react-native-firebase` is great? Please consider supporting the project with any of the below:

- 👉 Donate via [Open Collective](https://opencollective.com/react-native-firebase/donate)
- 👉 Follow [`React Native Firebase`](https://twitter.com/rnfirebase) and [`Invertase`](https://twitter.com/invertaseio) on Twitter
- 👉 Star this repo on GitHub ⭐️
- 👉 Contribute; see our [contributing guide](/CONTRIBUTING.md)

Issue #2995, opened 2019-12-11
APNS token is always null


theminerymike:
## Issue

My apns token is always null (on device) and as a result I never receive notification while my app is in the background on iOS.

I'm migrating from react-native-fcm. setBadge works, getting an FCM token works, onMessage works, and everything works on Android.

I've tried every combination of registering, configuring, etc, based off of all of the other issues I've seen related to this. The capabilities are on, and I tried turning them off and on again as well. These issues do carry over to my new repo made for testing this issue.

What I'm experiencing instead is a null apns token. I believe that's the main reason I'm not receiving these notifications. I am registering first, then trying to get the token second, as I've seen in other people's issues.

When I try using postman to post to `https://fcm.googleapis.com/fcm/send` I get back a successful response. My project is configured on Firebase with an APNs Auth Key.

Yes, "FirebaseAppDelegateProxyEnabled" is not set to "NO" in my Info.plist
Yes, I'm not trying in the simulator.
Yes, the APNS keys appear active on Apple's developer site.
Yes, I am very up to date on all of my libraries.
Yes, I have tried `react-native clean-project`.
Yes, I am prompted if I want to receive notifications.
Yes, I accept it.
Yes, everything looks good from notification settings.

---

## Project Files

https://github.com/theminery-hr4/firebasetest

I've excluded my GoogleService-Info.plist from this repo, but it exists during my testing.

Key files are Firebase.js, AppDelegate.m. Pretty much everything else is default react native.

#### `AppDelegate.m`:
```
@import Firebase;
#import "RNFirebaseNotifications.h"
#import "RNFirebaseMessaging.h"

#import "AppDelegate.h"

#import <React/RCTBridge.h>
#import <React/RCTBundleURLProvider.h>
#import <React/RCTRootView.h>
#import <React/RCTLinkingManager.h>

#if RCT_DEV
#import <React/RCTDevLoadingView.h>
#endif

@implementation AppDelegate

- (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions
{
  [FIRApp configure];
  [RNFirebaseNotifications configure];

  RCTBridge *bridge = [[RCTBridge alloc] initWithDelegate:self launchOptions:launchOptions];
  
#if RCT_DEV
  [bridge moduleForClass:[RCTDevLoadingView class]];
#endif
  
  RCTRootView *rootView = [[RCTRootView alloc]  initWithBridge:bridge
[...]

PR #3273, opened 2020-03-04
Fix 2657 error getting remote notifications tokens

elyalvarado:
### Summary

This PR fixes the logic on `registerForRemoteNotifications` and a race condition that caused issue #2657.

##### Change registerForRemoteNotifications in RNFirebase to skip the check on `isRegisteredForRemoteNotifications` 

The native call to `registerForRemoteNotifications` shouldn't be aborted even if `isRegisteredForRemoteNotifications` returns `true`. This is because after successful registration on the first boot, on successive boots the app `isRegisteredForRemoteNotifications` will return true, and then `registerForRemoteNotifications` won't be called again and therefore the APNSToken will never be set (because it is only set on the `didRegisterForRemoteNotificationsWithDeviceToken` swizzled by the `FIRMessagingRemoteNotificationsProxy`). If the APNSToken is null when calling getTokens, then the call to `[[FIRInstanceID instanceID] tokenWithAuthorizedEntity] ` on `getToken` fails.

##### Change RNFBMessagingAppDelegateInterceptor to set [FIRMessaging messaging].APNSToken

This is what the `FIRMessagingRemoteNotificationsProxy` does, and doing it here seems to avoid a race condition that happens when the `registerForRemoteNotification` method is resolved but the APNSToken is not yet set because the swizzling done by `RNFirebase` finishes first than the one done by the `FirebaseMessaging` library. Even though this race condition only seems to happen on iOS 12, It doesn't do any damage anyway to do it here first anyway.

### Checklist

- [x] Supports `Android`: *This change doesn't affect `Android`*
- [x] Supports `iOS`
- [x] `e2e` tests added or updated in packages/**/e2e: *can only be tested on actual devices*.
- [x] Flow types updated: *doesn't require type changes*
- [x] Typescript types updated: *doesn't require type changes*

### Test Plan

Tested on iOS 12 device by calling `registerForRemoteNotifications` followed by `getToken` on app start and restarting multiple times.

### Release Plan

[IOS][BUGFIX] [Messaging] - Fix race condition that prevented getting APNSToken and Firebase Token on iOS after first boot, and on iOS 12.

---

### DIFF
--- a/packages/messaging/ios/RNFBMessaging/RNFBMessagingAppDelegateInterceptor.m
+++ b/packages/messaging/ios/RNFBMessaging/RNFBMessagingAppDelegateInterceptor.m
@@ -44,6 +44,7 @@
 
 // called when `registerForRemoteNotifications` completes successfully
 - (void)application:(UIApplication *)application didRegisterForRemoteNotificationsWithDeviceToken:(NSData *)deviceToken {
+  [FIRMessaging messaging].APNSToken = deviceToken;
   if (_registerPromiseResolver != nil) {
     _registerPromiseResolver(@([RCTConvert BOOL:@([UIApplication sharedApplication].isRegisteredForRemoteNotifications)]));
     _registerPromiseResolver = nil;
--- a/packages/messaging/ios/RNFBMessaging/RNFBMessagingModule.m
+++ b/packages/messaging/ios/RNFBMessaging/RNFBMessagingModule.m
@@ -181,10 +181,6 @@ RCT_EXPORT_METHOD(registerForRemoteNotifications:
   (RCTPromiseResolveBlock) resolve
     : (RCTPromiseRejectBlock) reject
 ) {
-  if ([UIApplication sharedApplication].isRegisteredForRemoteNotifications == YES) {
-    return resolve(@([RCTConvert BOOL:@(YES)]));
-  }
-
   [[RNFBMessagingAppDelegateInterceptor sharedInstance] setPromiseResolve:resolve andPromiseReject:reject];
   [[UIApplication sharedApplication] registerForRemoteNotifications];
 }


elyalvarado:
Hi @Salakar, @Ehesp. Can you review this? It fixes #2657, without changing the API as #3130

mikehardy:
Certainly seems well-reasoned and is nice and small+targeted
Doesn't look like it would apply to v5 the code is all different there
The ci.ios.test failure appeared unrelated, something about firestore where clauses that seemed like a timeout to me, I restarted that task

elyalvarado:
The full explanation of what is going on with the current fixes and work arounds is in this [comment in issue #2657](https://github.com/invertase/react-native-firebase/issues/2657#issuecomment-594722675)

elyalvarado:
This also solves #2995 APNS Token is always null

ptgamr:
+1, let get this merge, please!

robertqin86:
@elyalvarado So I tried your PR, and `[FIRMessaging messaging].APNSToken = deviceToken;` works. But the removal of ```if ([UIApplication sharedApplication].isRegisteredForRemoteNotifications == YES) {
    return resolve(@([RCTConvert BOOL:@(YES)]));
  }``` still does not resolve the race condition that you talked about. I was still getting the error. 

Instead, the surefire way that this worked for me is the revert the fix in `packages/messaging/ios/RNFBMessaging/RNFBMessagingModule.m` and keep the fix in `packages/messaging/ios/RNFBMessaging/RNFBMessagingAppDelegateInterceptor.m`.

In addition, just add `[[UIApplication sharedApplication] registerForRemoteNotifications];` in the first line of `didFinishLaunchingWithOptions`. Would this not be a better solution? Obviously, the docs would have to be updated.

elyalvarado:
@robertqin86 well, adding register for remote notifications in the AppDelegate assumes an use case that might not be what everybody wants. Some apps require user approval before registering, or might not need to do it until later.
The condition in `RNFBMessagingModule.m` should still not be there, as I mentioned in native code you're free to call `registerForRemoteNotifications` whenever you want regardless of whether the app already did it or not, and in addition to that there is still the issue of `isRegisteredForRemoteNotifications` returning true after restarting the app, which will prevent it from being called.
What device do you have? And why do you say the race condition is not solved in your case? With the PR adding `registerForRemoteNotifications` in the AppDelegate is basically the same than calling `registerForRemoteNotifications` in react-native. Are you calling it before calling `getToken`, because if you don't do it it'll still give an error?

robertqin86:
Hi @elyalvarado, you're absolutely right. I didn't think of the use case where people might want to defer registerForRemoteNotifications 😓 . Sorry about this. 

I retried the fix again and removed my hard coded registerForRemoteNotifications  in the app delegate and the fix works now.  

// This was my fix
```
if (!messaging.isRegisteredForRemoteNotifications) {
        await messaging.registerForRemoteNotifications();
}
 const granted = await messaging.requestPermission();
      if (granted) {

elyalvarado:
@robertqin86 no need to be sorry about it. Thanks for taking the time to test the PR. The more people that test it the better for all of us.

CaptBoscho:
When will this fix be published?

mikehardy:
@Salakar as a side note, this is something that is still hard to figure out in v6 adoption - for v5 I could easily use patch-package or a local fork as a dependency, but for v6, how can I successfully rely on a local fork? I'm still not aware of anything except maybe https://github.com/ramasilveyra/gitpkg

JanSunavec:
Just working.. It really fix problem with missing apns token..

russellwheatley:
fixes #3292

JanSunavec:
What about new release to make this project sence again?

mikehardy:
@JanSunavec you are a strong open source contributor, you know these things (especially with large user bases where releasing too soon can cause support problems) take some time. The repo is on fire though right now - multiple PRs merging a day. I'm guessing v7 will have a release soon. In the meantime - patch-package? I use it while waiting for things, personally

perrosnk:
Any updates on when this will be merged? Thank you for the great work!

ricechenghk:
Thanks for merging the commit into master branch. I know that there are a lot of pull request pending. May I know how to check out the master branch in package.json instead of v6.3.4? I tried different ways but still cannot check out the master branch....

JanSunavec:
You don not need checkout master MikeHardy mentions patch-package solution. In postinstall event in  package.json apply patch to 6.3.4 patch is special diff file. So just put 6.3.4 into package.json and in post istall apply patch and during compilation you will use patched version.. That's it..

mikehardy:
Yes - patch-package is a very solid way to get pre-release fixes, it works across machines, in CI etc - it meets all the criteria of a proper solution in my opinion, for release engineering. That said for your specific question because RNFBv6 is in a monorepo the only way I'm aware of to actually use master via package.json is to do the intermediate step of building and "releasing" via "gitpkg" https://github.com/ramasilveyra/gitpkg - I have not tested this, I use patch-package myself

dbilgin:
And yet I still have this very issue.

ricechenghk:
Thanks guys. The patch-package way works like a charm!!!

bsonmez:
When is it going to be merged?

JeffGuKang:
This PR is applied to https://github.com/invertase/react-native-firebase/pull/3339. 
Try 6.4.0-rc1.

PR #3278, opened 2020-03-07
feat(remote-config): implement updated APIs (v7)

russellwheatley:
## RemoteConfig Module
```js
const remoteConfig = firebase.remoteConfig();
```
### Updates
-  added `ensureInitialized ` API.
- `console.warn()` if user tries to set `defaultConfig` which is part of web sdk.
- `console.warn()` if user tries to set `settings` which is part of web sdk.
- `console.warn()` if user tries to set `setLogLevel` which is part of web sdk.

## RemoteConfig.setConfigSettings()
```js
const remoteConfig = firebase.remoteConfig().setConfigSettings({});
```
### Updates
- can set 'minimumFetchIntervalMillis' in `setConfigSettings` to match web sdk.
- can set 'fetchTimeMillis' in `setConfigSettings` to match web sdk.

### Removed
- `isDeveloperModeEnabled` from config settings and console.warn() if tried to set.
- `minimumFetchInterval ` config setting and console.warn() if tried to set.

## RemoteConfig.getValue(key)
```js
const remoteConfig = firebase.remoteConfig().getValue('key');
```
### Updates
- `asBoolean()` resolves value to a boolean.
- `asNumber()` resolves value to a number.
- `asString()` resolves value to a string.
- `getSource()` source of the property.

### Removed
- `value` removed property. Not sure how to warn users it is no longer available
- `source` removed property. Again, not sure how to warn users.

## Internal Changes
- Switched to `setConfigSettingsAsync ` for Android, nothing similar for iOS.
- Switched to `fetchAndActivate` API for both platforms.
- Switched to async `activate` API for iOS. No changes needed for Android.
- Switched to multi-app support for both platforms.

## Comments
- I had to comment out a test for `lastFetchStatus` as it is coming back as `null` oddly enough as it should be 0 according to: https://firebase.google.com/docs/reference/android/com/google/firebase/remoteconfig/FirebaseRemoteConfig#LAST_FETCH_STATUS_NO_FETCH_YET
- Also had to update JS code for `lastFetchTime` as it was also coming back as `null` if no previous fetch had been made. 
- Test coverage is almost 100%. Not mocking the `console.warn()` calls is stopping full coverage.

### DIFF
--- a/docs/remote-config/quick-start.md
+++ b/docs/remote-config/quick-start.md
@@ -62,14 +62,14 @@ async function getValues() {
 
 In some cases you may want to fetch values from the console in the background without the process visibly impacting
 your application. To prevent any race conditions where values are being requested (i.e. via `getValue`) before they
-have been fetched and activated, it is recommended you set default values using the `defaultConfig` property:
+have been fetched and activated, it is recommended you set default values using the `setDefaults` property:
 
 ```js
 import remoteConfig from '@react-native-firebase/remote-config';
 async function bootstrap() {
-      firebase.remoteConfig().defaultConfig = {
-        experiment_enabled: false,
-      };
+  await remoteConfig().setDefaults({
+    experiment: false,
+  });
 }
 ```
 
--- a/docs/remote-config/quick-start.md
+++ b/docs/remote-config/quick-start.md
@@ -62,7 +62,7 @@ async function getValues() {
 
 In some cases you may want to fetch values from the console in the background without the process visibly impacting
 your application. To prevent any race conditions where values are being requested (i.e. via `getValue`) before they
-have been fetched and activated, it is recommended you set default values using the `setDefaults` property:
+have been fetched and activated, it is recommended you set default values using the `setDefaults` method:
 
 ```js
 import remoteConfig from '@react-native-firebase/remote-config';
--- a/docs/remote-config/quick-start.md
+++ b/docs/remote-config/quick-start.md
@@ -62,15 +62,14 @@ async function getValues() {
 
 In some cases you may want to fetch values from the console in the background without the process visibly impacting
 your application. To prevent any race conditions where values are being requested (i.e. via `getValue`) before they
-have been fetched and activated, it is recommended you set default values using `setDefaults`:
+have been fetched and activated, it is recommended you set default values using the `defaultConfig` property:
 
 ```js
 import remoteConfig from '@react-native-firebase/remote-config';
-
 async function bootstrap() {
-  await remoteConfig().setDefaults({
-    experiment: false,
-  });
+      firebase.remoteConfig().defaultConfig = {
+        experiment_enabled: false,
+      };
 }
 ```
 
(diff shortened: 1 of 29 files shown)


russellwheatley:
Thought I'd make this draft PR so I could ask for advice on `ensureInitialized` method in Java which became the bane of my life. I was tripping up on the idea that I needed to return this abstract class `FirebaseRemoteConfigInfo` with my own implementation. The penny has finally dropped that the Task actually returns an already implemented `FirebaseRemoteConfigInfo`. Anyway, got it working now. But I'll leave this draft PR here anyway 👍

packages/remote-config/lib/index.d.ts
```diff
@@ -380,49 +410,6 @@ export namespace FirebaseRemoteConfigTypes {
      * @param key A key used to retrieve a specific value.
      */
     getValue(key: string): ConfigValue;
-
```
Salakar: Just a friendly reminder these need adding back in 😅

docs/remote-config/quick-start.md
```diff
@@ -62,11 +62,10 @@ async function getValues() {
 
 In some cases you may want to fetch values from the console in the background without the process visibly impacting
 your application. To prevent any race conditions where values are being requested (i.e. via `getValue`) before they
-have been fetched and activated, it is recommended you set default values using `setDefaults`:
+have been fetched and activated, it is recommended you set default values using the `setDefaults` property:
```
Salakar:
```suggestion
have been fetched and activated, it is recommended you set default values using the `setDefaults` method:
```

packages/remote-config/ios/RNFBConfig/RNFBConfigModule.m
```diff
@@ -175,55 +215,42 @@ + (BOOL)requiresMainQueueSetup {
...
-- (NSDictionary *)resultWithConstants:(id)result {
+- (NSDictionary *)resultWithConstants:(id) result firebaseApp:(FIRApp *) firebaseApp {
   NSMutableDictionary *responseDict = [NSMutableDictionary new];
   responseDict[@"result"] = result;
-  responseDict[@"constants"] = [self getConstantsForApp];
+  responseDict[@"constants"] = [self getConstantsForApp:firebaseApp ];
   return responseDict;
 }
 
 // TODO only for default app, add support for multiple apps once SDK supports it
```
Salakar:
```suggestion
```

packages/remote-config/lib/index.js
```diff
@@ -79,12 +79,48 @@ function convertNativeConfigValues(configValues) {
...
+  }
+}
 
 class FirebaseConfigModule extends FirebaseModule {
   constructor(...args) {
     super(...args);
+    this._settings = {};
+    this._lastFetchStatus = null;
+    this._lastFetchTime = null;
     // TODO(salakar) iOS does not yet support multiple apps, for now we'll use the default app always
```
Salakar:
```suggestion
```

PR #3302, opened 2020-03-14
[BUGFIX][Notifications][Android] - fix false negative getInitialNotification call

kirillpisarev:
[BUGFIX][Notifications][Android] - fix false negative getInitialNotification call

Fixes the situation when `getInitialNotification` fires with an empty `notificationOpen` object even if the app was opened via notification press. It happens because in the native method implementation `getCurrentActivity` returns null. Under the hood mCurrentActivity is set on `Activity.onResume` event, so the fix is to wait for this event and try to reach the notification data once again.

### DIFF
--- a/android/src/main/java/io/invertase/firebase/notifications/RNFirebaseNotifications.java
+++ b/android/src/main/java/io/invertase/firebase/notifications/RNFirebaseNotifications.java
@@ -13,6 +13,7 @@ import android.util.Log;
 
 import com.facebook.react.bridge.ActivityEventListener;
 import com.facebook.react.bridge.Arguments;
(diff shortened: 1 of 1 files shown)


mikehardy:
Hi there @kirillpisarev ! Thank you! This looks reasonable and I have one other change I could do a v5.x.x release with, but we need the CLA signed

kirillpisarev:
Now the CLA is signed

russellwheatley:
Hey @kirillpisarev, thank you for your pull request. We’re no longer actively supporting `<= v5` development so I’m going to go ahead and close this. If this PR change still applies to the latest version of React Native Firebase then please feel free to raise a new PR targeting the latest version with your changes.

Issue #2592, opened 2019-09-17
🔥 Crash - Tried to start task RNFirebaseBackgroundMessage while in foreground, but this is not allowed.


vrozaev:
## Issue

🔥 Crash - Tried to start task RNFirebaseBackgroundMessage while in foreground, but this is not allowed.

Looks like this issue should be fixed in https://github.com/invertase/react-native-firebase/pull/1491 but we still can reproduce it.

The exception happens in this line of react-native code: https://github.com/facebook/react-native/blob/master/ReactAndroid/src/main/java/com/facebook/react/jstasks/HeadlessJsTaskContext.java#L98

Looks like that happen because reactContext.getLifecycleState() == LifecycleState.RESUMED condition is true.

Stacktrace: 

com.facebook.react.jstasks.HeadlessJsTaskContext.startTask
HeadlessJsTaskContext.java, line 95
java.lang.IllegalStateException: Tried to start task RNFirebaseBackgroundMessage while in foreground, but this is not allowed.
com.facebook.react.jstasks.HeadlessJsTaskContext.startTask HeadlessJsTaskContext.java:95
com.facebook.react.HeadlessJsTaskService$2.run HeadlessJsTaskService.java:127
android.os.Handler.handleCallback Handler.java:739
android.os.Handler.dispatchMessage Handler.java:95
android.os.Looper.loop Looper.java:158
android.app.ActivityThread.main ActivityThread.java:7224
java.lang.reflect.Method.invoke Method.java
com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run ZygoteInit.java:1230
com.android.internal.os.ZygoteInit.main ZygoteInit.java:1120

---

### iOS

<details><summary>Click To Expand</summary>
<p>

#### `ios/Podfile`:

- [ ] I'm not using Pods
- [x] I'm using Pods and my Podfile looks like:

```ruby
# N/A
```

#### `AppDelegate.m`:

```objc
// N/A
```

</p>
</details>

---

### Android

<details><summary>Click To Expand</summary>
<p>

#### Have you converted to AndroidX?

- [ ] my application is an AndroidX application?
- [ ] I am using `android/gradle.settings` `jetifier=true` for Android compatibility?
- [ ] I am using the NPM package `jetifier` for react-native compatibility?

#### `android/build.gradle`:

```groovy
// Top-level build file where you can add configuration options common to all sub-projects/modules.

buildscript {
    repositories {
        google()
        jcenter()
        maven {
[...]

PR #3311, opened 2020-03-17
fix(messaging): allow RNFirebaseMessagingHeadlessTask to run in foreground

daisy1754:
### Summary

This PR fixes https://github.com/invertase/react-native-firebase/issues/2592

#2592 is caused by race condition. While current logic [only executes this code after checking app is not in foreground](https://github.com/invertase/react-native-firebase/blob/f429699990b6d521d996479f2bed96b52bbdb8d8/packages/messaging/android/src/main/java/io/invertase/firebase/messaging/ReactNativeFirebaseMessagingService.java#L72), it's possible that app gets foreground when execution reached [here](https://github.com/facebook/react-native/blob/b55146f7763e4e9d6cc2fd54dde62d4e4289721f/ReactAndroid/src/main/java/com/facebook/react/jstasks/HeadlessJsTaskContext.java#L103) 

(See also 
https://github.com/invertase/react-native-firebase/issues/2592#issuecomment-584190439 ) for how to reproduce.

Fix is rather simple, we just need to specify `allowedInForeground` flag to true (documentation - https://github.com/facebook/react-native/blob/722feeb02b5dbc8aeb5d975feb61ccedf4d9e07e/ReactAndroid/src/main/java/com/facebook/react/jstasks/HeadlessJsTaskConfig.java#L48 ). Now instead of crashing app, service would just run in UI thread (which is better than crash :))

### Checklist

- [ x ] Supports `Android`
- [ ] Supports `iOS`
- [ ] Flow types updated

### Test Plan

You can test this PR by:
1. close app
2. send notification
3. open app
if you're unlucky you'll get crash with this PR. With this PR you won't crash

### Release Plan

[Android][BUGFIX] [FirebaseMessaging] - Fix crash #2592

### DIFF
--- a/packages/messaging/android/src/main/java/io/invertase/firebase/messaging/ReactNativeFirebaseMessagingHeadlessService.java
+++ b/packages/messaging/android/src/main/java/io/invertase/firebase/messaging/ReactNativeFirebaseMessagingHeadlessService.java
@@ -25,7 +25,7 @@ public class ReactNativeFirebaseMessagingHeadlessService extends HeadlessJsTaskS
       TASK_KEY,
       ReactNativeFirebaseMessagingSerializer.remoteMessageToWritableMap(remoteMessage),
       ReactNativeFirebaseJSON.getSharedInstance().getLongValue(TIMEOUT_JSON_KEY, TIMEOUT_DEFAULT),
-      false
+      true /* allow to run foreground */
     );
   }
 }


Ehesp:
Thanks for this, it makes sense. Although bear in mind this is for version 6, not 5.

mikehardy:
I would accept a version 5 PR and do a release there also, as stated before :) - thanks

Issue #3312, opened 2020-03-17
Content-available is missing from the remote message


fruedaCode:
I have made the following change in order to emit the notification from the native code to he RN listener.
https://github.com/invertase/react-native-firebase/blob/cf0bbdcaf06a3f1ba8c210ef1c03bcfff10b5522/packages/messaging/ios/RNFBMessaging/RNFBMessagingAppDelegateInterceptor.m#L67

Instead of content-avaiable I'm receiving @alert the data in an entry called alert.
![image](https://user-images.githubusercontent.com/22854367/76872973-6f860700-686d-11ea-824d-59c62330a135.png)

So the line should look like this for my case:
```Objective-C
if (userInfo[@"aps"] && ((NSDictionary *) userInfo[@"aps"]).count >= 1 && userInfo[@"aps"][@"alert"]) {
```

Do you think you can include this change please?

Thanks!

Ehesp:
Can you submit a PR so we can review please. Thanks

PR #3309, opened 2020-03-16
feature(messaging): android messaging upgrade

Ehesp:
This upgrade includes the following changes:

- Rename of `registerForRemoteNotifications` to `registerDeviceForRemoteMessages`. This will be a deprecation and will be removed in v7. Nothing internally is changing, however the current wording is just totally confusing.
- The background handler will be ALWAYS set, and if you dont provide a handler to `setBackgroundMessageHandler` it'll throw a better/custom warning (e.g. not a React Native one).
- `onMessage` will work for data-only, notification-only AND data-notification messages. Currently it only works for data-only.
- `setBackgroundMessageHandler` will now work when the app is both fully quit and backgrounded, for data-only, notification-only AND data-notification messages. 
- We're implementing both `getInitialNotification` & `onNotificationOpenedApp` into the messaging module. This won't require any native changes from users.
- If a message is sent with a notification payload, it'll be included within the `RemoteMessage` object passed around.

Fixes #3209, #3032, #3293, #3236

PR #3314, opened 2020-03-17
change aps user key content-available by alert

fruedaCode:
Issue created here:
https://github.com/invertase/react-native-firebase/issues/3312

### Checklist

- [ X] Supports `Android`
- [ X] Supports `iOS`
- [ ] Flow types updated


Ehesp:
Currently looking into messaging for iOS, I can see what you're trying to do here however it's not quite right. I may close this off as I'm changing a few things around, but generally this will be fixed. Instead of doing those checks, we're going to check only if it's a valid FCM message and send an event.

Salakar:
Quick note that #3309 (RC is also available on NPM - 6.4.0-rc2) now supersedes this.

Salakar:
Closing as #3339 which supersedes this has now been merged and is available on an RC release (6.4.0-rc4).

Thank you for your contribution

PR #3317, opened 2020-03-18
Fix for 5.x - allow RNFirebaseBackgroundMessagingService to run in foreground

daisy1754:
Same as https://github.com/invertase/react-native-firebase/pull/3311 but for v5.x branch. Below is copy from #3311 (class name is different but issue and fix is same)

### Summary

This PR fixes https://github.com/invertase/react-native-firebase/issues/2592

#2592 is caused by race condition. While current logic [only executes this code after checking app is not in foreground](https://github.com/invertase/react-native-firebase/blob/f429699990b6d521d996479f2bed96b52bbdb8d8/packages/messaging/android/src/main/java/io/invertase/firebase/messaging/ReactNativeFirebaseMessagingService.java#L72), it's possible that app gets foreground when execution reached [here](https://github.com/facebook/react-native/blob/b55146f7763e4e9d6cc2fd54dde62d4e4289721f/ReactAndroid/src/main/java/com/facebook/react/jstasks/HeadlessJsTaskContext.java#L103) 

(See also 
https://github.com/invertase/react-native-firebase/issues/2592#issuecomment-584190439 ) for how to reproduce.

Fix is rather simple, we just need to specify `allowedInForeground` flag to true (documentation - https://github.com/facebook/react-native/blob/722feeb02b5dbc8aeb5d975feb61ccedf4d9e07e/ReactAndroid/src/main/java/com/facebook/react/jstasks/HeadlessJsTaskConfig.java#L48 ). Now instead of crashing app, service would just run in UI thread (which is better than crash :))

### Checklist

- [ x ] Supports `Android`
- [ ] Supports `iOS`
- [ ] Flow types updated

### Test Plan

You can test this PR by:
1. close app
2. send notification
3. open app
if you're unlucky you'll get crash with this PR. With this PR you won't crash

### Release Plan

[Android][BUGFIX] [FirebaseMessaging] - Fix crash #2592

### DIFF
--- a/android/src/main/java/io/invertase/firebase/messaging/RNFirebaseBackgroundMessagingService.java
+++ b/android/src/main/java/io/invertase/firebase/messaging/RNFirebaseBackgroundMessagingService.java
@@ -22,7 +22,7 @@ public class RNFirebaseBackgroundMessagingService extends HeadlessJsTaskService 
         "RNFirebaseBackgroundMessage",
         messageMap,
         60000,
-        false
+        true /* allows to run in foreground */
       );
     }
     return null;


daisy1754:
cc @mikehardy (thanks for reminding me that there are 5.x branch :) )

mikehardy:
Got it! Thanks @daisy1754 - please use patch-package to incorporate this change in the meantime (before release) and you may be confident that the next v5.x release will have this change in it (patch-package will let you know, it's safe+easy to use)

SNWStaff:
Is there a v5 release to get the fix on V5 ? 

Thanks

mikehardy:
Sorry - been really busy on other things, I'll look through the v5.x.x tree and try to get a 5.6.1 out soon. patch-package *will* work though, you should not be blocked on it

Cirych:
@daisy1754 Thank you so much for your fix!

@mikehardy Are there any chances to get 5.6.1 in the nearest future? Should we install packet from the 5.x.x for now?

mikehardy:
Well, separate subjects but I will say that I can't imagine having a react-native project *without* patch-package integrated, there is always some patch for something that I am carrying around. So do that first to take any pressure off. It's a very well done package

Other than that I have to admit (though I did not come back and say so on this issue) I do not currently have plans to put out another 5.x.x release - contrary to my comment above sorry. If there were a big stop-work feature that someone implemented like backporting the switch to firebase crashlytics I could imagine doing it, but for easily patched / focused things no - there is enough to do with my work project and v6 I'd like to keep the focus on v6. Migrating to the new version is important, it should be scheduled and done is my opinion and this is a nudge to do so

PR #3075, opened 2020-01-07
feat(messaging:) use prominent iOS push notification permission

anfriis:
This PR solves the "Provisional push authorization should be optional on iOS" post on the Feedback site: https://invertase.io/oss/react-native-firebase/feedback/

The `requestPermissionWithoutProvisional()` requests push permission for iOS without the `UNAuthorizationOptionProvisional` auth option set, showing the native permission alert to the user.
The current `requestPermission()` has no option for leaving out the provisional auth option. 

### Summary
It is currently not possible to explicitly ask for push permission for iOS 12+, because the `UNAuthorizationOptionProvisional` flag is set making push notifications silent for the user until the user actively accepts prominent permission (more info on Provisional Authorization [here](http://info.localytics.com/blog/ios-12-brings-new-power-to-push-notifications)). In some cases, e.g. the commercial app I am currently working on, it is a requirement to explicitly ask for push permission. 
This PR adds the possibility to explicitly ask for push permission with `requestPermissionWithoutProvisional()` without changing the current functionality of `requestPermission()`.

### Checklist

- [x] Supports `Android` (doesn't affect Android)
- [x] Supports `iOS`
- [x] `e2e` tests added or updated in packages/**/e2e
- [x] Flow types updated
- [x] Typescript types updated

### Test Plan
The "\<app name\> Would Like To Send You Notifications" permission alert is shown on iOS when calling `requestPermissionWithoutProvisional()`. When calling `requestPermission()` instead no alert is shown on iOS 12+ since it has the `UNAuthorizationOptionProvisional` auth option set.

### Release Plan

 [IOS] [ENHANCEMENT] {Messaging} - Add prominent push notification permission option to Messaging

PR #3339, opened 2020-03-23
feat(messaging): ios & android messaging updates & fixes

### LINKED ISSUE #1814: 🔥 RNFirebaseMessaging is intercepting FCM delegate call
We are using FCM and expect the following function to be called

`AppDelegate's 
 - (void)messaging:(FIRMessaging *)messaging didReceiveRegistrationToken:(NSString *)fcmToken {`

However `RNFirebaseMessaging.m` 's same function (signature) gets called instead

When we shake device and reload the js, RNFirebaseMessaging.m 's init method gets called and it is setting the delegate to itself. 

When we run the app by running it from xcode, this sometimes happens and sometimes doesn't..  
There's no explicit code to call RNFirebaseMessaging's init anywhere in our code. (We want our AppDelegate to be the delegate of the FCM messaging ..

## Issue

---

#### `ios/Podfile`:

- [ ] I'm not using Pods
 
```ruby
platform :ios, '9.0'
 use_frameworks!
 target "littlehome" do
   rn_maps_path = '../node_modules/react-native-maps'

   pod 'React', :path => '../node_modules/react-native', :subspecs => [
     'Core',
     'RCTText',
     'RCTImage',
     'RCTNetwork',
     'RCTSettings',
     'RCTLinkingIOS',
     'DevSupport',
     'RCTAnimation',
     'RCTWebSocket', # needed for localhost testing of your app
     'CxxBridge',
     #'BatchedBridge',
     'RCTGeolocation'
     # Add any other subspecs you want to use in your project
   ]
   pod "yoga", :path => "../node_modules/react-native/ReactCommon/yoga"
   pod 'Firebase', '4.11.0'
   pod 'Firebase/Core', '4.11.0'
   pod 'Firebase/Messaging', '4.11.0'
   pod 'Firebase/DynamicLinks', '4.11.0'
   pod 'Firebase/AdMob', '4.11.0'
   pod 'Firebase/RemoteConfig', '4.11.0'

   #---https://github.com/facebook/react-native/issues/20031 end//

   # react-native-maps dependencies
   pod 'react-native-maps', path: rn_maps_path
   # pod 'react-native-google-maps', path: rn_maps_path  # Remove this line if you don't want to support GoogleMaps on iOS
   # https://github.com/react-community/react-native-maps/issues/1582
   pod 'GoogleMaps'  # Remove this line if you don't want to support GoogleMaps on iOS
   # pod 'Google-Maps-iOS-Utils' # Remove this line if you don't want to support GoogleMaps on iOS
   # https://github.com/googlemaps/google-maps-ios-utils/blob/master/Swift.md

   pod 'SentryReactNative', :path => '../node_modules/react-native-sentry'
   pod 'BugsnagReactNative', :path => '../node_modules/bugsnag-react-native'

   pod 'RNDeviceInfo', :path => '../node_modules/react-native-device-info'

 end


 post_install do |installer|
   installer.pods_project.targets.each do |target|
       if target.name == 'yoga'
           target.build_configurations.each do |config|
               config.build_settings['GCC_TREAT_WARNINGS_AS_ERRORS'] = 'NO'
               config.build_settings['GCC_WARN_64_TO_32_BIT_CONVERSION'] = 'NO'
           end
       end

     target.build_configurations.each do |config|
       config.build_settings['SWIFT_VERSION'] = '3.0'
     end
   end
 end


 pre_install do |installer|
   # workaround for https://github.com/CocoaPods/CocoaPods/issues/3289
   Pod::Installer::Xcode::TargetValidator.send(:define_method, :verify_no_static_framework_transitive_dependencies) {}
 end

 # [!] [Xcodeproj] Generated duplicate UUIDs:
 # https://github.com/CocoaPods/CocoaPods/issues/4370
 install! 'cocoapods',
          :deterministic_uuids => false

```

#### `AppDelegate.m`:

```objc
   [FIRApp configure];

   // fcm
   [FIRMessaging messaging].delegate = self;

   if ([UNUserNotificationCenter class] != nil) {
     // iOS 10 or later
     // For iOS 10 display notification (sent via APNS)
     [UNUserNotificationCenter currentNotificationCenter].delegate = self;
     UNAuthorizationOptions authOptions = UNAuthorizationOptionAlert |
       UNAuthorizationOptionSound | UNAuthorizationOptionBadge;
     [[UNUserNotificationCenter currentNotificationCenter]
       requestAuthorizationWithOptions:authOptions
                     completionHandler:^(BOOL granted, NSError * _Nullable error) {
         // ...
       }];
   } else {
     // iOS 10 notifications aren't available; fall back to iOS 8-9 notifications.
     UIUserNotificationType allNotificationTypes =
       (UIUserNotificationTypeSound | UIUserNotificationTypeAlert | UIUserNotificationTypeBadge);
     UIUserNotificationSettings *settings =
       [UIUserNotificationSettings settingsForTypes:allNotificationTypes categories:nil];
     [application registerUserNotificationSettings:settings];
   }

   [application registerForRemoteNotifications];
   // fcm
```

---

#### `android/build.gradle`:

```groovy
// N/A
```

```groovy
// N/A
```

```groovy
// N/A
```

#### `MainApplication.java`:

```java
// N/A
```

#### `AndroidManifest.xml`:

```xml
<!-- N/A -->
```

---

## Environment

  - [x] iOS
  - [ ] Android
  - [ ] Both
- **If known, the version of the platform are you experiencing the issue on:**
  - `ADD_SOMETHING_HERE` e.g. iOS 10 or Android API 28
- **Operating System:**
  - [x] MacOS, version: `N/A`
  - [ ] Windows, version: `N/A`
  - [ ] Other, please specify: `N/A`
- **Build Tools:**
  - Xcode 10.1
- **`React Native` version:**
  - `0.57.4`
- **`React Native Firebase` library version:**
  - `5.1.0`
  - [ ] **N/A**
  - [ ] Authentication
  - [ ] Analytics
  - [ ] Cloud **Firestore**
  - [x] Cloud **Messaging** (FCM)
  - [ ] Crashlytics
  - [ ] Dynamic **Links**
  - [ ] **Functions** Callable
  - [ ] Invites
  - [ ] Instance ID
  - [ ] Notifications
  - [ ] Performance Monitoring
  - [ ] Realtime **Database**
  - [ ] Remote **Config**
  - [ ] Storage
  - [x] No
  - [ ] Yes, version: `N/A`
- **Are you using Expo, e.g. `ExpoKit`?**
  - [x] No
  - [ ] Yes, I've _not_ ejected
  - [ ] Yes, but I **have** ejected to `ExpoKit`
  - [ ] Yes, but I **have** ejected to vanilla React Native
  - Expo version: `N/A`

---

- 👉 Contribute; see our [contributing guide](./../../CONTRIBUTING.md)

Ehesp:
This PR is a follow-on to the android only updates (https://github.com/invertase/react-native-firebase/pull/3309) now in `6.4.0-rc2`.

## Android & iOS

- [FEAT] Messaging now supports `onNotificationOpenedApp` & also `getInitialNotification` APIs. These are used to detect if a user opened the app via pressing a notification.

- [FEAT] The `RemoteMessage` API now includes a notification payload (if present) when being sent to the messaging APIs (e.g. `onMessage`).

## Android

- [FIX] Ensure a color is always present when user does not set one (https://github.com/invertase/react-native-firebase/pull/3309#issuecomment-602253815). 

- [FIX] If a background message is received but no handler (via `setBackgroundMessageHandler`) is registered, React Native would throw an unhelpful warning. This has now been reworked to provide better handling/messages in cases where no handler is set. Via #3309.

- [FIX] `onMessage` will work for data-only, notification-only AND data-notification messages. Currently it only works for data-only. Via #3309.

- [FIX] `setBackgroundMessageHandler` will now work when the app is both fully quit and backgrounded, for data-only, notification-only AND data-notification messages. Previously, when the app was fully quit, events were not being handled. Via #3309.

## iOS

We've spent a massive amount of time on iOS over the last two weeks, figuring out the quirks and going over the main issues on the library. iOS messaging is just overall super confusing compared to Android, especially the initial setup with certificates etc. Hopefully this PR will help drastically 🙏 

- [FEAT] `setBackgroundMessageHandler` support for iOS!

- [FEAT] The permissions API has been upgraded to now support custom permissions. The permission API selects sensible defaults however allows you to fully customise them if required. Please note, provisional permissions is now disabled by default (previously, it was enabled by default for iOS 12+ devices):

```js
// Default permissions:
await messaging().requestPermission({
  alert: true,
  announcement: false,
  badge: true,
  carPlay: true,
  provisional: false,
  sound: true,
});
```

- [FEAT] The `requestPermission` & `hasPermission` API now returns the current authorization status, rather than a boolean value. This generally won't break existing apps as any "truthy" logic checks should still work:

```js
const permission = await messaging().requestPermission();
// or const permission = await messaging().hasPermission();

if (permission === IOSAuthorizationStatus.NOT_DETERMINED) {
  // -1 =  No permission has been requested yet
}

if (permission === IOSAuthorizationStatus.DENIED) {
  // 0 =  User declined permission. They must enable permissions from the app settings
}

if (permission === IOSAuthorizationStatus.AUTHORIZED) {
  // 1 =  User granted permission
}

if (permission === IOSAuthorizationStatus.PROVISIONAL) {
  // 2 =  User accepted provisional permission
}
```

- [DEPRECATION] `registerForRemoteNotifications` has been deprecated in favour of `registerDeviceForRemoteMessages`. It will be removed in v7. Underlying functionality has not changed, just renamed to avoid confusion with messages vs notifications.

- [DEPRECATION] `isRegisteredForRemoteNotifications` has been deprecated in favour of `isDeviceRegisteredForRemoteMessages`. It will be removed in v7. Underlying functionality has not changed, just renamed to avoid confusion with messages vs notifications.

- [DEPRECATION] `unregisterForRemoteNotifications` has been deprecated in favour of `unregisterDeviceForRemoteMessages`. It will be removed in v7. Underlying functionality has not changed, just renamed to avoid confusion with messages vs notifications.

- [FIX] Calling `registerDeviceForRemoteMessages`/`registerForRemoteNotifications` was causing permissions to be requested before explicitly requesting them via the messaging API. 

- [FIX] Fixed an issue whereby registering the device was not being called if it was already registered internally. Devices should always register when `registerDeviceForRemoteMessages` is called as per Apple guidelines, irregardless of current registration status. Make sure you always call `registerDeviceForRemoteMessages` during your app initialization on iOS.

- [FIX] In cases where requesting an FCM with default scope/authorizedEntity, the   underlying code now uses the recommended `instanceIDWithHandler` vs `tokenWithAuthorizedEntity`. This fixes an issue where FCM would throw a `"The operation couldn’t be completed"` error. Extension of #3130.

- [FIX] Direct FCM connection is now fixed. When the app was in the foreground, data-only messages were not coming through. 

- [FIX] When in debug mode, the APNS token will be set as a "sandbox" key type as per the Apple guidelines.

- [FIX] The original APNS swizzling we implemented was not functioning correctly with `application:didReceiveRemoteNotification:fetchCompletionHandler:`. We added additional logic whereby this is executed in all scenarios (foreground/background/quit) and replaces a deprecated Apple API. This fixes issues with data-only messages not being handled by the device. 

- [FIX] When the application is in the foreground, notification messages were not being sent to the `onMessage` handler (via `willPresentNotification`). This is now fixed.

- [FIX] `FIRMessagingDelegate` methods added to your `AppDelegate.m` will also be called (fixes #1814)


Salakar:
Hey all, v6.4.0-rc1 has been published to NPM with all these changes, please try it out, thanks

tomsotte:
Ive tested the `v6.4.0-rc1` on iPhone 5C, iOS 10, RN 0.61.2, react-native-notifications 3.1.2.

I've tried so send a notification with this format:

```js
{
	data: {title, body},
	apns: {
		headers: {'apns-push-type': 'alert'},
		payload: {aps: {title, body}
	}
}
```

This is what I've found:

- on foreground the `onMessage` callback gets called endlessly and it seems to not cause the notifications to show anymore, when on previous version, this worked; when I lock the phone a notification appears on the lock screen and I can access the app.
- on background (app not focused or on phone locked) it now receives the notification and it display an alert correctly

This is what I've tested so far.

Ehesp:
@tomsotte thanks for testing! 

> on foreground the onMessage callback gets called endlessly and it seems to not cause the notifications to show anymore, when on previous version, this worked; when I lock the phone a notification appears on the lock screen and I can access the app.

Could you expand on "gets called endlessly"? When the device is in the foreground it won't show a notification (same behaviour on Android & iOS). FCM shows a notification only when the device is in the background or quit.

> on background (app not focused or on phone locked) it now receives the notification and it display an alert correctly

Great. You should be able to handle the user opening your app via `getInitialNotification`.

tomsotte:
> @tomsotte thanks for testing!
> 
> > on foreground the onMessage callback gets called endlessly and it seems to not cause the notifications to show anymore, when on previous version, this worked; when I lock the phone a notification appears on the lock screen and I can access the app.
> 
> Could you expand on "gets called endlessly"? When the device is in the foreground it won't show a notification (same behaviour on Android & iOS). FCM shows a notification only when the device is in the background or quit.
> 

I can see the `console.log` on the callback being called endlessly, showing the same last notification data as in: `onMessage {"data": {"body": "Prova iOS back", "silent": false, "title": "Prova iOS back"}}`.

All the callback does is `console.log` the notification and then `Notifications.postLocalNotification` (from `react-native-notifications`). The last command does not seems to get executed correctly because I does not show the alert anymore.

> > on background (app not focused or on phone locked) it now receives the notification and it display an alert correctly
> 
> Great. You should be able to handle the user opening your app via `getInitialNotification`.

I'll try this and report back. At the moment I'm using the other notification library to get the initial notification. Should I be using this one instead as it's more reliable or something?

Ehesp:
I haven't used `react-native-notifications` before, however there could be a conflict going on. If you comment out the code which displays a notification from that library, does it still keep logging?

ottob:
I tried rc1 in a typescript project and got:
```
Expected 1 arguments, but got 0.ts(2554)
index.d.ts(647, 23): An argument for 'IOSPermissions' was not provided.
```
using this code:
```
const permission = await messaging().requestPermission()
```
Is this expected? I thought the argument to requestPermission was optional if I wanted the defaults.

Ehesp:
Nope, that's a bug. Have just pushed a quick fix: https://github.com/invertase/react-native-firebase/pull/3339/commits/6de4ba83713b8d43425cd3e0e1f8d3dd6d46b550

I'd suggest editing locally or ts-ignoring for now until we get another release up. Thanks again for reporting.

tomsotte:
> I haven't used `react-native-notifications` before, however there could be a conflict going on. If you comment out the code which displays a notification from that library, does it still keep logging?

You're right, commenting the code to show the notification stopped the spam of calls. Although now I'm trying to show the notification as an alert because it does not seems to work that way anymore, on foreground. On background it shows.

I've tried to use the new `getInitialNotification` I get is always this object, either if I open the app from a notification or directly from home: `{"_40": 0, "_55": null, "_65": 0, "_72": null}`.

Ehesp:
> You're right, commenting the code to show the notification stopped the spam of calls. Although now I'm trying to show the notification as an alert because it does not seems to work that way anymore, on foreground. On background it shows.

I'll have to look into what intents that library listens to and see if we can handle it.

> I've tried to use the new getInitialNotification I get is always this object, either if I open the app from a notification or directly from home: {"\_40": 0, "\_55": null, "\_65": 0, "\_72": null}.

Are you waiting for the promise to resolve:

```js
const initialNotification = await messaging().getInitialNotification();
// or
messaging().getInitialNotification().then((initialNotification) => {...});
```

tomsotte:
> Are you waiting for the promise to resolve:
> 
> ```js
> const initialNotification = await messaging().getInitialNotification();
> // or
> messaging().getInitialNotification().then((initialNotification) => {...});
> ```

Oh my bad, I wasn't awaiting for it 😅 I tried to understand the changes but I did not figured it out as a Promise. I'll try it later on and report back!

Ehesp:
No worries. I'm working to get the new changes to the docs site live ASAP: https://rnfb-docs.netlify.com/messaging/usage

Loads has been documented for Messaging, however it's still WIP!

Salakar:
> * on foreground the `onMessage` callback gets called endlessly and it seems to not cause the notifications to show anymore, when on previous version, this worked; when I lock the phone a notification appears on the lock screen and I can access the app.

@tomsotte I've just pushed a fix for this - will publish RC2 shortly

packages/messaging/ios/RNFBMessaging/RNFBMessaging+UNUserNotificationCenter.m
```diff
@@ -53,16 +53,18 @@ - (nullable NSDictionary *)getInitialNotification {
...
+    // Don't send an event if contentAvailable is true - application:didReceiveRemoteNotification will send the event
+    // for us, we don't want to duplicate them
+    if (!notificationDict[@"contentAvailable"]) {
+      [[RNFBRCTEventEmitter shared] sendEventWithName:@"messaging_message_received" body:notificationDict];
+    }
 
-  // TODO in a later version allow customising completion options in JS code
-  completionHandler(UNNotificationPresentationOptionNone);
+    // TODO in a later version allow customising completion options in JS code
+    completionHandler(UNNotificationPresentationOptionNone);
```
Ehesp: Shouldn't this be outside of the if? not sure...

Salakar:
@tomsotte @ottob 6.4.0-rc2 is published with both those fixes (types & onMessage), please try it out - thanks

packages/messaging/ios/RNFBMessaging/RNFBMessaging+UNUserNotificationCenter.m
Salakar: We don't want to complete notifications that are presented from other libraries, if we complete them for other libraries with `UNNotificationPresentationOptionNone` then they won't show in the foreground if that is the users expected behaviour from other third party libs

Salakar: I'd imagine other libs would have their own `willPresentNotification` setup hopefully

ottob:
On a related note: it seems like the Xcode 11.4 simulator finally has some support for remote push notifications. Ive not yet played with it so can't say how well it works.
https://developer.apple.com/documentation/xcode_release_notes/xcode_11_4_release_notes#3530440

mikehardy:
> On a related note: it seems like the Xcode 11.4 simulator finally has some support for remote push notifications. Ive not yet played with it so can't say how well it works.
> https://developer.apple.com/documentation/xcode_release_notes/xcode_11_4_release_notes#3530440

That looks amazing! Thanks for the heads up @ottob !

ottob:
Typescript types works as expected in rc2, thanks.

My very basic use case seems to works as expected out of the box, great! I no longer need the getToken workaround and #3075. `onMessage` also triggers now, I never got that to work before.

One small thing I noticed was that I used to rely on the fact that `requestPermission` called `registerForRemoteNotifications` under the hood. This is no longer the case so my code broke. But it was trivial to fix this by calling `registerDeviceForRemoteMessages` before `getToken`.

Salakar:
@ottob glad to hear everything is working for you 🎉 

> One small thing I noticed was that I used to rely on the fact that `requestPermission` called `registerForRemoteNotifications` under the hood. This is no longer the case so my code broke. But it was trivial to fix this by calling `registerDeviceForRemoteMessages` before `getToken`.

We'll make this clearer on the release notes but I believe we did add an error message to `getToken` to make it clear that `registerDeviceForRemoteMessages` needs to be called first. 

Some background on why we made the change if you're curious;

 - Original implementation (v4 RNFB - v6) assumed `requestPermission` was also required to receive remote messages which is why it also called `registerForRemoteNotifications` - but this is not the case, `requestPermission` is only required to for ui alerts/visual interruptions - data only messages still come through even if `requestPermission` wasn't called.
 - Apple docs state that `registerForRemoteNotifications` should **always** be called during your app startup, irregardless of current registration status, we confirmed this behaviour as part of this PR - if you don't call it then your `AppDelegate` handlers such as `didReceiveRemoteNotification` would never get called, so `onMessage` etc would never trigger from App background/quit states. 
 - `requestPermission` for most apps should really be called at a point in your app where it makes the most sense rather than delaying startup, e.g. in a chat view and prompting to turn on notifications for chat messages 

With those points in mind it made sense that they should not be tied to each other.

packages/messaging/ios/RNFBMessaging/RNFBMessaging+UNUserNotificationCenter.m
anguyen1817: @Salakar Thank you for this pr! It's really helpful. I just have one question. For
```- (void)userNotificationCenter:(UNUserNotificationCenter *)center didReceiveNotificationResponse:(UNNotificationResponse *)response withCompletionHandler:(void (^)(void))completionHandler {
  NSDictionary *remoteNotification = response.notification.request.content.userInfo;
  if (remoteNotification[@"gcm.message_id"]) {
    NSDictionary *notificationDict = [RNFBMessagingSerializer remoteMessageUserInfoToDict:remoteNotification];
    [[RNFBRCTEventEmitter shared] sendEventWithName:@"messaging_notification_opened" body:notificationDict];
    _initialNotification = notificationDict;
  }
  completionHandler();
}
```
why, in this case, completionHandler is outside the if check?

ottob:
> but I believe we did add an error message to `getToken` to make it clear that `registerDeviceForRemoteMessages` needs to be called first.

Yeah, that made it really easy to fix :)

Thanks for the explanation. It's fascinating that something that could be seen as fairly simple like push messaging is so complex behind the scenes.

JeffGuKang:
I tested it with my applications focusing on iOS Remote Notification. 
First,  two iOS devices worked well, but another one was not.

- Worked devices
  - iPhone XR, OS: 13.3.1
  - iPhone  8+, OS: 13.3.
- Not worked device
  - iPhone 6s, OS: 13.2.3 => `getAPNSToken` is null. 

And I am trying bump os version of iPhone 6s not worked to 13.4 from 13.2.3. 

Is there any example project to run this?

otto-md:
@Ehesp I saw your [latest commit](https://github.com/invertase/react-native-firebase/pull/3339/commits/e54f4ec87f08eb62c8a44c6021352c1b1a95601c). Given that the Xcode 11.4 simulator has some support for remote notifications, see [post above](https://github.com/invertase/react-native-firebase/pull/3339#issuecomment-604027043). Perhaps this should be gated to only disable the feature in the simulator if iOS < 13.4 ?

Salakar:
> @Ehesp I saw your [latest commit](https://github.com/invertase/react-native-firebase/pull/3339/commits/e54f4ec87f08eb62c8a44c6021352c1b1a95601c). Given that the Xcode 11.4 simulator has some support for remote notifications, see [post above](https://github.com/invertase/react-native-firebase/pull/3339#issuecomment-604027043). Perhaps this should be gated to only disable the feature in the simulator if iOS < 13.4 ?

@otto-md the support added for the Xcode 1..4 simulators involves dragging & dropping an APS payload json file onto the simulator, there's still no 'remote' aspect of it - calling the method on simulator will still error as the iOS SDK code would need changing - I've not tested it yet though as I don't want to risk my productivity by upgrading to a new Xcode 😅

ottob:
> there's still no 'remote' aspect of it

I just assumed that remote notifications would work also. But, that makes sense that its local only. 

I restructured my code a little so I asked for permission in the simulator but not call registerDeviceForRemoteMessages() and now it works with an apns file dropped on the simulator. Neat.

Salakar:
> I restructured my code a little so I asked for permission in the simulator but not call registerDeviceForRemoteMessages() and now it works with an apns file dropped on the simulator. Neat.

@ottob / @otto-md I'll have the simulator change check up in an RC shortly so you should no longer need to conditionalise and can just call `registerDeviceForRemoteMessages` - will ping here when published

Salakar:
The final RC is up on NPM, 6.4.0-rc4, which includes the simulator checks and also the addition of the `isHeadless` React prop for iOS, see https://rnfb-docs.netlify.com/messaging/usage#background-application-state for information about this.

thanhtuyenpkk:
npm ERR! notarget No matching version found for react-native-firebase@6.4.0-rc4.
I change "react-native-firebase": "6.4.0-rc4" in package.json then get that error when reinstall

JeffGuKang:
@thanhtuyenpkk  Check the package name. May you use wrong name.

Mine 

```
"@react-native-firebase/app": "6.4.0-rc4",
"@react-native-firebase/messaging": "6.4.0-rc4",
```

thanhtuyenpkk:
How to use it? are you have any document guild for this version. 
My package.json

> "dependencies": {
>     "react": "16.8.6",
>     "react-native": "0.60.5",
>     "react-native-firebase": "^5.5.6"
>   }
If i change to 

> "dependencies": {
>     "react": "16.8.6",
>     "react-native": "0.60.5",
>     "react-native-firebase": "^5.5.6",
>     "@react-native-firebase/app": "6.4.0-rc4",
>     "@react-native-firebase/messaging": "6.4.0-rc4"
>   },
I think it must config again for IOS
My issue https://github.com/invertase/react-native-firebase/issues/3348

ugomarinelli1:
Hello, 
Is this fix working with "react-native-firebase": "^5.5.6", or should we upgrade to v6 ? 
Thanks !

JeffGuKang:
@thanhtuyenpkk @ugomarinelli1 
Try this: https://invertase.io/oss/react-native-firebase/migrating-to-v6

packages/messaging/ios/RNFBMessaging/RNFBMessagingModule.m
```diff
@@ -128,12 +138,22 @@ - (NSDictionary *)constantsToExport {
...
+      return;
+    }
+    #endif
     resolve([NSNull null]);
   }
 }
 
 RCT_EXPORT_METHOD(requestPermission:
-  (RCTPromiseResolveBlock) resolve
+  (NSDictionary *) permissions
```
kaushil111: Nice! Very much needed it. Thanks for championing! 🎉

ugomarinelli1:
Hey guys, thanks a lot for your work !! 
I was just wondering, I upgraded to v6, what is the listener replacing onNotificationOpened ? I tried setBackgroundMessageHandler but it is not called ... 
Thanks :)

Ehesp:
Please check out the work in progress docs: https://rnfb-docs.netlify.com/messaging/notifications#handling-interaction

thanhtuyenpkk:
@JeffGuKang Thanks you very much. I had try to v6. 
- In android it ok but in background when i click the notification. my application will open but not have any thing, exam Alert console ... vvv. In v5 i can was handle it.
- In IOS not have notification. I has deploy in real iphone but it not.

My App.js

> import React, { useEffect } from 'react';
> import {
>   View,
>   Text,
>   Alert,
>   Platform,
> } from 'react-native';
> 
> import messaging from '@react-native-firebase/messaging';
> const App = () => {
>   useEffect(() => {
>     checkPermission();
>     const unsubscribe = messaging().onMessage(async remoteMessage => {
>       console.log('FCM Message Data:', remoteMessage);
>       Alert.alert('A new FCM message arrived!', JSON.stringify(remoteMessage));
>     });
>   }, []);
> 
>   // 1
>   async function checkPermission() {
>     const enabled = await messaging().hasPermission();
>     if (enabled) {
>       if (Platform.OS = 'ios') {
>         await messaging().registerForRemoteNotifications();
>       }
>       await getToken();
>     } else {
>       await requestPermission();
>     }
>   }
> 
>   async function getToken() {
>     const fcmToken = await messaging().getToken();
>     console.log('fcmToken', fcmToken);
>     Alert.alert('fcmToken', fcmToken);
>   }
> 
>   async function requestPermission() {
>     const granted = messaging().requestPermission();
>     console.log('requestPermission', granted)
>     if (granted) {
>       console.log('User granted messaging permissions!');
>     } else {
>       console.log('User declined messaging permissions :(');
>     }
>   }
> 
>   return (
>     <View>
>       <Text>Step One</Text>
>     </View>
>   );
> };
> 
> export default App;

My index.js

> import {AppRegistry, Alert} from 'react-native';
> import App from './App';
> import {name as appName} from './app.json';
> import messaging from '@react-native-firebase/messaging';
> messaging().setBackgroundMessageHandler(async remoteMessage => {
>     console.log('Message handled in the background!', remoteMessage);
>     Alert.alert('A new FCM message arrived!', JSON.stringify(remoteMessage));
> });
> AppRegistry.registerComponent(appName, () => App);
> 

Is there have any problem?

Ehesp:
@thanhtuyenpkk you have 2 problems:

1. You can't do an alert from `setBackgroundMessageHandler`. The app has no access to the UI.

2. When a notification press triggers the app to open, you need to call the `onNotificationOpenedApp` (for foreground app opens) or `getInitialNotification` (for when the app is quit).

Check the link I posted above, specifically https://rnfb-docs.netlify.com/messaging/notifications#handling-interaction

thanhtuyenpkk:
@Ehesp thanks you. 
About this link https://invertase.io/oss/react-native-firebase/migrating-to-v6
Is right? This is different with your document.
Now i try uses version 6.4.0-rc4.
And your document use to version ???

ngregrichardson:
> @thanhtuyenpkk you have 2 problems:
> 
> 1. You can't do an alert from `setBackgroundMessageHandler`. The app has no access to the UI.
> 2. When a notification press triggers the app to open, you need to call the `onNotificationOpenedApp` (for foreground app opens) or `getInitialNotification` (for when the app is quit).
> 
> Check the link I posted above, specifically https://rnfb-docs.netlify.com/messaging/notifications#handling-interaction

Does this apply for if I want to display the notification using react-native-push-notification?

Ehesp:
Not sure what you mean sorry?

ngregrichardson:
I hadn't read enough into the docs, I understand what it does now. Sorry about that.

thanhtuyenpkk:
> @thanhtuyenpkk @ugomarinelli1
> Try this: https://invertase.io/oss/react-native-firebase/migrating-to-v6


> @thanhtuyenpkk you have 2 problems:
> 
> 1. You can't do an alert from `setBackgroundMessageHandler`. The app has no access to the UI.
> 2. When a notification press triggers the app to open, you need to call the `onNotificationOpenedApp` (for foreground app opens) or `getInitialNotification` (for when the app is quit).
> 
> Check the link I posted above, specifically https://rnfb-docs.netlify.com/messaging/notifications#handling-interaction

Following the instructions above, there is another guide to installing react-native-firebase.
And you also have another tutorial.
My problem occurred with ios not receiving notifcation. I have tried many ways according to the docs, I could not send notifcation on ios. Will react-native-firebase work on ios?.
I need a simple project template with react-native-firebase to try it. are you have it?

JeffGuKang:
@thanhtuyenpkk It works well on 6.4.0-rc4. I tested a lot. 

1. Use only `react-native-firebase/messging` first without another library as react native push notification.
2. Check your iOS is getting APNSToken. 
```
console.log(
    'fcm firebase.messaging().getAPNSToken()',
    await firebase.messaging().getAPNSToken(),
  )
```
3. Try push notification in cloud messaging of firebase console. - https://firebase.google.com/

pierregoutheraud:
I am getting this error when using phone auth on iOS which I did not have on `6.3.2`.
```
If app delegate swizzling is disabled, remote notifications received by UIApplicationDelegate need to be forwarded to FIRAuth's canHandleNotificaton: method.
```
Is swizzling (whatever that is) disabled now ?
Sould I do what the doc says here https://firebase.google.com/docs/auth/ios/phone-auth#appendix:-using-phone-sign-in-without-swizzling ?

JeffGuKang:
@pierregoutheraud Swizzling is a default option. You don't need to do that.

pierregoutheraud:
@JeffGuKang Okay, what should I do to fix the error message then ?

thanhtuyenpkk:
https://invertase.io/oss/react-native-firebase/migrating-to-v6
https://rnfb-docs.netlify.com/messaging/notifications#handling-interaction

I fixed my problem according to your instructions.
However, the notification has no config instructions.
I followed v5's instructions
https://rnfirebase.io/docs/v5.x.x/notifications/android
and fixed it on android. And ios is not.
Sincere thanks.

ronneldavis:
> I am getting this error when using phone auth on iOS which I did not have on `6.3.4`.
> 
> ```
> If app delegate swizzling is disabled, remote notifications received by UIApplicationDelegate need to be forwarded to FIRAuth's canHandleNotificaton: method.
> ```
> 
> I did not disable swizzling (whatever that is).
> Sould I do what the doc says here https://firebase.google.com/docs/auth/ios/phone-auth#appendix:-using-phone-sign-in-without-swizzling ?

I have been getting this error too using rc4.

vovka-s:
Hi guys!
I tried to use version 6.4.0-rc4 to handle my pushes on android and faced a problem:
all callbacks work fine except `onNotificationOpenedApp`.
Event is not fired if I press on push in app's background state.
Used https://rnfb-docs.netlify.com/messaging/notifications#handling-interaction
```
            messaging().onMessage(remoteMessage => {
                console.log('FCM message Data:', remoteMessage.data);
            });

            messaging().setBackgroundMessageHandler(async remoteMessage => {
                console.log('Message handled in the background!', remoteMessage);
            });

            messaging().onNotificationOpenedApp((remoteMessage) => {
                console.log('Notification caused app to open from background state:', remoteMessage.notification);
            });

                // Check whether an initial notification is available
            messaging().getInitialNotification()
                .then((remoteMessage) => {
                    if (remoteMessage) {
                        console.log('Notification caused app to open from quit state:', remoteMessage.notification);
                    }
                });
```
What may be wrong?

thanhtuyenpkk:
@vovka-s check file index.js function setBackgroundMessageHandler

kaushil111:
I think we are supposed to use `onMessage` for notifications when app is in background mode

vovka-s:
@thanhtuyenpkk 
sorry, I don't know what to check here. All seems logical
Screenshot from 2020-04-02 19-55-57

@kaushil111 
`onMessage` is not called for notifications when app is in background state
only `backgroundMessageHandler` works

Ehesp:
The new docs address this, however in a nutshell:

- `onMessage` called when a message is received & when the app is in the foreground.
- `setMessageBackgroundHandler` called when a message is received and the app is in the background or quit.
- `onNotificationOpenedApp` called when the user presses a notification and the app is in the foreground or background (as an event)
- `getInitialNotification` called when the user presses a notification but the app was quit

vovka-s:
@Ehesp 
Yes, I understand this terms.
The problem is that `onNotificationOpenedApp` is not called when I press a notice
Other callbacks are called correctly

alandotic:
I am having the same issue as @vovka-s , the following is my setup:
`const [message, setMessage] = useState<string>();`

`messaging().onNotificationOpenedApp(remoteMessage => {
      setMessage('onNotificationOpenedApp');
    });
    messaging().onMessage(remoteMessage => {
      setMessage('onMessage');
    });
    messaging()
      .getInitialNotification()
      .then(remoteMessage => {
        setMessage('getInitialNotification');
        if (remoteMessage) {
          setMessage('getInitialNotificationFromQuitState');
        }
      });`
and
`    <Layout>
      <Text style={{marginTop: 40, marginLeft: 40}}>{message}</Text>
    </Layout>`

I've tested the following scenario:

 Fore ground
1. leave the app opened, send a notification
2. Expected output: onMessage word shown in the screen
3. Actual output: onMessage word shown in the screen

Quit state
1. Kill the app, send a notification
2. Expected output: notification received, tapping the notification will show onNotificationOpenedApp
3: Actual output: notification received. tapping the notification shows onNotificationOpenedApp

Background state
1. Kill the app, send a notification
2. Expected output: notification received, tapping the notification will show onNotificationOpenedApp
3: Actual output: notification received. tapping the notification shows **getInitialNotificationFromQuitState** instead

it works fine on iOS as all the 3 scenarios have the expected output, only **onNotificationOpenedApp** doesn't not work as expected

vovka-s:
On release 6.4.0 `onNotificationOpenedApp` is still not called when I press a notice on Android

sndp07:
I recently upgraded to 6.4.0. Messaging works fine in Android, and iOS simulator, but it doesn't work on the real device. Am I missing any configuration specific to iOS real device?

LucasGarcez:
`onNotificationOpenedApp` does't called when open a notification on Android. And `getInitialNotification` always return null in both platform.

packages/messaging/ios/RNFBMessaging/RNFBMessagingModule.m
```diff
@@ -83,26 +81,38 @@ - (NSDictionary *)constantsToExport {
...
+    [[FIRInstanceID instanceID] instanceIDWithHandler:^(FIRInstanceIDResult *_Nullable result, NSError *_Nullable error) {
+      if (error) {
+        [RNFBSharedUtils rejectPromiseWithNSError:reject error:error];
+      } else {
+        resolve(result.token);
+      }
+    }];
+  } else {
+    NSDictionary *options = nil;
+    if ([FIRMessaging messaging].APNSToken) {
```
TPXP, line 104: If the APNSToken is not set (returning user), an 1001 error is likely to be thrown. Showing a warning message "did you forget to call registerForRemoteNotifications() ?" could help.

packages/messaging/ios/RNFBMessaging/RNFBMessagingModule.m
```diff
@@ -143,96 +163,78 @@ - (NSDictionary *)constantsToExport {
...
-  }
+  #if TARGET_IPHONE_SIMULATOR
+  resolve(@([RCTConvert BOOL:@(YES)]));
+  return;
+  #endif
   if (@available(iOS 10.0, *)) {
-    UNAuthorizationOptions authOptions;
-    if (@available(iOS 12.0, *)) {
-      authOptions = UNAuthorizationOptionProvisional | UNAuthorizationOptionAlert | UNAuthorizationOptionSound | UNAuthorizationOptionBadge;
+    if ([UIApplication sharedApplication].isRegisteredForRemoteNotifications == YES) {
```
TPXP, line 224: This condition is broken and is the _real_ root cause of the 1001 error code.
Diving into the code of the iOS SDK reveals the 1001 code is [kFIRInstanceIDErrorCodeMissingAPNSToken](https://github.com/firebase/firebase-ios-sdk/blob/42c5ae17fc71886b23ba008623f509db0ae5316a/Firebase/InstanceID/NSError%2BFIRInstanceID.h#L43), which is thrown by the `tokenWithAuthorizedEntity` method if it did not receive an APNS token, which is provided by the `didRegisterForRemoteNotificationsWithDeviceToken` callback: https://github.com/firebase/firebase-ios-sdk/blob/c049f741f0d3b7de955dff7c4173baedf3aa8355/Firebase/InstanceID/FIRInstanceID.m#L296

While `instanceIDWithHandler` has a caching system that will auto-fill the APNS token, which is why #3130 worked, we should **not** bypass registering for notifications in any case since iOS will reply with an updated token if needed.
According to the [official docs](https://developer.apple.com/documentation/usernotifications/registering_your_app_with_apns#2942135):
> You register your app and receive your device token **each time your app launches** using Apple-provided APIs.

15110011:
i used onMessage and its return data is an empty object. So, how do I fix that?

nicpro85:
> i used onMessage and its return data is an empty object. So, how do I fix that?

did you find out why ? I have this issue on android. 

{"collapseKey": "ch.iomedia.communes.cran", "data": {}, "from": "593590134815", "messageId": "0:1589793258145266%2e1fdf9e2e1fdf9e", "notification": {"android": {"imageUrl": "3877", "sound": "Default"}, "body": "my great title"}, "sentTime": 1589793258119, "ttl": 2419200}

andersonaddo:
Just putting it out there that the deprecated functions (like `registerForRemoteNotifications`) are still there :P
We're in v8 now haha

PR #3359, opened 2020-03-28
Update quick-start.md

JaioSkura:
We found this issue in many projects after upgrading to firebase 6 and choosing wix's library for local notification. It is not easy to realize what is happening because the only way to follow it is with the logs in logCat. We hope this will help people and safe their time.

### Checklist

- [x] Supports `Android`
- [ ] Supports `iOS`
- [ ] Flow types updated

### Release Plan

---

### DIFF
--- a/docs/messaging/quick-start.md
+++ b/docs/messaging/quick-start.md
@@ -156,6 +156,29 @@ packages available for React Native which allow you to display local notificatio
 
 - [Notifee](https://notifee.app/) - See [integrating with FCM](https://notifee.app/react-native/docs/integrations/fcm).
 - [wix/react-native-notifications](https://github.com/wix/react-native-notifications)
+    Only Android release mode:
+      By default when runing from an APK the wix/react-native-notifications library is handeling all the data notifications, @react-native-firebase/messaging does not receive any data, so it can't handle anything. To avoid this behaviour you can comment or remove everything inside <application> tag:
+   <application>
+       <!-- &lt;!&ndash;
+         A proxy-service that gives the library an opportunity to do some work before launching/resuming the actual application task.
+         &ndash;&gt;
+        <service android:name=".core.ProxyService"/>
+
+        <service
+            android:name=".fcm.FcmInstanceIdListenerService">
+            <intent-filter>
+                <action android:name="com.google.firebase.MESSAGING_EVENT" />
+                <action android:name="com.google.firebase.INSTANCE_ID_EVENT" />
+            </intent-filter>
+        </service>
+
+        <service
+            android:name=".fcm.FcmInstanceIdRefreshHandlerService"
+            android:exported="false" />-->
+    </application>
+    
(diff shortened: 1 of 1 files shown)


Ehesp:
Thanks for the PR. I believe however this should be something document on the Wix side of things, rather than this library. This library handles FCM messages as the underlying SDKs intended, whereas external local notification libraries are attempting to override that behaviour.

cc @Salakar for thoughts.

Also, your PR has a fair few grammar/typos on it which would need sorting if it were to be merged.

Salakar:
Also, can you try see if this behaviour happening on 6.4.0-rc4?

JaioSkura:
@Ehesp I have been thinking about this, but I really think that should be in this documentation, as long as react-native-firebase/messaging' does not provide local notification functionality and 'react-native-notifications' is recommended as a library for that purpose. Wix team should not be aware that we are using both libraries together.

Is just my point of view, maybe I am wrong and we can sugest to them to change their library to avoid having this behavior in producction. I don't know if this is posible.

Ehesp:
> Is just my point of view, maybe I am wrong and we can sugest to them to change their library to avoid having this behavior in producction. I don't know if this is posible.

I've no idea to be honest. I think something like this would be better placed in a guide/blog post somewhere? 

Also, which version are you using? We've just done a load of changes on 6.4 rc releases.

JaioSkura:
@Salakar  yeah, I sow this new PR announcement in Discord channel, but couldn`t try yet. I will do it and let you know. Anyway, in that PR there are not included local notifications, are they?  We should continue using wix/react-native-notifications, or any other library who provides that functionality, aren't we?

Ehesp:
You can display local notifications directly from FCM, only when the app is in the background (standard behaviour).

JaioSkura:
> > Is just my point of view, maybe I am wrong and we can sugest to them to change their library to avoid having this behavior in producction. I don't know if this is posible.
> 
> I've no idea to be honest. I think something like this would be better placed in a guide/blog post somewhere?
> 
> Also, which version are you using? We've just done a load of changes on 6.4 rc releases.

yes, I will test it and let you know. Thanks!

JaioSkura:
> You can display local notifications directly from FCM, only when the app is in the background (standard behaviour).

Yes I know, but I need to take only data and then decide if I show the notification or not. And apart from that, I also need to show notificactions even ir the app is in foreground. I will let you know what happens with 6.4.0-rc0.
Thanks a lot!

JaioSkura:
Hi again people!

I try again with  6.4.0-rc4, and it seems to be fixed, even if I don't see why. Do you know why? just curiosity...


By the way, so great job!!!

Salakar:
Probably fixed in RC4 as we now create our own `C2D_MESSAGE` listeners

If you're happy this is working for you now can you close this PR? Thanks for submitting it though!

PR #3360, opened 2020-03-28
fix(auth): user.metadata uses ISO strings

### LINKED ISSUE #2555: [v6][auth] UserMetadata times are milliseconds, should be ISO Date
@ehesp @Salakar

russellwheatley:
fixes #2555

### DIFF
--- a/packages/auth/e2e/user.e2e.js
+++ b/packages/auth/e2e/user.e2e.js
@@ -409,6 +409,9 @@ describe('auth().currentUser', () => {
       should(metadata.lastSignInTime).be.a.String();
       should(metadata.creationTime).be.a.String();
 
+      new Date(metadata.lastSignInTime).getFullYear().should.equal(new Date().getFullYear());
+      new Date(metadata.creationTime).getFullYear().should.equal(new Date().getFullYear());
+
       await firebase.auth().currentUser.delete();
     });
   });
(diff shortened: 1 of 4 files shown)


packages/auth/lib/User.js
```diff
@@ -19,6 +19,21 @@ import { isObject, isString } from '@react-native-firebase/app/lib/common';
 
 export default class User {
   constructor(auth, user) {
+    if (user) {
+      const { metadata } = user;
+
+      if (metadata && metadata.creationTime && metadata.lastSignInTime) {
```
Ehesp: I think this logic would be better placed in the getter?

packages/auth/e2e/user.e2e.js
```diff
@@ -396,4 +396,20 @@ describe('auth().currentUser', () => {
...
+    it("should have the properties 'lastSignInTime' & 'creationTime' which are ISO strings", async () => {
+      const random = Utils.randString(12, '#aA');
+      const email = `${random}@${random}.com`;
+
+      const { user } = await firebase.auth().createUserWithEmailAndPassword(email, random);
+
+      const { metadata } = user;
+
+      should(metadata.lastSignInTime).be.a.String();
+      should(metadata.creationTime).be.a.String();
```
Salakar: Could we do an additional check here to make sure the date is valid, e.g. new date the ISO date string and check the year is the same as the current dates year - this check ensure then that the number that comes from native also valid (seconds vs milliseconds)

```js
(new Date(metadata.lastSignInTime)).getFullYear().should.equal((new Date()).getFullYear());
(new Date(metadata. creationTime)).getFullYear().should.equal((new Date()).getFullYear());
```

PR #3385, opened 2020-04-01
feat(analytics): add & deprecate pre-defined analytics events

### LINKED ISSUE #3512: [analytics] ECOMMERCE_PURCHASE is now deprecated
See https://firebase.google.com/docs/reference/android/com/google/firebase/analytics/FirebaseAnalytics.Event#public-static-final-string-ecommerce_purchase

API needs changing to support `logPurchase` and deprecate the old API.

https://firebase.google.com/docs/reference/android/com/google/firebase/analytics/FirebaseAnalytics.Event#PURCHASE

russellwheatley:
fixes #3512 

## New API

New event methods added.

- logAddShippingInfo
- logPostScore
- logPurchase
- logRefund
- logSelectItem
- logSelectPromotion
- logViewCart
- logViewPromotion

## Removed API

API that now fires a `console.warn()` when called.

- logEcommercePurchase
- logPresentOffer
- logPurchaseRefund

## Unchanged or slight modifications 

Following API may have changed in minor ways
1. Where all properties are optional, I've made the event argument a default empty object so that users can make event function calls without passing any argument.
2. A number of API have `currency` & `value` properties which require the compound property check to ensure both are present if one is present.

- logAddToWishlist
- logAppOpen
- logBeginCheckout
- logCampaignDetails
- logEarnVirtualCurrency
- logGenerateLead
- logJoinGroup
- logLevelStart
- logLevelUp
- logLogin
- logSelectContent
- logRemoveFromCart
- logSearch
- logAddPaymentInfo
- logAddToCart
- logSetCheckoutOption
- logShare
- logSignUp
- logSpendVirtualCurrency
- logTutorialBegin
- logUnlockAchievement
- logViewItem
- logViewItemList
- logViewSearchResults

## Notes

- Moved all analytic's JS tests to `analytics.test.ts`.


packages/analytics/lib/index.js
```diff
@@ -195,21 +195,41 @@ class FirebaseAnalyticsModule extends FirebaseModule {
   resetAnalyticsData() {
     return this.native.resetAnalyticsData();
   }
+  // TODO test nested items objects in structs file. not here
```
Salakar, line 198: Whats this?

packages/analytics/lib/index.d.ts
```diff
@@ -50,143 +50,124 @@ import { ReactNativeFirebase } from '@react-native-firebase/app';
...
      */
+    currency?: string;
     value?: number;
+  }
+
+  export interface BeginCheckoutEventParameters {
     /**
      * Purchase currency in 3 letter [ISO_4217](https://en.wikipedia.org/wiki/ISO_4217#Active_codes) format. E.g. `USD`.
      */
+    //TODO if value is a param, so must currency: https://firebase.google.com/docs/reference/android/com/google/firebase/analytics/FirebaseAnalytics.Event#public-static-final-string-add_to_wishlist
```
Salakar, line 162: TODO here

packages/analytics/lib/index.js
```diff
@@ -380,37 +390,53 @@ class FirebaseAnalyticsModule extends FirebaseModule {
...
   }
 
-  logPresentOffer(object) {
-    if (!isObject(object)) {
-      throw new Error(
-        'firebase.analytics().logPresentOffer(*): The supplied arg must be an object of key/values.',
-      );
-    }
+  /**
+   * deprecated, use logRefundEvent instead:
```
Salakar, line 394:
```suggestion
   * Deprecated, use logRefundEvent instead:
```

packages/analytics/lib/index.js
```diff
@@ -380,37 +390,53 @@ class FirebaseAnalyticsModule extends FirebaseModule {
...
+   * https://firebase.google.com/docs/reference/android/com/google/firebase/analytics/FirebaseAnalytics.Event#public-static-final-string-present_offer
+   */
+  logPresentOffer() {
+    console.warn(
+      'firebase.analytics().logPresentOffer(), "PRESENT_OFFER" event is now deprecated. Please use firebase.analytics().logViewPromotion() instead',
+    );
+  }
+
+  /**
+   * deprecated, use logRefundEvent instead:
```
Salakar, line 404:
```suggestion
   * Deprecated, use logRefundEvent instead:
```

packages/analytics/lib/index.js
```diff
@@ -380,37 +390,53 @@ class FirebaseAnalyticsModule extends FirebaseModule {
...
+   */
+  logPurchaseRefund() {
+    console.warn(
+      'firebase.analytics().logPurchaseRefund(), "PURCHASE_REFUND" event is now deprecated. Please use firebase.analytics().logRefund() instead',
+    );
+  }
 
-    validateCompound(object, 'value', 'currency', 'firebase.analytics().logPresentOffer(*):');
+  /**
+   * deprecated, use logRefundEvent instead:
```
Salakar, line 414:
```suggestion
   * Deprecated, use logRefundEvent instead:
```

packages/analytics/lib/structs.js
```diff
@@ -14,45 +14,54 @@
  *  limitations under the License.
  */
 import struct from '@react-native-firebase/app/lib/common/struct';
+const Item = struct({
```
Salakar, line 17:
```suggestion

const Item = struct({
```

packages/analytics/lib/index.js
russellwheatley: This was for checking the validity of `Item` objects in an array of `Items` which has been implemented: https://github.com/invertase/react-native-firebase/blob/afe213b9d9b2ea5185badb6f9c6f5b65d3e20873/packages/analytics/lib/structs.js#L33. Will remove.

packages/analytics/lib/index.d.ts
russellwheatley: I'm not actually sure about this. If `currency` is a property on the object then `value` must also be present.

I think we will have to update the event types to have something like this: https://stackoverflow.com/a/51507473 what do you think?

The `superstruct` object check should probably ensure if one property is present, the other one should be as well I think. Let me know if you think otherwise.

russellwheatley:
I've had to temporarily stop the `ml-vision` tests to stop it from failing.

Issue #3253, opened 2020-02-28
[Have demo code] [iOS] So many installation issue: crashed on first open - RNFirebase latest version (@6.3.4)


luatnd:
## Issue

--- Issue 1: ---
App was crashed on the first open on iOS

Reproduction steps:
- clone above app
- Run `yarn ensureI`  to install node modules and pod
- Build and run on a real device

App was crash at first open.

- Rebuild and run again --> Work normally

For more details, plz see the **Project files** section, don't worry about secret keys, that is a public project.

--- Issue 2: ---
You can see in my Podfile, I comment a workaround for this issue:
> Fix: target has transitive dependencies that include statically linked binaries:

--- Issue 3: ---
You can see in my Podfile, I comment a workaround for this issue:
```
post_install do |installer|
  installer.pods_project.targets.each do |target|
    #
    # Cannot build with rnfirebase@6
    # https://github.com/invertase/react-native-firebase/issues/3024
    #
    if ['RNFBApp', 'RNFBMessaging' ].include?(target.name)
      target.add_system_framework(installer.pods_project.targets.find{|target| target.name == 'GoogleUtilities'})
     elsif ['RNFBAnalytics', 'RNFBAuth'].include?(target.name)
       target.add_system_framework(installer.pods_project.targets.find{|target| target.name == 'FirebaseAnalytics'})
       target.add_system_framework(installer.pods_project.targets.find{|target| target.name == 'GoogleAppMeasurement'})
       target.add_system_framework(installer.pods_project.targets.find{|target| target.name == 'nanopb'})
       target.add_system_framework(installer.pods_project.targets.find{|target| target.name == 'GoogleUtilities'})
    end
  end
end
```

--- Issue 4: ---
Here my xCode log content:
https://github.com/Pushdy/react-native-pushdy-example/blob/eb7be26a3667eadf5bf211b6529cea4c48528f25/tmp-xcode.log

I think this is abnormal:
> objc[7519]: Class FIRAnalytics is implemented in both /private/var/containers/Bundle/Application/9F4CBCCE-4A81-4E61-880B-537C69A894A3/reactNativePushdyExample.app/Frameworks/RNFBAnalytics.framework/RNFBAnalytics (0x1057a1200) and /var/containers/Bundle/Application/9F4CBCCE-4A81-4E61-880B-537C69A894A3/reactNativePushdyExample.app/reactNativePushdyExample (0x104d74a68). One of the two will be used. Which one is undefined.

[...]

PR #3388, opened 2020-04-02
feat(ios): allow static_framework usage via Podfile global

Salakar:
### Summary

First pass at addressing #3253 to allow release in 6.4.0 - there is no breaking changes as the default is currently `false`

For v7.0.0 we should make the default value `true` along with the RN min version set to 0.61.x +

### Checklist

- [x] Supports `iOS`

### Test Plan

`pod install` logs show flag being detected and used correctly;

image

---

### DIFF
--- a/packages/analytics/RNFBAnalytics.podspec
+++ b/packages/analytics/RNFBAnalytics.podspec
@@ -2,11 +2,6 @@ require 'json'
 package = JSON.parse(File.read(File.join(__dir__, 'package.json')))
 
 firebase_sdk_version = '~> 6.13.0'
-using_custom_firebase_sdk_version = defined? $FirebaseSDKVersion
-if using_custom_firebase_sdk_version
-  Pod::UI.puts "RNFBAnalytics: Using user specified Firebase SDK version '#{$FirebaseSDKVersion}'"
-  firebase_sdk_version = $FirebaseSDKVersion
-end
 
 Pod::Spec.new do |s|
   s.name                = "RNFBAnalytics"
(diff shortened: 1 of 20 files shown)


Salakar:
cc @luatnd for additional review

tests/ios/Podfile
```diff
@@ -1,7 +1,10 @@
 platform :ios, '9.0'
 
+# Allow using RNFirebase as static frameworks
+$RNFirebaseAsStaticFramework = true
+
 # Version override testing
-# $FirebaseSDKVersion = '6.8.1'
+# $FirebaseSDKVersion = '6.15.0'
```
mikehardy, line 7: Oh come on take it to 6.21.0 :-)
Default to 6.21.0 even... why not?

packages/analytics/RNFBAnalytics.podspec
```diff
@@ -22,8 +17,24 @@ Pod::Spec.new do |s|
...
   s.dependency          'RNFBApp'
-  s.static_framework    = false
+
+  if defined?($FirebaseSDKVersion)
+    Pod::UI.puts "#{s.name}: Using user specified Firebase SDK version '#{$FirebaseSDKVersion}'"
+    firebase_sdk_version = $FirebaseSDKVersion
+  end
+
+  # Firebase dependencies
+  s.dependency          'Firebase/Core', firebase_sdk_version
```
mikehardy, line 31: I could be wrong, but I think with the up to date Pods you can evict this dependency and it will be brought in transitively when needed ?

tests/ios/Podfile
Salakar: I wish! Will be for v7

packages/analytics/RNFBAnalytics.podspec
Salakar: You're right as RNFBApp brings this in of which all are dependant on - I originally left this in so its visually explicit event though it doesn't actually get added as it's already a dependency elsewhere.

Could just remove it though now I guess :D

Salakar: I think I'll go for removing these for the v7 release, have made a note

PR #3425, opened 2020-04-09
Fix (messaging) and signInWithPhoneNumber

mateusmirandaalmeida:
### Summary

Currently, when trying to use the phone login. conflicts with notifications

 Error:
NativeFirebaseError: [auth/notification-not-forwarded] If app delegate swizzling is disabled, remote notifications received by UIApplicationDelegate need to be forwarded to FIRAuth's canHandleNotificaton: method.
    at FirebaseAuthModule.signInWithPhoneNumber (http://192.168.0.167:8081/index.bundle?platform=ios&dev=true&minify=false:107323:28)


mateusmirandaalmeida:
This works for me and is already in my production application. 
The problem is that the whole "npm install" loses my change. 
Do you have a forecast to review this?

mikehardy:
Patch Package to the rescue: https://github.com/ds300/patch-package

Do not develop in react-native without it

Salakar:
This change needs updating to handle if the Auth package is not in use by the users application - FIRAuth class won't be found in that case and would cause a build failure

I can look at this though if needed

mateusmirandaalmeida:
what's the best way to do this?

tautvilas:
This fix works for me.

Salakar:
Have fixed this in https://github.com/invertase/react-native-firebase/commit/fb70797d5427d20b688dc33e2a0407b246d82616 with the FIRAuth class existence check

Thanks for sending this PR - it was helpful

Salakar:
(a release will be out later today)

Issue #3049, opened 2019-12-27
[v6] Change database URL during run time for same project


lisabaut:
## Issue

I am running my app on Android and have one firebase project with two databases.

**With version 5.x of RN-Firebase** I was able to change the database URL during runtime with simply calling the module with the URL string: 
```TS
firebase.database('https://path-to-database.url');
```

See older discussion on this implementation here: https://github.com/invertase/react-native-firebase/pull/649#issuecomment-375038938

**Since version 6 I get the error**
```TS
Error: No Firebase App ''https://path-to-database.url'' has been created 
- call firebase.initializeApp()
```

Initializing a new app with the exact same [DEFAULT] project's data and configuration would result in something like:
```TS
  // merge default app options with new databaseURL
  const options = {
    ...firebase.app().options,
    databaseURL: 'https://path-to-database.url'
  };

  // create new app with a custom name
  const app = await firebase.initializeApp(options, {
    name: 'can-be-any-name'
  });
  
 // get database instance
 database(app);
```

Now I am wondering, why since version 6 it is nessecary to initialize a new app again although it is the same app/project with the same configuration?
([This issue with a similar question](https://github.com/invertase/react-native-firebase/issues/2677) was closed unfortunately. [This issue with changing the module itself by hand](https://github.com/invertase/react-native-firebase/issues/2991) seems a bit of an overhead.)

Is there another way to simply point the database module to the correct database URL of my default app (like it was with version 5.x)?

Thanks a lot!

---

### Javascript

<details><summary>Click To Expand</summary>
<p>

#### `package.json`:

```json
    "@react-native-firebase/analytics": "^6.2.0",
    "@react-native-firebase/app": "^6.2.0",
    "@react-native-firebase/auth": "^6.2.0",
    "@react-native-firebase/database": "^6.2.0",
    "@react-native-firebase/messaging": "^6.2.0",
    "react": "16.9.0",
    "react-native": "0.61.5",
```

#### `firebase.json` for react-native-firebase v6:

```json
# N/A
```

</p>
</details>

---

### Android

<details><summary>Click To Expand</summary>
<p>

#### Have you converted to AndroidX?

- [ ] my application is an AndroidX application?
- [ ] I am using `android/gradle.settings` `jetifier=true` for Android compatibility?
- [x] I am using the NPM package `jetifier` for react-native compatibility?
</p>
</details>

---

## Environment

<details><summary>Click To Expand</summary>
<p>

**`react-native info` output:**

```
[...]

PR #3431, opened 2020-04-09
docs: update database usage with switching database URL

lisabaut:
### Summary
I struggled a while with finding out how to switch the database URL for the [DEFAULT] firebase instance during runtime.
With version 5 it was very easy with simply passing the databaseURL to the database module.

The docs did not help unfortunately, so I tried a lot of ways which are documented in this ticket:
https://github.com/invertase/react-native-firebase/issues/3049

This PR shall help other people with the same situation too.

Furthermore, I hope to get valid feedback if the proposed implementation documented in this PR is actually correct.

### Checklist

It's a documentation update

- [x] Supports `Android`
- [x] Supports `iOS`
- [x] `e2e` tests added or updated in packages/\*\*/e2e
- [x] Flow types updated

### DIFF
--- a/docs/database/usage/index.md
+++ b/docs/database/usage/index.md
@@ -320,9 +320,7 @@ Please follow the firebase Realtime Database documentation on [security](https:/
 
 # Using a secondary database
 
-## Switch the database URL of the default app
-
-If the default installed firebase instance needs to address a different database within the same project, call the database method on the default app with passing the database URL.
+If the default installed Firebase instance needs to address a different database within the same project, call the database method on the default app with passing the database URL.
 For example:
 
 ```js
--- a/docs/database/usage/index.md
+++ b/docs/database/usage/index.md
@@ -318,30 +318,12 @@ database and the new [`DataSnapshot`](/reference/database/datasnapshot) containi
 It is important that you understand how to write rules in your firebase console to ensure that your data is secure.
 Please follow the firebase Realtime Database documentation on [security](https://firebase.google.com/docs/database/security)
 
-# firebase.json
-
-## Disabling persistence
-
-By default the Realtime Database persists data on the user application, and is used by the SDKs for offline usage
-and caching. To disable this functionality, update the `database_persistence_enabled` key in the `firebase.json` file:
-
-```json
-// <project-root>/firebase.json
-{
-  "react-native": {
-    "database_persistence_enabled": false
-  }
-}
-```
-
-To enable persistence, view the [Offline Support](/database/offline-support) documentation.
-
 # Change the database URL
 
 ## Switch the database URL of the default app
 
-If the default installed firebase instance needs to address a different database within the same project, call the database method on the default app with passing the database URL. 
-For example: 
+If the default installed firebase instance needs to address a different database within the same project, call the database method on the default app with passing the database URL.
+For example:
 
 ```js
 import firebase from '@react-native-firebase/app';
@@ -370,3 +352,21 @@ const secondaryDatabase = database(secondaryApp);
 
 secondaryDatabase.ref();
 ```
+
+# firebase.json
+
+## Disabling persistence
+
+By default the Realtime Database persists data on the user application, and is used by the SDKs for offline usage
+and caching. To disable this functionality, update the `database_persistence_enabled` key in the `firebase.json` file:
+
+```json
+// <project-root>/firebase.json
+{
+  "react-native": {
+    "database_persistence_enabled": false
+  }
+}
+```
+
+To enable persistence, view the [Offline Support](/database/offline-support) documentation.
--- a/docs/database/usage/index.md
+++ b/docs/database/usage/index.md
@@ -324,8 +324,7 @@ If the default installed Firebase instance needs to address a different database
 For example:
 
 ```js
-import firebase from '@react-native-firebase/app';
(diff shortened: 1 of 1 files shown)


docs/database/usage/index.md
```diff
@@ -335,3 +335,38 @@ and caching. To disable this functionality, update the `database_persistence_ena
 ```
 
 To enable persistence, view the [Offline Support](/database/offline-support) documentation.
+
+# Change the database URL
```
Ehesp, line 339: Could you move this section about `# firebase.json`? Also, can the name be "Using a secondary database" instead please.

docs/database/usage/index.md
```diff
@@ -335,3 +335,38 @@ and caching. To disable this functionality, update the `database_persistence_ena
...
+
+## Switch the database URL of the default app
```
Ehesp, line 341: This can be removed with the header changes.

docs/database/usage/index.md
```diff
@@ -335,3 +335,38 @@ and caching. To disable this functionality, update the `database_persistence_ena
...
+
+If the default installed firebase instance needs to address a different database within the same project, call the database method on the default app with passing the database URL.
```
Ehesp, line 343: `firebase` -> `Firebase`

docs/database/usage/index.md
```diff
@@ -335,3 +335,38 @@ and caching. To disable this functionality, update the `database_persistence_ena
...
+# Change the database URL
+
+## Switch the database URL of the default app
+
+If the default installed firebase instance needs to address a different database within the same project, call the database method on the default app with passing the database URL. 
+For example: 
+
+```js
+import firebase from '@react-native-firebase/app';
+import '@react-native-firebase/database';
```
Ehesp, line 348: Lets go with the single import like the other examples:

```
import database '@react-native-firebase/database';
```

docs/database/usage/index.md
```diff
@@ -335,3 +335,38 @@ and caching. To disable this functionality, update the `database_persistence_ena
...
+
+const database = firebase.app().database('https://path-to-database.firebaseio.com');
```
Ehesp, line 350: Trying to reduce size/examples, this would be more practical I think:

```
const ref = database('https://path-to-database.firebaseio.com').ref();
```

docs/database/usage/index.md
lisabaut: done

docs/database/usage/index.md
lisabaut: done

docs/database/usage/index.md
lisabaut: done

docs/database/usage/index.md
lisabaut: This was the whole point for me and also the reason why I split this section into two different parts:

Changing the database URL for the default Firebase app is not working when calling the `database` module directly.
The `database` module accepts a Firebase instance only, not a simple URL string.
=> `database(FirebaseApp);`

If you want to point your default Firebase app **within the same project** to a different database you need to call the default app first, and then database. Only in this case `database` accepts a simple URL string.
=> `firebase.app().database('http://your-url-string.com')`;

See discussion here: #3049

Maybe I am missing something fundamental on using the lib correctly.
I am happy if you could correct me in this case. Thanks a lot.

docs/database/usage/index.md
lisabaut: Calling `firebase.app().database('http://your-url-string.com')` requires the general import of `import '@react-native-firebase/database';`
=> otherwise a warning is thrown: "You attempted to use "firebase.app('[DEFAULT]').database" but this module could not be found"

At least this happened to me. Is there another way to do it?

See the reason for this suggested implementation [in the comment below](https://github.com/invertase/react-native-firebase/pull/3431#discussion_r406701830)

docs/database/usage/index.md
Salakar: @lisabaut your original example is correct, can only change it from an app instance

docs/database/usage/index.md
Salakar: Can do:

```js
import { firebase } from '@react-native-firebase/database';
```

docs/database/usage/index.md
```diff
@@ -335,3 +335,38 @@ and caching. To disable this functionality, update the `database_persistence_ena
...
+
+## Connect to a database of a secondary app
+
+If you want to address a database from a different firebase project, you will need to create a secondary app first
+(Read more on creating a secondary app here: https://rnfirebase.io/app/usage).
+For example:
+
+```js
+import firebase from '@react-native-firebase/app';
+import database from '@react-native-firebase/database';
```
Salakar, line 343:
```suggestion
import database, { firebase } from '@react-native-firebase/database';
```

docs/database/usage/index.md
lisabaut: Thanks. Done.

docs/database/usage/index.md
lisabaut: Thanks. Done.

Salakar:
Thank you for this and for resolving the feedback items 🙃

Salakar:
Thanks again for this, really appreciated! It will be live on the site shortly once the build completes.

PR #3436, opened 2020-04-10
docs(messaging): Add java imports and fix typo

sunnysid3up:
### Summary

I got build errors when I followed the current docs. This PR adds more directions and fixes typos.

### Checklist

- [x] Supports `Android`
- [ ] Supports `iOS`
- [ ] `e2e` tests added or updated in packages/\*\*/e2e
- [ ] Flow types updated

### DIFF
--- a/docs/messaging/usage/installation/android.md
+++ b/docs/messaging/usage/installation/android.md
@@ -35,10 +35,19 @@ Import and apply the React Native Firebase module package to your `/android/app/
 
 ### 3.1 Import the package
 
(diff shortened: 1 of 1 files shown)


Ehesp:
Is this needed? The docs are lightweight to focus on RNFB only so this change would need to be reflected on all modules.

mikehardy:
> Is this needed? The docs are lightweight to focus on RNFB only so this change would need to be reflected on all modules.

IMHO no

PR #3455, opened 2020-04-12
fix(storage): bug when getting the root ref of a bucket

Adrxx:
A bug crashes the app when trying to list objects, root urls don't have a "/" so rangeOfString gives location == NSNotFound, if it goes unchecked rangeOfSlash.location + 5 creates range outside the bounds


Ehesp:
Could you give a code example of what this fixes?

Looking at the JS code, a `/` should always be sent over if something isn't specified.

Salakar:
cc @dackers86 @russellwheatley - one to try reproduce if you can, thanks

Adrxx:
Sorry, I forgot about this issue, it happened for me when trying to `listAll` files on a storage bucket on the root ref.

russellwheatley:
@Salakar @Ehesp, tried running this code for our RNFB Firebase project.

```
const sto = await storage()
      .ref('vision')
      .listAll();
```
Retrieved the 7 objects in that directory with no errors. 

If you could please provide a reproducible sample of code @Adrxx, that would be helpful.

Adrxx:
@russellwheatley 
```
const sto = await storage()
      .ref('vision')
      .listAll();
```
Here you are ref'ing 'vision', not the root directory of the storage bucket, I wanted to list all the elements on the root of the storage bucket, so I think I did something like this

```
const sto = await storage()
      .ref('')
      .listAll();
```

or maybe

```
const sto = await storage()
      .listAll();
```
Are you able to list the root directory by doing that?

mikehardy:
@russellwheatley user responded, may have repro case now

dackers86 (approved):

mikehardy:
Thanks for your patience, just published this one @Adrxx !

Adrxx:
Oh! Totally forgot about this. Sorry I couldn't give you a reproduction case. Glad I could help ☺️!

Issue #3463, opened 2020-04-14
:fire: getInitialNotification crashed in production - Android


jbohman:
## Issue

Hello react-native-firebase team,

I observed the following crash in production:

```
Fatal Exception: java.lang.NullPointerException: Attempt to invoke virtual method 'android.content.Intent android.app.Activity.getIntent()' on a null object reference
       at io.invertase.firebase.messaging.ReactNativeFirebaseMessagingModule.getInitialNotification(ReactNativeFirebaseMessagingModule.java:58)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.facebook.react.bridge.JavaMethodWrapper.invoke(JavaMethodWrapper.java:372)
       at com.facebook.react.bridge.JavaModuleWrapper.invoke(JavaModuleWrapper.java:151)
       at com.facebook.react.bridge.queue.NativeRunnable.run(NativeRunnable.java)
       at android.os.Handler.handleCallback(Handler.java:761)
       at android.os.Handler.dispatchMessage(Handler.java:98)
       at com.facebook.react.bridge.queue.MessageQueueThreadHandler.dispatchMessage(MessageQueueThreadHandler.java:27)
       at android.os.Looper.loop(Looper.java:156)
       at com.facebook.react.bridge.queue.MessageQueueThreadImpl$4.run(MessageQueueThreadImpl.java:226)
       at java.lang.Thread.run(Thread.java:776)
```

The log is from crashlytics, I've not been able to reproduce it locally yet. What else can I help with to debugging this issue further? I've filled in the form below with as much information I could.

Thanks

---

### Javascript

<details><summary>Click To Expand</summary>
<p>

#### `package.json`:

```json
{
  "name": "myapp",
  "version": "0.1.0",
  "private": true,
  "engines": {
    "node": "^10.16.0",
    "npm": "^6.11.3"
  },
  "devDependencies": {
    "@babel/core": "^7.6.2",
    "@babel/runtime": "^7.6.2",
    "@react-native-community/eslint-config": "^0.0.5",
    "babel-jest": "^24.9.0",
    "babel-plugin-transform-remove-console": "^6.9.4",
    "eslint": "^6.5.1",
    "jest": "^24.9.0",
    "metro-react-native-babel-preset": "0.58.0",
    "react-test-renderer": "16.11.0"
  },
  "scripts": {
    "start": "react-native start",
    "android": "react-native run-android",
    "ios": "react-native run-ios",
[...]

PR #3495, opened 2020-04-20
fix(messaging): add activity check to getInitialNotification

Ehesp:
Puts a safe check around #3463

### DIFF
--- a/docs/messaging/notifications.md
+++ b/docs/messaging/notifications.md
@@ -165,7 +165,7 @@ function App() {
 ```
 
 The call to `getInitialNotification` should happen within a React lifecycle method after mounting (e.g. `componentDidMount` or `useEffect`).
-If it's called too soon (e.g. within a class constructor or global scope), the notification data may be available.
+If it's called too soon (e.g. within a class constructor or global scope), the notification data may not be available.
 
 # Notifee - Advanced Notifications
 
--- a/docs/messaging/notifications.md
+++ b/docs/messaging/notifications.md
@@ -164,6 +164,9 @@ function App() {
 }
 ```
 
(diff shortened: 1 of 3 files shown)


docs/messaging/notifications.md
```diff
@@ -164,6 +164,9 @@ function App() {
 }
 ```
 
+The call to `getInitialNotification` should happen within a React lifecycle method after mounting (e.g. `componentDidMount` or `useEffect`).
+If it's called too soon (e.g. within a class constructor or global scope), the notification data may be available.
```
Salakar, line 168: May not be available

packages/messaging/android/src/main/java/io/invertase/firebase/messaging/ReactNativeFirebaseMessagingModule.java
```diff
@@ -55,23 +56,29 @@ public void getInitialNotification(Promise promise) {
...
+
+            if (remoteMessage != null) {
+              promise.resolve(ReactNativeFirebaseMessagingSerializer.remoteMessageToWritableMap(remoteMessage));
+              initialNotificationMap.put(messageId, true);
+              return;
+            }
           }
         }
+      } else {
+        Log.w(TAG, "Attempt to call getInitialNotification failed. The current activity is not ready, try calling the method later in the React lifecycle.");
```
Salakar, line 81: Resolve null here? Otherwise promise will hang. Same in other scenarios (e.g null intent) I guess - missing a resolve

Salakar: Ignore this comment,  promise resolve below

PR #3327, opened 2020-03-19
fix(functions): throw UNAVAILABLE code on network IO errors

justinwh:
Fixes #3326 

### Summary
Makes the error thrown by Functions when a network IO error occurs consistent between iOS and Android.

### Checklist

- [x] Supports `Android`
- [ ] Supports `iOS`
- [ ] `e2e` tests added or updated in packages/**/e2e
- [ ] Flow types updated
- [ ] Typescript types updated

### Test Plan
N/A

### Release Plan

[ANDROID][BUGFIX] [FirebaseFunctions] - change network IO errors to use code=UNAVAILABLE (matching iOS)

---

Think `react-native-firebase` is great? Please consider supporting the project with any of the below:

- 👉 Donate via [Open Collective](https://opencollective.com/react-native-firebase/donate)
- 👉 Follow [`React Native Firebase`](https://twitter.com/rnfirebase) and [`Invertase`](https://twitter.com/invertaseio) on Twitter
- 👉 Star this repo on GitHub ⭐️
- 👉 Contribute; see our [contributing guide](/CONTRIBUTING.md)

PR #3534, opened 2020-04-24
feat(functions): support function timeout

russellwheatley:
cloud functions timeout v7 implementation

- added timeout option property to functions which allows you to set the timeout response from a cloud function in milliseconds (native timeout is seconds for both platforms).


packages/functions/lib/index.d.ts
```diff
@@ -144,6 +144,24 @@ export namespace FirebaseFunctionsTypes {
...
+   *
+   * const instance = firebase.functions().httpsCallable('order', { timeout: 7000 });
+   * // The below will wait 7 seconds for a response from the cloud function before an error is thrown
+   * try {
+   *  const response = await instance({
+   *    id: '12345',
+   *  });
+   * } catch (e) {
+   *  console.log(e);
+   * }
```
Salakar, line 159:
```suggestion
   * ### Example
   *
   * ```js
   * const instance = firebase.functions().httpsCallable('order', { timeout: 7000 });
   * // The below will wait 7 seconds for a response from the cloud function before an error is thrown
   * try {
   *  const response = await instance({
   *    id: '12345',
   *  });
   * } catch (e) {
   *  console.log(e);
   * }
   * ```
```

packages/functions/lib/index.d.ts
```diff
@@ -144,6 +144,28 @@ export namespace FirebaseFunctionsTypes {
...
+   * ```
+   */
```
Salakar, line 164: Can we move this docs/example and the description about the timeout option to the timeout property, rather than on the `HttpsCallableOptions` interface?

Salakar:
image

Needs updating also;

```ts
httpsCallable(name: string, options?: HttpsCallableOptions): HttpsCallable;
```

Salakar:
Android CI test failing;

image


Error codes node matching up, most likely issue is that this logic needs tweaking to exclude `DEADLINE_EXCEEDED` from it;

https://github.com/invertase/react-native-firebase/blob/%40russell/functions-timeout/packages/functions/android/src/reactnative/java/io/invertase/firebase/functions/ReactNativeFirebaseFunctionsModule.java#L83-L86

packages/functions/ios/RNFBFunctions/RNFBFunctionsModule.m
```diff
@@ -54,6 +56,10 @@ @implementation RNFBFunctionsModule
 
     FIRHTTPSCallable *callable = [functions HTTPSCallableWithName:name];
 
+    if(options[@"timeout"]){
```
mikehardy, line 59:
```suggestion
    if (options[@"timeout"]) {
```

looked like there was an auto-quality checker setup, does it check whitespace? This one was inconsistent

I'm only peaking in here because functions timeouts is something I have patch-packaged into v5 (in a brutal/useless/hard-coded-value hack way - not worth sharing) so I'm interested in this in particular :-)

mikehardy: also maybe a prettier config and VSCode shared project setting that runs prettier on save (maybe) and as a husky pre-commit lint-staged type deal. Thinking out loud

Salakar: Unfortunately this is Obj-c code - so prettier won't run on this 😅I would love something for Obj-C/Java code though if you have any suggestions

Good spot though as this formatting does need fixing

packages/functions/lib/index.d.ts
```diff
@@ -144,6 +144,30 @@ export namespace FirebaseFunctionsTypes {
...
+     *  const instance = firebase.functions().httpsCallable('order', { timeout: 7000 });
+     *  const response = await instance({
+     *    id: '12345',
+     *  });
+     * } catch (e) {
+     *  console.log(e);
+     * }
+     * ```
+     */
+    timeout?: number;
```
mikehardy, line 168: The JS SDK is in millis https://github.com/firebase/firebase-js-sdk/blob/74870c7a3832ec0c1e6f93e4d45e9d0a3180dfb3/packages/functions-types/index.d.ts#L37

The iOS SDK PR that implemented it says it will be 60 seconds in the code but is actually 70 after a commit in the PR changed it from 60 to 70 without comment https://github.com/firebase/firebase-ios-sdk/pull/2329/files and is in seconds not millis (double 🤦 )

Android also does 70 seconds and in units of seconds https://github.com/firebase/firebase-android-sdk/blob/master/firebase-functions/src/main/java/com/google/firebase/functions/HttpsCallOptions.java

Can't believe they did that (seconds vs millis). Thank you in my case at least for using milliseconds so that it maintains JS consistency and works for me on web and mobile

packages/functions/ios/RNFBFunctions/RNFBFunctionsModule.m
mikehardy: For AnkiDroid we use Codacy service as a github check and it applies our checkstyle and PMD rules. I see Sonar has obj-c support and there is something called oclint https://github.com/oclint/oclint - I think you went with Sonar as a github check? and oclint could run command line as an action. Here's whitespace for objective-c? https://github.com/square/spacecommander

All sorts of stuff to waste time on 🤣

packages/functions/e2e/functions.e2e.js
```diff
@@ -298,7 +298,7 @@ describe('functions()', () => {
         await functions.httpsCallable(fnName, { timeout: 1000 })({ testTimeout: 3000 });
         return Promise.reject(new Error('Did not throw an Error.'));
       } catch (error) {
-        error.message.should.containEql('DEADLINE_EXCEEDED');
+        error.message.should.containEql('UNAVAILABLE');
```
Salakar, line 301: Any specific reason you changed this rather than fixing the underlying issue I mentioned here:

https://github.com/invertase/react-native-firebase/pull/3534#issuecomment-618955152

packages/functions/lib/index.d.ts
Salakar: JS SDK consistency is our motto

packages/functions/e2e/functions.e2e.js
russellwheatley: Ah, I interpreted that as a tweak to test logic because this code was approved a while back: https://github.com/invertase/react-native-firebase/pull/3327/files

Happy to change back if that's what you want?

Salakar:
Release of `@react-native-firebase/functions` v6.8.0 has been published with this change 🎉

PR #3557, opened 2020-04-27
tests(perf): migrate JS only tests to Jest

dackers86:


packages/perf/__tests__/perf.test.js
```diff
@@ -1,9 +1,107 @@
 import perf from '../lib';
+import firebase from '../../../node_modules/@react-native-firebase/app';
```
Salakar, line 2:
```suggestion
import perf, { firebase } from '../lib';
```

PR #3586, opened 2020-04-30
tests(admob): migrate JS only tests to Jest

dackers86:


Salakar:
@dackers86 could do with a sync with master - think there's some conflicts, could you update the title also to match the style mentioned in the contributing guide, thanks

cc @Ehesp for admob related review

PR #3240, opened 2020-02-26
feat(app): include play service utils

russellwheatley:
fixes this issue: https://invertase.canny.io/react-native-firebase/p/playservicesavailability. 

### Checklist

- [ ] Supports `Android`
- [ ] Supports `iOS`
- [ ] `e2e` tests added or updated in packages/**/e2e
- [ ] Flow types updated
- [ ] Typescript types updated

### Release Plan

[CATEGORY][type] [LOCATION] - Message

---

Think `react-native-firebase` is great? Please consider supporting the project with any of the below:

- 👉 Donate via [Open Collective](https://opencollective.com/react-native-firebase/donate)
- 👉 Follow [`React Native Firebase`](https://twitter.com/rnfirebase) and [`Invertase`](https://twitter.com/invertaseio) on Twitter
- 👉 Star this repo on GitHub ⭐️
- 👉 Contribute; see our [contributing guide](/CONTRIBUTING.md)

PR #3601, opened 2020-05-04
feat(app): add Play Services available utilities

russellwheatley:
fixes this issue: https://invertase.canny.io/react-native-firebase/p/playservicesavailability. 

Replaces: #3240

**NOTES**
- No tests for `firebase.utils().makePlayServicesAvailable()` or `firebase.utils().resolutionForPlayServices()`  as I'm not sure it's possible. Will update if there is a way.


russellwheatley:
> Side comment - out of the 8 files changed, 5 were changed in non-semantic / formatting only ways, is that expected behavior? Makes the diff really big and a bit scattered when functionally it's not so large. Not a big deal, just something I noticed

@mikehardy It should've probably been in its own PR to be honest, but I ran the `format:markdown` script which formatted those files.

Salakar:
> > Side comment - out of the 8 files changed, 5 were changed in non-semantic / formatting only ways, is that expected behavior? Makes the diff really big and a bit scattered when functionally it's not so large. Not a big deal, just something I noticed
> 
> @mikehardy It should've probably been in its own PR to be honest, but I ran the `format:markdown` script which formatted those files.

Sorted on master, so this should be clean now.


----

In terms of review - I noticed @russellwheatley that the documentation has been done in the messaging package for some reason? Shouldn't be in there

russellwheatley:
@Salakar I have no idea what that is doing in the messaging docs, prolly should've had a bigger slurp of coffee before that commit 😓 .

packages/app/lib/utils/index.js
```diff
@@ -31,6 +31,44 @@ class FirebaseUtilsModule extends FirebaseModule {
...
+        isAvailable: true,
+        status: 0,
+      };
+    }
+    return this.native.androidPlayServices;
+  }
+
+  getPlayServicesStatus() {
+    if (isIOS) {
+      return Promise.resolve();
```
Salakar, line 46: `playServicesAvailability` above returns a default object on iOS, so lets do the same for this one, Promise.resolve the default object, rather than undefined - the user won't need to conditionally switch then based on platform

Types and the test above will need updating though also

russellwheatley: sure. I've updated the return type and the docs.

packages/app/e2e/utils.e2e.js
```diff
@@ -29,4 +29,26 @@ describe('utils()', () => {
...
+
+  describe('getPlayServicesStatus', () => {
+    it('returns isAvailable and Play Service status', async () => {
+      const status = await firebase.utils().getPlayServicesStatus();
+      if (device.getPlatform() === 'android') {
+        should(status.isAvailable).equal(true);
+        should(status.status).equal(0);
+      }
+      if (device.getPlatform() === 'ios') {
+        should(status).equal(undefined);
```
Salakar, line 50: This is still incorrect, CI failing now on iOS 👀

russellwheatley: ah, sorry about that. Totally forgot that the test would have to be updated as well 😓

Issue #3576, opened 2020-04-29
🔥Boolean default remote config values are coerced to integers on iOS


vytautasvargonas:
## Issue
When setting default remote config values as booleans they are coerced to integers - `true` is coerced to `1` and `false` is coerced to `0` on iOS. No coercion happening on android.
```
import { firebase } from '@react-native-firebase/remote-config';

firebase.remoteConfig().setDefaults({ flag1: true, flag2: false });
console.log(firebase.remoteConfig().getAll());
// {flag1: {value: 1, source: "default"}, flag2: {value: 0, source: "default"}} on iOS
// {flag1: {value: true, source: "default"}, flag2: {value: false, source: "default"}} on android
```
**Expected behaviour**
On both - iOS and android - default remote config values should not be coerced.

---

### Javascript

<details><summary>Click To Expand</summary>
<p>

#### `package.json`:

```json
    "@react-native-firebase/app": "6.4.0",
    "@react-native-firebase/auth": "6.4.0",
    "@react-native-firebase/database": "6.4.0",
    "@react-native-firebase/firestore": "6.4.0",
    "@react-native-firebase/remote-config": "6.4.0",
```

#### `firebase.json` for react-native-firebase v6:

```json
# N/A
```

</details>

### iOS

<details><summary>Click To Expand</summary>
<p>

#### `ios/Podfile`:

- [ ] I'm not using Pods
- [x] I'm using Pods and my Podfile looks like:

```ruby
# N/A
```

#### `AppDelegate.m`:

```objc
// N/A
```

</p>
</details>

---

### Android

<details><summary>Click To Expand</summary>
<p>

#### Have you converted to AndroidX?

- [ ] my application is an AndroidX application?
- [ ] I am using `android/gradle.settings` `jetifier=true` for Android compatibility?
- [ ] I am using the NPM package `jetifier` for react-native compatibility?

#### `android/build.gradle`:

```groovy
// N/A
```

#### `android/app/build.gradle`:

```groovy
// N/A
```

#### `android/settings.gradle`:

```groovy
// N/A
```

#### `MainApplication.java`:

```java
// N/A
```

#### `AndroidManifest.xml`:

```xml
<!-- N/A -->
```

</p>
</details>

---

## Environment

<details><summary>Click To Expand</summary>
<p>

**`react-native info` output:**

```
 System:
    OS: macOS Mojave 10.14.6
[...]

PR #3611, opened 2020-05-05
fix(remote-config): remote config defaults

dackers86:
Bug fix for issue https://github.com/invertase/react-native-firebase/issues/3576.


Salakar:
Title should be something like:

`fix(remote-config): remote config defaults`

packages/remote-config/e2e/config.e2e.js
```diff
@@ -304,5 +304,13 @@ describe('remoteConfig()', () => {
       stringValue.should.be.equal('invertase');
       numberValue.should.be.equal(1337);
     });
+
+    it.only('can use boolean values', async () => {
```
Salakar, line 308:
```suggestion
    it('can use boolean values', async () => {
```

packages/remote-config/lib/index.js
```diff
@@ -47,6 +47,10 @@ const statics = {
 const namespace = 'remoteConfig';
 const nativeModuleName = 'RNFBConfigModule';
 
+function isBooleanValue(v) {
+  return v === 'true' || v === 'false' || v === null || v === '0' || v === '1';
```
Salakar, line 51:
```suggestion
  return v === 'true' || v === 'false' || v === '0' || v === '1';
```

Null not a boolean + also already excluding null in the if statement below.

Salakar:
No longer required since Remote Config rework

PR #3663, opened 2020-05-18
feat: upgrade native Firebase SDK version

russellwheatley:
Upgrade every package dependency to the latest Firebase SDK versions for iOS & Android.


Salakar:
Result of debugging `firebase.iid().delete()`: Looks like an iOS SDK issue, stepped through all Firebase iOS SDK code and fails at point of clearing the keychain of previous tokens.

We'll just mark this as a known issue for now as one method failing shouldn't hold up all other modules - not much else we can do; we'll work on a vanilla iOS project reproduction of the issue and raise the issue on the Firebase iOS SDK repo.

PR #3684, opened 2020-05-25
fix!(auth): confirm code returns User instead of UserCredential

### LINKED ISSUE #2543: SMS auth flow response of confirm verification code is invalid
##  Issue
Hello guys, I'm trying to implement signing with SMS provider and I found response mismatch.
If we try to confirm verification code we will get User object (like described in docs: https://rnfirebase.io/docs/v5.x.x/auth/reference/ConfirmationResult#confirm).
In my POV it's not correct, because it's not a added value and you're not passing more needed data like `isNewUser` flag.

As you can see in official Firebase docs this method should return `UserCredentials` (https://firebase.google.com/docs/reference/js/firebase.auth.ConfirmationResult#confirm)

I checked your native code and it's doable because you're passing only user object without `additionalUserInfo` (see the link to ios code: https://github.com/invertase/react-native-firebase/blob/master/packages/auth/ios/RNFBAuth/RNFBAuthModule.m#L728 and android: https://github.com/invertase/react-native-firebase/blob/master/packages/auth/android/src/main/java/io/invertase/firebase/auth/ReactNativeFirebaseAuthModule.java#L956)

Hope it does make sense to you!

gwenoleR:
BREAKING CHANGE: `confirm(verificationCode)` now return `UserCredentials`

### Description

Actually the Phone provider do not follow the Firebase Web SDK.
The [docs](https://firebase.google.com/docs/reference/node/firebase.auth.ConfirmationResult?authuser=0#confirm) said the confirm method should return an `UserCredential` and not only an `User`.
 
### Related issues
Related to [this issue](https://github.com/invertase/react-native-firebase/issues/2543)
Fixes #2543 

### Checklist

  - [x] Yes
  - [x] `Android`
  - [x] `iOS`
- My change includes tests;
- This is a breaking change;
  - [x] Yes
  - [ ] No

### Need help

The method now return a UserCredential, but i have a curious issue. When I log the result using for exemple there few lines of code : 
```
const result = await firebaseConfirmation.confirm(code)
console.log(result)
```
I have the expected result : 
```
{"additionalUserInfo": {"isNewUser": false, "profile": null, "providerId": "phone", "username": null}, "user": {"displayName": "Gwénolé", "email": null, "emailVerified": false, "isAnonymous": false, "metadata": [Object], "phoneNumber": "...", "photoURL": null, "providerData": [Array], "providerId": "firebase", "refreshToken": "...", "uid": "..."}}
```
But when I try to consume, for exemple the `isNewUser` I've got undefined..

I tried to list keys of the object and I have for result 
```
console.log(Object.keys(result))
> ["_auth", "_user"]
```

I think i missed something but i don't know what..
I can only consume the `isNewUser` using : 
```
const isNewUser = result._user.additionalUserInfo.isNewUser
```
🔥

### DIFF
--- a/packages/auth/android/src/main/java/io/invertase/firebase/auth/ReactNativeFirebaseAuthModule.java
+++ b/packages/auth/android/src/main/java/io/invertase/firebase/auth/ReactNativeFirebaseAuthModule.java
@@ -954,7 +954,7 @@ class ReactNativeFirebaseAuthModule extends ReactNativeFirebaseModule {
           TAG,
           "confirmationResultConfirm:signInWithCredential:onComplete:success"
         );
-        promiseWithUser(Objects.requireNonNull(task.getResult()).getUser(), promise);
+        promiseWithAuthResult(Objects.requireNonNull(task.getResult()), promise);
       } else {
         Exception exception = task.getException();
         Log.e(
--- a/packages/auth/e2e/phone.e2e.js
+++ b/packages/auth/e2e/phone.e2e.js
@@ -25,9 +25,9 @@ describe('auth() => Phone', () => {
       confirmResult.verificationId.should.be.a.String();
       should.ok(confirmResult.verificationId.length, 'verificationId string should not be empty');
       confirmResult.confirm.should.be.a.Function();
-      const user = await confirmResult.confirm(TEST_CODE_A);
-      user.should.be.instanceOf(jet.require('packages/auth/lib/User'));
-      user.phoneNumber.should.equal(TEST_PHONE_A);
+      const userCredential = await confirmResult.confirm(TEST_CODE_A);
+      userCredential.user.should.be.instanceOf(jet.require('packages/auth/lib/User'));
+      userCredential.user.phoneNumber.should.equal(TEST_PHONE_A);
     });
 
     it('errors on invalid code', async () => {
--- a/packages/auth/e2e/phone.e2e.js
+++ b/packages/auth/e2e/phone.e2e.js
@@ -18,8 +18,8 @@ describe('auth() => Phone', () => {
     }
   });
 
-  // TODO temporarily disabled tests, these are flakey on CI and sometimes fail - needs investigation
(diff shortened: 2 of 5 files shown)


Salakar:
Awesome, thanks for sorting this - once CI passes I can ship this straight away 🤓

gwenoleR:
Oh yes, `_setUserCredential` obviously does the trick.. 
Work nice on my side 👌

gwenoleR:
Hum, CI failed because of a timeout operation 🤔 
Anyone have any idea what's going on?

mikehardy:
@gwenoleR that's unexpected, of course. I asked Github to rerun all jobs. It was an auth change and the auth seemed to be the problem? So it might be a false-negative but may also be a real issue. We'll see if it reproduces

Salakar:
Shipped in https://github.com/invertase/react-native-firebase/commit/8b72876c749373af4424c68309afb2cea8dc2f07

Thanks :)

Manoj002:
Is the change applicable for older versions too, as im still getting only user object after
`.confirm(verificationCode)`
and error too sms expired.

can anyone please help

gwenoleR:
Hi @Manoj002, this fix make a breaking change, so you must use the v8.0.0 of the auth package

chellongCoder:
i am using **com.google.firebase:firebase-auth:19.3.1**, but when I received otp code, I click notification to move to messenger, but when comeback to app, my otp code is submited responsed the error **The sms verification code used to create the phone auth credential is invalid. Please resend the verification code sms and be sure use the verification code provided by the user.**
any fixed? :(( thanks

mikehardy:
@chellongCoder you are adding a comment not to an issue, but a pull request, and it's a merged/closed one at that.
Read the docs on the auth module on the main doc page where it talks about implementing the auth listener for android to handle the "android auto-verified my one time code" case. That one trips up a lot of people and results in errors like the one you mention. If it doesn't work, open a new issue and fill in the template for support

PR #3702, opened 2020-05-28
feat(firestore): support limitToLast query filter 

russellwheatley:
- set `limit` to `undefined` when `limitToLast` is set after `limit` and vice-versa.
- validate `get` & `snapshot` for presence of `order` if `limitToLast` is set. It is required for both.
- updated mocha tests for Firestore as nested tests weren't included.


packages/firestore/e2e/Query/limitToLast.e2e.js
```diff
@@ -0,0 +1,106 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+const { wipe } = require('../helpers');
+
+describe.only('firestore().collection().limitToLast()', () => {
```
Salakar, line 20:
```suggestion
describe('firestore().collection().limitToLast()', () => {
```

Salakar:
Native crash happening on Android Firestore - needs investigating first, the tests weren't all enabled originally - you had a `.only`

Issue #3714, opened 2020-05-29
[iOS][FCM] Cannot get new token when use 'messaging().deleteToken()'


kratuwu:
## Issue

I need to delete token and get renew token, I try to use `messaging().deleteToken()`
```javascript
await requestUserPermission();
const token1 = await messaging().getToken();
console.log(token1);

await messaging().deleteToken();

const token2 = await messaging().getToken();
console.log(token2);

```
For Android working find but iOS still log same value of token2 and token1
```json
{
  "@react-native-firebase/app": "^7.1.0",
  "@react-native-firebase/messaging": "^7.1.0",
  "react-native": "^0.62.2",
}
```
---

### Javascript

<details><summary>Click To Expand</summary>
<p>

#### `package.json`:

```json
{
  "@react-native-firebase/app": "^7.1.0",
  "@react-native-firebase/messaging": "^7.1.0",
  "react-native": "^0.62.2",
}
```

#### `firebase.json` for react-native-firebase v6:

```json
# N/A
```

</details>

### iOS

<details><summary>Click To Expand</summary>
<p>

#### `ios/Podfile`:

- [ ] I'm not using Pods
- [x] I'm using Pods and my Podfile looks like:

```ruby
platform :ios, '9.0'
require_relative '../node_modules/@react-native-community/cli-platform-ios/native_modules'

def add_flipper_pods!(versions = {})
  versions['Flipper'] ||= '~> 0.33.1'
  versions['DoubleConversion'] ||= '1.1.7'
  versions['Flipper-Folly'] ||= '~> 2.1'
  versions['Flipper-Glog'] ||= '0.3.6'
  versions['Flipper-PeerTalk'] ||= '~> 0.0.4'
  versions['Flipper-RSocket'] ||= '~> 1.0'

  pod 'FlipperKit', versions['Flipper'], :configuration => 'Debug'
  pod 'FlipperKit/FlipperKitLayoutPlugin', versions['Flipper'], :configuration => 'Debug'
  pod 'FlipperKit/SKIOSNetworkPlugin', versions['Flipper'], :configuration => 'Debug'
  pod 'FlipperKit/FlipperKitUserDefaultsPlugin', versions['Flipper'], :configuration => 'Debug'
  pod 'FlipperKit/FlipperKitReactPlugin', versions['Flipper'], :configuration => 'Debug'

  # List all transitive dependencies for FlipperKit pods
  # to avoid them being linked in Release builds
[...]

PR #3725, opened 2020-06-01
@dackers86/e2e authentication updates

dackers86:


mikehardy:
@dackers86 this is only tangentially related to this PR, but there's an active issue with auth().deleteToken right now (#3714) and when I look at the auth.e2e.js file in it's current state in this PR it does not look like the deleteToken API is exercised. Is there any way it could be added? It looks like there may be an underlying SDK bug with deleteToken and a subsequent token fetch and I'm curious if you see that in your environment as well, a test would show it

dackers86:
@mikehardy The issue flares up when introducing a new test scenario.  Looks like there is an issue with deleteToken on IOS

packages/auth/e2e/auth.e2e.js
```diff
@@ -39,9 +39,18 @@ describe('auth()', () => {
   });
 
   describe('applyActionCode()', () => {
+    xit('works as expected', async () => {
```
Ehesp, line 42: Is this still valid to be skipped?

packages/auth/e2e/auth.e2e.js
```diff
@@ -321,6 +336,22 @@ describe('auth()', () => {
       await firebase.auth().signOut();
       await Utils.sleep(50);
     });
+
+    //TODO Readd this
```
Ehesp, line 340: TODO?

packages/auth/e2e/auth.e2e.js
```diff
@@ -856,7 +887,7 @@ describe('auth()', () => {
   });
 
   // TODO temporarily disabled tests, these are flakey on CI and sometimes fail - needs investigation
```
Ehesp, line 889: TODO comment

packages/auth/e2e/emailLink.e2e.js
```diff
@@ -25,6 +25,13 @@ describe('auth() -> emailLink Provider', () => {
       };
       await firebase.auth().sendSignInLinkToEmail(email, actionCodeSettings);
     });
+
+    xit('should send email with defaults', async () => {
```
Ehesp, line 29: Guessing this is skipped because you dont know whether an email is sent?

packages/auth/e2e/phone.e2e.js
```diff
@@ -39,4 +40,110 @@ describe('auth() => Phone', () => {
...
+
+      await firebase
+        .auth()
+        .verifyPhoneNumber(TEST_PHONE_A)
+        .then($ => $);
+
+      return Promise.resolve();
+    });
+
+    it('succesffuly runs and adds emiters', async () => {
```
Ehesp, line 117: succesffuly > successfully

packages/auth/e2e/phone.e2e.js
```diff
@@ -39,4 +40,110 @@ describe('auth() => Phone', () => {
...
+
+      await confirmResult.confirm(TEST_CODE_A);
+
+      const obervserCb = () => {};
+
+      const errorCb = () => {};
+
+      const successCb = () => {};
+
+      await firebase
```
Ehesp, line 129: What's this actually testing?

packages/auth/e2e/auth.e2e.js
dackers86: Yeah, this is a new test that I added. But have no way of testing it at the moment. Will be needed in the future for full coverage at least

packages/auth/e2e/auth.e2e.js
dackers86: Removed in next commit

packages/auth/e2e/auth.e2e.js
dackers86: Removed.

packages/auth/e2e/phone.e2e.js
dackers86: This covers all event emitters can be added to the function, helps with code coverage but could probably do with some kind of  assertion that it has worked?

Ehesp: Yeah if we actually test they work that'd be good. Right how it's just causing code coverage to increase which isn't really a good idea.

packages/auth/e2e/phone.e2e.js
dackers86: Updated

packages/auth/e2e/emailLink.e2e.js
dackers86: I've got this as a broken test from the commit. Will look into why this has been changed...

dackers86: Yeah so on native, iut has success listeners. So I cannot test a response with it set as an email sign in.

```
firebaseAuth
      .createUserWithEmailAndPassword(email, password)
      .addOnSuccessListener(authResult -> {
        Log.d(TAG, "createUserWithEmailAndPassword:onComplete:success");
        promiseWithAuthResult(authResult, promise);
        Log.d(TAG, "createUserWithEmailAndPassword:onComplete:promiseResolved");
      })
      .addOnFailureListener(exception -> {
        Log.e(TAG, "createUserWithEmailAndPassword:onComplete:failure", exception);
        promiseRejectAuthException(promise, exception);
      });
```

packages/auth/e2e/phone.e2e.js
dackers86: Test updated to resolve on a successful emit

.github/workflows/tests_e2e.yml
```diff
@@ -243,7 +243,7 @@ jobs:
         shell: bash
 
       - name: Detox Test
-        timeout-minutes: 20
+        timeout-minutes: 40
```
mikehardy, line 246: changes to the actual workflow should go in separately I think
```suggestion
        timeout-minutes: 20
```

This effectively just a revert of the change here so it's a null diff

PR #3733, opened 2020-06-02
feat: disable direct channels API on iOS, fixes #3674

### LINKED ISSUE #3674: Firebase Messaging iOS SDK version 7.0 will be a breaking change where the SDK will no longer support iOS Direct Channel API
Hi, we have received this message from Firebase

> Firebase Messaging iOS SDK version 7.0 will be a breaking change where the SDK will no longer support iOS Direct Channel API

We are on Firebase/Core 6.13.0 and FirebaseMessaging (~> 4.1.9)
Is iOS Direct Channel used in this version ? 
Can we upgrade to 7.0 ?

Thank you

MrLoh:
### Description

As discussed in #3674 Direct Channel API on iOS was deprecated, so the SDK shouldn't set `shouldEstablishDirectChannel = YES` anymore. I tested that this doesn't affect the delivery of notifications on iOS.

### Related issues

Fixes #3674

### Release Summary

Remove support for deprecated iOS Direct Channel API (#3674)

### Checklist

  - [x] Yes
  - [ ] `Android`
  - [x] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

Tested that notifications (including those with data payloads) still arrive on iOS in all 3 states (killed, background, running)


MrLoh:
There is a comment related to shouldEstablishDirectChannel here: 

https://github.com/invertase/react-native-firebase/blob/4889c7a1c3071f955fbf34d51aa1160b084cd15b/packages/messaging/ios/RNFBMessaging/RNFBMessaging%2BFIRMessagingDelegate.m#L66-L68

I think it can just be removed, but maybe @Salakar or @Ehesp who have the git blame for this can advice.

mikehardy:
@Salakar  / @Ehesp  the related link has the discussion but the change is nearly trivial, if you take as given that direct channel is going away upstream (it is), then setting this as the default is the thing to do

PR #3743, opened 2020-06-04
Better handling of the auto SMS verification on Android

### LINKED ISSUE #464: Phone Auth - phoneAuthSnapshot with null values
### Issue

I was trying to setup phone auth (great work btw), and at first the only issue I was having was `onAuthStateChanged` not getting called after `firebase.auth().signInWithCredential(credential)`. My app is setup in a way that only `onAuthStateChanged` changes redux-store, after I restarted the app to try again, `onAuthStateChanged` got triggered and I logged in normally, but then when i tried to replicate, `phoneAuthSnapshot` was/is
```{verificationId: null, code: null, error: null, state: "verified"}```

Tried deleting the user directly from firebase console, doesn't work either.
Here is my code:
```
  onSubmit(values) {
    const { enterCode, verificationId } = this.state

    // Code was sent already
    if (enterCode) this.signIn(values.code, verificationId)
    else {
      const phoneNumber = `+${values.country.callingCode}${values.phoneNumber}`

      return firebase.auth()
        .verifyPhoneNumber(phoneNumber)
        .then(phoneAuthSnapshot => {
          switch (phoneAuthSnapshot.state) {
            case firebase.auth.PhoneAuthState.CODE_SENT:
              this.setState({ verificationId: phoneAuthSnapshot.verificationId, enterCode: true })
              break;

            case firebase.auth.PhoneAuthState.ERROR:
              toast.danger(phoneAuthSnapshot.error, 'top')
              this.reset()
              break;

            case firebase.auth.PhoneAuthState.AUTO_VERIFY_TIMEOUT:
              toast.danger('Verification timed out')
              this.reset()
              break;

            case firebase.auth.PhoneAuthState.AUTO_VERIFIED:
              this.signIn(phoneAuthSnapshot.code, phoneAuthSnapshot.verificationId)
              break;
          }
        })
        .catch(err => toast.danger(err, 'top'))
    }
  }

  signIn(code, id) {
    const verificationId = this.state.verificationId || id

    if (!verificationId || !code) return toast.danger('', 'top')

    const credential = firebase.auth.PhoneAuthProvider.credential(verificationId, code)
    return firebase.auth().signInWithCredential(credential).catch(() => toast.danger('', 'top'))
  }
```

### Environment

1. Application Target Platform: Android

2. Development Operating System: macOS Sierra 10.12.6

3. Build Tools: Android 23.0.2

4. React Native version: 0.48.4

5. RNFirebase Version: 3.0.0

6. Firebase Module: auth

### PR DESCRIPTION
**Pull request by aaronSig**
Better handling of the auto SMS verification case with custom internal auth credentials and a reusable cache.

### Description

Android suports automatically verifying a phone number. When this happens invoking:

`firebase.auth().verifyPhoneNumber(...)`

Calls back with a`verified` state however at the moment the `verificationId` and `code` are both null.

[Some work had been done](https://github.com/invertase/react-native-firebase/blob/master/packages/auth/android/src/main/java/io/invertase/firebase/auth/ReactNativeFirebaseAuthModule.java#L1379) to cache the credential and reuse it when a null authToken was supplied. However the null came in serialised as a string "null" and failed the check.   

Also the cached credential was removed after it was used. In testing I've seen the phone credential can be used more than once.

To solve I've added a hash map to store the PhoneAuthCredential and return a key derived from the phone number as the verificationId and code. Users can then use the tokens as they would normally, they will be substituted for the real credential at a lower level.

### Related issues

Fixes #464 #482

### Checklist

  - [x] Yes
  - [x] `Android`
  - [ ] `iOS`
- My change includes tests;
- [ ] I have updated TypeScript types that are affected by my change.  N/A
- This is a breaking change;
  - [ ] Yes
  - [x] No (debatable as it wasn't working as recommended before)

### CODE REVIEW

File: packages/auth/android/src/main/java/io/invertase/firebase/auth/ReactNativeFirebaseAuthModule.java [hunk 1]
```diff
@@ -995,14 +997,15 @@ public void verifyPhoneNumber(
     }
 
     // Reset the credential
-    mCredential = null;
+    String newCode = URLEncoder.encode(phoneNumber);
+    String  key = "RNFATOKEN" + newCode;
+    mPhoneAutoVerifyCredentials.remove(phoneNumber);
```
Ehesp, line 1002: I'm a bit confused about this line - when does `mPhoneAutoVerifyCredentials` get added with a credential with the phone number? It gets added with a `key` (String + Phone number), so this seems it'd never do anything?

Ehesp: Also a more simpler way to do thing could be `credential.hashCode().toString()` and storing by the hash code?

mikehardy:
@aaronSig I'm interested in collaborating with you to get this merged, but there are review questions pending a response from you

mikehardy:
Lack of response

PR #3751, opened 2020-06-07
fix(app): convert NativeFirebaseError.getStackWithMessage to static to fix crash

BlaShadow:
### Description

After receiving an error from AdMob, the framework was trying to create a `NativeFirebaseError` but it was not able to call `getStackWithMessage` from the constructor. 

### Checklist

  - [ ] Yes
  - [ ] `Android`
  - [ ] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [ ] No

### Test Plan

---

### DIFF
--- a/tests/ios/Podfile.lock
+++ b/tests/ios/Podfile.lock
@@ -195,26 +195,26 @@ PODS:
     - GoogleToolboxForMac/Defines (= 2.3.0)
     - "GoogleToolboxForMac/NSString+URLArguments (= 2.3.0)"
   - "GoogleToolboxForMac/NSString+URLArguments (2.3.0)"
-  - GoogleUserMessagingPlatform (1.3.0)
-  - GoogleUtilities/AppDelegateSwizzler (7.1.0):
+  - GoogleUserMessagingPlatform (1.4.0)
+  - GoogleUtilities/AppDelegateSwizzler (7.1.1):
     - GoogleUtilities/Environment
     - GoogleUtilities/Logger
     - GoogleUtilities/Network
-  - GoogleUtilities/Environment (7.1.0):
+  - GoogleUtilities/Environment (7.1.1):
     - PromisesObjC (~> 1.2)
-  - GoogleUtilities/ISASwizzler (7.1.0)
-  - GoogleUtilities/Logger (7.1.0):
+  - GoogleUtilities/ISASwizzler (7.1.1)
+  - GoogleUtilities/Logger (7.1.1):
     - GoogleUtilities/Environment
-  - GoogleUtilities/MethodSwizzler (7.1.0):
+  - GoogleUtilities/MethodSwizzler (7.1.1):
     - GoogleUtilities/Logger
-  - GoogleUtilities/Network (7.1.0):
+  - GoogleUtilities/Network (7.1.1):
     - GoogleUtilities/Logger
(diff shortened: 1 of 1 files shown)


mikehardy:
I believe the CI failure was unrelated, I am restarting it - the same thing occurred on #4078 so I think this is a flaky test or something else must have merged which causes a problem

```

  1) firestore().doc().onSnapshot()
       unsubscribes from further updates:
     AssertionError: expected Function { name: '' } to be called twice but was called thrice
    spy([object Object], null) at handleSuccess (http://localhost:8081/index.bundle?platform=android&dev=true&minify=false&sourceMapURL=true:187724:11)
    spy([object Object], null) at handleSuccess (http://localhost:8081/index.bundle?platform=android&dev=true&minify=false&sourceMapURL=true:187724:11)
    spy([object Object], null) at handleSuccess (http://localhost:8081/index.bundle?platform=android&dev=true&minify=false&sourceMapURL=true:187724:11)
  
```

mikehardy:
I looked at this again and actually chatted with a colleague here and we can't figure out what the original issue was exactly or why this change would fix it.

We will need to see a simple App.js that reproduces the problem before we can move forward

thomasonweb:
Hi @mikehardy 

We are seeing the same issue in the Google Play crashes. I'm unable to reproduce it myself but it happens quite frequently. Currently our top crash.

I'm seeing "undefined is not a function" on this line:

`this.stack = this.getStackWithMessage(`NativeFirebaseError: ${this.message}`);`

in `app/lib/internal/NativeFirebaseError.js`

As I'm unable to reproduce it myself I can't really pinpoint or debug why it is happening. Our team is still investigating so if we would find the cause I'm happy to share it. However, turning them static as @BlaShadow is proposing would most likely resolve them.

mikehardy:
@thomasonweb (or @BlaShadow ) If you implement this locally via patch-package and report success, post the exact patch up (so I can compare to verify - as we don't have a solid reproduction, right?) then I could reopen and merge this

I love fixing crash bugs but obviously I need to be careful chasing things without good reproductions...

thomasonweb:
Hi @mikehardy 

To give some context: we used to be on an older version of Firebase and a few weeks ago we started upgrading to v8.3.0 which was now released in production. Because we upgraded we started seeing this crash happening. I've applied a patch where we made the function static and did a rollout into production. I can confirm the crash is gone but we were never been able to find what caused it as we could not reproduce it ourselves.

I've applied the patch on v8.3.0 but the file looks similar in v10.

[@react-native-firebase+app+8.3.0.txt](https://github.com/invertase/react-native-firebase/files/5597136/%40react-native-firebase%2Bapp%2B8.3.0.txt)

BlaShadow:
The first time I got this error my Admob was kind of new and it was failing to retrieve ads (throwing an error). I don't know how to reproduce it now, because the account it's not failing anymore.

mikehardy:
So strange - thanks for reporting back. As mentioned I do love fixing crash bugs but I am trying to be careful here. I'll investigate and see if I can get this in, it does seem like at worst it is harmless, and at best if could fix something.

mikehardy:
This appears to pass my local testing (I pulled the branch and tested it)
I rebased the PR and pushed the result to a fresh PR to make sure it got a clean CI run from the most up to date code and will merge from there #4619

PR #3753, opened 2020-06-08
feat: CLI

Phaen:
# React Native Firebase CLI

A plugin that extends the React Native CLI with Firebase commands to help initialise and diagnose React Native Firebase projects.
<ins>It currently only supports Android.</ins>

## `react-native firebase init`

Initialises Firebase for a project to get it completely up and running.

#### Firebase details

The tool will first try to obtain access to a Firebase account, either through already saved credentials, or through browser authentication. It will use this account throughout the process to communicate with the [Firebase Management API](https://firebase.google.com/docs/projects/api/reference/rest) to setup the project. Once it has access, it asks the user to select a Firebase Project in that account to proceed with. If no projects are available, it will guide the user to making one and quit.

#### `firebase.json`

Based on the dependencies of the project, the tool will compile a list of Firebase settings that are required for those packages. It will then proceed to validate the project's `firebase.json` based on that list. For example, if admob is added as a dependency, the tool will verify that both the keys `admob_ios_app_id` and `admob_android_app_id` exist and their values have the correct format. If a setting fails to verify, the tool will attempt to fix it. In this example, it will guide the user towards the page to obtain the AppID and proceed to prompt for it.

### Android

If the platform `'android'` or `'all'` was passed, or Android was selected in the starting prompt, the tool will proceed to setup Firebase for Android.

#### App

Using the selected Firebase Project, the tool will attempt to find a matching Android App for the current project. If no matching Android App is found, it will attempt to create one itself (**TODO**).

#### `google-services.json`

If a `google-services.json` file already exists, the tool will try to verify that it matches the selected Android App and warn the user if it doesn't. It will then in either case prompt the user to download a new one to replace the current file with. If the file doesn't already exist, the tool will automatically download a new one instead.

#### build.gradle

Based on the dependencies of the project, the tool will compile a list of plugins that are necessary as a Gradle dependency. It will then check whether those plugins are added as a dependency and if they are, whether they are the latest version. If a required dependency is missing or outdated, the tool will fix that.

#### App build.gradle

The tool will again compile a list of plugins that need to be applied in this file, based on the dependencies of the project. Additionally, it also takes in account whether plugins should be applied at the top or bottom of the file. For all plugins in the list, it will attempt to register any that are and move any that are misplaced.

#### Signing Certificate

The tool will ask the user whether they want to add a signing certificate to the Android App in their Firebase Project. If so, it will prompt for its fingerprint, verify that it's the correct format and either SHA-1 or SHA-256 and then add it to the Android App (**TODO**). (**TODO**: Obtain the signing certificate for the project ourselves automatically.)

### iOS

If the platform `'ios'` or `'all'` was passed, or iOS was selected in the starting prompt, the tool should proceed to setup Firebase for iOS. **TODO**

## `react-native firebase doctor`

This command is essentially a stripped down `init`, calling just the validation features and compiling the results into a list, similar to `react-native doctor`. It additionally also lists all the RNFB and RN packages used as a dependency and their versions. This command does not modify any files whatsoever.

As of now, it does not implement any fix functionality like the original RN command.

### Safety checks

Before running any commands that may modify any files, the tool first checks whether the repository is clean. If there are any pending changes, it will simply display an error and quit. This behaviour can be bypassed with the `--force` option, displaying a warning for doing so instead.

### Gradle plugin versions

The tool keeps track of the versions for the Gradle plugins to use as a dependency in the `plugin-versions.json` file, at the root of the CLI package. Any plugin listed in here will have its version automatically updated by running `yarn update-plugin-versions`. It does so by scraping the [Maven Repository](https://mvnrepository.com/) for each plugin, to obtain its latest version. After running this command, you should take care to verify that the versions are indeed the latest and compatible with RNFB.

# Todo

## Resource creation

Creating new resources is currently not functional. All of the implementation is there, save for the ID of the new resource, that the Firebase Management API seems to require. It seems to me that you first need to create a resource to obtain its ID, which is however required to create a new resource. I cannot find any documentation on how this is supposed to work.

## Signing Certificate

Creating a signing certificate and obtaining its SHA fingerprint can be complicated, so it would be really nice if we could automate that process. As of now, it completely relies on the user to obtain it.

## iOS

I'm too poor to afford an iOS machine to work on this part myself. If someone could pick this up, that'd be great. With no Gradle shenanigans on the iOS side, there should not be that much to do.


packages/cli/src/commands/doctor.ts
```diff
@@ -1,5 +1,63 @@
...
+
+  log.info('# iOS');
+  const iosGoogleServicesFile = await file.readIosGoogleServices(iosProjectConfig);
+  if (iosGoogleServicesFile) {
+    log.success('A "GoogleService-Info.plist" has been found.');
+  } else {
+    log.warn(
+      'No "GoogleService-Info.plist" file has been found, Firebase for iOS has not been setup.',
+    );
+  }
```
Salakar, line 106: Can we have the doctor command print the following also, useful for issue triage;

 - `@react-native-firebase/*` NPM packages being used & their versions (could read from package.json)
 - `react-native` version being used (could read from package.json)

Salakar: cc @mikehardy for additional thoughts

mikehardy: Very interesting... I like the SHA check, should check debug and release somehow but I'm not sure how feasible that is. Thinking of where all the issue traffic comes from, I think missing an auth listener while you are using the auth package is a red flag especially if a phone auth confirm code is directly checked. Integrating react-native-splash-screen as LAUNCHER activity instead of MainActivity is a red flag. So much messaging stuff - putting a background handler outside of App.js/index.js is a warning ? Modules in package.json that are not in Podfile.lock (pod install failure).

dackers86:
Code cleaned and merged. Current version provides results similar to...

image

mikehardy:
That is nifty! I'm traveling right now but I'm excited to check this one out

dackers86:
# Doctor Pre Release Checklist / Testing (WIP)

## Setup / Installation

- [x] Successfully finds current Firebase account.
- [x] Successfully logs into Firebase account when selected.
- [ ] Checks for any changes in the workspace before running any cli commands.
- [ ] Checks for any changes in the workspace and successfully overrides with-- force flag

## Common
- [ ] Correct RNFB packages and versions are listed

## Android

- [ ] Lists current Firebase app
- [x] States `Not found` if Firebase app is not setup correctly.
- [ ] Lists successful google-services.json
- [x] Lists `Unable to validate` if invalid google-services.json
- [ ] Successfully lists `App` build.gradle plugins.

## iOS

- [ ] GoogleService-Info.plist is correctly validated.
- [x] GoogleService-Info.plist is correctly in-validated.

## Web

- [ ] Not supported

dackers86:
# Init Pre Release Checklist / Testing (WIP)

## Setup / Installation

- [ ] Checks for any changes in the workspace before running any cli commands.
- [ ] Checks for any changes in the workspace and successfully overrides with-- force flag
- [ ] Can run through `npx`

Issue #2753, opened 2019-10-20
v6 Storage refFromUrl does not work correctly


ThomasFindlay:
## Issue

Describe your issue here

v6 refFromURL does not work on the url below. v5 worked without problems.

```
firebase.storage().refFromURL(https://firebasestorage.googleapis.com/v0/b/<project-name>.appspot.com/o/<image-path>)
```

In v6 it throws an error that object was not found even though it exists. I was able to delete the file when using .ref(path).delete(). I don't think that the stack trace is really useful, but here it is.

```
code:"storage/object-not-found"
jsStack:"Error↵    at Object.delete (blob:file:///06ce93ab-5900-4f65-9a07-f6d7415cdc5e:97535:23)↵    at StorageReference._delete (blob:file:///06ce93ab-5900-4f65-9a07-f6d7415cdc5e:106533:43)↵    at _removeProfilePic$ (blob:file:///06ce93ab-5900-4f65-9a07-f6d7415cdc5e:348611:59)↵    at tryCatch (blob:file:///06ce93ab-5900-4f65-9a07-f6d7415cdc5e:5488:19)↵    at Generator.invoke [as _invoke] (blob:file:///06ce93ab-5900-4f65-9a07-f6d7415cdc5e:5663:24)↵    at Generator.prototype.(anonymous function) [as next] (blob:file:///06ce93ab-5900-4f65-9a07-f6d7415cdc5e:5531:23)↵    at tryCatch (blob:file:///06ce93ab-5900-4f65-9a07-f6d7415cdc5e:5488:19)↵    at invoke (blob:file:///06ce93ab-5900-4f65-9a07-f6d7415cdc5e:5564:22)↵    at blob:file:///06ce93ab-5900-4f65-9a07-f6d7415cdc5e:5594:13↵    at tryCallTwo (blob:file:///06ce93ab-5900-4f65-9a07-f6d7415cdc5e:49272:7)"
message: "[storage/object-not-found] No object exists at the desired reference."
namespace: "storage"
nativeErrorCode: null
nativeErrorMessage: null
[...]

PR #3775, opened 2020-06-11
fix(storage): Changed refFromUrl regex to exclude appspot.com

dackers86:
Fix for issue https://github.com/invertase/react-native-firebase/issues/2753.

When extracting the url, regex currently excludes appspot.com. This has now been included so that the domain is included.

Format changed from {bucket} to {bucket}.appspot.com

### DIFF
--- a/packages/storage/e2e/StorageTask.e2e.js
+++ b/packages/storage/e2e/StorageTask.e2e.js
@@ -435,7 +435,7 @@ describe('storage() -> StorageTask', () => {
 
       task.on('state_changed', {
         error: error => {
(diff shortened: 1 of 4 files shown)


mikehardy (approved):
@dackers86 finally unwinding some of the PR stack! This one looks fine, I de-merge-committed it, stripped it down to just the commits doing real changes and rebased it against master so it was clean+easy-to-review and it seems to pass everything locally

This is the one that re-enables the storage tests so I'm excited to get it in prior to the big breaking change about to land

Cheers

Issue #3671, opened 2020-05-20
Current user's creationtime,lastsignintime is wrong


erhan355:
## Issue

When you get current user's info, creationtime is different than firebase console value.It seems when you create multiple users on the same android emulator creation time is never updated.
---

### Javascript
 const { currentUser } = auth();
    await currentUser.reload();
let creationTime=currenUser.metadata.creationTime // Returns 2020-04-07T22:28:20Z (UTC)
//In fact I created this user today.
<details><summary>Click To Expand</summary>
<p>

#### `package.json`:

```json
# N/A
```

#### `firebase.json` for react-native-firebase v6:

```json
# N/A
```

</details>

### iOS

<details><summary>Click To Expand</summary>
<p>

#### `ios/Podfile`:

- [ ] I'm not using Pods
- [x] I'm using Pods and my Podfile looks like:

```ruby
# N/A
```

#### `AppDelegate.m`:

```objc
// N/A
```

</p>
</details>

---

### Android

<details><summary>Click To Expand</summary>
<p>

#### Have you converted to AndroidX?

- [ ] my application is an AndroidX application?
- [ ] I am using `android/gradle.settings` `jetifier=true` for Android compatibility?
- [ ] I am using the NPM package `jetifier` for react-native compatibility?

#### `android/build.gradle`:

```groovy
// N/A
```

#### `android/app/build.gradle`:

```groovy
// N/A
```

#### `android/settings.gradle`:

```groovy
// N/A
```

#### `MainApplication.java`:

```java
// N/A
```

#### `AndroidManifest.xml`:

```xml
<!-- N/A -->
```

</p>
</details>

---

## Environment
+-- @react-native-firebase/app@6.3.4
+-- @react-native-firebase/auth@6.3.4
<details><summary>Click To Expand</summary>
<p>

**`react-native info` output:**

```
System:
    OS: Windows 10
    CPU: (8) x64 Intel(R) Core(TM) i7-3630QM CPU @ 2.40GHz
    Memory: 4.47 GB / 15.89 GB
  Binaries:
    Node: 12.9.1 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.15.2 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
    npm: 6.13.7 - C:\Program Files\nodejs\npm.CMD
  SDKs:
    Android SDK:
      API Levels: 23, 27, 28
      Build Tools: 27.0.3, 28.0.3, 29.0.3
[...]

PR #3776, opened 2020-06-11
tests(auth): added tests for checking creation time across multiple created users

dackers86:
Investigation into creation times that are not updating when creating multiple users on an emulator. Please see https://github.com/invertase/react-native-firebase/issues/3671#event-3366923083


packages/auth/e2e/auth.e2e.js
```diff
@@ -58,6 +58,47 @@ describe('auth()', () => {
...
+      firstUser.metadata.creationTime.should.not.equal(secondUser.metadata.creationTime);
+    });
+
+    it('Meta data returns as expected with email and password sign in', async () => {
+      const random = Utils.randString(12, '#aA');
+      const email1 = `${random}@${random}.com`;
+      const pass = random;
+
+      await firebase.auth().createUserWithEmailAndPassword(email1, pass);
+      await Utils.sleep(100);
```
Ehesp, line 84: What's this sleep needed for?

dackers86: Current user was unavailable, it appeared that the test was too quick. Seemed a bit of an anomaly at the time.

Happy to retest...?

dackers86: PR updated

PR #3790, opened 2020-06-16
fix(ios): handle nil messaging token

TPXP:
# Description

On iOS, calling `[FIRInstanceID instanceID] deleteIDWithHandler` causes a crash. I quite like this method since it enables me to drop all tokens generated until now, and make sure we won't send any more messages to this device (use case: user logout).

### Related issues
Not reported.

### Release Summary
[messaging] Fixed a crash if the fetched token is null (typically when deleting the cloud messaging instance).

### Checklist

  - [x] Yes
  - [x] `iOS`
- My change includes tests: **NO** (but maybe it should!)
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan
You can try to call `[FIRInstanceID instanceID] deleteIDWithHandler` in your app code (exposing it as a JS API would be welcome by the way!).

Thanks a lot for this project!

### DIFF
--- a/packages/messaging/ios/RNFBMessaging/RNFBMessaging+FIRMessagingDelegate.m
+++ b/packages/messaging/ios/RNFBMessaging/RNFBMessaging+FIRMessagingDelegate.m
@@ -49,8 +49,9 @@
 
 // JS -> `onTokenRefresh`
 - (void)messaging:(FIRMessaging *)messaging didReceiveRegistrationToken:(NSString *)fcmToken {
-  if(fcmToken == nil) // Don't crash when the token is reset
+  if (fcmToken == nil) { // Don't crash when the token is reset
     return;
+  }
   [[RNFBRCTEventEmitter shared] sendEventWithName:@"messaging_token_refresh" body:@{
       @"token": fcmToken
   }];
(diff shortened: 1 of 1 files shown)


packages/messaging/ios/RNFBMessaging/RNFBMessaging+FIRMessagingDelegate.m
```diff
@@ -49,6 +49,8 @@ - (void)observe {
 
 // JS -> `onTokenRefresh`
 - (void)messaging:(FIRMessaging *)messaging didReceiveRegistrationToken:(NSString *)fcmToken {
+  if(fcmToken == nil) // Don't crash when the token is reset
+    return;
```
mikehardy, line 53: Just a whitespace nit - adding a JS API might not be so hard, I think the reason it doesn't exist is because it doesn't have any corresponding APIs on the javascript API and the public API surface area for this package is usually defined by the JS API. Not sure what the policy is on non-JS interfaces but as long as it was coherent across iOS and Android to match the rest of the Instance ID API (none of which has corresponding JS interfaces) it would probably go in

```suggestion
  if (fcmToken == nil) { // Don't crash when the token is reset
    return;
  }
```

Salakar:
Re the JS api, its already available through the Instance ID package: https://rnfirebase.io/reference/iid#delete

Salakar:
LGTM if this fixes a crash. Though one minor point; should we not be sending `null` tokens to JS so you know when a token has been invalidated? I think Android sends null - the JS SDK Web types don't really indicate what should happen here and is just typed as `any` 🙈

PR #3795, opened 2020-06-17
fix(auth,android): too-many-requests code now correctly returned

dackers86:
*Testing*

Update quote for testing purposes only (reverted once testing has finished)

image

to

image

Tests results ran locally return the following...

image

### DIFF
--- a/packages/auth/android/src/main/java/io/invertase/firebase/auth/ReactNativeFirebaseAuthModule.java
+++ b/packages/auth/android/src/main/java/io/invertase/firebase/auth/ReactNativeFirebaseAuthModule.java
@@ -33,6 +33,7 @@ import com.google.android.gms.tasks.OnCompleteListener;
 import com.google.firebase.FirebaseApp;
 import com.google.firebase.FirebaseException;
 import com.google.firebase.FirebaseNetworkException;
+import com.google.firebase.FirebaseTooManyRequestsException;
 import com.google.firebase.auth.ActionCodeResult;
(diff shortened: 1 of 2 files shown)


packages/auth/android/src/main/java/io/invertase/firebase/auth/ReactNativeFirebaseAuthModule.java
```diff
@@ -1781,6 +1782,10 @@ private WritableMap getJSError(Exception exception) {
       } else if (exception instanceof FirebaseNetworkException) {
         code = "NETWORK_REQUEST_FAILED";
       }
+      else if (exception instanceof FirebaseTooManyRequestsException) {
+        code = "too-many-requests";
```
Salakar, line 1786:
```suggestion
      } else if (exception instanceof FirebaseTooManyRequestsException) {
        code = "TOO_MANY_REQUESTS";
```

packages/auth/e2e/user.e2e.js
```diff
@@ -283,6 +283,32 @@ describe('auth().currentUser', () => {
       // Clean up
       await firebase.auth().currentUser.delete();
     });
+
+    xit('should throw too many requests when limit has been reached', async () => {
```
Salakar, line 287:
```suggestion
    // Can only be run/reproduced locally with Firebase Auth rate limits lowered on the Firebase console.
    xit('should throw too many requests when limit has been reached', async () => {
```

Issue #3719, opened 2020-05-30
[Error: [storage/unknown] Permission denied] line: 101, column: 244, sourceURL: 'index.android.bundle'


pclacs:
## Issue
[Error: [storage/unknown] Permission denied] line: 101, column: 244, sourceURL: 'index.android.bundle'

I am getting this when using utils.FilePath.DOCUMENT_DIRECTORY and utils.FilePath.PICTURES_DIRECTORY on Android 10.0 and Android 8.1 devices (the error messages are on both devices and are permissions related).

If I change the value to EXTERNAL_DIRECTORY everything works, but this makes the files visible to android phone users. The app was originally running on RN 0.59.8 and RN firebase v5 and the DOCUMENT_DIRECTORY location did not cause any problems and still works when downloaded from the appstore.

The code is included here. I have arrowed where the permission denied error is being reported in the code below.

```
import { Platform } from 'react-native';
import { firebase, utils } from '@react-native-firebase/app';
import storage from '@react-native-firebase/storage';
import functions from '@react-native-firebase/functions';
import fs from 'react-native-fs';

type AvailableDownloads = {|
  data: {
    fileDownloads: { [filename: string]: string },
  },
|};
type DownloadErrorAction = { type: 'DOWNLOAD_ERROR', err: any };
type DownloadStartAction = { type: 'DOWNLOAD_START' };
type DownloadSuccessAction = { type: 'DOWNLOAD_SUCCESS' };
export type DownloadAction =
  | DownloadErrorAction
  | DownloadStartAction
  | DownloadSuccessAction;

export const DOWNLOAD_ERROR = 'DOWNLOAD_ERROR';
export const DOWNLOAD_START = 'DOWNLOAD_START';
export const DOWNLOAD_SUCCESS = 'DOWNLOAD_SUCCESS';

const BASE_DIR = utils.FilePath.DOCUMENT_DIRECTORY;

export const download = () => async (dispatch: any => void) => {
  dispatch(downloadStart());

  try {
    // Generate the Firebase Functions callable function
    const callable = firebase.app().functions('europe-west1').httpsCallable('FILELIST');
    
    const response: AvailableDownloads = await callable({});

      const downloadsAndroid = [];
      Object.keys(response.data.fileDownloads).forEach((filename: string) => {

          const downloadTo = `${BASE_DIR}/${filename}`
          const downloadFrom = '/audio/' + filename

          const audioDownload = storage()
              .ref(downloadFrom)
[...]

PR #3803, opened 2020-06-19
bug(storage): deprecated storage paths investigation

dackers86:
Investigation into https://github.com/invertase/react-native-firebase/issues/3719#event-3453380108


Salakar:
Doesn't seem to be any changes required here anymore as they're all already on master.

PR #3811, opened 2020-06-20
tests: bump soloader dependency to fix ABI lib load issues

mikehardy:
soloader 0.8.0 has problems loading ABI-specific libraries in
universal APKs, this workaround appears effective and allows the
test APK to run on x86 emulators in developer environments

Can likely be removed when moving to react-native 0.63.0

https://github.com/facebook/react-native/issues/25923#issuecomment-594047945

### DIFF
--- a/tests/android/app/build.gradle
+++ b/tests/android/app/build.gradle
@@ -117,6 +117,13 @@ android {
   }
 }
 
+// Can probably remove this when tests upgrades to react-native 0.63
+configurations.all {
+    resolutionStrategy {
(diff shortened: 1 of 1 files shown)


mikehardy:
@Salakar remember your fit with github actions loading the DSOs correctly in the emulator for tests?
I think this is the "fix" (with alternative / future fix being that RN0.63+ includes the soloader version sans bug w/o specially requiring it)

Salakar:
That was 2 days of fun I spent on that 😭 glad the actual cause has been found at least, thanks!

Issue #3614, opened 2020-05-05
onSnapshot hanging after killing iOS and restarting


dsessoms:
## Issue
Using onSnapshot works when first installing an iOS app with a release scheme, but once the app is killed and restarted a couple of times (1-2 on average), onSnapshot fails to return data or throw an error. 

After extensive testing I found that setting `includeMetadataChanges` to true prevented this issue, but does lead to more events than desired. 

```typescript
useEffect(() => {
    if (!uid || uid === '') {
      log(`uid: ${uid}`);
      setBoordUser(undefined);
      return;
    }
    log(`uid: ${uid}`);
    const subscriber = firestore()
      .collection('users')
      .doc(uid)
      .onSnapshot(
        (snapshot) => {
          if (!snapshot || !snapshot.exists) {
            setBoordUser(undefined);
          } else {
            setBoordUser(snapshot.data() as IBoordUser);
          }
        },
        (error) => {
          log(JSON.stringify(error));
        },
      );
    return () => subscriber();
  }, [uid]);
```

---

### Javascript

<details><summary>Click To Expand</summary>
<p>

#### `package.json`:

```json
{
  "version": "0.0.1",
  "private": true,
  "scripts": {
    "android": "npx react-native run-android --verbose",
    "android-release": "npx react-native run-android --variant=release",
    "ios": "react-native run-ios",
    "start": "react-native start",
    "code-push": "ts-node ./bin/code-push.ts",
    "bundle-android": "cd android && ./gradlew bundleRelease",
    "start-android-release": "npx react-native run-android --variant=release",
    "test": "jest",
    "compile": "tsc --noEmit",
    "lint": "eslint . --ext .js,.jsx,.ts,.tsx",
    "dev-tools": "react-devtools",
    "postinstall": "npx jetify"
  },
  "dependencies": {
    "@expo/react-native-action-sheet": "^3.6.0",
    "@invertase/react-native-apple-authentication": "^0.2.0",
    "@react-native-community/cameraroll": "^1.5.2",
    "@react-native-community/google-signin": "^4.0.0",
    "@react-native-community/masked-view": "^0.1.7",
    "@react-native-firebase/analytics": "6.7.1",
    "@react-native-firebase/app": "6.7.1",
    "@react-native-firebase/auth": "6.7.1",
    "@react-native-firebase/database": "6.7.1",
    "@react-native-firebase/dynamic-links": "6.7.1",
    "@react-native-firebase/firestore": "6.7.1",
[...]

PR #3881, opened 2020-07-05
feat(android,ios): upgrade native SDK versions

andersonaddo:
I did this mostly to solve #3614 for everyone. Is this all that's required?

I think this should be fine, since the [tests already override to SDK v6.27.0](https://github.com/invertase/react-native-firebase/blob/65042d08ed1233cfdb35b6c2d51d87be2d305d3c/tests/ios/Podfile#L7).

Looking at the [changelog](https://firebase.google.com/support/release-notes/ios#cloud-firestore), I think the only changes that might change the dev experience are:

[Installations change 3](https://firebase.google.com/support/release-notes/ios#installations_2)

[Remote Config change 1](https://firebase.google.com/support/release-notes/ios#remote-config_1)

[Storage change 2](https://firebase.google.com/support/release-notes/ios#storage)


andersonaddo:
Doesn't look like there's been a new BoM for Android since 25.5.0

Salakar:
Thanks for this, LGTM, though could you change https://github.com/invertase/react-native-firebase/blob/master/packages/app/package.json#L70 as well, thanks

andersonaddo:
Added the Android upgrade.
Note that this drops support for Gradle v4

mikehardy:
I think gradle is on v6 now, so that's okay, react-native hasn't been on v4 for a while? Should be fine?
Ah wait! https://github.com/facebook/react-native/commit/b4017a9923b09fed4b693a8e4cfadd30ce34c88d

react-native 0.60 was the first to use it. So there are a few people using RNFBv6+ with RN59-, they need a warning that they will have to override things, so this is a major version bump for android :-)

But version numbers are free! Just needs a changelog entry warning IMHO

Salakar:
LGTM

I wonder if we even need to major version this, I thought v6 was RN 60 min with 59 as "it works" - can't find the support table anywhere though 🙈 so can't confirm

mikehardy:
I think you are right, if RN59 is not officially supported, you could just put a very obvious change note for RN59 people. That seems friendly enough. It's a fail-fast as well they'll find out quite quickly, not like an obscure prod only runtime issue

PR #3892, opened 2020-07-07
docs: setBackgroundMessageHandler isn't Android-only

andersonaddo:
setBackgroundMessageHandler isn't an Android-only function.


Salakar:
Nice catch, thanks

PR #3895, opened 2020-07-07
tests(firestore): additional tests for numeric where clauses

dackers86:
Added additional testing to investigates issue querying numeric values.

Cases include 'in' and 'array-contains' operators


packages/firestore/e2e/Query/where.e2e.js
```diff
@@ -410,4 +410,32 @@ describe('firestore().collection().where()', () => {
...
+    const items = [];
+    await ref
+      .where('status', 'in', [1, 2])
+      .get()
+      .then($ => $.forEach(doc => items.push(doc.data())));
+
+    items.length.should.equal(1);
+  });
+
+  it.only('should correctly query integer values with array-contains operator', async () => {
```
russellwheatley, line 428: remove `only` from test.

dackers86: Good spot. Removed!

dackers86: In terms of reference there is a before function running...

```js

before(() => wipe());

```

Which should take care of it!

Issue #3836, opened 2020-06-25
Firebase Realtime Database snapshot.forEach was not display suggestions


Eumhongin:
<img width="518" alt="스크린샷 2020-06-25 오후 11 04 56" src="https://user-images.githubusercontent.com/23710025/85735105-6e038f00-b738-11ea-92db-c819e33c737b.png">

Maybe Functions is Problem..

[https://github.com/invertase/react-native-firebase/issues/3823](url)
i think same problem

This is Firebase Database Admin ts
<img width="549" alt="스크린샷 2020-06-25 오후 11 10 09" src="https://user-images.githubusercontent.com/23710025/85735929-0863d280-b739-11ea-948e-737337dc8bb0.png">

Then i Fixed it. suggestion work.
<img width="558" alt="스크린샷 2020-06-25 오후 11 13 32" src="https://user-images.githubusercontent.com/23710025/85736496-8031fd00-b739-11ea-9d10-c7a4bb65f2f0.png">

how can i push this code?..

andersonaddo:
Go ahead and make a pull request!

https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/creating-a-pull-request

https://opensource.com/article/19/7/create-pull-request-github

PR #3963, opened 2020-07-17
Suggestion Update

Eumhongin:
Realtime Database Suggestion Fix. #3836

### DIFF
--- a/packages/database/lib/index.d.ts
+++ b/packages/database/lib/index.d.ts
@@ -167,9 +167,13 @@ export namespace FirebaseDatabaseTypes {
     child(path: string): Reference;
 
     /**
-     * The last part of the Reference's path.
-     * For example, "ada" is the key for https://<DATABASE_NAME>.firebaseio.com/users/ada.
-     * The key of a root Reference is null.
+     * Returns an `OnDisconnect` object - see
+     * {@link
+     *   https://firebase.google.com/docs/database/web/offline-capabilities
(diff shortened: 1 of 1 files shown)


packages/database/lib/index.d.ts
```diff
@@ -167,9 +167,13 @@ export namespace FirebaseDatabaseTypes {
...
-     * The last part of the Reference's path.
-     * For example, "ada" is the key for https://<DATABASE_NAME>.firebaseio.com/users/ada.
-     * The key of a root Reference is null.
+     * Returns an `OnDisconnect` object - see
+     * {@link
+     *   https://firebase.google.com/docs/database/web/offline-capabilities
+     *   Enabling Offline Capabilities in JavaScript} for more information on how
+     * to use it.
+     *
+     * @return An `OnDisconnect` object .
```
mikehardy, line 176: This does not seem correct?

The rest seem like good changes - typing the function return more strongly is a benefit I think

Eumhongin: so.. What can I do?

this is my first pull request and i am begginer in github..

mikehardy: The documentation here in this part is saying it will return an `onDisconnect` object etc, but is connected to the `key` property. So the documentation is not valid and needs correction

mikehardy:
Unfortunately, closed due to inactivity. If you would like to get this in please rebase to master (there has been work in the area while this sat idle), address review comments, and re-submit

PR #4030, opened 2020-07-31
fix(messaging): added missing `from` property in Remote Message type

emeraldsanto:
### Description

I believe the `from` property was accidentally omitted from the `FirebaseMessagingType.RemoteMessage` interface. I've added it following the standards set by the rest of the properties.

### Checklist

  - [x] Yes
  - [x] `Android`
  - [x] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### DIFF
--- a/packages/messaging/lib/index.d.ts
+++ b/packages/messaging/lib/index.d.ts
@@ -89,7 +89,6 @@ export namespace FirebaseMessagingTypes {
     * The topic name or message identifier.
     */
     from?: string;
-    
     /**
      * The address for the message.
      */
--- a/packages/messaging/lib/index.d.ts
+++ b/packages/messaging/lib/index.d.ts
@@ -85,7 +85,7 @@ export namespace FirebaseMessagingTypes {
      */
     messageType?: string;
     /**
-    * The topic name or message identifier.
+     * The topic name or message identifier.
     */
     from?: string;
     /**
--- a/packages/messaging/lib/index.d.ts
+++ b/packages/messaging/lib/index.d.ts
@@ -86,7 +86,7 @@ export namespace FirebaseMessagingTypes {
     messageType?: string;
     /**
      * The topic name or message identifier.
-    */
+     */
     from?: string;
     /**
      * The address for the message.
--- a/packages/messaging/lib/index.d.ts
+++ b/packages/messaging/lib/index.d.ts
@@ -84,7 +84,6 @@ export namespace FirebaseMessagingTypes {
      * The message type of the message.
      */
     messageType?: string;
-    
     /**
     * The topic name or message identifier.
     */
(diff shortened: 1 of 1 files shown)


emeraldsanto:
I hadn't seen this but I can confirm I am using this property in production to differentiate between Firebase topics.

```typescript
/**
 * Handles an incoming data-only message from Firebase
 * @param {object} message The message sent by Firebase
 */
export function handleRemoteMessage(message: FirebaseMessagingTypes.RemoteMessage) {
    if (!message.from || !message.data) {
        return;
    }

    // The `from` property is in the `/topics/XXX` format
    const topicName = message.from.slice(message.from.lastIndexOf("/") + 1);

    switch (topicName) {
        case FirebaseTopic.REMOTE_CONFIG_UPDATE:
            EncryptedStorage.setItem(
                StorageKey.REMOTE_CONFIG_STATE,
		message.data.configState ?? RemoteConfigState.STALE
            );
            break;

        default:
            log("Notifications", `Received remote message with unknown topic ${topicName}`, "warn");
            break;
    }
}
```

emeraldsanto:
Yeah maybe on my end as well, I'm pretty sure it used to exist in V5 otherwise I would never have known about it.

packages/messaging/lib/index.d.ts
```diff
@@ -84,7 +84,12 @@ export namespace FirebaseMessagingTypes {
      * The message type of the message.
      */
     messageType?: string;
-
+    
+    /**
+    * The topic name or message identifier.
+    */
+    from?: string;
+
```
Salakar, line 92:
```suggestion

```

packages/messaging/lib/index.d.ts
```diff
@@ -84,7 +84,12 @@ export namespace FirebaseMessagingTypes {
      * The message type of the message.
      */
     messageType?: string;
-
+
```
Salakar, line 87:
```suggestion

```

packages/messaging/lib/index.d.ts
```diff
@@ -84,7 +84,12 @@ export namespace FirebaseMessagingTypes {
      * The message type of the message.
      */
     messageType?: string;
-
+    
+    /**
+    * The topic name or message identifier.
```
Salakar, line 89:
```suggestion
     * The topic name or message identifier.
```

packages/messaging/lib/index.d.ts
```diff
@@ -84,7 +84,12 @@ export namespace FirebaseMessagingTypes {
...
+    */
```
Salakar, line 90:
```suggestion
     */
```

Salakar:
@mikehardy - gets created on Native side only; https://github.com/invertase/react-native-firebase/blob/master/packages/messaging/android/src/main/java/io/invertase/firebase/messaging/ReactNativeFirebaseMessagingSerializer.java#L67-L69 does it exist though!

PR #4043, opened 2020-08-04
chore(auth): removed obsolete auth directory for android

dackers86:
Remove obsolete directory for Android React-Native Auth module


Salakar:
> LGTM - but of course the E2E test fails (flaky), but not on the compile and surely the compile would fail if this were invalid as Java is strongly typed...

😭 CI. Pushed to disable those flakey tests - seen them come up a few times, needs investigating.

mikehardy:
@Salakar the flakiness I see is usually a complete failure of the packager/detox/app-on-emulator stack to connect and start running anything, resulting in timeout fail

Salakar:
> @Salakar the flakiness I see is usually a complete failure of the packager/detox/app-on-emulator stack to connect and start running anything, resulting in timeout fail

This particular issue for Android is also happening on FlutterFire annoyingly, so imo it's a general issue with Android emulators on GitHub actions, most likely resource constraints given the low spec of the current GH Action machines and that the APK takes 2-3 minutes to install and then timeouts out when trying to launch (Flutter & RN). There's customisable GitHub action runner sizes in beta at the moment according to the GitHub roadmap, so hopefully we'll have that in GA soon and we can try bump up the machine size. That or look at moving E2E only to a different CI provider.

mikehardy:
Hmmm are you sure APK takes 2-3mins to install? I used to see that with non-accelerated emulators (ARM emus running on x86 VM hosts) and it was problematic, but on accelerated VM hosts (like the macOS GH actions executors you use) it takes seconds and should never fail. The APK install would actually fail if it took that long anyway as adb has internal timers that inspire and are noisy about it

I noticed this same packager/detox/app-on-emu failure locally which is accelerated emulators and I know the APK is installing quickly

My hypothesis is that it's something different unless you're 100% on APK install timeouts.

PR #4068, opened 2020-08-11
fix(remote-config): error calling getValue() before fetch

AlexeyVlasenko:
### Description

remoteConfig().getValue() errored when it invokes before fetch values success/failed, because this._values is undefined.
Added check for the correct type of "target" to hasOwnProperty method to prevent this issue.

### Checklist

  - [x] Yes
  - [x] `Android`
  - [x] `iOS`
- My change includes tests;
  - [-] `e2e` tests added or updated in `packages/\*\*/e2e`
  - [-] `jest` tests added or updated in `packages/\*\*/__tests__`
- [-] I have updated TypeScript types that are affected by my change.
- This is a breaking change;
  - [ ] Yes
  - [-] No


AlexeyVlasenko:
> Interesting - this looks good to me but since this is a common function I'd like a second pair of 👀

As an option, we can check is ```this._values !== undefined``` inside `getValue()` method.

Salakar:
> > Interesting - this looks good to me but since this is a common function I'd like a second pair of 👀
> 
> As an option, we can check is `this._values !== undefined` inside `getValue()` method.

I'd prefer we did this rather than change a function in core 😅 would you mind making this change instead?

AlexeyVlasenko:
> > > Interesting - this looks good to me but since this is a common function I'd like a second pair of 👀
> > 
> > 
> > As an option, we can check is `this._values !== undefined` inside `getValue()` method.
> 
> I'd prefer we did this rather than change a function in core 😅 would you mind making this change instead?

Not a problem!

Salakar:
Thank you :) LGTM, waiting for CI then I'll merge.

Salakar:
LGTM GIF

PR #4069, opened 2020-08-11
use @import Firebase; instead of #import <Firebase.h>

Lucas-Geitner:
Continue to use @import Firebase instead of #import <Firebase.h> who failed to archive on IOS

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -104,7 +104,7 @@ To do this, open your `/ios/{projectName}/AppDelegate.m` file, and add the follo
 At the top of the file, import the Firebase SDK:
 
 ```
-#import <Firebase.h>
+@import Firebase;
 ```
 
 Within your existing `didFinishLaunchingWithOptions` method, add the following to the top of the method:


Lucas-Geitner:
Yes exactly, when i Archive (with fastlane) it throw an error. It doesn't with the old way.

mikehardy:
Very interesting. With my demo script (using `@import` style) I do not attempt to archive, so I don't have experience trying it.

With my work project I use `#import <Firebase.h>` and it archives fine, via fastlane also, so my experience does not match yours - mine works this way and yours fails.

Why the difference? :thinking:

Lucas-Geitner:
Hello @mikehardy, 

Thanks for you fast answer; 
I woudn't know neither, i'm not an expert in native language;

It's a production app, unfortunatly, i cannot share the source code with you, i have recently built it with a fresh new react-native 63 app.

This is the error i got: 

`use of undeclared identifier 'FIRApp'`

Tell me if i can help in any way

Lucas-Geitner:
I found why it didn't work, 
it have nothing to do with React Native Firebase, 

I get the same issue then https://github.com/zoontek/react-native-bootsplash/issues/139

-> There is now a condition in appDelegate.m  in react-native 63. my import was after this condition. 

I think we can close this issue or improved the documentation if you think you will have people like me, 

Thanks !

```
 #import "AppDelegate.h"
#import "RNBootSplash.h" 
#import <Firebase.h> // <- add the header import should be before  ifdef FB_SONARKIT_ENABLED


#import <React/RCTBridge.h>
#import <React/RCTRootView.h>

#ifdef FB_SONARKIT_ENABLED  //  <- THE NEW CONDITION
#import <FlipperKit/FlipperClient.h>
#import <FlipperKitLayoutPlugin/FlipperKitLayoutPlugin.h>
#import <FlipperKitUserDefaultsPlugin/FKUserDefaultsPlugin.h>
#import <FlipperKitNetworkPlugin/FlipperKitNetworkPlugin.h>
#import <SKIOSNetworkPlugin/SKIOSNetworkAdapter.h>
#import <FlipperKitReactPlugin/FlipperKitReactPlugin.h>
```

mikehardy:
That makes a lot more sense, whew! I'm not great at Objective-C but I couldn't understand this one. There is a tips-and-tricks section on rnfirebase.io, you might add a new case "My app no longer compiles after upgrading to react-native 0.63, can't find firebase" (or the specific error, whatever it is) and then the explanation

Issue #4058, opened 2020-08-06
E2E Github Action is flaky


mikehardy:
This is a tracking issue for a work effort I intend to undertake personally

I will greatly appreciate any collaboration with anyone that has thoughts or help to offer, so feel free to chime in!

Observation: The E2E android test fails frequently, unknown root cause

Recent example:
```
detox[4758] INFO:  [DetoxServer.js] server listening on localhost:49713...
detox[4758] ERROR: [exec.js/EXEC_FAIL, #23] ""/Users/runner/Library/Android/sdk/platform-tools/adb" -s emulator-5554 shell "ps | grep \"com\.invertase\.testing$\""" failed with code = 1, stdout and stderr:

detox[4758] ERROR: [exec.js/EXEC_FAIL, #23] 
detox[4758] ERROR: [exec.js/EXEC_FAIL, #23] 
  1) "before all" hook in "{root}"
detox[4758] INFO:  at e2e/init.js:53:11 
  ✨ Tests Complete ✨ 
detox[4758] ERROR: at internal/process/warning.js:25:20 
 (node:4758) UnhandledPromiseRejectionWarning: DetoxRuntimeError: Failed to run application on the device
```

I have noticed that it fails sometimes locally as well, unknown cause.

I'd like to figure out exactly why.

Initial Thoughts:
- reproduce locally, fix (ideally)
- Removing the jet dependency (and possibly using cavy.app) is a stretch goal

mikehardy:
For reference a successful run currently looks like this at the start of the 'Detox Test' step

```
[Sat Aug 08 2020 15:12:55.914]  BUNDLE  ./index.js 

[Sat Aug 08 2020 15:13:11.365]  LOG      Running "testing" with {"rootTag":1}
[...]

PR #4076, opened 2020-08-12
tests: use Detox as compile reference, ignore UIManager queues

mikehardy:
### Description

Still working on de-flaking Android E2E

Even with the tap stream delayed, the extra logging from the previous merge showed the React Native UIManager's dispatch queues are sometimes permanently non-empty, blocking Detox from signalling it is ready on the app/detox-testee side.

This commit switches Detox to a compile dependency (from AAR) so we can patch-package it, then I hack out the UIManager espresso idle resource waiter completely.

This would potentially be a disaster in normal apps, but for our testing app is inconsequential since we perform no UI activities (and thus they should *never* block)

### Related issues

#4058

### Checklist

  - [ ] Yes
  - [ ] `Android`
  - [ ] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [ ] No

### Test Plan

---

### DIFF
--- a/tests/android/app/build.gradle
+++ b/tests/android/app/build.gradle
@@ -1,14 +1,7 @@
-// Add the Firebase Crashlytics plugin.
-
 apply plugin: 'com.android.application'
 apply plugin: 'com.google.gms.google-services'
 apply plugin: 'com.google.firebase.crashlytics'
-
-
-
-
 apply plugin: "com.google.firebase.firebase-perf"
-
 
 import com.android.build.OutputFile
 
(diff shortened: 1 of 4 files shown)


mikehardy:
This should be most of the systemic Android E2E flakiness, the rest will likely be per-test stuff but I think that's pretty solid right now.

iOS still flakes but not nearly as much, I'll probably leave it sit for now except for perhaps a PR to increase logging for diagnostics to already be available when/if needed

mikehardy:
I've been re-running this over and over on the PR against my own fork and this PR and have about 10 runs in now with no failures. I know of at least 2 other failure modes so it won't be 100% but already this is an improvement, definitely worth a merge for now

PR #4100, opened 2020-08-17
fix(database): bug when passing invalid keys as a reference

dackers86:
Current error is shown as a server error for Database. This adds JS validation to provide a better and more appropriate error message

Fixes #4100


Salakar (approved):

mikehardy:
Okay - going to unblock this PR and move the underlying topic of "should we allow bubble-up-and-crash or catch-and-pass-to-completion-handler" to a specific issue as that does seem like it needs to be addressed - confirmed via chat with Mike Diarmid

smaccoun:
I just encountered this issue #4177. I will say that I'm not fully aware of all the surrounding context of this PR, but on a more philosophical note I think trusting server side validation is better than trying to client side validate, especially like in this case when dealing with a 3rd part API (and in general it is my preference). The server knows best what it's allowed, and so long as we can get back an error to deal with in our client code I'd prefer to deal with it as best fits my usecase in client code. Just my 2 cents, but I'm curious what the desire is in doing client validation error? Is it just that the server errors are too cryptic to respond to?

mikehardy:
@smaccoun the server side error turns into an application crash because of a related issue #4168 - I'm quite certain that's not an expected outcome, so here we are

PR #4133, opened 2020-08-25
test(firestore): simplify firestore tests

russellwheatley:
### Description

Cleaned up rules so that only a single collection (`firestore`) & a single collection group (`collectionGroup`) is used for tests.

### Related issues

As per this [comment](https://github.com/invertase/react-native-firebase/pull/3977#issuecomment-674394585)

### Checklist

  - [X] Yes
  - [X] `Android`
  - [X] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [X] No

---


mikehardy:
This looks good in general but is failing in CI:

```


  1) firestore.collection().add()
       adds a new document:
     NativeFirebaseError: [firestore/permission-denied] The caller does not have permission to execute the specified operation.
      at FirestoreDocumentReference.set (http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false&sourceMapURL=true:188225:39)
```

russellwheatley:
There are two bugs associated with this PR that conspired to throw me off the scent. One is a Firestore emulator UI bug that I've filed [here](https://github.com/firebase/firebase-tools-ui/issues/371). 

The other is a reset of the Firestore settings at this [point](https://github.com/invertase/react-native-firebase/blob/master/packages/app/e2e/config.e2e.js#L39) which meant that, for iOS at least, we were using the remote database 😅 .

packages/app/e2e/config.e2e.js
```diff
@@ -39,6 +39,13 @@ describe('config', () => {
       await NativeModules.RNFBAppModule.preferencesClearAll();
     });
 
+    // NOTE: "preferencesClearAll" clears Firestore settings. Set DB as emulator again.
```
mikehardy, line 42: That is fiendish + subtle by the way, nice catch

tests/ios/Podfile.lock
```diff
@@ -836,25 +837,25 @@ SPEC CHECKSUMS:
...
+  RNFBMLNaturalLanguage: 34da64ba56998ff42419f3e1a59abb8db6f10813
+  RNFBMLVision: eca56297eba2b22f6b5c99669204c7bab7558be8
+  RNFBPerf: 6f9172724e67e094bcd33334ab1bda8fe81c508c
+  RNFBRemoteConfig: 4bd40833203d1d458663d12805c8a5ebdb6e2080
+  RNFBStorage: b54e785a0ddecfc941eb69756a623a8b53535c92
   Yoga: 3ebccbdd559724312790e7742142d062476b698e
 
 PODFILE CHECKSUM: 51062414d4f230ae385561d26f3c88e6c8a3ce28
 
-COCOAPODS: 1.10.0.beta.2
```
mikehardy, line 860: not terribly relevant, but cocoapods has been stuck on 1.10.0-rc.1 for 29 days now, thought for sure it would be out. Anyway, this change doesn't really matter one way or the other and is not incorrect, just a comment

mikehardy (approved):
This looks great and I suppose sets us up to more easily implement the new firestore operators they just added (there's always something to do... 😅 ) - thanks, I'll shepherd it through CI and get it merged

packages/app/e2e/config.e2e.js
mikehardy: @russellwheatley this caused me a problem that I just figured out, at least part way

- I have never been able to run the firestore e2e tests as is locally after this change
- I presume then that there is a difference between firestore in the cloud and firestore emulator that impacts the tests
- I check the errors I get and they all have to do with errors not being thrown into the sinon.spy when we expect them to be
- I check the firestore-debug.log from the emulator and I see:

```
[warn] ⚠  firestore: The emulator will default to allowing all reads and writes. Learn more about this option: https://firebase.google.com/docs/emulator-suite/install_and_configure#security_rules_configuration. {"metadata":{"emulator":{"name":"firestore"},"message":"The emulator will default to allowing all reads and writes. Learn more about this option: https://firebase.google.com/docs/emulator-suite/install_and_configure#security_rules_configuration."}}

```

I infer that the firestore emulator is not running with the same security rules somehow? I don't have insight into the firebase-in-the-cloud rules (they are not in the firestore.json that exists in the tests project) so I don't know what the difference could be.

Question: do you have access to the firestore security rules for the cloud project? Could you paste them in here?

But I have to revert this change at least temporarily in order to pass tests

russellwheatley: @mikehardy Woops, just spotted this. The firestore rules are [here](https://github.com/invertase/react-native-firebase/blob/master/.github/workflows/scripts/firestore.rules). When we run this [script](https://github.com/invertase/react-native-firebase/blob/master/.github/workflows/scripts/start-firestore-emulator.sh) in the [CI](https://github.com/invertase/react-native-firebase/blob/master/.github/workflows/tests_e2e.yml#L54), it ought to use those rules 🤔

mikehardy: Perfect thanks for the pointers! I'll see what I can do

mikehardy: @russellwheatley just to close the loop - Mike D pointed out that it might be exactly how the script was started, specifically it requires you start it in the correct working directory, if you start it outside the directory with the rules they do not apply and you see the problem I saw. I fixed up the CI / package.json scripts so it will be reliable for everyone, and you may already have been aware but just in case you weren't: careful with the working directory of the firebase emulator :-)

PR #4145, opened 2020-08-26
feat(analytics): add 'logScreenView' API and deprecate `setCurrentScreen` API.

russellwheatley:
WIP. I'll request a review once I have a bit more feedback.

### Description

- deprecated `setCurrentScreen` API. [iOS](https://firebase.google.com/docs/reference/ios/firebaseanalytics/api/reference/Classes/FIRAnalytics#+setscreenname:screenclass:) & [Android](https://firebase.google.com/docs/reference/android/com/google/firebase/analytics/FirebaseAnalytics#public-void-setcurrentscreen-activity-activity,-string-screenname,-string-screenclassoverride) 
- added `logScreenView` 

### Checklist

  - [X] Yes
  - [X] `Android`
  - [X] `iOS`
- My change includes tests;
  - [X] `jest` tests added or updated in `packages/\*\*/__tests__`
- This is a breaking change;
  - [ ] Yes
  - [X] No

### Test Plan

---


packages/analytics/lib/index.d.ts
```diff
@@ -651,26 +662,6 @@ export namespace FirebaseAnalyticsTypes {
      * @param enabled A boolean value representing whether Analytics collection is enabled or disabled. Analytics collection is enabled by default.
      */
     setAnalyticsCollectionEnabled(enabled: boolean): Promise<void>;
-
```
Salakar, line 654: Need to keep this in otherwise it's a hard deprecation for TS users. Add a `@deprecated` line on the code comments

packages/analytics/lib/index.js
```diff
@@ -102,20 +102,10 @@ class FirebaseAnalyticsModule extends FirebaseModule {
...
-    }
-
-    if (!isUndefined(screenClassOverride) && !isString(screenClassOverride)) {
-      throw new Error(
-        "firebase.analytics().setCurrentScreen(_, *) 'screenClassOverride' expected a string value.",
-      );
-    }
-
-    return this.native.setCurrentScreen(screenName, screenClassOverride);
+  setCurrentScreen() {
```
Salakar, line 105: Can this after the warning just pass along the args to `logScreenView` method in the JS class, so it still continues to function?

docs/analytics/screen-tracking.md
```diff
@@ -38,15 +40,18 @@ documentation on the React Navigation website.
...
+[`logScreenView`](/reference/analytics#logScreenView) method the Analytics library provides:
 
 ```js
 import analytics from '@react-native-firebase/analytics';
 import { Navigation } from 'react-native-navigation';
 
 Navigation.events().registerComponentDidAppearListener(({ componentName, componentType }) => {
   if (componentType === 'Component') {
-    analytics().setCurrentScreen(componentName, componentName);
+    analytics().logScreenView({
```
Salakar, line 51: Should probably make both the example functions async and await the call of logScreenView 😅

packages/analytics/lib/index.d.ts
```diff
@@ -324,6 +324,17 @@ export namespace FirebaseAnalyticsTypes {
     transaction_id?: string;
   }
 
+  export interface ScreenViewParameters {
+    /**
+     * Screen name the user is currently viewing.
```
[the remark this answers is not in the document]
russellwheatley: @mikehardy Good catch. Question: Do you think we should change it to "Set to `null` to clear the current screen name" to keep it inline with iOS & Android? Web [docs](https://firebase.google.com/docs/reference/js/firebase.analytics.Analytics#eventname:-screen_view) don't reveal if setting to `null` clears the current screen unfortunately.

packages/analytics/lib/index.d.ts
```diff
@@ -324,6 +324,17 @@ export namespace FirebaseAnalyticsTypes {
...
+     */
+    screen_name?: string;
+    /**
+     * Current class associated with the view the user is currently viewing.
```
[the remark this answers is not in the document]
russellwheatley: Same question as above.

packages/analytics/lib/index.d.ts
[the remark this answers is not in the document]
mikehardy: Good point - better to be explicit and do null vs undefined (a, javascript...) to match exactly. I rooted around through JS SDK and it was inconclusive as well - I think that might just be because the function call with a null to clear in the web context simply makes no sense, it doesn't map to the paradigm. On native you might want to clear it because they auto-log screen events if you switch activities or view etc but (on react-native) you never switch thus the manual calls are needed, whereas on web you always have to call it or it never gets called? Not sure what the summary is there other than I like the idea of specific null to clear, matching both natives, and if web doesn't do exactly the same it's probably fine because logging a null screen is silly on web to begin with?

packages/analytics/lib/index.d.ts
```diff
@@ -662,6 +662,26 @@ export namespace FirebaseAnalyticsTypes {
...
+     * ```
+     *
+     * > Whilst screenClassOverride is optional, it is recommended it is
+     * always sent as your current class name. For example on Android it will always
+     * show as 'MainActivity' if you do not specify it.
+     *
+     * @param screenName A screen name, e.g. Product.
+     * @param screenClassOverride On Android, React Native runs in a single activity called
+     * @deprecated
+     * 'MainActivity'. Setting this parameter overrides the default name shown on logs.
```
mikehardy:
```suggestion
     * 'MainActivity'. Setting this parameter overrides the default name shown on logs.
     * @deprecated
```

deprecated and remainder of original comment were inverted

packages/analytics/lib/index.d.ts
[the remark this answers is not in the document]
russellwheatley: @mikehardy Makes sense to me 😄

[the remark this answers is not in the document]
russellwheatley: I think it's best to leave as it is then? 🤔

docs/analytics/screen-tracking.md
```diff
@@ -38,15 +40,18 @@ documentation on the React Navigation website.
...
 import { Navigation } from 'react-native-navigation';
 
-Navigation.events().registerComponentDidAppearListener(({ componentName, componentType }) => {
+Navigation.events().registerComponentDidAppearListener(async ({ componentName, componentType }) => {
   if (componentType === 'Component') {
-    analytics().setCurrentScreen(componentName, componentName);
+    await analytics().logScreenView({
+        screen_name: componentName,
+        screen_class: componentName,
+      });
```
Salakar, line 54:
```suggestion
    await analytics().logScreenView({
      screen_name: componentName,
      screen_class: componentName,
    });
```

packages/analytics/lib/index.js
```diff
@@ -103,19 +103,13 @@ class FirebaseAnalyticsModule extends FirebaseModule {
...
-    }
-
-    if (!isUndefined(screenClassOverride) && !isString(screenClassOverride)) {
-      throw new Error(
-        "firebase.analytics().setCurrentScreen(_, *) 'screenClassOverride' expected a string value.",
-      );
-    }
-
-    return this.native.setCurrentScreen(screenName, screenClassOverride);
+    console.warn(
```
Salakar, line 106: Could do with suppressing eslint for this line

packages/analytics/e2e/analytics.e2e.js
```diff
@@ -56,16 +56,6 @@ describe('analytics()', () => {
     });
   });
 
-  describe('setCurrentScreen()', () => {
```
Salakar, line 59: Can we keep this test in, just to make sure the old method is called during tests and nothing blows up

packages/analytics/lib/index.d.ts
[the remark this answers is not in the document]
mikehardy: That's a tough one. I think that's actually a documentation fail on google's part as I bet the underlying behavior is the same as previously documented, but better to conform to their documentation - leave as is I think

packages/analytics/lib/index.d.ts
```diff
@@ -667,10 +678,10 @@ export namespace FirebaseAnalyticsTypes {
      *
      * @param screenName A screen name, e.g. Product.
      * @param screenClassOverride On Android, React Native runs in a single activity called
+     * @deprecated
      * 'MainActivity'. Setting this parameter overrides the default name shown on logs.
```
mikehardy, line 682: This one is still out of order - the lines just need to be flipped so `@deprecated` doesn't break up the multi-line param from above

```suggestion
     *        'MainActivity'. Setting this parameter overrides the default name shown on logs.
     * @deprecated
```

mikehardy:
Still WIP or good-to-go @russellwheatley :-) ? I could give it a final review pass and if it looks good maybe get it out...

russellwheatley:
It should pass I think.  Android tests flaked out last time so fingers crossed this time 🤞

mikehardy:
Flaked again, on one I've seen - tracking flake cause here and rerunning https://github.com/invertase/react-native-firebase/issues/4058#issuecomment-674347815

MohamadKh75:
Hey
Thanks for your efforts!
I just see the new docs and for React Navigation, we should use just `class_name` but we use both of `screen_name` and `screen_class` for React Native Navigation!
That is the correct behavior or I should create a PR for that missing line :grin:?

cc @mikehardy

mikehardy:
@MohamadKh75 I'm not sure screen class is that useful. But it is a valid parameter. How did it work when you tried it? (what did it look like in analytics, in other words)

MohamadKh75:
Well, I just upgraded to the newest version and use both `screen_name` and `screen_class`! I should wait some time to get the analytics report.

mikehardy:
I think it will just work. I don't believe there will be "novel" information. It should be the same screen tracking events, just from a different API call. Honestly, it is supposed to be an exceptionally boring change, just forward-porting to new underlying API calls.

MohamadKh75:
So, should I create a PR to add that to React Navigation?

mikehardy:
That is up to you? If React Navigation has any sort of documentation that cross-links with react-native-firebase here, and it needs an update, then sure. It's a different / unrelated repo so nothing I say is really authoritative, right?

MohamadKh75:
I mean in [here](https://rnfirebase.io/analytics/screen-tracking#react-navigation) :grin:

mikehardy:
@MohamadKh75 Sure! DIdn't realize it was actually our example :-)

PR #4203, opened 2020-09-03
fix: [Firebase message][Android] issue open old notify

### LINKED ISSUE #4052: Android: getInitialNotification/onNotificationOpenedApp not triggered for old notifications if app restarts
## Issue
Before i was using firebase 5+ and everything was working just fine. But few months ago i updated firebase to 7.0.1 (it was latest version) and over time users have started to report issues that they can't open specific page from push notification. After investigation i found steps to reproduce this bug, here it is:

1. Completely close the app;
2. Send more than 1 notification (e.g 3, can be the same data or different - does not matter);
3. Open app from notification (everything still fine, listeners are working);
4. Again completely close the app (we still have 2 more notifications);
5. Open app from any of left notifications (App should open) and here is a bug happen: any of listeners wont trigger even if you try to open them from background/foreground or quit state.

It looks like they become expired or something, since you already opened app from another notification. Can that be possible ?

I'll update issue with code snippets if necessary, but seems like the problem is not in the code.

Versions:

"react-native": "0.61.2",
"@react-native-firebase/messaging": "^7.0.1",
"@react-native-firebase/app": "^7.0.1",

vannt1991:
### Related issues

Fixes #4052

### Checklist

  - [x] Yes
  - [x] `Android`
  - [ ] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

---

### DIFF
--- a/packages/messaging/android/src/main/java/io/invertase/firebase/messaging/ReactNativeFirebaseMessagingModule.java
+++ b/packages/messaging/android/src/main/java/io/invertase/firebase/messaging/ReactNativeFirebaseMessagingModule.java
@@ -69,7 +69,9 @@ public class ReactNativeFirebaseMessagingModule extends ReactNativeFirebaseModul
           // only handle non-consumed initial notifications
           if (messageId != null && initialNotificationMap.get(messageId) == null) {
             RemoteMessage remoteMessage = ReactNativeFirebaseMessagingReceiver.notifications.get(messageId);
-
+            if (remoteMessage == null) {
+              remoteMessage = ReactNativeFirebaseMessagingStoreHelper.getInstance().getMessagingStore().getFirebaseMessage(messageId);
+            }
             if (remoteMessage != null) {
               promise.resolve(ReactNativeFirebaseMessagingSerializer.remoteMessageToWritableMap(remoteMessage));
               initialNotificationMap.put(messageId, true);
--- a/packages/messaging/android/src/main/java/io/invertase/firebase/messaging/ReactNativeFirebaseMessagingModule.java
+++ b/packages/messaging/android/src/main/java/io/invertase/firebase/messaging/ReactNativeFirebaseMessagingModule.java
@@ -70,7 +70,9 @@ public class ReactNativeFirebaseMessagingModule extends ReactNativeFirebaseModul
           if (messageId != null && initialNotificationMap.get(messageId) == null) {
             RemoteMessage remoteMessage = ReactNativeFirebaseMessagingReceiver.notifications.get(messageId);
             if (remoteMessage == null) {
-              remoteMessage = ReactNativeFirebaseMessagingStoreHelper.getInstance().getMessagingStore().getFirebaseMessage(messageId);
+              ReactNativeFirebaseMessagingStore messagingStore = ReactNativeFirebaseMessagingStoreHelper.getInstance().getMessagingStore();
+              remoteMessage = messagingStore.getFirebaseMessage(messageId);
+              messagingStore.clearFirebaseMessage(messageId);
             }
             if (remoteMessage != null) {
               promise.resolve(ReactNativeFirebaseMessagingSerializer.remoteMessageToWritableMap(remoteMessage));
--- a/packages/messaging/android/src/main/java/io/invertase/firebase/messaging/ReactNativeFirebaseMessagingStoreImpl.java
+++ b/packages/messaging/android/src/main/java/io/invertase/firebase/messaging/ReactNativeFirebaseMessagingStoreImpl.java
@@ -1,0 +1,38 @@
+package io.invertase.firebase.messaging;
+
+import com.google.firebase.messaging.RemoteMessage;
+
+import org.json.JSONException;
+import org.json.JSONObject;
+
+import io.invertase.firebase.common.UniversalFirebasePreferences;
+
+import static io.invertase.firebase.messaging.JsonConvert.jsonToReact;
+import static io.invertase.firebase.messaging.JsonConvert.reactToJSON;
+import static io.invertase.firebase.messaging.ReactNativeFirebaseMessagingSerializer.remoteMessageFromReadableMap;
+import static io.invertase.firebase.messaging.ReactNativeFirebaseMessagingSerializer.remoteMessageToWritableMap;
+
+public class ReactNativeFirebaseMessagingStoreImpl implements ReactNativeFirebaseMessagingStore {
+
+  @Override
+  public void storeFirebaseMessage(RemoteMessage remoteMessage) {
+    try {
+      UniversalFirebasePreferences.getSharedInstance().setStringValue(remoteMessage.getMessageId(), reactToJSON(remoteMessageToWritableMap(remoteMessage)).toString());
+    } catch (JSONException e) {
+      e.printStackTrace();
+    }
+  }
+
+  @Override
+  public RemoteMessage getFirebaseMessage(String remoteMessageId) {
+    String remoteMessageString = UniversalFirebasePreferences.getSharedInstance().getStringValue(remoteMessageId, null);
+    if (remoteMessageString != null) {
+      try {
+        return remoteMessageFromReadableMap(jsonToReact(new JSONObject(remoteMessageString)));
+      } catch (JSONException e) {
+        e.printStackTrace();
+      }
+    }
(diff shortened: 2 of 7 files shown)


pavel-corsaghin (dismissed):

mikehardy:
Awesome! I will review as soon as I can

We must have the CLA signed before we can merge it

Anyone can use these auto generated patches from the PR for testing - I would love independent confirmation it works for someone else? https://github.com/invertase/react-native-firebase/suites/1137916996/artifacts/16295093

packages/messaging/android/src/main/java/io/invertase/firebase/messaging/ReactNativeFirebaseMessagingModule.java
```diff
@@ -69,7 +69,11 @@ public void getInitialNotification(Promise promise) {
           // only handle non-consumed initial notifications
           if (messageId != null && initialNotificationMap.get(messageId) == null) {
             RemoteMessage remoteMessage = ReactNativeFirebaseMessagingReceiver.notifications.get(messageId);
-
+            if (remoteMessage == null) {
+              ReactNativeFirebaseMessagingStore messagingStore = ReactNativeFirebaseMessagingStoreHelper.getInstance().getMessagingStore();
+              remoteMessage = messagingStore.getFirebaseMessage(messageId);
+              messagingStore.clearFirebaseMessage(messageId);
```
mikehardy, line 75: I'm worried about this. The messagingStore you've created could grow unbounded if messages aren't deterministically cleared. Here they will only be cleared for notifications that are tapped. Notifications that are dismissed or not tapped will persist, leading to unbounded growth.

They either need to be deleted in some sort of FIFO manner (maybe any notifications more than 10 deep are just assumed dead and are then cleared...) or by implementing the NotificationListener and deleting all notifications from messagingStore that aren't active (https://developer.android.com/reference/android/service/notification/NotificationListenerService#getActiveNotifications() - works for API18 and up at least)

Doing a FIFO queue is probably the easiest. On insert, if there are more than 9 notifications trim the oldest ones until there are only 9, then insert ?

packages/messaging/android/src/main/java/io/invertase/firebase/messaging/ReactNativeFirebaseMessagingStoreImpl.java
```diff
@@ -0,0 +1,52 @@
...
+import static io.invertase.firebase.messaging.ReactNativeFirebaseMessagingSerializer.remoteMessageToWritableMap;
+
+public class ReactNativeFirebaseMessagingStoreImpl implements ReactNativeFirebaseMessagingStore {
+
+  @Override
+  public void storeFirebaseMessage(RemoteMessage remoteMessage) {
+    try {
+      String remoteMessageString = reactToJSON(remoteMessageToWritableMap(remoteMessage)).toString();
+      Log.d("storeFirebaseMessage", remoteMessageString);
+      UniversalFirebasePreferences.getSharedInstance().setStringValue(remoteMessage.getMessageId(), remoteMessageString);
```
mikehardy, line 25: @Salakar - what do you think of this as a design? The goal is to persist notification state across app restarts in case the notification is tapped at some future date, so getInitialNotification works in all cases. So this turns the notification into JSON and stuffs it into preferences for later retrieval.

packages/messaging/android/src/main/java/io/invertase/firebase/messaging/ReactNativeFirebaseMessagingModule.java
mikehardy: @vannt1991 I will need your thoughts on this one. A separate question is out to the creator of the package to see their thoughts on where this is stored

vannt1991: @mikehardy yeahh, it is very good improvement, i think we can using NotificationListener to deleting all notifications from messagingStore that aren't active. that is good point 👍 
for FIFO queue is easiest but it will not cover all case, maybe we will lost important notification when user tapped at notification bar. Let me look deep into using NotificationListener. Thanks

mikehardy: That's the most correct way to do it I think - the only reason I hesitated was that it requires a new Manifest permission on Android and I'm not sure if a firebase app with messaging already has it normally or not (opening an .apk in Android Studio opens "APK Analyzer" and it will show the combined manifest I think, I just haven't looked). If it does not already include I'm not sure if it implies we need an extra permission or something, or if it ends up not mattering. Only research and doing a proof of concept will tell I think. I am definitely curious to see the results!

vannt1991: @mikehardy yep, i known that, we need notification access permission, it is very does not convenient for user. And for now, we can using FIFO queue to fast implement. Any suggestion for using FIFO to store all notification message? @mikehardy

mikehardy: I don't have any great answers. 

I have two thoughts - one is that you use one preference key to store a map of message ids in storage (or maybe message ids and a timestamp set in as value, separate by a pipe or something), so you can figure out what messages exist. Another is to add a "keySet" method to the preferences object so you could iterate over all entries, and prefix all messages with something that lets you find them, and put the message timestamp in the message (if it isn't there already). Both are not great as they involve hitting data storage but, but either would work.

Either way once you have the list of current messages in storage and their timestamps, order by timestamp and remove the oldest prior to insert of the newest.

packages/messaging/android/src/main/java/io/invertase/firebase/messaging/ReactNativeFirebaseMessagingStoreImpl.java
```diff
@@ -46,7 +56,39 @@ public RemoteMessage getFirebaseMessage(String remoteMessageId) {
...
-    UniversalFirebasePreferences.getSharedInstance().remove(remoteMessageId);
+    UniversalFirebasePreferences preferences = UniversalFirebasePreferences.getSharedInstance();
+    preferences.remove(remoteMessageId);
+    // check and remove old notifications message
+    String notifications = preferences.getStringValue(S_KEY_ALL_NOTIFICATION_IDS, "");
+    if (!notifications.isEmpty()) {
+      notifications = notifications.replace(remoteMessageId + ",", ""); // remove from list
+    }
+    List<String> allNotificationList = convertToArray(notifications);
+    if (allNotificationList.size() > MAX_SIZE_NOTIFICATIONS) {
```
mikehardy, line 67: Ah - so a "ring buffer", limited to 100 now (seems like a lot! but it is bounded at least, that is the most important part) - the second part of the logic though, where you check if there are too many and we need to remove one, should be done on storeFirebaseMessage above shouldn't it?

This approach seems like it can work though...

vannt1991: @mikehardy as you known, we can receive notification at many times, but we don't known what time notification item will be removed from status bar. then i will store all messages id we received from fcm and check and removed old notification when user trigger tap any item from status bar.

mikehardy: But this is the exact overflow case... It is brutal, but we must not overflow, so the limit needs to apply on message ingress I think

100 visible notifications should be more than sufficient for a well behaved app? And the API should just carry a note that you will lose notification attribution if you have more than that, so the app handles it.

If you just remove on delete, it's possible for an ill-behaved app to overflow in the current style

mikehardy:
@vannt1991 this looks much closer! I need to review it really closely now with the idea I can merge it, however we still need the CLA signed or I can't merge no matter what :pray:

vannt1991:
@mikehardy 
i have changed to this request, pls help me approve it, many thanks 👍 
[https://github.com/invertase/react-native-firebase/pull/4317](https://github.com/invertase/react-native-firebase/pull/4317)

Issue #4229, opened 2020-09-08
[🐛] Crashlytics 8.4.2 not collecting on iOS with pods 6.31?


abdullahizzuddiin:
**[EDITED by @mikehardy 20200913]: it appears the pods 6.28 may work fine with our `@react-native-firebase/crashlytics@^8.4.2`, but that pods 6.31 does not. This is an area of active investigation. But you may try overriding pods to 6.28 for now. https://rnfirebase.io/#ios**

Original issue follows:

----

## Issue

I just upgraded `@react-native-firebase/crashlytics` from `6.4.0` to `8.4.2`. Crash report from iOS didn't displayed on Firebase Console Dashboard.

**What I've Done**  
- Read https://github.com/invertase/react-native-firebase/issues/3879
- removed all fabric related code/config from `xschme`, `project.pbxproj`,  `podfile`, `podfile.lock`
- added `firebase.json`
- ran `crashlytics().setCrashlyticsCollectionEnabled(true);` in app code
- triggered crash using `crashlytics().crash()`
- Built on Debug and Release mode. 
- Last but not least, wait 8 hours

After all things i've done, nothing resulted to displaying error on my Firebase Console.

This log displayed immediately after app builded and ran on my phone.

```
[Firebase/Crashlytics] Version 4.4.0
6.30.0 - [Firebase/Crashlytics][I-CLS000000] [Crashlytics] Running on iPhone9,3, 13.7.0 (17H35)
6.30.0 - [Firebase/Crashlytics][I-CLS000000] Root: /var/mobile/Containers/Data/Application/7529E6A5-A2B7-42E4-929B-AF26A1E622AA/Library/Caches/com.crashlytics.data/id.flip.debug
6.30.0 - [Firebase/Crashlytics][I-CLS000000] [Crashlytics:Crash:Reports:Event] Registered Firebase Analytics event listener
6.30.0 - [Firebase/Crashlytics][I-CLS000000] Automatic data collection is disabled.
6.30.0 - [Firebase/Crashlytics][I-CLS000000] [Crashlytics:Crash] There are no unsent reports.
```

Sometimes, `1 unsent reports are available. Checking for upload permission.` displayed on console. But, nothing happened after I trigger `crashlytics().sendUnsentReports();`

```
[Firebase/Crashlytics] Version 4.4.0
6.30.0 - [Firebase/Crashlytics][I-CLS000000] [Crashlytics] Running on iPhone9,3, 13.7.0 (17H35)
[...]

PR #4275, opened 2020-09-17
fix(ios, podspec): depend on React-Core instead of React

mrousavy:
### Description

On Xcode 12 some wild build errors occur when trying to build a React Native app. This is because the `'React'` pod is only an umbrella pod. See: https://github.com/facebook/react-native/issues/29633

### Related issues

See: https://github.com/facebook/react-native/issues/29633

### Release Summary

* Uses 'React-Core' instead of 'React' pod in podspecs

### Checklist

  - [x] Yes
  - [ ] `Android`
  - [x] `iOS`
- My change includes tests;
- This is a breaking change;
  - [x] Yes for React < 0.60 (because of autolinking)
  - [ ] No

### DIFF
--- a/packages/admob/RNFBAdMob.podspec
+++ b/packages/admob/RNFBAdMob.podspec
@@ -26,7 +26,7 @@ Pod::Spec.new do |s|
   s.source_files        = 'ios/**/*.{h,m}'
 
   # React Native dependencies
-  s.dependency          'React'
+  s.dependency          'React-Core'
   s.dependency          'RNFBApp'
 
   # Other dependencies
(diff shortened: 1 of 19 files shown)


mikehardy (approved):
Thanks for this and I appreciate the collaboration over on the related issue since I'm having trouble reproducing it myself.

mikehardy:
For anyone else that needs this but is having problem upgrading to current because of for instance #4229 - you may depend on the patches we generate for every PR https://github.com/invertase/react-native-firebase/actions/runs/259924803 - applied via patch-package they can be used in place of updating, as a temporary measure while root causes are resolved

PR #4312, opened 2020-09-26
docs(types, dynamic-links): ShortLinkType is accessible via statics

mikehardy:
Using the previously documented example style results in a symbol not found

---


Salakar (approved):

mikehardy:
@Salakar worth noting that without this change in my work project (using the 'SHORT' enum symbol vs the default which is I suppose 'UNGUESSABLE') I was getting a different link every link build, which of course means you can't pull analytics on the links because they're always different. Now I have tiny URLs and stable stats for my non-privacy related shares. What a battle that was! And don't even ask about pulling the stats from the REST API, the default admin token for firebase functions doesn't contain the correct scope :disappointed:

PR #4324, opened 2020-09-29
fix(messaging, ios): ensure completion handler run in main thread v5.x.x

troublediehard:
fix https://github.com/invertase/react-native-firebase/issues/3944


mikehardy:
As mentioned on the related issue we aren't merging v5 changes anymore but for anyone else that sees this directly (vs from the related issue) this has confirmation it appears to work. If for whatever reason you are stuck on v5 for the moment I encourage the use of patch-package to integrate this change, patch-package works fantastically in my experience: https://github.com/ds300/patch-package

rukmanary:
> As mentioned on the related issue we aren't merging v5 changes anymore but for anyone else that sees this directly (vs from the related issue) this has confirmation it appears to work. If for whatever reason you are stuck on v5 for the moment I encourage the use of patch-package to integrate this change, patch-package works fantastically in my experience: https://github.com/ds300/patch-package

hey can you show me step by step to do that? Like this one: https://github.com/facebook/react-native/issues/29279#issuecomment-657284028 Because I don't really understand how it works
Thank you

troublediehard:
@rukmanary here the video example of how to apply patch package https://youtu.be/YEb4jNv1rfE , probably could help you.

Issue #4343, opened 2020-10-01
🔥 [🐛] NPM Publish actions are interrupted/incomplete


davidgovea:
## Issue

Re-opening of issue #4283 
#4330 attempted to fix, but did not.

Hey @mikehardy -- I discovered while working on #4331 that npm does not have the latest version of each package. It looks like the publish action still has some issues under certain concurrency scenarios.

* `npm info @react-native-firebase/database` => `latest: 7.5.10`
* In-repo version: `7.5.11`: https://github.com/invertase/react-native-firebase/blob/master/packages/database/CHANGELOG.md

---

Interestingly, the `app`, `analytics`, and `admob` are up-to-date on npm. They are alphabetically first, so I'm thinking we have an interrupted-publish issue.

Looking at the [Publish actions](https://github.com/invertase/react-native-firebase/actions?query=workflow%3APublish), here is my understanding of what happened:

1. `Publish` was manually triggered: https://github.com/invertase/react-native-firebase/actions/runs/281108519
   * Since the `app` module was updated, the peerDependency of each package was updated.
   * The new commits were pushed to master: https://github.com/invertase/react-native-firebase/runs/1189876219?check_suite_focus=true#step:6:104
1. The new master push triggers **another** `Publish` run: https://github.com/invertase/react-native-firebase/runs/1189883874?check_suite_focus=true
   * This job actually publishes the latest versions of `analytics` and `admob` before [being interrupted by a "You cannot publish over the previously published versions" `app` publish error](https://github.com/invertase/react-native-firebase/runs/1189883874?check_suite_focus=true#step:6:427)
1. The version-bumps are pushed to master, but the **action is skipped** https://github.com/invertase/react-native-firebase/runs/1189891921?check_suite_focus=true

---

### Javascript
N/A

### iOS
N/A

### Android
N/A

---

## Environment
N/A

---

- 👉 Check out [`React Native Firebase`](https://twitter.com/rnfirebase) and [`Invertase`](https://twitter.com/invertaseio) on Twitter for updates on the library.

mikehardy:
Yes - actually I was going to take this up with @Salakar - I'm not sure what's going on here and I don't have the push-to-master authority to conclusively un-fuddle it unfortunately. One hypothesis is the previous push breaks put things in a state the scripts can't self-correct out of. Another hypothesis is that it's something unknown (sounds silly but that hypothesis is always on my list of things to expect, keeps me honest). 

[...]

Issue #4283, opened 2020-09-19
[🐛] NPM package '@react-native-firebase/messaging' depends on '@react-native-firebase/app' v8.4.2 but found v8.4.3


lorenc-tomasz:
NPM package '@react-native-firebase/messaging' depends on '@react-native-firebase/app' v8.4.2 but found v8.4.3

mikehardy:
Yes! Sorry about that. Something isn't quite right with the publish from #4275 - I'm not sure what happened.

Please know that it shouldn't be a problem *this time* but yes it should be fixed, that's on me, I'll sort it out when I get a moment but it will likely be a day or three

lorenc-tomasz:
@mikehardy Thank you for answer :)) And sorry for that short description.

mikehardy:
haha it was nevertheless complete. Still hoping to get this early next week, launching other stuff today

sangameshrently:
@mikehardy : is that just warning ? for me ios app is getting crashed on launch.

mikehardy:
@sangameshrently This is just a warning. Your crash will be unrelated.

cristiankmb:
I have the same problem. When i run **npx react-native run-android** the app run without problems en real device, but when i run **npx react-native run-android --variant=release** to build an android release return build filed with

**Task :app:packageRelease FAILED**
**ReactNativeFirebase WARNING: NPM package '@react-native-firebase/storage' depends on '@react-native-firebase/app' v8.4.2 but found v8.4.3, this might cause build issues or runtime crashes.**

ArrudaWillyan:
> I have the same problem. When i execute **npx react-native run-android** the app run without problems en real device, but when i execute **npx react-native run-android --variant=release** to build an android release return build filed with
> 
> **Task :app:packageRelease FAILED**
> **ReactNativeFirebase WARNING: NPM package '@react-native-firebase/storage' depends on '@react-native-firebase/app' v8.4.2 but found v8.4.3, this might cause build issues or runtime crashes.**
> 
> @mikehardy Can this apply to the same issue or should be another section?

I have the same problem, does anyone have any solutions?

mikehardy:
THIS IS A WARNING. It is not the source of your problem.
If app:packageRelease FAILED you are looking for a FAILURE, an ERROR, not a WARNING

[...]

PR #4344, opened 2020-10-01
chore(ci): when generating patches, keep package.json from latest npm version

### LINKED ISSUE #4331: CI patch-package task failing to generate full set of patches
@mikehardy It looks like the automatic patch creation is having issues: patches were only generated for `app`, `analytics`, and `admob` modules.

The others have errors like: `Error: Can't find lockfile entry for @react-native-firebase/ml-vision`

_Originally posted by @davidgovea in https://github.com/invertase/react-native-firebase/pull/4306#issuecomment-698998034_

davidgovea:
### Related issues

Fixes #4331

### Description

:wave: Hi @mikehardy!
The patch-package failure was being caused by incomplete npm publish tasks. (repo versions higher than `npm` versions) (Related: #4283)
Looks like #4330 did not fix the issue

I've detailed that in a new issue #4343, and I suppose if that is fixed, then #4331 becomes a non-issue, but still, I've taken a self-contained approach based on the following assumptions:

1. The generated patches should **always** be for the latest npm version, **regardless of repo state**.
2. patch-package does not work well with `package.json`, so we can safely ignore package.json changes in the patches.

In the patch generation task, the `package.json` from the "live" version is now preserved.

### Release Summary

N/A

### Checklist

  - [x] Yes
  - `Android` N/A
  - `iOS` N/A
  - [x] CI onlu
- My change includes tests;
  - N/A
- [ ] I have updated TypeScript types that are affected by my change. N/A
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

I made a thrash-branch in my fork, adding bogus changes to each package to test the package generation.
It's very messy, but here are 2 Action runs:

- Bad (observe lock issues, only several generated patches): https://github.com/davidgovea/react-native-firebase/pull/1/checks?check_run_id=1195395418#step:7:1291
- Good (all patches generated once the package.json-preservation in place: https://github.com/davidgovea/react-native-firebase/pull/1/checks?check_run_id=1195543383#step:7:1174

### DIFF
--- a/.github/workflows/create_test_patches.yml
+++ b/.github/workflows/create_test_patches.yml
@@ -73,8 +73,9 @@ jobs:
             echo "Installing package $PACKAGE into fresh template app, then clobbering with PR version"
             yarn add @react-native-firebase/$PACKAGE
             pushd node_modules/@react-native-firebase
-            \rm -fr $PACKAGE
             tar -zxf $HOME/packages/react-native-firebase-${PACKAGE}-v*
+            mv $PACKAGE/package.json package/
+            \rm -fr $PACKAGE
             mv package $PACKAGE
             popd
             npx patch-package @react-native-firebase/$PACKAGE || true


mikehardy:
I was literally just going to sit down and look at this with the idea the publish issue broke it, and here it appears you've done it ahead of me, you have my sincere gratitude as that leaves a lot more time for other things, like I think I'll have enough time to finish the latest stable release of AnkiDroid, so you're indirectly helping 1.7M people become doctors or fluent in a second language :). Most excellent

By the way - my hope for this approach was to package it up as an action to make it generally available for react-native projects so any PR got a patch-package. If you have any interest in becoming "Patch-Package Action Person" you have my blessing and best wishes, otherwise I'll keep it on my backlog. Just needs generalization to handle monorepos and non-monorepos

mikehardy (approved):
Nice surgical fix, I like the underlying reasoning as well. Thanks

mikehardy:
the only downside to this one - a rare thing, but... - if there is a necessary change to package.json, like a new devDependency or something - no patch is generated

davidgovea:
@mikehardy I've never been able to apply a patch to `package.json` using patch-package successfully. I noted this in assumption 2.

There are workarounds to `--exclude <nothing>` in the patch creation, to side-step `patch-package`'s default exclusion of `package.json`.. some related issues:
https://github.com/ds300/patch-package/issues/250
https://github.com/ds300/patch-package/issues/179

`npm` mangles the `package.json`s of installed dependencies. Perhaps `yarn` does not? That could explain the inconsistent issue reports. If that were the case, then npm-users still could not use the generated patches.

mikehardy:
Oh interesting - thanks for the extra info there. I'd say "no change" here then, and we'll just treat it as a known thing. I don't imagine most user-generated PRs will change those, and my biggest goal with the patch generation was that users could post a PR and with no effort from maintainers other users could easily test - thus reflecting the testing need back to the community but in an empowered way instead of just begging. The current capability serves that fine

Issue #4220, opened 2020-09-06
Documentation website not linking correctly


mikehardy:
I have noticed a linking issue in the website rnfirebase.io

For some reason the navigation bar becomes incorrectly rooted sometimes and you have to force it all the way back to rnfirebase.io then you can navigate again. I'll look into that hopefully early next week if @andersonaddo (who fixed the last website issue, actually :-) ) doesn't beat me to it

_Originally posted by @mikehardy in https://github.com/invertase/react-native-firebase/issues/4219#issuecomment-687900205_

mikehardy:
closed in #4365 we think

PR #4365, opened 2020-10-07
fix(docs): prevent trailing slash removal from links to root

### LINKED ISSUE #4357: [Docs] "Getting Started" sidebar link broken
It appears that the "Getting Started" link on the documentation page has an empty `href`, and thus it will always go to the page the user is on.
image

DoctorJohn:
This PR fixes #4357 and #4220

Before links pointing to "/" were converted to links pointing to "" which rendered them useless. This was caused by the custom link component which is supposed to remove trailing slashes from internal links. This PR adds a check for the "/" edge case.

### DIFF
--- a/website/src/components/Link.tsx
+++ b/website/src/components/Link.tsx
@@ -69,7 +69,7 @@ const Link = ({
       to,
     };
 
-    if (gatsbyLinkProps.to.endsWith('/')) {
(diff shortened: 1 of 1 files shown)


mikehardy (approved):
I really appreciate this @DoctorJohn - it was very strange behavior for folks that are usually our newest users - small but high impact fix

PR #4391, opened 2020-10-14
docs(setup): Add instructions for using Firebase in an emulator

apfritts:
### Description

While troubleshooting `InternalFirebaseAuth.FIREBASE_AUTH_API is not available on this device`, it seems you need Google Play services installed. As a React Native newbie, I don't recall seeing this included in any instructions and I think this might be the best place.

### Related issues

None in this repository, but there were a lot of random results when I [searched](https://www.google.com/search?q=firebase_auth_api+is+not+available+on+this+device) for the error.

### Release Summary

Documentation change

### Checklist

  - [x] Yes
  - [x] `Android`
  - [ ] `iOS`
- My change includes tests;
  - Not applicable to documentation change
  - ~~[ ] `e2e` tests added or updated in `packages/\*\*/e2e`~~
  - ~~[ ] `jest` -tests added or updated in `packages/\*\*/__tests__`~~
- ~~[ ] I have updated TypeScript types that are affected by my change.~~
  - Not applicable to documentation change
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

Screen Shot 2020-10-14 at 09 02 36

:fire:

### DIFF
--- a/docs/faqs-and-tips.md
+++ b/docs/faqs-and-tips.md
@@ -39,7 +39,7 @@ This happens to be a known problem with the upstream Analytics SDKs. The Firebas
 
 ### I'm receiving `InternalFirebaseAuth.FIREBASE_AUTH_API is not available on this device`. How do I fix this?
 
-To use some Firebase services (like auth) in an emulator, you need an Android devices with Google Play and associated services installed. Check this [StackOverflow post](https://stackoverflow.com/a/46246782/2275865) for instructions on creating a new Android device with the necessary APIs installed.
+To use some Firebase services (like auth) in an emulator, you need an Android virtual device with Google Play services installed. Check this [StackOverflow post](https://stackoverflow.com/a/46246782/2275865) for instructions on creating a new Android virtual device with the necessary APIs installed.
 
 Alternatively, you can install the Google Play Services SDK. In Android Studio:
 
--- a/docs/faqs-and-tips.md
+++ b/docs/faqs-and-tips.md
@@ -39,7 +39,7 @@ This happens to be a known problem with the upstream Analytics SDKs. The Firebas
 
 ### I'm receiving `InternalFirebaseAuth.FIREBASE_AUTH_API is not available on this device`. How do I fix this?
 
-To use some Firebase services (like auth) in an emulator, you need an Android virtual device with Google Play services installed. Check this [StackOverflow post](https://stackoverflow.com/a/46246782/2275865) for instructions on creating a new Android virtual device with the necessary APIs installed.
+To use some Firebase services (like auth) in an emulator, you need an Android virtual device with Google Play services installed. Check this [Stack Overflow post](https://stackoverflow.com/a/46246782/2275865) for instructions on creating a new Android virtual device with the necessary APIs installed.
 
 Alternatively, you can install the Google Play Services SDK. In Android Studio:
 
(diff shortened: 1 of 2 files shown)


docs/index.md
```diff
@@ -74,6 +74,18 @@ apply plugin: 'com.android.application'
 apply plugin: 'com.google.gms.google-services' // <- Add this line
 ```
 
+#### Using Firebase services in an emulator
+
+To use Firebase services in an emulator, you need to install the Google Play services SDK. From Android Studio:
```
mikehardy, line 79: This is only partially true, it's a lot more subtle than this

In order to use *parts* of the underlying firebase-android-sdk, like auth, a device (or emulator) must have Play Services installed. Other parts that need it I think are things that rely on instance ids like messaging, and remote config. Even with those there is some movement to make those possible without play services since devices like Huawei won't have them going forward - in other words it's actually an area of active development.

Here's auth without play services for instance https://github.com/FirebaseExtended/auth-without-play-services

There are many parts that do not require play services though. Analytics and Firestore I think don't need it for instance.

Additionally, I don't think the SDK has anything to do with it, so I believe the following instructions to be inaccurate.

I believe the important thing is to use an *emulator* (not the SDK) that has Play APIs or Play Store installed. That's what has always made the difference for me at least, and could perhaps use a mention

Finally - I think the better location for this is in the tips and tricks area with the header for the tip being the actual error message - https://rnfirebase.io/faqs-and-tips

mikehardy (changes requested):

docs/index.md
apfritts: Thanks for the tips! I've moved it to FAQs and also found another alternative as well. I left them both documented. Let me know if you think I should clarify when this happens more.

docs/faqs-and-tips.md
```diff
@@ -37,6 +37,19 @@ This may be fixed by creating a `firebase.json` file at the root of your project
...
+
+Alternatively, you can install the Google Play Services SDK. In Android Studio:
+
+1. Select `Tools -> SDK Manager`
+1. Search for `SDK Tools` and select the `SDK Tools` tab under `Android SDK`
+1. Install `Google Play services`
+
+You may need to delete your Android devices to pickup Play services.
+
+
```
mikehardy, line 52:
```suggestion
```

I'm still not sure how installing the SDK works to solve this at all. That's a development installation for compilation etc. I believe if you install the SDK but attempt to run on a device without play services you will suffer the indicated error message, conversely without the SDK installed if you run on a device with play services it will work. But without it won't.

So for me it is 100% an emulator issue, nothing to do with the SDK. I never install that SDK, for instance. I can't recommend any one else do something that doesn't work and that I never do, as maintainer

mikehardy (changes requested):

docs/faqs-and-tips.md
apfritts: I initially fixed it by installing the SDK and other folks have also fixed it by doing the same.

https://stackoverflow.com/a/17276663/2275865

I'm not sure everyone outside the US has access to Google Android images but happy to accept your suggestion.

mikehardy: I live outside the US :-), the android images are available to anyone without geo restriction. I'm very confident in suggesting it as a workaround. How odd though - I had no idea the SDK could be used as a workaround, I guess you learn something every day! I think the current version is still best though - clean and easy

Keep an eye on this space in general though, as "Firebase without Play Services" will become a hot topic with the popularity of Huawei devices in combination with inability of Google to work with them via sanctions

mikehardy (approved):
A thing of beauty 😄 - and this should save people some hassle, thanks @apfritts

Issue #3220, opened 2020-02-17
firestore query validators


russellwheatley:
## Issue

[firestore]Firebase web version allows the use of `where` query to be invoked before and after `orderBy` query. On RNFB, if a user invokes the `where` query after the `orderBy` query, the validation is bypassed. Perhaps merging the validation for `where` and `orderBy` query to ensure it is consistent.

dackers86:
@russellwheatley Just to clarify, what do you mean by validation?

Does this mean to say the where clause is not executed?

PR #4459, opened 2020-10-27
fix(firestore): Added test case for orderby before where query

dackers86:
Query validators are ineffective when using a where clause after an order by clause. This PR investigates and adds tests cases to prove the logic.

See issue https://github.com/invertase/react-native-firebase/issues/3220


mikehardy (dismissed):
LGTM 👍

mikehardy:
@dackers86 I just brought this one up to date with main branch, and the test was still `it.only` plus was not using it's own collection so it was muddled with other results when not `it.only` so I altered the collection ref to be unique, then it passed -> merging!

PR #4536, opened 2020-11-11
fix(analytics): added missing quantity parameter to the Item structure

Danchoys:
### Description

This pull request addresses the problem of not being able to pass the quantity parameter when trying to log any of these events:
- add_to_cart
- add_to_wishlist
- remove_from_cart
- view_cart
- refund

## Motivation

According to the [documentation](https://firebase.google.com/docs/analytics/measure-ecommerce#swift_7) for the aforementioned events one must or can pass the quantity parameter, but currently the library will reject such parameters, since quantity is not mentioned in the Item struct.

### Checklist

  - [X] Yes
  - [X] `Android`
  - [X] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [X] No

:fire:

### DIFF
--- a/packages/analytics/ios/RNFBAnalytics/RNFBAnalyticsModule.m
+++ b/packages/analytics/ios/RNFBAnalytics/RNFBAnalyticsModule.m
@@ -44,21 +44,7 @@
         rejecter:
         (RCTPromiseRejectBlock) reject) {
     @try {
-      if (params[kFIRParameterItems]) {
-        NSMutableDictionary *newParams = [params mutableCopy];
-        NSMutableArray *newItems = [NSMutableArray array];
-        [(NSArray *)params[kFIRParameterItems] enumerateObjectsUsingBlock:^(id  _Nonnull obj, NSUInteger idx, BOOL * _Nonnull stop) {
-          NSMutableDictionary *item = [obj mutableCopy];
-          if (item[kFIRParameterQuantity]) {
-            item[kFIRParameterQuantity] = @([item[kFIRParameterQuantity] integerValue]);
-          }
-          [newItems addObject:[item copy]];
-        }];
-        newParams[kFIRParameterItems] = [newItems copy];
-        [FIRAnalytics logEventWithName:name parameters:[newParams copy]];
-      } else {
-        [FIRAnalytics logEventWithName:name parameters:params];
-      }
+      [FIRAnalytics logEventWithName:name parameters:[self cleanJavascriptParams:params]];
     } @catch (NSException *exception) {
       return [RNFBSharedUtils rejectPromiseWithExceptionDict:reject exception:exception];
     }
@@ -147,6 +133,25 @@
         (RCTPromiseRejectBlock) reject) {
     // Do nothing - this only exists in android
     return resolve([NSNull null]);
+  }
+
+#pragma mark -
+#pragma mark Private methods
+
+  - (NSDictionary *)cleanJavascriptParams:(NSDictionary *)params {
+    NSMutableDictionary *newParams = [params mutableCopy];
+    if (newParams[kFIRParameterItems]) {
+      NSMutableArray *newItems = [NSMutableArray array];
+      [(NSArray *)newParams[kFIRParameterItems] enumerateObjectsUsingBlock:^(id  _Nonnull obj, NSUInteger idx, BOOL * _Nonnull stop) {
+        NSMutableDictionary *item = [obj mutableCopy];
+        if (item[kFIRParameterQuantity]) {
+          item[kFIRParameterQuantity] = @([item[kFIRParameterQuantity] integerValue]);
+        }
+        [newItems addObject:[item copy]];
+      }];
+      newParams[kFIRParameterItems] = [newItems copy];
+    }
+    return [newParams copy];
   }
 
 @end
--- a/packages/analytics/ios/RNFBAnalytics/RNFBAnalyticsModule.m
+++ b/packages/analytics/ios/RNFBAnalytics/RNFBAnalyticsModule.m
@@ -44,7 +44,21 @@
         rejecter:
         (RCTPromiseRejectBlock) reject) {
     @try {
-      [FIRAnalytics logEventWithName:name parameters:params];
+      if (params[kFIRParameterItems]) {
+        NSMutableDictionary *newParams = [params mutableCopy];
+        NSMutableArray *newItems = [NSMutableArray array];
+        [(NSArray *)params[kFIRParameterItems] enumerateObjectsUsingBlock:^(id  _Nonnull obj, NSUInteger idx, BOOL * _Nonnull stop) {
+          NSMutableDictionary *item = [obj mutableCopy];
+          if (item[kFIRParameterQuantity]) {
+            item[kFIRParameterQuantity] = @([item[kFIRParameterQuantity] integerValue]);
+          }
+          [newItems addObject:[item copy]];
+        }];
+        newParams[kFIRParameterItems] = [newItems copy];
+        [FIRAnalytics logEventWithName:name parameters:[newParams copy]];
+      } else {
+        [FIRAnalytics logEventWithName:name parameters:params];
+      }
     } @catch (NSException *exception) {
       return [RNFBSharedUtils rejectPromiseWithExceptionDict:reject exception:exception];
     }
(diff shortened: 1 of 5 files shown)


mikehardy:
Clearly a potential parameter in the item object in javascript https://firebase.google.com/docs/analytics/measure-ecommerce#add_remove_product


It appears we are also missing item price (!), item coupon code, and quite a few others, here is the type definition from firebase-js-sdk https://github.com/firebase/firebase-js-sdk/blob/47b182b91879d6e4bcce07609c517938d9354c9f/packages/analytics-types/index.d.ts#L197-L228 and the guide https://firebase.google.com/docs/reference/js/firebase.analytics.Item

A question for verification @Danchoys - I assume you have tested this change, when you do so does the quantity param actually travel through correctly and show up on the analytics dashboard? You can hack right in to node_modules to try it if you have not yet (use patch-package to make the patch manageable if it works and you want to keep it...)

The reason I ask is because the Java reference docs match the item-specific params with what we have, https://firebase.google.com/docs/reference/android/com/google/firebase/analytics/FirebaseAnalytics.Param#QUANTITY - quantity does not seem item-specific. Sometimes there are things in the javascript APIs that don't map to Java/iOS so we end up with implementation differences. This may be one of them, and make this change have no effect

iOS mirrors Java in that it also does not have quantity so I'm iffy on this. It looks like the only item-specific things we are missing are "list id" and "list name"

🤔

Danchoys:
@mikehardy In fact I did test it and I thought that it worked, but decided to double check it and got something that I didn't expect to see at all. First of all the quantity does show up:

Screen Shot 2020-11-11 at 19 31 46

But as you can see it is zero. Then I played with the native code and actually got it to work by using the following trick:
```ObjC
        if ([name isEqualToString:@"add_to_cart"]) {
            NSMutableDictionary *item = [params[@"items"][0] mutableCopy];
            item[@"quantity"] = [NSNumber numberWithInteger:[item[@"quantity"] integerValue]]; // Magic!
            NSDictionary *dict = @{ @"items" : @[ [item copy] ] };
            [FIRAnalytics logEventWithName:name parameters:dict];
        } else {
            [FIRAnalytics logEventWithName:name parameters:params];
        }
```
Using NSNumber literal (e.g. `@5`) also does the trick. It seems that firebase doesn't like the number that arrives from javascript...

Here is the corresponding JS code:

```js
firebase
      .analytics()
      .logAddToCart({
        items: [
          {
            item_name: 'Test item',
            item_id: '123123123',
            item_category: 'test_category',
            quantity: 5,
          },
        ],
      })
      .catch(() => {});
```
😩

mikehardy:
Wow! This will require some deeper thought...

1. you were able to get it to go through and show up: that's great! It means that the other item parameters from firebase-js-sdk that are not specifically called out in firebase-ios-sdk and firebase-android-sdk as being item-specific may go through too!

That implies that we are potentially missing quite a few useful parameters...

2. the code there at the native layer needs corresponding work on the android java side, and it should probably be keyed more to what exists in the item object vs the event in use? Or is there a mapping of what item data can be sent depending on what event is logged? If some item data (e.g., quantity) may only be sent on some event types (e.g., add to card, refund etc) then we should do it similar to how you did it, but with only the right things put in the event depending on type

I'm willing to help with this - getting the code shaped up and merged, offering pointers - but I won't have time to do the research on item data element validity vs event type, or the implementation testing. There is a need to change the documentation and the e2e tests to support the new parameters and also they should be validated ([I think quantity is 1-200 inclusive](https://developers.google.com/analytics/devguides/collection/analyticsjs/field-reference#(product)_quantity))

Danchoys:
1. Yeah, there may really be other fields missing as well. I just bumped into the quantity specifically since I was migrating my code to v6. Shall I extend this PR with the other fields or shall it be done separately?

2. That was just an attempt to get it to work. Of course checking for the event name like this is not a production-ready solution. I will further investigate why the numbers are acting the way they do... I’ve also tested Android and it seems to be plagued by the same problem.

Regarding the tests and docs - sure, will update these too.

mikehardy:
If you've got the time for the other ones sure - that would be fantastic! I understand that as a sort of "do it because you need it" contribution you may not be motivated but the more the merrier, from my maintainer perspective, given some proof they are working of course. This strikes me as kind of a documentation miss on the firebase part as well, that's something I could log with them upstream once we're sure that the other things shown as product level in the main analytics docs actually work via firebase if attached to items

packages/analytics/ios/RNFBAnalytics/RNFBAnalyticsModule.m
```diff
@@ -44,7 +44,21 @@ - (dispatch_queue_t)methodQueue {
...
         (RCTPromiseRejectBlock) reject) {
     @try {
-      [FIRAnalytics logEventWithName:name parameters:params];
+      if (params[kFIRParameterItems]) {
+        NSMutableDictionary *newParams = [params mutableCopy];
+        NSMutableArray *newItems = [NSMutableArray array];
+        [(NSArray *)params[kFIRParameterItems] enumerateObjectsUsingBlock:^(id  _Nonnull obj, NSUInteger idx, BOOL * _Nonnull stop) {
+          NSMutableDictionary *item = [obj mutableCopy];
+          if (item[kFIRParameterQuantity]) {
+            item[kFIRParameterQuantity] = @([item[kFIRParameterQuantity] integerValue]);
```
mikehardy, line 53: Just to make sure I understand, this is effectively a type conversion (from number which I guess goes to float instead of integer?) in native code because you need it to be an native integer (as discovered by testing) for it to show up in the event on the analytics console?

Danchoys: Exactly. `item[kFIRParameterQuantity]` is of type `NSNumber` and its `integerValue` method gives back a pure whole number. Then the value is wrapped with `@()`, which is a shorthand for creating an NSNumber, equivalent would be:
`[NSNumber numberWithInteger:[item[kFIRParameterQuantity] integerValue]]`.

mikehardy: Great! My Obj-C is weak but apparently (finally!!!) getting stronger if I got that correct :-). Thank you for confirming

Danchoys:
I have updated the PR with some refactoring on iOS plus what I think would be a solution for Android, but since I'm not really a Java developer, would highly appreciate if someone checks the code. I have tested both platforms and it seems to work for me. I have also checked if there are any specific restrictions on the range and was able to successfully report quantity of 1000, so I bet 200 limit does not apply here.

Finally I checked the existing unit and e2e tests and frankly speaking I'm not sure what I could add there... Firebase itself wasn't complaining about the number having the floating point, it was simply appearing as 0 afterwards. I guess mocking out the whole native FirebaseAnalytics module and checking the incoming values would do the trick, but I feel it is going way too far.

Danchoys:
Update: I decided to add the quantity param to one of the e2e tests so that the new code path is executed, which will make sure that at least it doesn't crash anything.

mikehardy:
Thanks for adding a little bit of test cover - it is always a tension on how much to test etc but at least sending new values through to make sure the code is actually exercised with no crashing is good. Most of the jest testing we do is just input validation / throw on invalid input. I've just been *all* over that stuff with #4552 - if there's anything to add that's trivial I'll add it as a review comment, or if it's non-trivial I can post up a follow-on PR

I'm weak at Obj-C but Android was my whole world prior to react-native so I'm comfortable giving those a once over

Goes without saying - thanks for working through this, we've advanced the general knowledge on the subject and you've implemented something really important for the specific events.

packages/analytics/ios/RNFBAnalytics/RNFBAnalyticsModule.m
```diff
@@ -135,4 +135,23 @@ - (dispatch_queue_t)methodQueue {
...
+        NSMutableDictionary *item = [obj mutableCopy];
+        if (item[kFIRParameterQuantity]) {
+          item[kFIRParameterQuantity] = @([item[kFIRParameterQuantity] integerValue]);
+        }
+        [newItems addObject:[item copy]];
+      }];
+      newParams[kFIRParameterItems] = [newItems copy];
+    }
+    return [newParams copy];
+  }
```
Danchoys, line 155: Alternative solution would be to transform the incoming dictionary into a JSON string, search and replace all numbers with zeroes after the dot with whole numbers and transform it back into an object, but obviously this may have some unexpected impact on other things...

Danchoys:
I'm also happy to contribute :) Regarding the tests - I can see that on Android one test is failing in some irrelevant module (messaging) due to timeout. Doubt though that it is related to the changes from this PR.. and also the tests do pass locally.

mikehardy:
Yep - I just switched that test in #4552  so that it no longer runs in CI but does run locally (setBackgroundMessageHandler) as it's flakey due to performance reasons of nested virtualization in CI, no worries about that one

mikehardy (approved):
LGTM - working through the pending items stack and heading for a release shortly - thanks for this @Danchoys

PR #4550, opened 2020-11-13
fix(database): fixed typescript type definitions

### LINKED ISSUE #4545: [🐛] Type Definition Error in database Types
## Issue

The type definition for database `Query.on` has an incorrect type.

Current Type:
```
on(
      eventType?: EventType,
      callback?: Function,
      cancelCallbackOrContext?: Record<string, any>,
      context?: Record<string, any> | null,
    ): Function;
``` 

`cancelCallbackOrContext` is describe as : `@param cancelCallbackOrContext An optional callback that will be notified if your event subscription is ever canceled because your client does not have permission to read this data (or it had permission but has now lost it). This callback will be passed an `Error` object indicating why the failure occurred.` but has a type of `Record<string, any>`. 

I believe this type should be similar to `Query.once`'s `failureCallbackContext` which is `((a: Error) => void) | Record<string, any> | null`. I do wonder why either of them is a `Record` in the first place if they are just meant as a function but I haven't been working with the lib in ts for more than a few minutes now (converted js -> ts).

Happy to make a quick PR but thought I would post here before assuming the right answer.

SCasarotto:
### Description

See Discussion in #4545  for context. Simply fixing up some typescript type definitions to match the js lib.

### Related issues
Fixes #4545  

### Checklist

  - [x] Yes
  - [x] `Android`
  - [x] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

---

🔥

### DIFF
--- a/packages/database/lib/index.d.ts
+++ b/packages/database/lib/index.d.ts
@@ -647,7 +647,7 @@ export namespace FirebaseDatabaseTypes {
     on(
       eventType?: EventType,
       callback?: (a: DataSnapshot, b?: string | null) => any,
-      cancelCallbackOrContext?: ((a: Error) => any) | Object | null,
+      cancelCallbackOrContext?: ((a: Error) => any) | Record<string, any> | null,
       context?: Record<string, any> | null,
     ): (a: DataSnapshot, b?: string | null) => any;
 
@@ -676,8 +676,8 @@ export namespace FirebaseDatabaseTypes {
     once(
       eventType: EventType,
       successCallback?: (a: DataSnapshot, b?: string | null) => any,
-      failureCallbackContext?: ((a: Error) => void) | Object | null,
-      context?: Object | null,
+      failureCallbackContext?: ((a: Error) => void) | Record<string, any> | null,
+      context?: Record<string, any> | null,
     ): Promise<DataSnapshot>;
 
     /**
--- a/packages/database/lib/index.d.ts
+++ b/packages/database/lib/index.d.ts
@@ -646,10 +646,10 @@ export namespace FirebaseDatabaseTypes {
      */
     on(
       eventType?: EventType,
-      callback?: Function,
-      cancelCallbackOrContext?: Record<string, any>,
+      callback?: (a: DataSnapshot, b?: string | null) => any,
+      cancelCallbackOrContext?: ((a: Error) => any) | Object | null,
       context?: Record<string, any> | null,
-    ): Function;
+    ): (a: DataSnapshot, b?: string | null) => any;
 
     /**
      * Listens for exactly one event of the specified event type, and then stops listening.
@@ -670,12 +670,14 @@ export namespace FirebaseDatabaseTypes {
      * @param eventType One of the following strings: "value", "child_added", "child_changed", "child_removed", or "child_moved."
      * @param successCallback A callback that fires when the specified event occurs. The callback will be passed a DataSnapshot. For ordering purposes, "child_added", "child_changed", and "child_moved" will also be passed a string containing the key of the previous child by sort order, or `null` if it is the first child.
        @param failureCallbackContext An optional callback that will be notified if your client does not have permission to read the data. This callback will be passed an Error object indicating why the failure occurred.
+       @param context If provided, this object will be used as this when calling your callback(s).
      */
 
     once(
       eventType: EventType,
       successCallback?: (a: DataSnapshot, b?: string | null) => any,
-      failureCallbackContext?: ((a: Error) => void) | Record<string, any> | null,
+      failureCallbackContext?: ((a: Error) => void) | Object | null,
+      context?: Object | null,
     ): Promise<DataSnapshot>;
 
     /**


mikehardy:
Following on the discussion from the issue, do you mind if I peg this one as pending on resolution of your upstream documentation request, to make sure we don't waffle on the types if they go a different direction? (you never know!)

In the meantime, so that isn't an inconvenient stance for you I highly recommend https://github.com/ds300/patch-package if you haven't seen it, and you should note that we generate patches for every PR so they are easy to integrate + test :-), just follow the "details" link above on the 'Create Test Patches' check and you'll find a little hard-to-see dropdown for "artifacts" where you can snag the zip file with the patches. Or obviously you can make them yourself

SCasarotto:
Yeah I think thats totally fine. This isn't really a blocker and I can easily locally override the types. Having said that, ultimately its their docs site and not their working code that is at fault, so I would be surprised my request there has any impact on the actual js lib. Ultimately no rush. 

Have been using this lib for years (ever since the v5/v6 conversion) and am just glad I get to contribute. 😄  🚀

SCasarotto:
That has now been merged into the firebase docs have been updated thus this PR should be allowed to be merged. It does look like there is now a merge conflict. It looks like there have been changes to the types slightly. These types that are currently merged don't directly reflect the firebase docs (which I thought was our standard here). Below are the differences:

Current Master Code:
```
      callback?: (a: DataSnapshot, b?: string | null) => any,
      cancelCallbackOrContext?: ((a: Error) => any) | Record<string, any> | null,
      context?: Record<string, any> | null,
    ): (a: DataSnapshot, b?: string | null) => any;
```

My Code:
```
      callback?: (data: DataSnapshot, previousChildKey?: string) => void,
      cancelCallbackOrContext?: Record<string, any>,
      context?: Record<string, any> | null,
    ): () => void;
```

Firebase Code ([ref](https://github.com/firebase/firebase-js-sdk/blob/5b2d1073a7a4e1366b0d807510f2bded713ed7bc/packages/database-types/index.d.ts#L84-L95)):
```
    callback: (a: DataSnapshot, b?: string | null) => any,
    cancelCallbackOrContext?: ((a: Error) => any) | Object | null,
    context?: Object | null
  ): (a: DataSnapshot, b?: string | null) => any;
```

As I am not deeply in tune with this communities thoughts I don't know if I am the right person to resolve these conflicts. I think it can make sense to name the variables something more meaningful. There doesn't seem to be a lot of harm in that. However, some of the actual types are different and that seems to go against trying to keep them in sync. 

Any thoughts?

Other Context:
- Firebase have updated their docs since this PR was made to match their code. 
  - Docs: https://firebase.google.com/docs/reference/js/firebase.database.Query#on
  - PR: firebase/firebase-js-sdk#4250 (comment)
- Some comments in the issue related to this PR https://github.com/invertase/react-native-firebase/issues/4545

mikehardy:
The type deviation was a result of me doing #4711 a large-scale depedency update that flagged a ton of lint and then the corresponding de-linting pass where eslint typescript rules were complaining about the 'Object' and 'Function' definitions - https://github.com/invertase/react-native-firebase/pull/4711/commits/1beb5b5f688e07a5abe3bc7177329d81f135d0d7?diff=unified&w=1#diff-bf8976d56b60a0e9ceb9258c0027de145b69fa134f8ec1bde7701ffb6788a9a9

The firebase-js-sdk typings should be our standard, my intention is not to deviate from them but I dropped the ball on this one

I was moving quickly through there and completely forgot I was constrained by firebase-js-sdk API typing :facepalm: - at the same time I was referring to the actual javascript implementation as I was modifying the types, so I'm not sure we would accept or generate the full set of firebase-js-sdk type signature options for some of the parameters / return values etc.

I have these goals then 
- I need [`yarn lint`](https://github.com/invertase/react-native-firebase/blob/b4fb97698eb2579da3937bc6fed5dc4267f56aa1/package.json#L10) to pass, and I can accept tsconfig changes or ts-ignore directives if needed to make it so
- I *don't* want things added to the signature that don't appear to be valid in our codebase - even if that represents a slight deviation (like, some of the possible parameter types in a positional argument aren't valid for us
- I *do* want any types we *do* accept to match firebase-js-sdk if at all possible

I think that gives us the best possible outcome: we'll fail to compile things that aren't valid here but it will be possible to write code that is perfectly portable between react-native-firebase and firebase-js-sdk

I'm running a little low on sleep - so to confirm: Does that sound reasonable? Do you think a mild re-shape of this PR while de-conflicting can meet all those goals, at least for this function? (possibly for the file, if you're feeling ambitious)

SCasarotto:
@mikehardy This totally makes sense. Thanks for the detailed reply. There is a lot on my plate for the next couple weeks but will try to circle back to help on this stuff.

mikehardy:
Hey @SCasarotto no pressure, I understand contributing to open source projects does not frequently move to the top of the priority pile, but I'm curious if you made any progress on this? Sorry for messing up the types in between the original analysis and the final merge to firebase-js-sdk, that definitely set things back

SCasarotto:
I haven't made any progress. We also are not longer using this part of the RNF package so its dropped even lower priority for me. I likely won't get back to this. At this point master looks fine and wouldn't have caused me an issue. Id probably just close this pr.

mikehardy:
No worries - and I totally understand. I'll see what I can do via a scan-and-salvage on it is I do like the types to line up personally - thanks for the original effort I do appreciate it!

mikehardy:
#5141 carries this work into the repo, thanks @SCasarotto !

SCasarotto:
Very welcome. Glad I could contribute even if only a little.

Issue #4551, opened 2020-11-14
[🐛] Bug Report Title - Auth Emulator


jonathanroze:
Hi, 

On the documentation it's explain how to use firestore locally by using FireStore Emulator but I can't find anything about use Auth locally.

It's a missing part in docs or it's just not possible ? Thanks

mikehardy:
Just merged auth emulator support, will be in the next release which should be tomorrow

You can see examples of how to use it by examining our test app:

https://github.com/invertase/react-native-firebase/blob/master/packages/auth/e2e/helpers.js
https://github.com/invertase/react-native-firebase/blob/master/.github/workflows/scripts/start-firebase-emulator.sh


https://github.com/invertase/react-native-firebase/blob/4c688413cb6516ecfdbd4ea325103d0d8d8d84a8/packages/auth/e2e/auth.e2e.js#L24-L41

mikehardy:
v10, just released, has auth emulator support! Enjoy

PR #4552, opened 2020-11-15
feat(auth, emulator): implement useEmulator with jest + e2e testing

mikehardy:
### Description

Implement `firebase.auth().useEmulator`

I'm about to release a major version anyway so I included the removal of deprecated analytics methods

I intend to rebase on to master so these will stay as separate commits and will be picked up correctly for release

It should be reviewed commit by commit for clarity

Phone flow is not implemented yet - may get that in a follow-on commit here or in a follow-on PR. It is possible but you have to interact with the REST API on the emulator to do it.

### Related issues

#4461 
#4463 
#4510 
#4551 

### Checklist

  - [x] Yes
  - [x] `Android`
  - [x] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

This enabled me to turn on auth e2e testing again, and I implemented jest tests even though they were really difficult because of the native mocking need

---


packages/app/e2e/helpers.js
```diff
@@ -0,0 +1,12 @@
...
+  return 'react-native-firebase-testing';
+};
+
+// Android requires a re-mapped hostname during testing sometimes
+exports.getE2eEmulatorHost = function getE2eEmulatorHost() {
+  // This appears to work from localhost even on android. Different execution context?
+  // if (platform === 'android') {
+  //   return '10.0.2.2';
+  // }
+  return 'localhost';
```
Salakar, line 11: This works as localhost due to; https://github.com/invertase/react-native-firebase/pull/4552/files#diff-17651da2be070111f5b3f961a4e6fda957ed253e301e14f8a1da4a5ce95c44efR343-R351

We dod this on all emulator apis for RNFB so the users don't need to worry about it

packages/auth/e2e/auth.e2e.js
```diff
@@ -15,7 +15,41 @@
...
+    // Need a way to create the user but have it disabled immediately
+    // If we had the functions emulator running as well, it would work:
+    // https://stackoverflow.com/questions/43140441/how-to-disable-user-accounts-on-firebase-authentication-automatically#comment73363725_43141780
+    // try {
+    //   const userCredential = await firebase
+    //     .auth()
+    //     .createUserWithEmailAndPassword(DISABLED_EMAIL, DISABLED_PASS);
+    // } catch (e) {
+    //   // they may already exist, that's fine
+    // }
```
Salakar, line 43: Can hold off for now and hopefully in future something like this becomes available on the emulator 🤞

packages/auth/e2e/helpers.js
```diff
@@ -0,0 +1,91 @@
+/* eslint-disable no-console */
+const { getE2eTestProject, getE2eEmulatorHost } = require('../../app/e2e/helpers');
+const http = require('http');
+
+exports.getRandomPhoneNumber = function getRandomPhoneNumber() {
+  return '+593' + Utils.randString(9, '#19');
+};
+
+exports.clearAllUsers = async function clearAllUsers() {
```
Salakar, line 9: Off topic but curious if a disable user API could be reverse engineered of the Emulator Web UI requests as well, does the emulator web ui have an option to disable a user?

Salakar (approved):

packages/app/e2e/helpers.js
mikehardy: Yeah, I got that part from the functions emulator implementation and it works great. This is separate though. This helper function is returning a string without any intervening translation to a higher layer of helper function which constructs a raw URL out of it with no other translation, and only localhost works. It makes me think the e2e javascript is running in the context of jet on the host as opposed to the react-native JS executor in the emulator. I verified that you need 10.0.2.2 if you try it on the emulator (using `adb shell` and then `nc` in that shell to pretend I was a web client).

TL:DR; which you probably know: the e2e javascript is calling functions in the emulator but the regular javascript, is it executing in jet on the host?

packages/auth/e2e/helpers.js
mikehardy: Oh it does! This might solve the problem. I will check if it's possible

mikehardy: Looks like a POST to the standard APIs works against the emulator! The docs were not clear on that, they have an "emulator specific" setting but don't mention if the non-specific APIs work, apparently they do:

```
scheme | http
-- | --
host | localhost:9099
filename | /identitytoolkit.googleapis.com/v1/accounts:update

{"disableUser":true,"localId":"tUkcA81QgKn0l6n1CXl3EFW1kR1x"}
```

Needs the same 'Authorization: Bearer owner' trick from account purge REST API call, but I already have the infrastructure to do REST API calls to the emulator so this is easy peasy now.

Salakar: Nice!

packages/app/e2e/helpers.js
Salakar: Oh ye sorry I misread this, e2e JS does run in Node.js on the host machine and not the emulator - so this makes sense.

mikehardy: perfect, I will update that comment as "where things run" is vital to know, especially since we (as mentioned) do the translation to 10.0.2.2 in other places.

Side note: I do the equivalent of rebase -i for these PRs so the commits can stay separate, I've taken care that they'll even survive (with correct success/fail) if applied partially but in sequence so git bisect will work (something we learned to be careful about in AnkiDroid). Not the standard I would hold most contributors too but I think these commits serve best if merged in separated, similar to my iosv7/androidv26 forward port

PR #4565, opened 2020-11-17
feat(admob, ios): add adaptive banner support

dylancom:
### Description

This pull request adds support for adaptive banners. https://github.com/invertase/react-native-firebase/discussions/4276

### Release Summary

Added adaptive banner support for iOS.

### Checklist

  - [x] Yes
  - [ ] `Android`
  - [x] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No


packages/admob/lib/index.d.ts
```diff
@@ -237,6 +237,11 @@ export namespace FirebaseAdMobTypes {
      */
     SMART_BANNER: 'SMART_BANNER';
 
+    /**
+     * A (next generation) dynamically sized banner that is full-width and auto-height.
+     */
+    APAPTIVE_BANNER: 'APAPTIVE_BANNER';
```
mikehardy, line 243:
```suggestion
    ADAPTIVE_BANNER: 'ADAPTIVE_BANNER';
```

packages/admob/lib/index.d.ts
```diff
@@ -237,6 +237,11 @@ export namespace FirebaseAdMobTypes {
      */
     SMART_BANNER: 'SMART_BANNER';
 
+    /**
+     * A (next generation) dynamically sized banner that is full-width and auto-height.
```
mikehardy, line 241: If there is only an iOS implementation - should mention it is iOS only?

```suggestion
     * A (next generation) dynamically sized banner that is full-width and auto-height.
     *
     * iOS only
```

mikehardy (changes requested):
Hey thanks for this! Looks like a typo and maybe could use a note for iOS only, but otherwise if you have tested this (specifically the CFGloat / size calculations) and it works, that works for me

dylancom:
> Hey thanks for this! Looks like a typo and maybe could use a note for iOS only, but otherwise if you have tested this (specifically the CFGloat / size calculations) and it works, that works for me

I have tested it and it works. I will fix the typo and add an iOS note. In the future I will probably have time to also add Android support.

mikehardy (approved):
Looks great! Thanks!
Until release, you can go to the "patch packages" action and find a zip file with the patch you can apply locally (using the always amazing https://github.com/ds300/patch-package). I mention that because I just did a couple of very big releases and I want those to marinate for a bit before I do another one

Issue #4600, opened 2020-11-23
Vercel deploy failing on normal PR test builds and merges


mikehardy:
https://vercel.com/invertase/react-native-firebase/l8xtr6lmy

Not sure why?

> Command "cd website && yarn && yarn run build" exited with 1

```

 
--
10:30:05.113 | $ gatsby build
10:30:08.374 | success open and validate gatsby-configs - 0.268s
10:30:08.977 | success load plugins - 0.602s
10:30:09.055 | success onPreInit - 0.076s
10:30:09.059 | success delete html and css files from previous builds - 0.003s
10:30:09.066 | success initialize cache - 0.006s
10:30:09.097 | success copy gatsby files - 0.031s
10:30:09.115 | success onPreBootstrap - 0.018s
10:30:09.163 | success createSchemaCustomization - 0.001s
10:30:10.358 | /vercel/workpath0/tsconfig.json
10:30:10.673 | Using TypeScript 3.9.7 from /vercel/workpath0/website/node_modules/typescript/lib
10:30:17.970 | Error: /vercel/workpath0/node_modules/@types/react-native/node_modules/@types/react/index.d.ts(2980)
10:30:17.970 | Duplicate identifier 'LibraryManagedAttributes'.
10:30:17.971 | Error: /vercel/workpath0/node_modules/@types/react-native/node_modules/@types/react/index.d.ts(2991)


```

andersonaddo:
Let the entropy-induced headaches commence.

mikehardy:
Yeah, entropy literally never stops. I released useful energy as pointless heat just by transmitting this message, and the heat probably broke something in the internet somewhere.

mikehardy:
I have a rough root cause analysis on this one. 'react' as a dependency has moved from v16 to v17, and there are some types dependencies that are loose (`*` specified for `@types/react`) while others are pinning a version.

So multiple react types are coming in at the same time causing it to blow up. I can reproduce it in a clean RNFB checkout locally by doing `yarn && cd website && yarn && yarn build`) but I don't have a fix quite yet

https://github.com/DefinitelyTyped/DefinitelyTyped/issues/25145

PR #4618, opened 2020-11-25
chore: make ReactNativeFirebaseMessagingSerializer + events builder functions public

alexisbronchart:
### Description

I am in a situation where I need to manually send events through `ReactNativeFirebaseEventEmitter`. See discussion https://github.com/invertase/react-native-firebase/discussions/4577 
In order to properly do so, I'd like to use the functions present in `ReactNativeFirebaseMessagingSerializer`. 
This PR just adds the public access modifier to that class and to the functions that build events.

### Checklist

  - [x] Yes
  - [x] `Android`
  - [ ] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### DIFF
--- a/packages/messaging/android/src/main/java/io/invertase/firebase/messaging/ReactNativeFirebaseMessagingSerializer.java
+++ b/packages/messaging/android/src/main/java/io/invertase/firebase/messaging/ReactNativeFirebaseMessagingSerializer.java
@@ -11,7 +11,7 @@ import io.invertase.firebase.common.SharedUtils;
 import java.util.Map;
 import java.util.Set;
 
-class ReactNativeFirebaseMessagingSerializer {
+public class ReactNativeFirebaseMessagingSerializer {
   private static final String KEY_TOKEN = "token";
   private static final String KEY_COLLAPSE_KEY = "collapseKey";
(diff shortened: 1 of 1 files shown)


mikehardy (approved):
We have some history of doing this in Firebase classes, and this follows the exact precedent

We may be *slightly* less careful with regards to breaking changes coming through as this now represents public API surface area that is not normal for us to consider but we'll try to semver those correctly too (if they ever change, unlikely, but just a heads up - it'll be a compile-time thing anyway so will fail fast even if we mess up semver here)

So these LGTM

mikehardy:
I have a medium-size stack of changes pending release now, plus some CI work to do, which is to say that I'm expecting to release these shortly (I want to get that stack out...) but it may take a day or three (the CI work is also important prior to release, there are a couple things not working well and I don't want to have a quality error). So please have patience but also you can click the "patch package" CI check from this PR run and you'll find an artifact associated with that CI check. It has the official patch-package files you can use in the meantime, knowing that they'll match the future release exactly so you don't have to wait on me

Cheers

mikehardy:
CI triage / fix completed in #4600 - this should be good to go

Issue #4598, opened 2020-11-23
Bump Github Actions versions used in workflows


mikehardy:
the retry Action at least has a v2 which appears fine (but needs to be implemented and then watched once, of course)

There may be others so all should be checked

Should be done at same time as #4595 since they're in the same location

mikehardy:
It is extremely difficult to work on github actions without just pushing them to master, but for anyone following along at home know that I watched every.single.check go after updating these, and they all passed - the associated commit was good :+1:

PR #4620, opened 2020-11-26
test(ci): Test non-semantic PR title PR. Do not merge. Will close/delete.

mikehardy:
The semantic pull request PR action was not resolving after #4598 - should be fixed now but only one way to test...


mikehardy:
The semantic pull request workflow check is now correct and working.
Very strange that `@v2` did not work as a dependency for the action but `@v2.1.0` did.

PR #4664, opened 2020-12-09
fix(remote-config,ios): fixed fetch throttling issue

Danchoys:
### Description

As per documentation, apps must be very gentle when it comes to fetching from the Firebase Remote Config storage. By default, it happens not more often than once in 12 hours. For that the library provides a selector: `fetchWithCompletionHandler:` that doesn't have the time interval parameter and uses the value from settings and the second one `fetchWithExpirationDuration:completionHandler:` that allows to override this configuration. In JS it is exposed as a single method of signature `fetch(expirationDurationSeconds?: number): Promise<void>;`. When `expirationDurationSeconds` is undefined, -1 is passed to the native module and then treated as follows:
```
  if (expirationDuration == @(-1)) {
    [[FIRRemoteConfig remoteConfigWithApp:firebaseApp] fetchWithCompletionHandler:completionHandler];
  } else {
    [[FIRRemoteConfig remoteConfigWithApp:firebaseApp] fetchWithExpirationDuration:expirationDuration.doubleValue completionHandler:completionHandler];
  }
```

The problem here is that an `NSNumber` instance is compared by reference with another `NSNumber` instance (yet holding the same value) which always fails, leading to the `else` path always executed. And internally in the FirebaseRemoteConfig SDK `-1` is going to be treated same as `0` in such a way that no limitation on fetching will be applied, resulting in almost immediate throttling unleashed by the server.

This PR mitigates this issue and allows to fetch the remote config without the risk of throttling.

### Checklist

  - [X] Yes
  - [ ] `Android`
  - [X] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [X] No

:fire:

### DIFF
--- a/packages/remote-config/ios/RNFBConfig/RNFBConfigModule.m
+++ b/packages/remote-config/ios/RNFBConfig/RNFBConfigModule.m
@@ -119,7 +119,7 @@ RCT_EXPORT_METHOD(fetch:
     }
   };
 
-  if (expirationDuration == @(-1)) {
+  if (expirationDuration.integerValue == -1) {
     [[FIRRemoteConfig remoteConfigWithApp:firebaseApp] fetchWithCompletionHandler:completionHandler];
   } else {
     [[FIRRemoteConfig remoteConfigWithApp:firebaseApp] fetchWithExpirationDuration:expirationDuration.doubleValue completionHandler:completionHandler];


mikehardy (approved):
Nice catch! I think you win my available bounty of internet points for the day :1st_place_medal:  - I would never have seen this.

I'm hoping to get a release done today, but the patch-package Github check will have a patch set available as an attached artifact in the meantime

PR #4665, opened 2020-12-09
chore(ci): removed sonar cloud code coverage

dackers86:
Sonar cloud is currently not utilised in our CI. 

This can be removed as code coverage is covered with Codecov


mikehardy (approved):
interesting, that's it? Just one properties file? And indeed, a `grep -ri sonarcloud *` int he source tree comes up empty. Cool!

PR #4676, opened 2020-12-15
fix(ios, storage): handle nil file extension from ios14 M1 emulators

### LINKED ISSUE #4661: [Crash] putFile will crash when extension in unknown
I'm running xcode12 on Apple Silicon and it seems that pathExtension which used to return the extensions of the file in path is returning nil on iOS14 simulators (for gif files - does not happen on iOS13). This is probably a simulator issue, however, this code does not check if UTI is null and therefor crashes when trying to release:

`CFRelease(UTI);`

https://github.com/invertase/react-native-firebase/blob/b5357573e97ccc3bfdf6a190713a7df66e462c4b/packages/storage/ios/RNFBStorage/RNFBStorageCommon.m#L62-L66

It should be something like: 

`if (UTI) {
CFRelease(UTI);
}`

Return value can be null according to the [documentation](https://developer.apple.com/documentation/coreservices/1448939-uttypecreatepreferredidentifierf):
> Return Value
> A new CFStringRef containing a uniform type identifier, or NULL if inTagClass is not a known tag class

Maxoos:
### Description

This fix resolves a crash that occurs when pathExtension returns NULL, failing to detect file extension. This was detected as a result of a simulator bug with Apple Silicon Macs where pathExtension returns NULL only iOS14 (works as expected on iOS13).

This bug would probably surface on any device if file extension could not be detected when calling mimeTypeForPath, this could be used to verify the fix on intel Macs.

### Related issues

Fixes #4661 
### Checklist

  - [ ] Yes
  - [ ] `Android`
  - [x] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

Verified mimeTypeForPath function on Apple Silicon with iOS14 sim to return @"application/octet-stream". Verified CFRelease is being called on iOS13 sim.

---

### DIFF
--- a/packages/storage/ios/RNFBStorage/RNFBStorageCommon.m
+++ b/packages/storage/ios/RNFBStorage/RNFBStorageCommon.m
@@ -63,7 +63,7 @@
       (__bridge CFStringRef) [localFilePath pathExtension],
       NULL);
   CFStringRef mimeType = UTTypeCopyPreferredTagWithClass(UTI, kUTTagClassMIMEType);
-  CFRelease(UTI);
+  if (UTI) { CFRelease(UTI); }
 
   if (!mimeType) {
     return @"application/octet-stream";


mikehardy:
Thanks @Maxoos - it looks like the CLA still isn't signed? If you can sign that I can work this through the QA process and merge it

mikehardy:
Hi @Maxoos ! Unfortunately without the CLA signed (you can follow the instructions above or click the 'details' link on the license/cla check) we cannot merge this and I will have to close it

Maxoos:
Sorry @mikehardy was a bit full on before the holidays. Should be all sorted now!

mikehardy (approved):

Issue #4784, opened 2021-01-18
[🐛] Analytics not reporting on Android


WadhahEssam:
## Issue

Describe your issue here
- Logging analytics events are not reported after the addition of the new `analytics_auto_collection_enabled` feature in 10.4.1.
- This happens when you don't set a value for `analytics_auto_collection_enabled` in the firebase.json, which is not labeled as a mandatory field in the docs, so I  was led to believe it wasn't.
---

## what is the problem: 
- in file `analytics/android/build.gradle` the check for `analytics_auto_collection_enabled` checks if the value is true, else it will just set it to false.
<img width="617" alt="Screen Shot 2021-01-18 at 9 10 01 AM" src="https://user-images.githubusercontent.com/24798045/104878588-4c385680-596d-11eb-813b-50c11afb74fa.png">

## suggested solution:
- add a function in `app/android/firebase-json.gradle` to check if the value exists or not, without checking its value.
<img width="477" alt="Screen Shot 2021-01-18 at 9 11 28 AM" src="https://user-images.githubusercontent.com/24798045/104878638-607c5380-596d-11eb-8606-5e18a1af45ce.png">

- add a checking for the value existence in `analytics/android/build.gradle` before checking its value
<img width="617" alt="Screen Shot 2021-01-18 at 9 10 08 AM" src="https://user-images.githubusercontent.com/24798045/104878692-725df680-596d-11eb-8d9e-2ec5c315064f.png">

WadhahEssam:
For those who are looking for an immediate solution:


- set `analytics_auto_collection_enabled` in the firebase.json to true
<img width="433" alt="Screen Shot 2021-01-18 at 9 16 11 AM" src="https://user-images.githubusercontent.com/24798045/104878883-d4b6f700-596d-11eb-9c7e-b84e161f1f96.png">


- OR call setAnalyticsCollectionEnabled();
```
import { firebase } from '@react-native-firebase/analytics';
// ...
await firebase.analytics().setAnalyticsCollectionEnabled(true);
```

WadhahEssam:
For those who are looking for an immediate solution:


- set `analytics_auto_collection_enabled` in the firebase.json to true

- OR call setAnalyticsCollectionEnabled();
```
import { firebase } from '@react-native-firebase/analytics';
// ...
await firebase.analytics().setAnalyticsCollectionEnabled(true);
```

mikehardy:
[...]

PR #4786, opened 2021-01-18
fix(analytics,android): fixed analytics reporting in android

WadhahEssam:
### Description

- Fixed analytics issue where events are not reported if the value of `analytics_auto_collection_enabled` is not set in firebase.json file.

### Related issues

#4784 

### Checklist

  - [x] Yes
  - [x] `Android`
  - [ ] `iOS`
- My change includes tests; (not needed)
- [ ] I have updated TypeScript types that are affected by my change. (not needed)
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

---

### Note
I am not really that familiar with contributing to open source libraries, hope that was helpful.

### DIFF
--- a/packages/analytics/android/build.gradle
+++ b/packages/analytics/android/build.gradle
@@ -68,8 +68,10 @@ apply from: file("./../../app/android/firebase-json.gradle")
 def autoCollectionEnabled = "true"
 
 if (rootProject.ext && rootProject.ext.firebaseJson) {
-  if (rootProject.ext.firebaseJson.isFlagEnabled("analytics_auto_collection_enabled") == false) {
-    autoCollectionEnabled = "false"
+  if (rootProject.ext.firebaseJson.doesFlagExist("analytics_auto_collection_enabled") == true) {
+    if (rootProject.ext.firebaseJson.isFlagEnabled("analytics_auto_collection_enabled") == false) {
+      autoCollectionEnabled = "false"
+    }
   }
 }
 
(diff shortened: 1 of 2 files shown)


mikehardy:
Thanks for posting this up! I need to check iOS now as well to see how it is handling this, might have the equivalent problem, I'm scanning my comms queue from overnight to see if anything more important is waiting but otherwise this is the highest priority actual task, should be shortly

WadhahEssam:
I have published an app in both platform, and noticed that android events are not reported, but iOS worked just fine, ( I had to republish the app) and now I have many users who are in a version without any analytics enabled, so thanks for taking this seriously

mikehardy:
Yeah I am in the same boat - I have an android release build live on the store now in a work project that has no analytics at the moment.

This is triggered when you have a firebase.json, but that firebase.json does not have the key present. In that case the groovy method from firebase-json.gradle will always return false, which is incorrect.

I went a different direction with a fix though - rather than add another method I copied the stringValue method style from just below your proposed method, and enforced that callers send a default value to return if it doesn't exist and I modified the 3 callers in the various packages. This is a bit more compact in usage because some things default to true (analytics collection) and some to false (admob delay init). Now it is explicit in the code what the default is because you send it in, and you don't need a separate "does it exist?" check in calling code

ios works differently because it is additive, not replacing - if there is no value, nothing is added so the SDK default "wins". The Android style is replacing a placeholder in an AndroidManifest.xml so if there was no value the implementation silently "failed" to a false and put that in the placeholder spot.

I probably wouldn't have noticed this problem for a while regardless, all the internet points are yours!

WadhahEssam:
makes total sense, I just didn't want to mess with a code outside of this issue scope, Thanks @mikehardy

mikehardy:
Yeah - that thought is normally correct, you mention you are relatively new to open source, your idea there - keep the change very small - is actually the exact right place to start

Normally I would have made comments and asked for / given explicit approval for widening the change a bit and worked with you for the change but this is a big deal so I wanted to try to move faster so I could make sure to get a release out today

PR #4791, opened 2021-01-18
fix(app, android): require default firebase.json boolean key

### LINKED ISSUE #4784: [🐛] Analytics not reporting on Android

mikehardy:
### Description

This fixes the case where firebase.json existed but the key being
query was missing. Previously the config system would silently fail on
the missing key and return false even though the default for some keys
should be true

In particular this caused analytics auto collection to be disabled
if you had a firebase.json but no analytics_auto_collection_enabled = true entry,
even though the default should have been true

Credit to @WadhahEssam for discovery, diagnosis, and general solution

### Related issues

Fixes #4784 
Obsoletes #4786 (different API shape, same solution though)

### Checklist

  - [x] Yes
  - [x] `Android`
  - [ ] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

I tested:
- firebase.json present, key missing, default value false: got false
- firebase.json present, key missing, default value true: got true
- firebase.json present, key present and true, default value false: got true
- firebase.json present, key present and false, default value true: got false

Seems to work fine although I did discover some problems and was able to break my tests - the tests I was doing were testing it and I proved it locally by toggling back and forth

the logic is not engaged if firebase.json is not present

---

### DIFF
--- a/packages/admob/android/build.gradle
+++ b/packages/admob/android/build.gradle
@@ -74,7 +74,7 @@ def firebaseJSONAdmobDelayAppMeasurementInitBool = false
 
 if (rootProject.ext.firebaseJson) {
   firebaseJSONAdmobAppIDString = rootProject.ext.firebaseJson.getStringValue("admob_android_app_id", "")
-  firebaseJSONAdmobDelayAppMeasurementInitBool = rootProject.ext.firebaseJson.isFlagEnabled("admob_delay_app_measurement_init")
+  firebaseJSONAdmobDelayAppMeasurementInitBool = rootProject.ext.firebaseJson.isFlagEnabled("admob_delay_app_measurement_init", false)
 }
 
 if (!firebaseJSONAdmobAppIDString) {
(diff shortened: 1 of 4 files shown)


mikehardy:
As soon as this passes CI I will merge and release 10.4.2 @WadhahEssam

mikehardy:
Actually @WadhahEssam - since I know you are watching this :-), it will be 10.5.0 because I'm going to merge the feature / minor bump to the underlying SDKs when they finish too #4792

PR #4815, opened 2021-01-25
docs(auth, ios): clarified phone auth app verification setup

benallfree:
Doc update

### DIFF
--- a/docs/auth/phone-auth.md
+++ b/docs/auth/phone-auth.md
@@ -19,7 +19,11 @@ Ensure the "Phone" sign-in provider is enabled on the [Firebase Console](https:/
 
 # iOS Setup
 
-Ensure that all steps in https://firebase.google.com/docs/auth/ios/phone-auth have been followed, particularly the custom URL scheme. See https://github.com/invertase/react-native-firebase/issues/557 for more information.
+Ensure that all parts of step 1 and 2 from [the official firebase iOS phone auth docs](https://firebase.google.com/docs/auth/ios/phone-auth#enable-phone-number-sign-in-for-your-firebase-project) have been followed, noting in particular that you may need to re-download your firebase GoogleService-Info.plist file and for the reCAPTCHA flow to work you must make sure you have added your custom URL scheme to your project plist file.
+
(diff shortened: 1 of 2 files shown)


docs/auth/phone-auth.md
```diff
@@ -17,6 +17,10 @@ end-user consent prior to using the Firebase Authentication phone number sign-in
 
 Ensure the "Phone" sign-in provider is enabled on the [Firebase Console](https://console.firebase.google.com/project/_/authentication/providers).
 
+# iOS Setup
+
+Ensure that all steps in https://firebase.google.com/docs/auth/ios/phone-auth have been followed, particularly the custom URL scheme. See https://github.com/invertase/react-native-firebase/issues/557 for more information.
```
mikehardy, line 22: I'd rather point upstream (official firebase docs) rather than an issue here (downstream) where it isn't clear which part of the extensive comments the user needs

```suggestion
Ensure that all parts of step 1 and 2 from [the official firebase iOS phone auth docs](https://firebase.google.com/docs/auth/ios/phone-auth#enable-phone-number-sign-in-for-your-firebase-project) have been followed, noting in particular that you may need to re-download your firebase GoogleService-Info.plist file and for the reCAPTCHA flow to work you must make sure you have added your custom URL scheme to your project plist file.

Phone auth requires app verification, and the automatic app verification process uses data-only firebase cloud messages to the app. Data-only cloud messaging only works on real devices where the app has background refresh enabled. If background refresh disabled, or if using the Simulator, app verification uses the fallback reCAPTCHA flow allowing you to check if it is configured correctly.

For reliable automated testing, you may want to disable both automatic and fallback reCAPTCHA app verification for your app. To do this, [you may disable app verification in AuthSettings](https://rnfirebase.io/reference/auth/authsettings#appVerificationDisabledForTesting) prior to calling any phone auth methods.
```

mikehardy (approved):
Thanks for this! I like the idea of clarification but I'd rather have the docs be authoritative vs referring to a long/tangled issue thread. So I put all the useful bits of the thread with links upstream in place instead

PR #4849, opened 2021-02-02
fix(admob): improve defense logic to prevent multiple calls

ifsnow:
### Description

**`loaded: Whether the advert is loaded and can be shown.`**

If `_loaded` is changed to `true` when `load()` is called, the meaning of `loaded` getter becomes ambiguous.

If the goal is to prevent multiple calls, it would be better to use other variables for this.

### Checklist

  - [X] Yes
  - [X] `Android`
  - [X] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [X] No

### DIFF
--- a/packages/admob/lib/ads/InterstitialAd.js
+++ b/packages/admob/lib/ads/InterstitialAd.js
@@ -45,11 +45,11 @@ export default class InterstitialAd extends MobileAd {
 
   load() {
     // Prevent multiple load calls
-    if (this._loaded) {
+    if (this._loaded || this._isLoadCalled) {
       return;
     }
 
-    this._loaded = true;
+    this._isLoadCalled = true;
     this._admob.native.interstitialLoad(this._requestId, this._adUnitId, this._requestOptions);
   }
 
--- a/packages/admob/lib/ads/MobileAd.js
+++ b/packages/admob/lib/ads/MobileAd.js
@@ -28,6 +28,7 @@ export default class MobileAd {
     this._requestOptions = requestOptions;
 
     this._loaded = false;
+    this._isLoadCalled = false;
     this._onAdEventHandler = null;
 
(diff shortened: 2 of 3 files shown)


mikehardy (commented):
Should it set loadCalled to false here as the event to set loaded == false is handled?

https://github.com/invertase/react-native-firebase/pull/4849/files#diff-82dfca6ec9ac0ffb90446c057bb3e58807ae9249382e2e85e3ea953f8122b6fbR48

mikehardy:
...I should add - thanks for the PR :-)

mikehardy:
Hey @ifsnow - I had a comment on this PR, can you take a look?

ifsnow:
@mikehardy Sorry for being late. If `load()` can be called again after receiving `CLOSED` event, it will need to be initialized here.

```js
_handleAdEvent(event) {
    const { type, error, data } = event.body;

    ...

    if (type === AdEventType.CLOSED || type === RewardedAdEventType.CLOSED) {
      this._loaded = false;
      this._isLoadCalled = false; // <= here
    }

    ....
}
```

Let me know if this is what you said. I'll fix it right away.

mikehardy:
Yes! I think that's the spot. I'm not sure if it could happen but it seems better to match the states precisely?

ifsnow:
@mikehardy That's a good point. My commit was modified. Thanks!

mikehardy:
Great! @ifsnow last thing left is to sign the CLA so we can merge the change? You can follow the details link on the github actions check to do it - I think it's this https://cla-assistant.io/invertase/react-native-firebase?pullRequest=4849

mikehardy (approved):
looks good to me!

ifsnow:
@mikehardy Did it. Thanks

mikehardy:
v10.6.3 is newly published with this change :rocket:

Issue #4708, opened 2020-12-24
Publishing process does not bump and publish app every time?


mikehardy:
Something about the publish process did not quite work for the v10.3.1 release

All of the sub-packages were bumped to v10.3.1 and published but it appears the app package did not commit + publish as 10.3.1

mikehardy:
On minor version changes (e.g., 10.2.x to 10.3.x) app is published correctly: https://github.com/invertase/react-native-firebase/runs/1574459729?check_suite_focus=true

On patch version changes app is not necessarily published correctly: https://github.com/invertase/react-native-firebase/runs/1577737380?check_suite_focus=true

It may be that on minor version changes app *itself* changes (usually we alter underlying SDKs for those) so it is being detected then, whereas on the last few patch version changes app has not changed

Why is not being bumped every single time though? Needs triage

Note the main changelog has entries for patch version https://github.com/invertase/react-native-firebase/blob/master/CHANGELOG.md

app CHANGELOG does not have the entries though https://github.com/invertase/react-native-firebase/blob/master/packages/app/CHANGELOG.md
admob CHANGELOG does get entries though https://github.com/invertase/react-native-firebase/blob/master/packages/admob/CHANGELOG.md

PR #4870, opened 2021-02-04
feat(perf): support "perf_auto_collection_enabled" flag in firebase.json

michaelgmcd:
### Description

More or less clones https://github.com/invertase/react-native-firebase/commit/9a24ecd2826bfa8ab30657287432ccaeff8b7c7c
 
### Checklist

  - [X] Yes
  - [X] `Android`
  - [X] `iOS`
- My change includes tests;
- This is a breaking change;
  - [X] Yes - Theoretically if someone had this flag set beforehand, it would not have been accurately reflected. This could be a breaking change in the sense that it would change someone's app configuration.
  - [ ] No

### Test Plan

---

Checking that firebase performance is disabled on app start when this flag is set.

### DIFF
--- a/packages/perf/android/build.gradle
+++ b/packages/perf/android/build.gradle
@@ -63,9 +63,22 @@ project.ext {
   ])
 }
 
+apply from: file("./../../app/android/firebase-json.gradle")
+
+def autoCollectionEnabled = "true"
+
+if (rootProject.ext && rootProject.ext.firebaseJson) {
+  if (rootProject.ext.firebaseJson.isFlagEnabled("perf_auto_collection_enabled", true) == false) {
+    autoCollectionEnabled = "false"
+  }
+}
+
 android {
   defaultConfig {
     multiDexEnabled true
+    manifestPlaceholders = [
+      firebaseJsonAutoCollectionEnabled: autoCollectionEnabled
+    ]
   }
   lintOptions {
     disable 'GradleCompatible'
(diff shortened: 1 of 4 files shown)


packages/perf/android/build.gradle
```diff
@@ -63,9 +63,22 @@ project.ext {
   ])
 }
 
+apply from: file("./../../app/android/firebase-json.gradle")
+
+def autoCollectionEnabled = "true"
+
+if (rootProject.ext && rootProject.ext.firebaseJson) {
+  if (rootProject.ext.firebaseJson.isFlagEnabled("perf_auto_collection_enabled") == false) {
```
mikehardy, line 71: The code you're cloning had a "fun" problem! #4784 that I fixed by altering the gradle API a bit #4791

This clones the fix for the implementation clone: 
https://github.com/invertase/react-native-firebase/pull/4791/files#diff-4e4bf90ffe1ed3e1706bc42bb35a7816f0db7fe5cff54d5c17f761be59b165b0R71

```suggestion
  if (rootProject.ext.firebaseJson.isFlagEnabled("perf_auto_collection_enabled", true) == false) {
```

I'm surprised this built for you, as I think that gradle API change requires the second argument to isFlagEnabled, it should have kicked out some sort of gradle error?

mikehardy (changes requested):
This looks in general fantastic, thank you!
I had one change I propose - with a question attached - that I'd like your feedback on

packages/perf/android/build.gradle
michaelgmcd: Good catch! I've updated the PR.

mikehardy (approved):
This looks great - thank you!

I'm a little bit back-logged on publishing a new version but 10.6.0 with this change in it should release "really soon now" (a day? three? less than a week)

michaelgmcd:
All good. We'll use a workaround until it's ready. Thanks for the quick review!

mikehardy:
I had a moment and fixed the bottleneck problem #4708 and got 10.6.1 out with all peerDependencies between the modules correct and everything :sweat_smile:  - good to go!

PR #4886, opened 2021-02-09
fix(auth, android): do not timezone offset when getting UTC timestamp

kouheiszk:
### Description

In Android, JWT's iat and exp are now converted to correctly UTC time in strings.
Unnecessary offsets were being added to the timestamps, causing them to be converted to the wrong UTC time in strings.

### Checklist

  - [x] Yes
  - [x] `Android`
  - [ ] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

I have tested the following:

- The JWT iat was correctly converted to UTC time.
- The JWT exp was correctly converted to UTC time.

---

### DIFF
--- a/packages/app/android/src/reactnative/java/io/invertase/firebase/common/SharedUtils.java
+++ b/packages/app/android/src/reactnative/java/io/invertase/firebase/common/SharedUtils.java
@@ -97,12 +97,10 @@ public class SharedUtils {
   }
 
   public static String timestampToUTC(long timestamp) {
-    Calendar calendar = Calendar.getInstance();
     long millisTimestamp = timestamp * 1000;
-    Date date = new Date(millisTimestamp + calendar.getTimeZone().getOffset(millisTimestamp));
-    SimpleDateFormat format = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss'Z'", Locale.US);
+    SimpleDateFormat format = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss'Z'");
     format.setTimeZone(TimeZone.getTimeZone("UTC"));
-    return format.format(date);
+    return format.format(millisTimestamp);
   }
 
   /**
--- a/packages/auth/e2e/user.e2e.js
+++ b/packages/auth/e2e/user.e2e.js
@@ -51,8 +51,8 @@ describe('auth().currentUser', function () {
   });
 
   describe('getIdTokenResult()', function () {
-    it("should return a valid IdTokenResult Object", async function() {
-      const random = Utils.randString(12, "#aA");
+    it('should return a valid IdTokenResult Object', async function () {
+      const random = Utils.randString(12, '#aA');
       const email = `${random}@${random}.com`;
 
       const { user } = await firebase.auth().createUserWithEmailAndPassword(email, random);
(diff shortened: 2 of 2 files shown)


packages/app/android/src/reactnative/java/io/invertase/firebase/common/SharedUtils.java
```diff
@@ -97,12 +97,10 @@ public static WritableMap getExceptionMap(Exception exception) {
   }
 
   public static String timestampToUTC(long timestamp) {
-    Calendar calendar = Calendar.getInstance();
     long millisTimestamp = timestamp * 1000;
-    Date date = new Date(millisTimestamp + calendar.getTimeZone().getOffset(millisTimestamp));
```
mikehardy, line 102: Interesting - this method is also used by storage:

```
packages/app/android/src/reactnative/java/io/invertase/firebase/common/SharedUtils.java:  public static String timestampToUTC(long timestamp) {
packages/auth/android/src/main/java/io/invertase/firebase/auth/ReactNativeFirebaseAuthModule.java:            SharedUtils.timestampToUTC(Objects.requireNonNull(tokenResult).getAuthTimestamp()),
packages/auth/android/src/main/java/io/invertase/firebase/auth/ReactNativeFirebaseAuthModule.java:            SharedUtils.timestampToUTC(tokenResult.getExpirationTimestamp()),
packages/auth/android/src/main/java/io/invertase/firebase/auth/ReactNativeFirebaseAuthModule.java:            SharedUtils.timestampToUTC(tokenResult.getIssuedAtTimestamp()),
packages/storage/android/src/main/java/io/invertase/firebase/storage/ReactNativeFirebaseStorageCommon.java:    metadata.putString(KEY_TIME_CREATED, SharedUtils.timestampToUTC(storageMetadata.getCreationTimeMillis() / 1000));
packages/storage/android/src/main/java/io/invertase/firebase/storage/ReactNativeFirebaseStorageCommon.java:    metadata.putString(KEY_UPDATED, SharedUtils.timestampToUTC(storageMetadata.getUpdatedTimeMillis() / 1000));
```

...and this change is semantic, the timestamps *will* be different now for storage.

To be safe the previous behavior might be best preserved, with a new parameter added to the function `boolean useOffset` or similar where storage calls it with `true` and that creates the Calendar instance and does the offset, while Auth does not?

Otherwise the storage code assumptions must be examined to make sure that it can tolerate the new non-local-offset timestamps

mikehardy (changes requested):
Hi there! Thanks for this, I think you are on to something but we need to be careful with shared APIs

I left a comment related to the other consumers of the shared API you modified - can you have a look at it and tell me what you think?

packages/app/android/src/reactnative/java/io/invertase/firebase/common/SharedUtils.java
kouheiszk: Thanks!

I think we should be able to get the same time on both iOS and Android, so I investigated the usage in storage.

In storage, it is used to string the creation and update dates of storageReference.
iOS or Android without considering the time zone, I think the behavior will remain the same after this change. The others will be able to get the correct values, but it is a disruptive change.

What do you think?

mikehardy: I did not want to "prejudice" your check of the storage code by saying - before you looked at it - that I think essentially the same thing. I think this is more correct even though it is a disruptive change, and I like the change. We could preserve the old behavior but I think the change is actually an improvement. I'll release this in a little bit - I want to see if there is anything else queued up I could get in the release as well.

Thanks!

mikehardy (approved):

mikehardy (approved):

mikehardy (approved):

PR #4977, opened 2021-03-02
fix(analytics, android): change param access from ITEMS to ITEM_LIST to fix crash

muhammad-asad-26:
### Description
The application crashes while running on android in React Native and changing this to ITEM_LIST fix the issue

### DIFF
--- a/packages/analytics/android/src/reactnative/java/io/invertase/firebase/analytics/ReactNativeFirebaseAnalyticsModule.java
+++ b/packages/analytics/android/src/reactnative/java/io/invertase/firebase/analytics/ReactNativeFirebaseAnalyticsModule.java
@@ -125,7 +125,7 @@ public class ReactNativeFirebaseAnalyticsModule extends ReactNativeFirebaseModul
     if (bundle == null) {
       return null;
     }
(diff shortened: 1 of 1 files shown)


mikehardy:
A crash fix, thank you!
Could you post any relevant documentation and the stack trace for this one to make sure I understand it?
The reason I ask is that when @Danchoys originally added this, he specifically added something to the e2e tests that *should* have triggered a crash if it were going to crash

https://github.com/invertase/react-native-firebase/pull/4536/files#diff-84f708ec167c2368f1ffd15b498db692ae249426885ee1d64bb9c055b18fc067R144

So I'm confused how this can crash here :thinking: - perhaps the e2e test needs to be augmented to show the crash ?

Danchoys:
I would not recommend changing it to ITEM_LIST as an attempt to fix the issue. The docs state:
```
String | ITEM_LIST | This constant is deprecated. Use ITEM_LIST_NAME instead.
```
It seems that you have populated the items array in your event with something that makes it crash.

@muhammad-asad-26 Can you please post the code that sends the event? 

@mikehardy Yeah, there was an e2e test for that thing, so it shouldn't (shouldn't :D) be causing a crash.

mikehardy:
@muhammad-asad-26 this will be closed if there's no response, I'll give it a few more days, thanks!

PR #4981, opened 2021-03-03
perf: increase task throughput in Android using thread pool executor

bernard-kms:
### Description

Related discussion: https://github.com/invertase/react-native-firebase/discussions/4969

This PR aims to increase task execution throughput in Android with making minimal behavioral changes.

- As before, transactional tasks execute with single-threaded executor.
  - This includes "write" actions such as `set`, `update`, `remove` at the database/firestore and `putFile` at the storage.
- Non-transactional tasks execute with configured `ThreadPoolExecutor`.
  - If thread pool size exceeded, the task execute with the single-threaded executor which has unbounded queue size.
  - configuration reference: [Android `AsyncTask` module](https://github.com/aosp-mirror/platform_frameworks_base/blob/453966a20bc305693fb63488e5543d94f570a7cb/core/java/android/os/AsyncTask.java#L199)

I haven't run precise benchmarks yet, but typical use case applications should have improved performance.

**further considerations:**
- Is transactional tasks cleanly, and correctly divided?
- Should transaction scenario cover with test?

### Checklist

  - [x] Yes
  - [x] `Android`
  - [ ] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

---


mikehardy:
Hoping to look at this today or tomorrow - please have patience if it takes a little bit, but thank you very much for posting this

I will mention that on all PRs we generate patch-package patch sets for the PR as artifacts, you can use those in the meantime if you like:

here is the patch-package CI job https://github.com/invertase/react-native-firebase/pull/4981/checks?check_run_id=2021948264

and the artifact (not sure if the URL is stable) https://github.com/invertase/react-native-firebase/suites/2168204522/artifacts/44535863

bernard-kms:
As @Salakar pointed out in the discussion thread(https://github.com/invertase/react-native-firebase/discussions/4969#discussioncomment-427576), realtime update event should be handled in serial per query, which is currently(in this PR) not. This will be addressed in further works.

mikehardy:
Ok! I'll wait for real review until that is done but I can still give this a scan for structural items etc and have that queued this morning. I think this will be a big performance benefit if the special cases can be handled and am still positive on the idea, for what it's worth :-)

bernard-kms:
Now it processes realtime event in serial per listener. The approach taken here is to separate each event listener into a single threaded executor.

When a new listener is created, new single threaded executor is created and registered with the listener identifier(`eventRegistrationKey` in database and `listenerId` in firestore). And when it's unsubscribed, the executor is removed accordingly.

packages/app/android/src/main/java/io/invertase/firebase/common/UniversalFirebaseModule.java
```diff
@@ -68,12 +70,21 @@ private ExecutorService getExecutor(boolean isTransactional) {
...
+      return Executors.newSingleThreadExecutor();
     } else {
-      return new ThreadPoolExecutor(0, Integer.MAX_VALUE, 60L, TimeUnit.SECONDS, new SynchronousQueue<Runnable>());
+      ThreadPoolExecutor threadPoolExecutor = new ThreadPoolExecutor(0, MAXIMUM_POOL_SIZE, KEEP_ALIVE_SECONDS, TimeUnit.SECONDS, new SynchronousQueue<Runnable>());
+      threadPoolExecutor.setRejectedExecutionHandler(executeInFallback);
+      return threadPoolExecutor;
     }
   }
 
+  private RejectedExecutionHandler executeInFallback = new RejectedExecutionHandler() {
```
mikehardy, line 82: Is the RejectedExecutionHandler intended to solve a problem where the multi-thread executor was in an "all threads busy" state, and rejected tasks that would have had to wait but otherwise executed just fine in the prior SingleThreadExecutor?

If so, I like this as a fallback strategy, however, it means that if the multi-thread executor shuts down first, all of it's tasks will spill over to the SingleThreadExecutor, only to immediately be shutdown there:

https://developer.android.com/reference/java/util/concurrent/RejectedExecutionHandler#rejectedExecution(java.lang.Runnable,%20java.util.concurrent.ThreadPoolExecutor)

> Method that may be invoked by a ThreadPoolExecutor when execute cannot accept a task. This may occur when no more threads or queue slots are available because their bounds would be exceeded, **or upon shutdown of the Executor.**

(emphasis mine)

I think it is possible to stop a cascade of shutdown rejection from executor to executor if you gate task resubmission to the SingleThreadExecutor on a check of the rejecting executor's isTerminating/isShutdown/isTerminated status

https://developer.android.com/reference/java/util/concurrent/ThreadPoolExecutor#isTerminating() (and the other two)

packages/app/android/src/main/java/io/invertase/firebase/common/UniversalFirebaseModule.java
```diff
@@ -21,12 +21,19 @@
...
+import java.util.concurrent.RejectedExecutionHandler;
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.SynchronousQueue;
 
 import javax.annotation.OverridingMethodsMustInvokeSuper;
 
 public class UniversalFirebaseModule {
+  private static final int MAXIMUM_POOL_SIZE = 20;
+  private static final int KEEP_ALIVE_SECONDS = 3;
```
mikehardy, line 36: How did you arrive at these constants?
- what happens if they are much smaller, or much larger?
- is there a possibility that with certain workloads these values could exhaust resources and cause a crash?
- is it possible that different workloads might benefit from different values?

I had a separate thought with regard to this implementation too - that is related - in the case there are some errors with regard to threading, even if the underlying implementation is "correct" (gets the intended executor for the intended firebase task pathway), is there a way to toggle off the multi-thread behavior?

I *think* if these were a tunable in firebase.json, especially the pool size, and it was set to 0, then that threadpool would always reject and everything would go single-thread? (though I'm not sure you can create a threadpool with a zero limit...)

packages/app/android/src/reactnative/java/io/invertase/firebase/common/ReactNativeFirebaseModule.java
```diff
@@ -74,19 +81,65 @@ public ReactContext getContext() {
...
+      return threadPoolExecutor;
+    }
+  }
+
+  private RejectedExecutionHandler executeInFallback = new RejectedExecutionHandler() {
+    public void rejectedExecution(Runnable r, ThreadPoolExecutor executor) {
+      ExecutorService fallbackExecutor = getTransactionalExecutor("");
+      fallbackExecutor.execute(r);
+    };
+  };
```
mikehardy, line 119: The duplication between this and UniversalFirebaseModule needs a solution - having this vital logic duplicated makes no sense to me as I look at it. Is there any way to have the logic in Universal and just used by ReactNative ?

On a quick read of the code I don't see a way around it using inheritance or object access - they appear quite distinct :thinking: 

Maybe split the implementation itself out into a task  handling Util or something...

mikehardy (changes requested):
Hi there! I just read through this and thought about it this morning - not ready for approval yet - I left some comments (some prescriptive, some more questions...) in various spots

Allowing the executor limit to be tunable in particular, with the ability to have the whole system go back to a single thread if needed is important to me in order to allow users to self-heal if there are problems. In fact the first version might ship with it tuned to 1 in order to not change the default without opt-in field testing first

Here is a PR that implements a firebase.json feature: https://github.com/invertase/react-native-firebase/pull/4870/files - then I *think* with it in AndroidManifest.xml you may access the value in code and use it for pool size tuning?

mikehardy:
On re-read my comment seems 100% critical whereas it's important to recognize this is quite the fancy piece of work in my opinion. I think it's clever, and it's actually pretty close to merge, and should help a lot of people, and that's worth saying :-). Cheers.

packages/app/android/src/main/java/io/invertase/firebase/common/UniversalFirebaseModule.java
bernard-kms: > it means that if the multi-thread executor shuts down first, all of it's tasks will spill over to the SingleThreadExecutor, only to immediately be shutdown there:

Yeah, it's true! Your suggestion is correct and looks straightforward.

packages/app/android/src/main/java/io/invertase/firebase/common/UniversalFirebaseModule.java
mikehardy: I think the biggest win by allowing tuning is to tune it to 0 (if that's possible?) such that everything spills to single thread executor and we then have a possible fallback behavior of "works just like it did before"

That work up front could stop a lot of possible frantic work later if our assumptions about multi-thread being okay in some cases is wrong :-)

bernard-kms:
Thanks for the careful review! 

> In fact the first version might ship with it tuned to 1 in order to not change the default without opt-in field testing first

Indeed, I think it's important to not make any breaking changes in the first release of this improvement.

I will add commits to implement tunable pool sizing, and addressing review comments.

Salakar:
Pending the points by @mikehardy, this seems like a nice approach and looks great - thanks for this @bernard-kms

bernard-kms:
Added commits to implement tunable thread pool executor with the abstraction. And also added some documentation. (am I doing it right?)

Default behavior is almost same as in the previous release, except for differences in the realtime event listeners(has independent executor). Setting `android_task_executor_maximum_pool_size` to `0` completely disable any performance-wise implementation, and each module will have only one single thread executor as before.

mikehardy:
Thanks for your patience on this @bernard-kms - I'm still really excited for this change but I want to release it by itself so we get clear testing feedback, and react-native 0.64 was released in the meantime which required an immediate reaction #5042 - I'm releasing that work today which opens the tree up again for this change, I should be able to do final review and hopefully merge later today after releasing #5042

packages/app/android/src/main/java/io/invertase/firebase/common/TaskExecutorService.java
```diff
@@ -0,0 +1,115 @@
...
+    String executorIdentifier = maximumPoolSize != 0 ? identifier : "";
+    return getExecutor(true, executorIdentifier);
+  }
+
+  public ExecutorService getExecutor(boolean isTransactional, String identifier) {
+    String executorName = getExecutorName(isTransactional, identifier);
+    ExecutorService existingExecutor = executors.get(executorName);
+    if (existingExecutor != null) return existingExecutor;
+    ExecutorService newExecutor = getNewExecutor(isTransactional);
+    executors.put(executorName, newExecutor);
```
mikehardy, line 67: I wonder if this should be synchronized on the executors object? Right now the method isn't synchronized and the state being mutated is the static set - is it possible for multiple getExecutor() calls to race and end up returning two different executors for use in calling code, even if the Set de-duplicates them?

I think by the nature of react-native being single-threaded it is not possible normally - but that may change in the future, and it means that lock contention at least should not be an issue


I recognize using a static object as a lock may represent a performance bottleneck of it's own but I consider that an implementation detail in that there are ways to optimize it but more importantly if it is required for correctness it must be done regardless

packages/app/android/src/main/java/io/invertase/firebase/common/TaskExecutorService.java
```diff
@@ -0,0 +1,115 @@
...
+    });
+  }
+
+  public void removeExecutor(String executorName) {
+    ExecutorService existingExecutor = executors.get(executorName);
+    if (existingExecutor != null) {
+      existingExecutor.shutdownNow();
+      executors.remove(executorName);
+    }
+  }
```
mikehardy, line 114: these should probably be synchronized on the executors object as well

docs/index.md
```diff
@@ -226,6 +226,26 @@ setting present in `/android/gradle.properties`:
...
+  "react-native": {
+    "android_task_executor_maximum_pool_size": 10,
+    "android_task_executor_keep_alive_seconds": 3,
+  }
+}
+```
+
+| Key          | Description                                     |
+| ------------ | ----------------------------------------------- |
+| `android_task_executor_maximum_pool_size` | Maximum pool size of ThreadPoolExecutor. Defaults to `1`. Larger value typically improve performance when executing large numbers of asynchronous tasks, e.g. Firestore queries. Setting this value to `0` completely disables the pooled executor and all tasks execute in serial per module. |
```
mikehardy, line 246:
```suggestion
| `android_task_executor_maximum_pool_size` | Maximum pool size of ThreadPoolExecutor. Defaults to `1`. Larger values typically improve performance when executing large numbers of asynchronous tasks, e.g. Firestore queries. Setting this value to `0` completely disables the pooled executor and all tasks execute in serial per module. |
```

packages/app/firebase-schema.json
```diff
@@ -65,6 +65,14 @@
         "perf_auto_collection_enabled": {
           "description": "Disable or enable auto collection of performance monitoring data collection.\n This is useful for opt-in-first data flows, for example when dealing with GDPR compliance.\nThis can be overridden in JavaScript.",
           "type": "boolean"
+        },
+        "android_task_executor_maximum_pool_size" : {
+          "description": "Maximum pool size of ThreadPoolExecutor used by RNFirebase for Android. Defaults to `1`.\n Larger value typically improve performance when executing large numbers of asynchronous tasks, e.g. Firestore queries.",
```
mikehardy, line 70:
```suggestion
          "description": "Maximum pool size of ThreadPoolExecutor used by RNFirebase for Android. Defaults to `1`.\n Larger values typically improve performance when executing large numbers of asynchronous tasks, e.g. Firestore queries.",d
```

mikehardy (commented):
Thank you again for your patience, this month has been busier than I thought. It's worth stating again I'm really excited about this.

Also, the code and documentation are excellent.

The only thing I can think of at all is maybe a need for synchronizing on the executors object to protect against concurrent access during gets that result in sets, or gets / shutdowns / removes happening at the same time

I'm curious for your thoughts there.

That said, I'm going to integrate this patch set into my work app for a test drive

packages/app/firebase-schema.json
```diff
@@ -65,6 +65,14 @@
         "perf_auto_collection_enabled": {
           "description": "Disable or enable auto collection of performance monitoring data collection.\n This is useful for opt-in-first data flows, for example when dealing with GDPR compliance.\nThis can be overridden in JavaScript.",
           "type": "boolean"
+        },
+        "android_task_executor_maximum_pool_size" : {
+          "description": "Maximum pool size of ThreadPoolExecutor used by RNFirebase for Android. Defaults to `1`.\n Larger values typically improve performance when executing large numbers of asynchronous tasks, e.g. Firestore queries.",d
+          "type": "number"
+        },
+        "android_task_executor_keep_alive_seconds" : {
+          "description": "Keep-alive time of ThreadPoolExecutor used by RNFirebase for Android, in seconds. Defaults to `3`.\n Excess threads in the pool executor will be terminated if they have been idle for more than the keep-alive time.",
```
mikehardy, line 74:
```suggestion
        "android_task_executor_maximum_pool_size": {
          "description": "Maximum pool size of ThreadPoolExecutor used by RNFirebase for Android. Defaults to `1`.\n Larger values typically improve performance when executing large numbers of asynchronous tasks, e.g. Firestore queries.",
          "type": "number"
        },
        "android_task_executor_keep_alive_seconds": {
          "description": "Keep-alive time of ThreadPoolExecutor used by RNFirebase for Android, in seconds. Defaults to `3`.\n Excess threads in the pool executor will be terminated if they have been idle for more than the keep-alive time.",
```

packages/app/android/src/main/java/io/invertase/firebase/common/TaskExecutorService.java
bernard-kms: I think you've got a fairly good point. As you mentioned, the multi-threading race issue might be hardly happened right now, but it also means that use of synchronized block hardly be a bottleneck. I see the synchronization with the executors object would eliminates the concern almost at no cost.

bernard-kms:
I agree with your concern, and added a commit to prevent race condition during getting/setting executors. 

I also ran a test track with my app, and there seems to be no problem.

mikehardy (approved):
I just re-read the whole thing again but with a focus on the changes of course. Still looks good to me, and I have this integrated with my work project as a set of patches, working well also. I think it's good to go! :rocket:  - I think this will be a big help to some people @bernard-kms - thanks

mikehardy:
This finally launched in v11.3.0 just now! Thanks for your patience and for making this change it should be a big help

bernard-kms:
Good! Thanks for the hard work to maintain this wonderful library. It's my pleasure to be a contributor :)

zholmes1:
@mikehardy @bernard-kms

Do either of you think that this upgrade would fix this intermittent C++ crash that some of my users are experiencing when `storage.putFile` is called?

```
SIGABRT 0x0000000000000000
libc.so
... 61 more ...
libcutils.so
```

My app is currently on `rnfb@10.5.1`, and I'm considering upgrading everything to `12.0.0` as a possible way to address this issue. Alternatively, I could do away with `storage.putFile` altogether and create a function that accepts the image in an HTTPS request, and uploads the file to storage from the Node function code. Although that brings with itself other potential points of failure. So I'd like to stay with the storage package, if possible.

Any help would be appreciated, and looking at the changelog for the storage package, this is the only PR that would potentially address my issue.

mikehardy:
@zholmes1 commenting on random pull requests is odd from a maintainer's perspective. Your question is not related to the PR at all really

If I were suffering from a native crash I would always update to current stable before doing anything, literally before even looking at the logs again, as it's likely a waste of time because either it's fixed or (like us) anyone you ask about will just say "update then reproduce then ask again", because it could be fixed.

If you only consult our change logs here in react-native-firebase you're potentially missing any changes from underlying firebase-ios-sdk / firebase-android-sdk and all of their transitive dependencies (a lot) all of which move at a steady pace and have been updated between your version and current stable. Any of them could have fixed it

zholmes1:
Well it's not really random, since it's the only applicable commit between my version and the current version.

I'm just concerned about potential regression, as I think we all have experiences with that when upgrading packages. But good point about the issue possibly being in the Android SDK, I hadn't considered that. I'll upgrade to `12.0.0` and if I still see this crash, I'll open an issue.

Thanks for the quick reply.

mikehardy:
the commits that bump SDK versions are all applicable as is anything in `@app` really but I also see your point if you're thinking super narrowly on storage.

zholmes1:
True, I honestly wan't even thinking about bumping SDK versions. Benefit of using awesome cross-platform packages like this one!

PR #4999, opened 2021-03-08
TypeScript migration exploration

russellwheatley:
### Description

Exploratory PR for migrating RNFB to TypeScript.

Hey @dackers86:

### Broad Overview
1. Continue Updating `app/src` package types so that the code will compile. Don't change implementation. If you're not sure, use `//@ts-ignore`. For instance, private members are accessed outside of their class instance which is a TS error. I've just ignored like below for now:
```js
//@ts-ignore
  const config = module._config;
```

2. Once confirmed TS works, will look at creating an abstract platform interface class to separate native from web. 
3. Checkout/Implement builder-bob package for different targets.

### Particulars

1. I've kept the `lib` file with existing JS code, created a `src` with TS code, when compiled, it will output into `dist` directory.
2. I've setup a `types.ts` file in root to share interfaces & types. Use this whenever there is duplicate types that needs to be shared. Use the existing `index.d.ts` for type inspiration. It doesn't exactly map 1-to-1. These are the public facing interfaces, put private types on the class itself. Ultimately, `index.d.ts` file will be eventually replaced as the types will be generated by the `.ts` files themselves. 
3. I've implemented types for `src/FirebaseApp.ts` & `src/internal/registry/app.ts` which you can use for inspiration moving forward.
4. Use the naming convention outlined [here](https://github.com/invertase/react-native-firebase/pull/4999#issuecomment-793647538)
 

### Checklist

  - [ ] Yes
  - [ ] `Android`
  - [ ] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [ ] No

### Test Plan

---


Salakar:
I know this is still in progress but I noticed at least 3 types defined for FirebaseApp; https://github.com/invertase/react-native-firebase/blob/922bbe298e8bc989e618dc7e94defbef9241f50f/packages/app/src/types.ts

packages/app/package-lock.json
```diff
@@ -0,0 +1,102 @@
+{
```
mikehardy, line 1: we use yarn, there should be no package-lock.json files?

packages/app/src/index.d.ts
```diff
@@ -0,0 +1,550 @@
+/*
```
mikehardy, line 1: if we are coding in typescript, there should be no commited .d.ts file anymore, should there? There should be types files in a dist/ dir, but those are derived and should not be committed to source, only rendered out during build for packaging + upload to npmjs on release I think?

packages/app/src/internal/constants.ts
```diff
@@ -0,0 +1,42 @@
...
+  'database',
+  'inAppMessaging',
+  'firestore',
+  'functions',
+  'iid',
+  'indexing',
+  'storage',
+  'dynamicLinks',
+  'messaging',
+  'naturalLanguage',
```
mikehardy, line 37: not related to your conversion, but naturallanguage should be removed, missed that when dropping ml packages oops!

packages/app/src/version.ts
```diff
@@ -0,0 +1,2 @@
+// generated by genversion
+export default '10.8.1';
```
mikehardy, line 2: generated files shouldn't be committed I don't think? may need to update .gitignore for new .ts version of file?

tsconfig.json
```diff
@@ -1,12 +1,11 @@
 {
   "include": ["packages/**/lib/*.d.ts", "packages/**/type-test.ts"],
   "compilerOptions": {
-    "noEmit": true,
+    // "noEmit": true,
```
mikehardy, line 4: be bold, delete the line! :laughing:

packages/app/package.json
```diff
@@ -3,8 +3,8 @@
   "version": "10.8.1",
   "author": "Invertase <oss@invertase.io> (http://invertase.io)",
   "description": "A well tested, feature rich Firebase implementation for React Native, supporting iOS & Android. Individual module support for Admob, Analytics, Auth, Crash Reporting, Cloud Firestore, Database, Dynamic Links, Functions, Messaging (FCM), Remote Config, Storage and more.",
-  "main": "lib/index.js",
-  "types": "lib/index.d.ts",
+  "main": "dist/index.js",
+  "types": "src/index.d.ts",
```
mikehardy, line 7: I think the types should still point to dist/ ? and be defined in the actual implementation files?

packages/app/src/internal/registry/namespace.ts
```diff
@@ -0,0 +1,292 @@
+/*
```
mikehardy, line 1: why are there diffs to registry/namespace.js and registry/namespace.ts in this set of diffs?

Related:  was expecting to see deletes of each associated .js file, then I think github will recognize them largely as file moves and show things as a diff between the two files vs whole-new-file-green / while-new-file-red for easier review

packages/app/src/types.ts
```diff
@@ -0,0 +1,145 @@
...
+
+  utils(): any;
+}
+
+export interface FirebaseModuleNamespaceImpl {
+  statics: any; //this prop will have to be removed & extended for each namespace as each package has its own statics type
+  version: string;
+  namespace: NamespaceTypes;
+  nativeModuleName: any; //this prop will have to be removed & extended for each namespace as each package has its own statics type
+  nativeEvents: any; //this prop will have to be removed & extended for each namespace as each package has its own statics type
```
Salakar, line 43: On the previous types we had https://github.com/invertase/react-native-firebase/blob/master/packages/app/lib/index.d.ts#L234 which allowed custom types and statics e.g. auth does: https://github.com/invertase/react-native-firebase/blob/master/packages/auth/lib/index.d.ts#L1690

packages/app/src/internal/constants.ts
Salakar: notifications & indexing too

packages/app/src/version.ts
dackers86: You're correct, added to gitignore. Version `TS` files removed.

tsconfig.json
dackers86: Removed 😄

packages/app/src/types.ts
dackers86: Just clarify, it is ok to delete this `def`?

packages/app/src/internal/registry/namespace.ts
dackers86: We still have some  build errors at the moment so using for reference. 

I assume we are just going to delete the JS folders and repoint the project to the new `dist` folder?

packages/app/package-lock.json
dackers86: Removed package json 👍

packages/app/src/internal/registry/namespace.ts
mikehardy: Yeah, all the libs that use typescript can be used as a pattern, e.g. react-native-device-info, microsoft/reactxp etc - they have `src/` or `lib/` with all the `.ts` files, and then in package.json they point at `dist/` and in the community packages at least bob the builder :construction_worker:  is used in a package.json script to do the transpile

building from source in place then (like a `patch-package` case where a user needs the source to patch) involves a postinstall step *after* patch-package where they will have to run the build script locally after patch-package runs  since we shouldn't commit dist/ we just ship it in releases

PR #5004, opened 2021-03-10
feat(functions): support to set tenantID for firebase auth

xianlinbox:
### Description
When I working on the firebase auth, I found I can't set tenant id which is quite critical for working with Google Identity platform. In the repository discussion channel. there are 2 questions related with this also.

1. https://github.com/invertase/react-native-firebase/discussions/4462
2. https://github.com/invertase/react-native-firebase/discussions/4462

Then  I checked the[ Firebase iOS SDK,](https://github.com/firebase/firebase-ios-sdk/pull/6142) which actually provided tenantId for user to set it. So I created this PR to open this capability to React native Lib users.

### Related issues

No

### Release Summary

Support set tenant it in Firebase auth module.

### Checklist

  - [x] Yes
  - [x] `Android`
  - [x] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan
`yarn tests:jest`

Test Suites: 8 passed, 8 total
Tests:       1 skipped, 94 passed, 95 total
Snapshots:   0 total
Time:        7.393 s, estimated 14 s

---

### DIFF
--- a/packages/auth/__tests__/auth.test.ts
+++ b/packages/auth/__tests__/auth.test.ts
@@ -43,4 +43,13 @@ describe('Auth', function () {
       expect(bar).toEqual(['10.0.2.2', 9099]);
     });
   });
+
+  describe('tenantId', function () {
+    it('should can set tenantId ', function () {
+      const auth = firebase.app().auth();
+      auth.setTenantId('test-id').then(() => {
+        expect(auth.tenantId).toBe('test-id');
+      });
+    });
+  });
 });
--- a/packages/auth/ios/RNFBAuth/RNFBAuthModule.m
+++ b/packages/auth/ios/RNFBAuth/RNFBAuthModule.m
@@ -900,6 +900,15 @@ RCT_EXPORT_METHOD(setLanguageCode:
   
 }
 
+RCT_EXPORT_METHOD(setTenantID:
+  (FIRApp *) firebaseApp
+    :(NSString *) tenantID
+) {
+    if(tenantID){
+        [FIRAuth authWithApp:firebaseApp].tenantID = tenantID;
+    }
+}
+
 RCT_EXPORT_METHOD(useDeviceLanguage:
   (FIRApp *) firebaseApp
 ) {
--- a/packages/auth/lib/index.d.ts
+++ b/packages/auth/lib/index.d.ts
@@ -1202,6 +1202,16 @@ export namespace FirebaseAuthTypes {
    */
   export class Module extends FirebaseModule {
     /**
+     * Returns the current tenant Id.
+     *
+     * #### Example
+     *
+     * ```js
+     * const tenantId = firebase.auth().tenantId;
+     * ```
+     */
+    tenantId: string;
+    /**
      * Returns the current language code.
      *
      * #### Example
@@ -1228,6 +1238,20 @@ export namespace FirebaseAuthTypes {
      * > It is recommended to use {@link auth#onAuthStateChanged} to track whether the user is currently signed in.
      */
     currentUser: User | null;
+
+    /**
+     * Sets the tenant id.
+     *
+     * #### Example
+     *
+     * ```js
+     * await firebase.auth().setTenantId('tenant-123');
+     * ```
+     *
+     * @param tenantId the tenantID current app bind to.
+     */
+    setTenantId(tenantId: string | null): Promise<void>;
+
     /**
      * Sets the language code.
      *
--- a/packages/auth/lib/index.js
+++ b/packages/auth/lib/index.js
@@ -69,6 +69,7 @@ class FirebaseAuthModule extends FirebaseModule {
     this._settings = null;
     this._authResult = false;
     this._languageCode = this.native.APP_LANGUAGE[this.app._name];
+    this._tenantId = null;
 
     if (!this.languageCode) {
       this._languageCode = this.native.APP_LANGUAGE['[DEFAULT]'];
@@ -99,6 +100,10 @@ class FirebaseAuthModule extends FirebaseModule {
 
   get languageCode() {
     return this._languageCode;
+  }
+
+  get tenantId() {
+    return this._tenantId;
   }
 
   get settings() {
@@ -148,6 +153,16 @@ class FirebaseAuthModule extends FirebaseModule {
     } else {
       this._languageCode = code;
     }
+  }
+
+  async setTenantId(tenantId) {
+    if (!isString(tenantId) && !isNull(tenantId)) {
+      throw new Error(
+        "firebase.auth().setTenantId(*) expected 'tenantId' to be a string or null value",
+      );
+    }
+    await this.native.setTenantId(tenantId);
+    this._tenantId = tenantId;
   }
 
   _parseListener(listenerOrObserver) {
--- a/packages/auth/android/src/main/java/io/invertase/firebase/auth/ReactNativeFirebaseAuthModule.java
+++ b/packages/auth/android/src/main/java/io/invertase/firebase/auth/ReactNativeFirebaseAuthModule.java
@@ -1628,6 +1628,19 @@ class ReactNativeFirebaseAuthModule extends ReactNativeFirebaseModule {
   }
 
   /**
+   * setTenantId
+   *
+   * @param appName
+   * @param tenantId
+   */
+  @ReactMethod
+  public void setTenantId(String appName, String tenantId) {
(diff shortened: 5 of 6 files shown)


packages/auth/__tests__/auth.test.ts
```diff
@@ -43,4 +43,29 @@ describe('Auth', function () {
       expect(bar).toEqual(['10.0.2.2', 9099]);
     });
   });
+
+  describe('tenantId', function () {
+    it('should can set tenantId ', function () {
```
mikehardy, line 48:
```suggestion
    it('should be able to set tenantId ', function () {
```

packages/auth/__tests__/auth.test.ts
```diff
@@ -43,4 +43,29 @@ describe('Auth', function () {
...
+      const auth = firebase.app().auth();
+      auth.setTenantId('test-id').then(() => {
+        expect(auth.tenantId).toBe('test-id');
+      });
+    });
+    it('should throw error when set tenantId to null ', function () {
```
mikehardy, line 54:
```suggestion
    it('should throw error if tenantId is null ', function () {
```

packages/auth/lib/index.d.ts
```diff
@@ -1201,6 +1201,16 @@ export namespace FirebaseAuthTypes {
...
+    /**
+     * Returns the current tenant Id.
+     *
+     * #### Example
+     *
+     * ```js
+     * const tenantId = firebase.auth().tenantId;
+     * ```
+     */
+    tenantId: string;
```
mikehardy, line 1213: Can this be null if it has never been set? if so:
```suggestion
    tenantId: string | null;
```

packages/auth/lib/index.d.ts
```diff
@@ -1228,6 +1238,20 @@ export namespace FirebaseAuthTypes {
...
+     *
+     * #### Example
+     *
+     * ```js
+     * await firebase.auth().setTenantId('tenant-123');
+     * ```
+     *
+     * @param tenantId the tenantID current app bind to.
+     */
+    setTenantId(tenantId: string | null): Promise<void>;
```
mikehardy, line 1253: But in the test you wrote, it throws an error if null is sent in? Also the native methods do not take promises, they are synchronous, so I think it should be this (takes string only, returns void)?

```suggestion
    setTenantId(tenantId: string): void;
```

packages/auth/lib/index.js
```diff
@@ -150,6 +155,16 @@ class FirebaseAuthModule extends FirebaseModule {
     }
   }
 
+  async setTenantId(tenantId) {
+    if (!isString(tenantId) && !isNull(tenantId)) {
+      throw new Error(
+        "firebase.auth().setTenantId(*) expected 'tenantId' to be a string or null value",
+      );
+    }
+    await this.native.setTenantId(tenantId);
```
mikehardy, line 164: Hmm - the native methods do not take a Promise as an argument though, do they? They appear to be synchronous. So I'm not sure we should 'await' here if they keep their synchronous API signatures, and this method should not be async

packages/auth/ios/RNFBAuth/RNFBAuthModule.m
```diff
@@ -900,6 +900,15 @@ - (void)invalidate {
   
 }
 
+RCT_EXPORT_METHOD(setTenantID:
+  (FIRApp *) firebaseApp
+    :(NSString *) tenantID
+) {
+    if(tenantID){
+        [FIRAuth authWithApp:firebaseApp].tenantID = tenantID;
+    }
```
mikehardy, line 909: If you are checking for not-null and that it is a valid string in JS, I think you can skip the if(tenantId) here?

```suggestion
    [FIRAuth authWithApp:firebaseApp].tenantID = tenantID;
```

mikehardy (changes requested):
@xianlinbox thank you so much for this! We have had some interest in it, but have not had time to implement it.

I made some comments about the shape of the API as I think it can be fully synchronous based on the underlying native APIs being synchronous.

Also I made some grammar comments but please understand I know how tough it is to speak multiple languages (you should hear me speak Spanish...), I hope you don't mind the suggestions

Finally, it appears the User object also has a read-only tenantId property that you could easily expose, do you think you could add that? It is apparently related to the app tenantId but could be independent (you can change app tenantId after users are associated with another tenantId, and I think multi-tenant users might need to know the user's tenantId)

https://firebase.google.com/docs/reference/swift/firebaseauth/api/reference/Classes/User#tenantid
https://firebase.google.com/docs/reference/android/com/google/firebase/auth/FirebaseUser#public-abstract-string-gettenantid

I think user.tenantId may already be populated in the current user naturally, so it may just be a matter of copying it to the JS object from native and adding it to the types? Worst case, as a user is returned from native methods, you may need to call the tenantId getter in native code, then return it with the rest of the user information?

Finally, I really really appreciate that you added test code along with the change. That helps us so much. 

Cheers

mikehardy:
Maintainer note: I verified that the multi-tenant support was added prior to our current native SDK minimum, this should be good to merge as feature release / non-breaking change once it is in shape

packages/auth/__tests__/auth.test.ts
xianlinbox: agree, updated.

packages/auth/__tests__/auth.test.ts
xianlinbox: Thanks for typescript, we don't allow to call this method with  null. so no test for this scenario any more, deleted it.

packages/auth/lib/index.js
xianlinbox: Here I just following same pattern with LanguageCode, and test to make this one as sync method, Then I met the issue can't find method tenantId.

packages/auth/lib/index.d.ts
```diff
@@ -1201,6 +1201,16 @@ export namespace FirebaseAuthTypes {
    * TODO @salakar missing updateCurrentUser
    */
   export class Module extends FirebaseModule {
+    /**
+     * Returns the current tenant Id.
```
mikehardy, line 1205:
```suggestion
     * Returns the current tenant Id or null if it has never been set
```

packages/auth/lib/index.js
mikehardy: That is a fair point about following setLanguageCode - I accept that, thank you!

mikehardy (commented):
I suggested a different wording of the documentation a little bit to explain why null might be returned, I will commit that one

The only other question I have is about this:

> Finally, it appears the User object also has a read-only tenantId property that you could easily expose, do you think you could add that? It is apparently related to the app tenantId but could be independent (you can change app tenantId after users are associated with another tenantId, and I think multi-tenant users might need to know the user's tenantId)
>
> https://firebase.google.com/docs/reference/swift/firebaseauth/api/reference/Classes/User#tenantid
> https://firebase.google.com/docs/reference/android/com/google/firebase/auth/FirebaseUser#public-abstract-string-gettenantid
>
> I think user.tenantId may already be populated in the current user naturally, so it may just be a matter of copying it to the JS object from native and adding it to the types? Worst case, as a user is returned from native methods, you may need to call the tenantId getter in native code, then return it with the rest of the user information?

mikehardy:
Oh - and the CLA is not signed yet - we will need the CLA signed (https://github.com/invertase/react-native-firebase/pull/5004#issuecomment-794737661), it can merge now or if you have time to add user.tenantId we can merge after that, thanks @xianlinbox !

xianlinbox:
> Oh - and the CLA is not signed yet - we will need the CLA signed ([#5004 (comment)](https://github.com/invertase/react-native-firebase/pull/5004#issuecomment-794737661)), it can merge now or if you have time to add user.tenantId we can merge after that, thanks @xianlinbox !

Thanks for the feedbacks and correct wording. I have clicked the link and signed the CLA. but I don't know why the PR didn't reflect that.
image

mikehardy:
per the comment above it is typically when commit email address is different than github email address, and you have not associated the commit email address with your github account - this happens for some people

> Xianning Liu seems not to be a GitHub user. You need a GitHub account to be able to sign the CLA. If you have already a GitHub account, please add the email address used for this commit to your account.

xianlinbox:
@mikehardy  as I need to updated the commit author to correct my email address,  so I created a new PR (https://github.com/invertase/react-native-firebase/pull/5019) for this change. the code is same as this one. Can we move to that PR?

PR #5047, opened 2021-03-17
feat(crashlytics): flag fatal errors for crashlytics and analytics

mikehardy:
### Description

Crashlytics backend is working on better integration with interpreted languages

This change should work with changes in the backend to transform errors to a style that enables better reporting in both crashlytics and analytics

### Release Summary

The commit title

### Checklist

  - [x] Yes
  - [x] `Android`
  - [x] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

There is a specific `__DEV__` test which means this will never work in dev mode.
Analytics and Crashlytics must both be installed for a full test.
There is no easy way to mock out the functions and verify the attribute is correct

So a manual unhandled javascript crash should be attempted in release mode and then the results should be verified in the crashlytics and analytics consoles on the web I think

---

### DIFF
--- a/packages/crashlytics/lib/handlers.js
+++ b/packages/crashlytics/lib/handlers.js
@@ -15,9 +15,12 @@
  *
  */
 
+import { firebase } from '@react-native-firebase/app';
 import { isError, once } from '@react-native-firebase/app/lib/common';
 import tracking from 'promise/setimmediate/rejection-tracking';
 import StackTrace from 'stacktrace-js';
+
+export const FATAL_FLAG = 'com.firebase.crashlytics.reactnative.fatal';
 
 export function createNativeErrorObj(error, stackFrames, isUnhandledRejection, jsErrorName) {
   const nativeObj = {};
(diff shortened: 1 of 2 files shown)


packages/crashlytics/lib/handlers.js
```diff
@@ -78,6 +81,21 @@ export const setGlobalErrorHandler = once(nativeModule => {
     if (nativeModule.isErrorGenerationOnJSCrashEnabled) {
       try {
         const stackFrames = await StackTrace.fromError(error, { offline: true });
+
+        // Flag the Crashlytics backend that we have a fatal error, they will transform it
+        await nativeModule.setAttribute(FATAL_FLAG, Math.floor(Date.now() / 1000)); // "current UNIX time"
+
+        // Notify analytics, if it exists
+        const analytics = firebase.app().analytics();
```
Salakar, line 89: I think this will throw if its not actually there? I could be wrong though, we used to check NativeModules export from RN instead to check existence of a module to get around the throw

mikehardy:
Observations during testing:

1. **react-native-firebase generates two traces for each JS crash** - one crash, one non-fatal - we hand off to the default react-native handler. Our trace is (was) non-fatal with a JS stack, then the default handler gets it and logs a crash with no JS stack. This was partially addressed by #4904 but I think that went the wrong way and the system in general is deficient in a lot of ways. 

   **react-native-firebase action item, I will raise this separately**

2. **The custom key in crashlytics is invisible** - this may be expected

   **Crashlytics team to confirm custom key should be invisible**

3. **I am unsure of the timestamp format** - it was given as "current Unix time" for crashlytics, and as a Java code reference for analytics. "Unix time" conventionally is *seconds* since 0 UTC 19700101 but the Java code reference seems based on java.util.Date.getTime() which is *milliseconds* in unix epoch. 

   **Crashlytics team to specify the time format unambiguously**. For now I will use milliseconds for both.

4. **Fatals with the flag will likely be logged** It is likely (see 1 above) that a fatal crash with the custom key set will be logged. 

   **Crashlytics to confirm it is okay to over-send the flag - that is: sending the custom key set on a fatal is fine**

5. **FirebaseAnalytics does not let `_ae` through** - at least on Android the firebase-android-sdk will not allow the logging of this event name:

```
03-18 18:51:55.334 11416 11446 E FA      : Name must start with a letter. Type, name: event, _ae
03-18 18:51:55.334 11416 11446 E FA      : Invalid public event name. Event will not be logged (FE): _ae
```

  **FirebaseCrashlytics team to liaise with FirebaseAnalytics to confirm it will pass through in firebase-android-sdk and firebase-ios-sdk or advise other mechanism**


Example of the custom key timestamp format in crashlytics:

Crashlytics_Key_Timestamp_Example

I am unable to provide an example of our analytics event because of the above issue but I confirm I'm sending the same `{ fatal: 1, timestamp: Date.now() + '' }` data through so it should be okay if item 5 is resolved

Salakar:
> 2. **The custom key in crashlytics is invisible** - this may be expected
>    **Crashlytics team to confirm custom key should be invisible**

Invisible is correct.

> 3. **I am unsure of the timestamp format** - it was given as "current Unix time" for crashlytics, and as a Java code reference for analytics. "Unix time" conventionally is _seconds_ since 0 UTC 19700101 but the Java code reference seems based on java.util.Date.getTime() which is _milliseconds_ in unix epoch.

Value should be unix time in seconds - rounded up if converting from milliseconds.

> 4. **Fatals with the flag will likely be logged** It is likely (see 1 above) that a fatal crash with the custom key set will be logged.

This is fine.

> 5. **FirebaseAnalytics does not let `_ae` through** - at least on Android the firebase-android-sdk will not allow the logging of this event name

Can we try `app_exception` as the event name.

mikehardy:
Everything appears to work in the current implementation now except the `app_exception` analytics event.
We are logging the event per request - exactly as requested - but it may need an underlying SDK change for it to work fully. Going to merge as is since fatal transform is correct and exception log is to spec

PR #5113, opened 2021-04-01
fix: Disable Crashlytics in debug mode by default

### LINKED ISSUE #4793: [🐛] Crashlytics - sending crashes in debug mode
I have used crashlytics library and it is sending crashes in debug mode.

According to documentation in debug crashes should not be sent.
https://rnfirebase.io/crashlytics/usage#enable-debug-crash-logs

i have set flag `crashlytics_debug_enabled` to false and not working.

I have seen that in this commit 
https://github.com/invertase/react-native-firebase/commit/cad58e178b43dea461e17fa4a0a3fecd507ba68a
the debug flag code is completely removed and this has broken the functionality.

waqas19921:
### Related issues

Fixes #4793

### Checklist

  - [x] Yes
  - [x] `Android`
  - [x] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

---

### DIFF
--- a/packages/analytics/lib/index.js
+++ b/packages/analytics/lib/index.js
@@ -40,7 +40,7 @@ const ReservedEventNames = [
   'ad_reward',
   'app_background',
   'app_clear_data',
-  'app_exception',
+  // 'app_exception',
   'app_remove',
   'app_store_refund',
   'app_store_subscription_cancel',
--- a/packages/crashlytics/lib/handlers.js
+++ b/packages/crashlytics/lib/handlers.js
@@ -15,9 +15,12 @@
  *
  */
 
+import { firebase } from '@react-native-firebase/app';
 import { isError, once } from '@react-native-firebase/app/lib/common';
 import tracking from 'promise/setimmediate/rejection-tracking';
 import StackTrace from 'stacktrace-js';
+
+export const FATAL_FLAG = 'com.firebase.crashlytics.reactnative.fatal';
 
 export function createNativeErrorObj(error, stackFrames, isUnhandledRejection, jsErrorName) {
   const nativeObj = {};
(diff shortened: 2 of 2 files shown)


packages/crashlytics/ios/RNFBCrashlytics/RNFBCrashlyticsInitProvider.m
```diff
@@ -47,6 +47,13 @@ + (BOOL)isCrashlyticsCollectionEnabled {
     enabled = [RNFBMeta getBooleanValue:KEY_CRASHLYTICS_AUTO_COLLECTION_ENABLED defaultValue:YES];
     DLog(@"isCrashlyticsCollectionEnabled via RNFBMeta: %d", enabled);
   }
+  
+#ifdef DEBUG
+  enabled = NO;
+  if ([[RNFBJSON shared] contains:KEY_CRASHLYTICS_DEBUG_ENABLED]) {
+    enabled = [[RNFBJSON shared] getBooleanValue:KEY_CRASHLYTICS_DEBUG_ENABLED defaultValue:NO];
```
mikehardy, line 54: 🤔 This will override the auto_collection_enabled tests above and is a different logical test than the Java one (which appears correct

At this point enabled might already be no because auto collection is off, and we don't want to turn it on if auto collection is off even if debug collection is on

I think the test needs to be more specific similar to the java test

```suggestion
  if ([[RNFBJSON shared] contains:KEY_CRASHLYTICS_DEBUG_ENABLED] &&
     ![[RNFBJSON shared] getBooleanValue:KEY_CRASHLYTICS_DEBUG_ENABLED defaultValue:NO]) {
    enabled = NO;  
```

I haven't tested this locally - I'm attempting to produce Objective-C code in a web text area so I may not have gotten the syntax / typing correct for the booleans etc but I think it expresses the idea?

mikehardy (changes requested):
@waqas19921 this is great thank you!
I have one comment about a difference in behavior between how you implemented the android vs ios feature - could you check that?

packages/crashlytics/ios/RNFBCrashlytics/RNFBCrashlyticsInitProvider.m
waqas19921: @mikehardy Your solution will not work as it will enable crashlytics by default as user will not specify KEY_CRASHLYTICS_DEBUG_ENABLED by default. In this case crashlytics will automatically be enabled.

mikehardy: I don't understand, I used `&&` and the first part of the clause checks if the key exists. I believe that addresses your concern?

The larger point is the logic needs to be the same Android and your Objective-C solution so far is not. It will potentially overwrite a "wait before collecting / not automatic" false setting with "forget that, we're in debug let's enable it". The settings should have the same precedence behavior on both platforms and on Android if it is set to be not automatic, debug will not enable it. Objective-C should mirror that.

mikehardy: Okay - I took a deeper look into this, and the check for contains key is actually unnecessary in this part of the logic because we aren't trying to have different configuration areas take precedence like the auto init settings above, and the default value you provide will be returned if the JSON does not contain the key

So that simplifies things a little bit

Then to match the android implementation I propose *not* setting the value with any defaults in this specific debug-only block, we just set it to false if it is false, that way if it was true or false before from the auto init settings it will start out staying with that value but switch to false unless specifically turned on in debug

so crashes will not be handled in debug by default and auto init settings won't be overridden

I pushed that change to this branch, and after seeing CI go and reviewing it again - if it all still looks good I will merge

mikehardy:
Oops - mis-handled the push when trying to fix, and it closed this, apologies.
I've carried your commits and my last commit over to #5117 - @waqas19921 could you look there and let me know what you think? I believe the platforms will have consistent behavior now

waqas19921:
@mikehardy Ok no problem.

Issue #4706, opened 2020-12-23
[🐛🔥] NullPointerException Firebase Crashlytics


trizin:
## Issue
Hello! I'm using react native firebase in my app. I've discovered the issue while viewing crashlytics reports. Also some user reported that the app is crashing. The issue occurs on a few amount of devices.

Crashlytics logs:
`Error: Exception in HostObject::get(propName:RNFBUtilsModule): java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.String java.io.File.getAbsolutePath()' on a null object reference, stack: <unknown>@787:648 assign@-1 v@653:1922 N@653:1993 N@-1 createModuleNamespace@653:464 <unknown>@786:1265 v@2:1473 <unknown>@785:220 v@2:1473 <unknown>@403:368 v@2:1473 <unknown>@6:57 v@2:1473 d@2:875 global code@795:3`
```
Fatal Exception: com.facebook.react.common.JavascriptException: 
Error: Exception in HostObject::get(propName:RNFBUtilsModule): java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.String java.io.File.getAbsolutePath()' on a null object reference, stack:
<unknown>@787:648
assign@-1
v@653:1922
N@653:1993
N@-1
createModuleNamespace@653:464
<unknown>@786:1265
v@2:1473
<unknown>@785:220
v@2:1473
<unknown>@403:368
v@2:1473
<unknown>@6:57
v@2:1473
d@2:875
global code@795:3

       at com.facebook.react.modules.core.ExceptionsManagerModule.reportException(ExceptionsManagerModule.java:83)
       at java.lang.reflect.Method.invoke(Method.java)
       at java.lang.reflect.Method.invoke(Method.java:372)
       at com.facebook.react.bridge.JavaMethodWrapper.invoke(JavaMethodWrapper.java:372)
       at com.facebook.react.bridge.JavaModuleWrapper.invoke(JavaModuleWrapper.java:151)
       at com.facebook.react.bridge.queue.NativeRunnable.run(NativeRunnable.java)
       at android.os.Handler.handleCallback(Handler.java:739)
       at android.os.Handler.dispatchMessage(Handler.java:95)
       at com.facebook.react.bridge.queue.MessageQueueThreadHandler.dispatchMessage(MessageQueueThreadHandler.java:27)
       at android.os.Looper.loop(Looper.java:135)
[...]

PR #5118, opened 2021-04-03
fix(android, utils): fix rare crash getting documents directory

LittoCats:
[NullPointerException Firebase Crashlytics](https://github.com/invertase/react-native-firebase/issues/4706)

### Description

The exported constant value for key KEY_DOCUMENT_DIRECTORY will be missed in ReactNativeFirebaseUtilsModule when Build.VERSION.SDK_INT >= Build.VERSION_CODES.KITKAT and  externalDirectory == null

### Related issues

Fixes [#4706](https://github.com/invertase/react-native-firebase/issues/4706)

### Checklist

  - [x] Yes
  - [x] `Android`
  - [ ] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

---

### DIFF
--- a/packages/app/android/src/reactnative/java/io/invertase/firebase/utils/ReactNativeFirebaseUtilsModule.java
+++ b/packages/app/android/src/reactnative/java/io/invertase/firebase/utils/ReactNativeFirebaseUtilsModule.java
@@ -159,10 +159,16 @@ public class ReactNativeFirebaseUtilsModule extends ReactNativeFirebaseModule {
     constants.put(KEY_TEMP_DIRECTORY, context.getCacheDir().getAbsolutePath());
     constants.put(KEY_CACHE_DIRECTORY, context.getCacheDir().getAbsolutePath());
 
+    File externalDirectory = context.getExternalFilesDir(null);
     if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.KITKAT) {
-      File externalDirectory = context.getExternalFilesDir(null);
       if (externalDirectory != null) {
         constants.put(KEY_DOCUMENT_DIRECTORY, externalDirectory.getAbsolutePath());
+      } else {
+        // The external directory may be null if it is truly external *and*
+        // the device's external storage environment changes. We will use the regular
+        // Files directory as a backup and note in the documentation that the directory may
+        // vary under rare conditions
+        constants.put(KEY_DOCUMENT_DIRECTORY, context.getFilesDir().getAbsolutePath());
       }
     }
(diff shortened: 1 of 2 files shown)


mikehardy:
Hi @LittoCats ! I think this was already addressed in #4990 - when I look at the original diagnosis you did it seems correct but it's based on an old version of the code. In v11.1.0+ It already has a null check and so will not crash here

LittoCats:
Yes, I notice that last night, but I found, the constant value exported to js will be not exists for KEY_DOCUMENT_DIRECTORY when `Build.VERSION.SDK_INT >= Build.VERSION_CODES.KITKAT and externalDirectory == null`

```
    File externalDirectory = context.getExternalFilesDir(null);
    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.KITKAT) {
      if ( externalDirectory != null ) {
        constants.put(KEY_DOCUMENT_DIRECTORY, externalDirectory.getAbsolutePath());
      }
    } else {
      constants.put(KEY_DOCUMENT_DIRECTORY, context.getFilesDir().getAbsolutePath());
    }
```

mikehardy:
That's true and I thought of that as I compared the two types of fix for the issue. Your fix will cause the directory to change between runs though, so the value will exist in your version but it won't be stable as external media is changed.

I think that may just be an unsolvable problem though - maybe something just to put in the docs here vs attempt to solve in code

The doc is here https://github.com/invertase/react-native-firebase/blob/c94546d8127606dca5bfd09ef92ec32eec333f19/packages/app/lib/index.d.ts#L285 and if there was just a note that there is a small chance if a user uses removable external storage on Android and that storage configuration changes or is not available, the path returned could change, then this seems fine.

LittoCats:
Thanks for your explaination.

mikehardy:
Wait! @LittoCats I actually like this PR, I think the code itself is both correct and necessary to correct the null pointer exception - I am sorry I wasn't clear, what I think it needs is just an addition to the documentation *as well as the code fix*, to warn developers that this is an external directory which *could change* (making it appear as if documents appear or disappear) *if the users external storage environment changes*

Hopefully that's more clear :-). I actually really want to get this fix in, you have solved a real crash!

mikehardy:
@LittoCats I had time today to look at this - I pulled your branch, fixed the compile error and then altered the documentation to explain the conditions when documents directory could be internal vs external. 

I think this should work and fix the crash - can you take a look? I can merge it and release it if it looks good

PR #5143, opened 2021-04-09
docs(auth, social): Update google sign in docs

rawatnaresh:
### Description

I was trying out google signing with a fresh new project, pasted everything from the docs but it didn't work out. Thought the issue is because of the react-native-firebase but later found out I should be using `@react-native-google-signin/google-signin` instead of `react-native-google-signin` if `RN >= 0.60`.

### DIFF
--- a/docs/auth/social-auth.md
+++ b/docs/auth/social-auth.md
@@ -199,7 +199,7 @@ with the new authentication state of the user.
 
 ## Google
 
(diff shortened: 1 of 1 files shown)


mikehardy (approved):
Nice thank you @rawatnaresh - that package moved somewhat recently and the pointer was stale yes. I can merge this as soon as you sign the CLA - there are details on how to make that work next to the failing CLA check or on the CLA-bot comment above.

rawatnaresh:
@mikehardy done! 👍

PR #5206, opened 2021-04-23
fix(android): Resolve a crash under Android 7

ifsnow:
### Description

Related to : #4981 

[TaskExecutorService](https://github.com/invertase/react-native-firebase/blob/v11.3.2/packages/app/android/src/main/java/io/invertase/firebase/common/TaskExecutorService.java#L104) is using API level 24 Collection functions such as `removeIf`, `forEach`.
This can cause build failure or crash under Android 7.

I'm sure it needs to be modified to work with the lower minSdkVersion.

### Checklist

  - [X] Yes
  - [ ] `Android`
  - [ ] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [X] No

### Test Plan

None

### DIFF
--- a/packages/app/android/src/main/java/io/invertase/firebase/common/TaskExecutorService.java
+++ b/packages/app/android/src/main/java/io/invertase/firebase/common/TaskExecutorService.java
@@ -101,12 +101,13 @@ public class TaskExecutorService {
 
   public void shutdown() {
     Set<String> existingExecutorNames = executors.keySet();
-    existingExecutorNames.removeIf((executorName) -> {
-      return executorName.startsWith(name) == false;
-    });
-    existingExecutorNames.forEach((executorName) -> {
-      removeExecutor(executorName);
-    });
+    for (String executorName : existingExecutorNames) {
+      if (executorName.startsWith(name) == false) {
+        existingExecutorNames.remove(executorName);
+      } else {
+        removeExecutor(executorName);  
+      }
+    }
   }
 
   public void removeExecutor(String executorName) {


mikehardy:
Oh my! Thank you for letting us know, this is a new feature and is a great performance enhancement - when it's working - but we have no intention of requiring minSdkVersion 24 - I'll check into this and sorry for the bother. The releases prior to this (11.2.x) have no problems with them, it should be fine to revert until we've fixed

mikehardy:
Having tried this in another project I will likely go another direction, it has worked well for us:

https://github.com/ankidroid/Anki-Android/pull/8361/files

https://developer.android.com/studio/write/java8-support-table
- removeIf is supported
- forEach is supported

It has not appreciably increased build times either

A related failure (obviously) is that this was not caught and stopped in our CI, I'll alter our CI to use our minSdkVersion for the Android emulator (or the min possible - not sure 16 is still a reliable emulator, but something under 24 is, 21 at least I know works) vs the version it is on now

ifsnow:
@mikehardy I don't think it's the best way to support just 2 functions.

That's because using `desugaring` increases the App file size by 120kb. (APK file - release build)

If it is not necessary for the whole project and is not used frequently, you need to consider whether it is necessary to use it.

It's just my opinion.

mikehardy:
@ifsnow that's a reasonable point, I understand the internal storage constraints of a lot of devices and am sensitive to that, plus we're a library not an app so we would be effectively making this decision (programming clarity vs apk size) for all consumers vs leaving it to them.

Your code looks correct on first review and I'll look into the CI failure on the PR (likely unrelated), and merge it as is if it passes though. I want to alter our CI to reproduce the failure first though (API21 is available for testing in CI, as an emulator) so this can't happen again. Errors like this should never get past CI.

mikehardy:
Still working through CI flakiness so I can't merge this just yet, and I'll be traveling next 5hrs so this will likely be merged released overnight your time. Just giving you the update since it's nearly end of day there and this is a pretty serious issue. It'll be fixed by your next day though, thanks again for reporting this and posting the fix, I really appreciate it.

mikehardy:
v11.3.3 just went out with this fix, and I merged a CI change that will make sure it never happens again!

PR #5208, opened 2021-04-23
test(android, lint): enable java lint check in CI

mikehardy:
### Description

If we say we support minSdkVersion 16 we should check it somehow

It turns out that getting a non-API30 emulator is somewhat difficult in CI, but...

Java lint was not being checked anywhere, and the NewApi check in particular
would have stopped a crash. So add full lint configuration tuned for current status

Cannot enable it on the modules or our modules may fail library-consumer's builds,
but we may configure it on the sub-modules from our test app then check that

Tested and working and would have correctly failed the NewApi violation shown in #5206

### Related issues

Related #5206 - Android 7 crash from use of Java8 features in #4981

### Release Summary

Conventional commit message

### Checklist

  - [x] Yes
  - [x] `Android`
  - [ ] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

The change is literally a test :-)

---


mikehardy:
API21 emulator is *not* easily attainable on github CI runners, probably have to do a local update of SDK tools to get it's definition, and may need to install API21 API as well (unsure) but I did not have success with Detox getting it to run unfortunately, so this needs a lot more investigation

Alternative (or additive) solution for this is to enable a lintRelease check in CI that checks for `NewApi` (if I recall correctly). This one *should* be done no matter what but it turns out we are specifically configured the opposite to what I would expect right now, e.g. in the app module:

https://github.com/invertase/react-native-firebase/blob/1510502f9fd5509c102d29b343418e6363b951ad/packages/app/android/build.gradle#L58-L61

Instead I think we should have a common lint.gradle and all the modules should refer to it, with settings for NewApi to fail - when I toggle abortOnError to true in the file referenced above, this is correctly flagged as a build failure.

mikehardy:
The most important part - and the only part I changed - was to add lintDebug to the android e2e run, and I verified the log it was perfect. Merging

PR #5262, opened 2021-05-04
fix(ios): admob shows only non-personalized ads

dylancom:
### Description

On iOS the actual value of requestNonPersonalizedAdsOnly is never checked (it just checks if the key was set), which results in AdMob showing only non-personalized ads.
On Android it's already checking for the actual Boolean value of requestNonPersonalizedAdsOnly.

### Checklist

  - [x] Yes
  - [ ] `Android`
  - [x] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No


mikehardy (approved):
Looks like a :facepalm: issue. Nice catch

mikehardy:
iOS E2E is not working because it's based on an old version of my buildcache work #5239 before I used tokens to fetch the release in #5257 which I merged after this forked - I pulled it locally and it's :100: - merging - thanks again!

Issue #5087, opened 2021-03-26
🔥 Cloud Firestore Transactions .then is not a function


TPXP:
## Issue
Hello there, thanks again for this great project <3

I'm using a cloud firestore transaction in my app to handle a write requiring document data to match the one of another document in the database. For this purpose, we're using a transaction.

According to the [example usage shown in RNFirebase's documentation](https://rnfirebase.io/firestore/usage#transactions), this should work:
```
    await transaction.update(postReference, {
      likes: postSnapshot.data().likes + 1,
    });
```

Which matches with the [official documentation for the web SDK](https://firebase.google.com/docs/firestore/manage-data/transactions), all good so far!

Except, when trying to run this snippet, the JS engine gets confused and tries to do something with `.then()`, which leads to an error.

```
 [TypeError: transaction.update(postReference, {Data}).then is not a function.
``` 

---

### Javascript

<details><summary>Click To Expand</summary>
<p>

#### `package.json`:
(relevant part only)
```json
{
  "dependencies": {
    "@react-native-firebase/analytics": "^10.2.0",
    "@react-native-firebase/app": "^10.2.0",
    "@react-native-firebase/auth": "^10.2.0",
    "@react-native-firebase/crashlytics": "^10.2.0",
    "@react-native-firebase/firestore": "^10.2.0",
    "@react-native-firebase/functions": "^10.2.0",
    "@react-native-firebase/messaging": "^10.2.0",
    "@react-native-firebase/remote-config": "^10.2.0",
    "@react-native-firebase/storage": "^10.2.0"
  }
}
```

Outdated, I know. Still, I looked at the changelog and I believe the relevant code has not changed.

#### `firebase.json` for react-native-firebase v6:

```json
{ "react-native": {
  "crashlytics_debug_enabled": true
} }
```

</details>

### iOS

<details><summary>Click To Expand</summary>
<p>

#### `ios/Podfile`:

- [ ] I'm not using Pods
- [x] I'm using Pods and my Podfile looks like:

```ruby
platform :ios, '10.0'

require_relative '../node_modules/@react-native-community/cli-platform-ios/native_modules'
require_relative '../node_modules/react-native-unimodules/cocoapods'
require_relative '../node_modules/react-native/scripts/react_native_pods'

target 'MyApp' do
  config = use_native_modules!
  use_unimodules!

  use_react_native!(:path => config["reactNativePath"])

  # Enables Flipper - not for CI builds as it takes some build time
  #
[...]

PR #5273, opened 2021-05-06
docs(firestore): remove await before transaction.update

TPXP:
### Description

`transaction.set` does not return a promise, there is no need to `await` it. This could lead to a confusion among users since `await` usually indicates a promise.

### Related issues
#5087 

### Release Summary
doc(firestore): remove await before transaction.set

### Checklist

  - [x] Yes
  - [ ] `Android`
  - [ ] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan
This is a doc-only change, no need for testing

### DIFF
--- a/docs/firestore/usage/index.md
+++ b/docs/firestore/usage/index.md
@@ -546,7 +546,7 @@ function onPostLike(postId) {
       throw 'Post does not exist!';
     }
 
-    await transaction.update(postReference, {
+    transaction.update(postReference, {
       likes: postSnapshot.data().likes + 1,
     });
   });


mikehardy (approved):
You are right - good catch - upstream doc ref here https://firebase.google.com/docs/reference/js/firebase.firestore.Transaction#update - and I agree with you, if I see await, I'm thinking promise, but this one actually should return the transaction object again for chaining - totally different direction and would be quite unexpected with await in the doc

PR #5328, opened 2021-05-17
fix: add database get method

nes123:
### Description

If persistence is enabled SingleEvent Snapshot will read from the local database. Very non-intuitive and making migration of an online to offline somehow complicated.

A good solution to this problem will be to use the new method "get" in the SDK. This method gets the server values for the query. Updates the cache and raises events if successful. If not connected, falls back to a locally-cached value.

This is the required behavior in most cases.

### Related issues

Fixes #724012604

### Checklist

  Yes
  `Android`
- This is a breaking change;
   No

### DIFF
--- a/packages/database/android/src/reactnative/java/io/invertase/firebase/database/ReactNativeFirebaseDatabaseQuery.java
+++ b/packages/database/android/src/reactnative/java/io/invertase/firebase/database/ReactNativeFirebaseDatabaseQuery.java
@@ -53,6 +53,14 @@ public class ReactNativeFirebaseDatabaseQuery {
       }
     }
   }
+  
+  /**
+   * Gets the server values for this query. Updates the cache and raises events if successful. If
+   * not connected, falls back to a locally-cached value.
+   */
+  public void get() {
+    query.get();
+  }
 
   /**
    * Attaches a single single value event listener to the query
--- a/packages/database/lib/index.d.ts
+++ b/packages/database/lib/index.d.ts
@@ -691,6 +691,12 @@ export namespace FirebaseDatabaseTypes {
       failureCallbackContext?: ((a: Error) => void) | Record<string, any> | null,
     ): Promise<DataSnapshot>;
 
+    
+    /**
+     * Get the server-value for this query, or return a cached value if not connected.
+     */
+    get(): Promise<DataSnapshot>;
+  
     /**
      * Generates a new `Query` object ordered by the specified child key.
      *


mikehardy:
Hey! Fantastic @nes123 - I can work with you to get a feature going.

First step is to get your local environment set up like the https://github.com/invertase/react-native-firebase/blob/master/tests/README.md 

before you make any changes, just running from the current master branch, you should be able to successfully run both the android and ios e2e tests successfully. Occasionally it flakes, but with a couple re-runs it should run, that means you have your development environment set up correctly for testing.

Then step one is to locate the e2e test that is closest to your area, probably in this directory but I would make a new file called `get.e2e.js` and start by copying the contents of here https://github.com/invertase/react-native-firebase/blob/master/packages/database/e2e/query/query.e2e.js in but change `describe(` at the top to `describe.only(` then run `yarn tests:android:test-reuse` and `yarn tests:ios:test-reuse`

What's the point? Now you have a little focused test harness set up and you can reproducibly re-run it pretty quickly.

Then adding an API is like this:

1- add the API in `packages/database/index.d.ts` to match the firebase-js-sdk interface *exactly*
2- add the API in `packages/database/index.js` that implements the index.d.ts interface but just console log for now, and add a call to it as the first real test in `packages/database/e2e/query/get.e2e/js` and you should see it execute and log your message.
3 - now do either Java or iOS first - adding a new react method like some of the others, but just have it Promise.reject at first, and hook that up to the index.js - re-test but now that you have made native changes use `yarn test:android:test` (not `-reuse`) or the same for ios and you should see the javascript e2e test show an error with your rejection error message from native
4- now you have a native API wired up to javascript with a test that hits it. Once all that plumbing is done, you have to do the actual native implementation. For there I usually look at the upstream reference docs for firebase-ios-sdk or firebase-android-sdk and go from there

Most of the steps above are fairly mechanical so once you have an actual test going from JS to native and back on both platforms definitely commit again and push to this PR for a quick check - that will make sure that you have the structure right, then we may focus on the actual implementation

nes123:
When trying to run this command on my mac: brew tap wix/brew

i get the following error:

Invalid formula: /usr/local/Homebrew/Library/Taps/wix/homebrew-brew/applesimutils.rb

I wasted two hours on that without luck. Any ideas?

nes123:
I wasted a lot of time on that but setting up the testing environment does not work for me. I will try to get back to it in few weeks

mikehardy:
Oh that's frustrating - I'm sorry, that is supposed to be a trivial step, not a hard one. You can see in our Actions runs for the iOS E2E how everything works, including the CI runner's versions of everything - here's the latest example - maybe comparing will help? https://github.com/invertase/react-native-firebase/runs/2618565533?check_suite_focus=true

mikehardy:
No problem reopening this later if/when you get time

PR #5473, opened 2021-07-02
feat: Include actionIdentifier on the notification data sent to RN when on iOS (closes #4883)

nabilfreeman:
## Description

Hi there, as discussed in #4883 , for situations where the developer has decided to add Quick Actions directly onto a remote notification there is no way to determine if a notification was tapped/swiped or an action button was pressed in React Native Firebase.

Motivations for doing this range from code simplicity to inconsistent handling of background pushes due to iOS performance management using the silent remote / rich local pattern.

The information is readily available on the native side and can be passed to RN by adding a single property to the NSDictionary.

This would not be a breaking change.

When a quick action is pressed, it triggers the `FirebaseMessaging().onNotificationOpenedApp` handler.

An example of how this could be handled by a user:

```
// open handler
notification => {
    switch(notification.actionIdentifier) {
        case 'YES':
        case 'NO':
            // a quick action was pressed
        default:
            // unrecognised action, or the notification was opened in the normal way - fall back to normal behaviour
    }
}
```
### Related issues

#4883 - originally proposed by @TNicholson11 and @mikehardy 

### Release Summary

If you choose to register quick actions for specific notification categories on iOS, you will now be able to detect when these are pressed with the `.actionIdentifier` key in your `FirebaseMessaging().onNotificationOpenedApp` handler.

### Checklist

  - [x] Yes
  - [ ] `Android`
  - [x] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

IMG_4131
IMG_4128
IMG_4129
IMG_4130

IMG_4132
IMG_4133

---

### DIFF
--- a/packages/messaging/ios/RNFBMessaging/RNFBMessaging+UNUserNotificationCenter.m
+++ b/packages/messaging/ios/RNFBMessaging/RNFBMessaging+UNUserNotificationCenter.m
@@ -98,6 +98,7 @@ struct {
     // if there is an actionIdentifier on this notification, also add that in.
     [mergedDict setObject:response.actionIdentifier forKey:@"actionIdentifier"];
 
+    // send payload to JS space
     [[RNFBRCTEventEmitter shared] sendEventWithName:@"messaging_notification_opened" body:mergedDict];
     _initialNotification = notificationDict;
   }
--- a/packages/messaging/ios/RNFBMessaging/RNFBMessaging+UNUserNotificationCenter.m
+++ b/packages/messaging/ios/RNFBMessaging/RNFBMessaging+UNUserNotificationCenter.m
@@ -88,7 +88,17 @@ struct {
   NSDictionary *remoteNotification = response.notification.request.content.userInfo;
   if (remoteNotification[@"gcm.message_id"]) {
     NSDictionary *notificationDict = [RNFBMessagingSerializer remoteMessageUserInfoToDict:remoteNotification];
-    [[RNFBRCTEventEmitter shared] sendEventWithName:@"messaging_notification_opened" body:notificationDict];
+
+    // create a merged dictionary that will hold the serialized userInfo data, as well as any actionIdentifier
+    NSMutableDictionary * mergedDict = [NSMutableDictionary dictionary];
+
+    // copy the values of notificationDict into mergedDict
+    [mergedDict addEntriesFromDictionary:notificationDict];
+
+    // if there is an actionIdentifier on this notification, also add that in.
+    [mergedDict setObject:response.actionIdentifier forKey:@"actionIdentifier"];
+
+    [[RNFBRCTEventEmitter shared] sendEventWithName:@"messaging_notification_opened" body:mergedDict];
     _initialNotification = notificationDict;
   }
 


nabilfreeman:
If a user wanted to add quick actions to notification cateory, they could use this code:

AppDelegate.m
```
{
  ...
  [self registerActionableNotifications];
  ...
}

NSString * const NotificationCategoryIdentifier  = @"STATUS_CHALLENGE";
NSString * const NotificationYesActionIdentifier = @"YES";
NSString * const NotificationNoActionIdentifier  = @"NO";

- (void)registerActionableNotifications {
    NSLog(@"Setting up Yes/No actions for matching push notifications...");
  
    UIMutableUserNotificationAction *yesAction;
    yesAction = [[UIMutableUserNotificationAction alloc] init];
    [yesAction setActivationMode:UIUserNotificationActivationModeBackground];
    [yesAction setTitle:@"Yes"];
    [yesAction setIdentifier:NotificationYesActionIdentifier];
    [yesAction setDestructive:NO];
    [yesAction setAuthenticationRequired:NO];

    UIMutableUserNotificationAction *noAction;
    noAction = [[UIMutableUserNotificationAction alloc] init];
    [noAction setActivationMode:UIUserNotificationActivationModeBackground];
    [noAction setTitle:@"No"];
    [noAction setIdentifier:NotificationNoActionIdentifier];
    [noAction setDestructive:NO];
    [noAction setAuthenticationRequired:NO];

    UIMutableUserNotificationCategory *actionCategory;
    actionCategory = [[UIMutableUserNotificationCategory alloc] init];
    [actionCategory setIdentifier:NotificationCategoryIdentifier];
    [actionCategory setActions:@[yesAction, noAction]
                    forContext:UIUserNotificationActionContextDefault];

    NSSet *categories = [NSSet setWithObject:actionCategory];
    UIUserNotificationType types = (UIUserNotificationTypeAlert|
                                    UIUserNotificationTypeSound|
                                    UIUserNotificationTypeBadge);

    UIUserNotificationSettings *settings;
    settings = [UIUserNotificationSettings settingsForTypes:types
                                                 categories:categories];

    [[UIApplication sharedApplication] registerUserNotificationSettings:settings];
}
```

App.js
```
componentDidMount() {
  this.removeOnNotificationOpened = FirebaseMessaging().onNotificationOpenedApp(
    event => this.onNotificationOpened(event),
  );
}

onNotificationOpened(event) {
  // title, data, actionIdentifier etc.
}
```

The above code is confirmed working in iOS 14.5.

packages/messaging/ios/RNFBMessaging/RNFBMessaging+UNUserNotificationCenter.m
```diff
@@ -88,7 +88,17 @@ - (void)userNotificationCenter:(UNUserNotificationCenter *)center didReceiveNoti
   NSDictionary *remoteNotification = response.notification.request.content.userInfo;
   if (remoteNotification[@"gcm.message_id"]) {
     NSDictionary *notificationDict = [RNFBMessagingSerializer remoteMessageUserInfoToDict:remoteNotification];
-    [[RNFBRCTEventEmitter shared] sendEventWithName:@"messaging_notification_opened" body:notificationDict];
+
+    // create a merged dictionary that will hold the serialized userInfo data, as well as any actionIdentifier
+    NSMutableDictionary * mergedDict = [NSMutableDictionary dictionary];
```
mikehardy, line 93: I wonder - and this could be me just not understanding Objective-C - do we need to dealloc this dictionary anywhere / somehow? Is it already handled or is it automagically cleaned up?
I have a feeling that it will be orphaned and result in a memory leak

mikehardy (commented):
This is really useful - thank you for proposing this PR!
I had one question mostly about memory management but I'm not sure if memory was handled correctly to begin with. I'm curious for your thoughts though.
Regardless, enabling the CI run now and we'll see how that goes

packages/messaging/ios/RNFBMessaging/RNFBMessaging+UNUserNotificationCenter.m
```diff
@@ -88,7 +88,18 @@ - (void)userNotificationCenter:(UNUserNotificationCenter *)center didReceiveNoti
...
+
+    // copy the values of notificationDict into mergedDict
+    [mergedDict addEntriesFromDictionary:notificationDict];
+
+    // if there is an actionIdentifier on this notification, also add that in.
+    [mergedDict setObject:response.actionIdentifier forKey:@"actionIdentifier"];
```
mikehardy, line 99: I think this needs an addition to the RemoteMessage types so that actionIdentifier shows up here?

https://rnfirebase.io/reference/messaging/remotemessage#properties

https://github.com/invertase/react-native-firebase/blob/daaa72d82df6b9e5e2c1247c10792d4b12683541/packages/messaging/lib/index.d.ts#L69

mikehardy (changes requested):
Actually (still thinking through it - sorry) the types definitely need an update to include this

packages/messaging/ios/RNFBMessaging/RNFBMessaging+UNUserNotificationCenter.m
nabilfreeman: Funny I wondered the same thing, not sure what happens to the other `NSDictionary` instances and if there is any mem difference between `NSDictionary` and `NSMutableDictionary`.

There are alternatives to cloning this dict but I decided they were worse for code quality:
1) Modify `[RNFBMessagingSerializer remoteMessageUserInfoToDict:remoteNotification];` to take another argument for `actionIdentifier` (they are siblings in the parent object). Decided against because I would be butchering a very specifically named function to do something else, and possibly break the larger fn that parses the notification.
2) Make `[RNFBMessagingSerializer remoteMessageUserInfoToDict:remoteNotification];` return an NSMutableDict so I can add the `actionIdentifier` in without cloning. Almost did this but again, really hesitant to modify that other function as it is quite an important one for other handlers.

I think I could dealloc the `mergedDict` after firing `sendEventWithName`?

Also think `_initialNotification` needs to have `mergedDict` instead of `notificationDict` for cold starts to work

packages/messaging/ios/RNFBMessaging/RNFBMessaging+UNUserNotificationCenter.m
nabilfreeman: Ok I will sort this!

packages/messaging/ios/RNFBMessaging/RNFBMessaging+UNUserNotificationCenter.m
mikehardy: > I think I could dealloc the mergedDict after firing sendEventWithName?

This idea scares me as I think the JS side (being async) could consume it at some indeterminate future time so we can't safely dealloc until we *know* the JS side consumed.

For that reason I'm leaning towards merging even without resolving this, and lodging a separate issue to investigate (effectively peeling this discussion off and punting on it for the moment)

> Also think _initialNotification needs to have mergedDict instead of notificationDict for cold starts to work

You've definitely got the use case here and a test app to exercise, I'll defer to your thinking here

packages/messaging/ios/RNFBMessaging/RNFBMessaging+UNUserNotificationCenter.m
```diff
@@ -88,7 +88,18 @@ - (void)userNotificationCenter:(UNUserNotificationCenter *)center didReceiveNoti
...
+
```
TNicholson11, line 100: First, thanks very much for doing this! Our team literally just finalized Q3 planning which included investigating just this type of PR!

As an aside, would you also consider including the [TextInputNotificationResponse](https://developer.apple.com/documentation/usernotifications/untextinputnotificationresponse) for completeness? This is for notifications that are configured to receive user text input e.g. a reply message to an incoming notification.

Something to the effect of:
```
  if ([response isKindOfClass:UNTextInputNotificationResponse.class]) {
    [mergedDict setValue:[(UNTextInputNotificationResponse *)response userText] forKey:@"userInput"];
  }
```

If not, I totally understand. We can submit a supplemental PR (or consider using patch-package, etc)

Thanks again!

mikehardy: ...along with the same wish for a change to the index.d.ts for that key as well :-), but I'd merge that yes.
@TNicholson11 - any thoughts on an android implementation for parity? or is it not needed

TNicholson11: Happy to be wrong on this one (since it would be very helpful for our product) but as best I can tell, you cannot set Android actions on a remote message therefore there would be no response to pass along. On iOS, the notification categories are pre-registered so the iOS does it's magic on the incoming remote message. On Android, that is seemingly only supported for local notifications via Notifee (for example).

The response from a local notification on Android is included in the clickAction property of the [android payload](https://rnfirebase.io/reference/messaging/notification#android)

Long way of saying, to the best of my knowledge there isn't an Android solution required at this time :)

nabilfreeman: Yes I can back that up @TNicholon11, Android can only present action buttons via the silent remote / rich local flow. Which is bad news for us (because we need Quick Actions on our Android app), but good news for this PR 😅

I'll do the type definition for both `actionIdentifier` and `userText` and push here, if I have trouble getting a prototype of text input to work I'll post here

mikehardy: On the plus side, Android data-only delivery is reliable in my experience? Well, as long as the vendor hasn't destroyed it (https://dontkillmyapp.com) but in general I have no issues on Android.

packages/messaging/ios/RNFBMessaging/RNFBMessaging+UNUserNotificationCenter.m
mikehardy: Gentle comment: this seems like an unresolved thought here

packages/messaging/ios/RNFBMessaging/RNFBMessaging+UNUserNotificationCenter.m
mikehardy: gentle reminder: this needs a type update here

mikehardy:
abandoned, but always happy to reopen and collaborate to get this in if it's a need

PR #5480, opened 2021-07-05
feat: Added Expo config plugin

barthap:
### Description

Expo released [development clients](https://docs.expo.io/clients/introduction/) and rolled out an interface called [config plugins](https://docs.expo.io/guides/config-plugins/). These allow users to add native modules which aren't included in Expo Go.

This PR adds a config plugin, which allows users for painless setup of RNFB, without touching any native code. Until now, I've been maintaining the plugin as a separate npm package: [`with-rn-firebase`](https://www.npmjs.com/package/with-rn-firebase) and had some positive feedback. Now it's time to integrate it into the upstream library 😄  For more context, see the discussion at #5386. 

Basically I've copied my code from https://github.com/barthap/with-rn-firebase, but I've split it into separate plugins for each affected package to avoid additional config like setting `enableCrashlytics` etc.

**RNFB repo related changes:**
- Added `build:plugin` and `lint:plugin` npm scripts for affected packages
- The `build:plugin` is run during `npm prepare` phase
- Added `plugin/__tests__` directories with some basic tests, they work when running `npm run tests:jest`
- I had to modify global `tsconfig.json` to add `esModuleInterop: true` - otherwise the plugin tests failed. I haven't seen any regressions and hope it's not breaking - please correct me if it is.
- Added dependency `@expo/config-plugins` and devDependency with plugin tsconfig base.
- Added `packages/**/plugin/build` dir to eslint and prettier ignore - it's a generated directory
- Added some values to `sdkVersions` section in `app/package.json`

For the plugin to work, the package should have a `app.plugin.js` file in its root folder - in this case it only delegates to the `plugin/build` directory. The `build` directory is gitignored and generated during `npm build:plugin` from TypeScript files inside `plugin/src`. 

#### The plugins

As for now, to reflect current `with-rn-firebase` functionality, I've added plugins for 3 packages:
- **App**:
   Its purpose is to copy `google-services.json` and `GoogleServices-Info.plist` into correct places, modify `AppDelegate.m`, and add Gradle dependency and plugin - basically the installation steps from the homepage.
  
   The Google Services files by default are copied from the paths specified by `expo.ios|android.googleServicesFile` in `app.json`.
- **Crashlytics**:
   Installs the required Gradle dependencies for Android. It doesn't yet support the optional release NDK installation step.
- **Performance Monitoring**:
   Installs the required Gradle dependency for Android.

#### Usage / How it works

When using Expo, replace the `yarn add`/`npm install` with `expo install`:
```
expo install @react-native-firebase/app
expo install @react-native-firebase/crashlytics
```
Then in the `app.json` the `plugins` section should be updated automatically. When not using `expo install`, you have to add the plugins on your own`.

Remember to configure the `expo.android.googleServicesFile` and `expo.ios.googleServicesFile` in `app.json`. The paths are relative to the project root.

The `app.json` file may look like this (non-related parts omitted):
```json
{
  "expo": {
    "android": {
      "googleServicesFile": "./google-services.json"
    },
    "ios": {
       "googleServicesFile": "./GoogleService-Info.plist"
    },
    "plugins": [
      "@react-native-firebase/app",
      "@react-native-firebase/crashlytics"
    ]
  }
}
```

#### Further steps

After this PR is merged:

- Update the [docs `Expo` section](https://rnfirebase.io/#expo) and possibly the installation steps for the packages: `app`, `perf`, `crashlytics`

These were not included in my original plugin, they're not implemented yet:
- Investigate the need of config plugins for the `FCM` module (see the discussion in https://github.com/barthap/with-rn-firebase/issues/6)
- Add option to install NDK Crashlytics in release build
- Investigate the need of adding plugin for other packages. Afaik there are some configuration options requiring the `Podfile` changes, which are not trivial for config plugins.

### Related issues

Discussion #5386 

### Release Summary

- **app, crashlytics, perf**: added Expo config plugins

### Checklist

  - [x] Yes
  - [x] `Android`
  - [x] `iOS`
- My change includes tests;
  - [x] `jest` tests added or updated in `packages/\*\*/__tests__` (_added tests for config plugins_)
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

Steps that anyone should be able to reproduce:

1. Installed npm RNFB packages to a separate directory and applied my plugin changes - I've attached the packages to this post:
   [patched-packages.zip](https://github.com/invertase/react-native-firebase/files/6782106/patched-packages.zip)
2. In another directory `expo init app_name_here` - selected managed template
   > You can use `npx crna -t with-dev-client` instead to try it out with [Expo development client](https://docs.expo.io/clients/introduction/)
3. Run `expo prebuild` on this project
4. `git commit` your changes
5. `yarn link` all 3 RNFB packages into Expo app project
4. Add `plugins` section to `app.json`. If you have [Expo Tools](https://marketplace.visualstudio.com/items?itemName=byCedric.vscode-expo) VSCode extension, the intellisense should propose them automatically 
   Screenshot 2021-07-08 at 08 32 56
5. Copy or create blank `google-services.json` and `GoogleService-Info.plist` (e.g. in app root dir) and specify the `expo.android.googleServicesFile` and `expo.ios.googleServicesFile` fields in `app.json`.
6. `git commit` again to have a checkpoint to easy reset
7. Run `expo prebuild` and observe the git diff. Experiment with different combinations of `plugins` in app.json
   > You can build and run your app with `expo run:ios` / `expo run:android`


GertSallaerts:
From the best practices in [Expo's config plugins documentation](https://docs.expo.io/guides/config-plugins/#plugin-development-best-practices):

> __Leverage built-in plugins__: Account for built-in plugins from the prebuild config. Some features are included for historical reasons, like the ability to automatically copy and link Google services files defined in the Expo config. If there is overlap, then maybe recommend the user uses the built-in types to keep your plugin as simple as possible.

Wouldn't it then make more sense to direct users to fill in the `config.android.googleServicesFile` and `config.ios.googleServicesFile` entries in their `app.json`?

Or if you really want to provide overrides/defaults do something like this in your plugin to still leverage the existing functionality somewhat:

```ts

const DEFAULT_ANDROID_GOOGLE_SERVICES_PATH = './google-services.json';
const DEFAULT_IOS_GOOGLE_SERVICES_PATH = './GoogleService-Info.plist';

const withRnFirebase: ConfigPlugin<PluginProps> = (
  config,
  { androidGoogleServicesPath, iosGoogleServicesPath } = {},
) => {

  config.android = config.android || {};
  config.android?.googleServicesFile =
    androidGoogleServicesPath ||
    config.android?.googleServicesFile ||
    DEFAULT_ANDROID_GOOGLE_SERVICES_PATH;

  config.ios = config.ios || {};
  config.ios?.googleServicesFile =
    iosGoogleServicesPath ||
    config.ios?.googleServicesFile ||
    DEFAULT_IOS_GOOGLE_SERVICES_PATH;

  return withPlugins(config, [
    // iOS
    withFirebaseAppDelegate,
  
    // Android
    withBuildscriptDependency,
    withApplyGoogleServicesPlugin,
  ]);
};
```

mikehardy:
Sorry I haven't had time to review yet - however! I built some infrastructure in the past which may help - if you follow the details link on the patches github actions check, you'll find a "patches.zip" artifact that is a full set of patch-package compatible files from last release to current main branch + this PR, so anyone can test it easily. I don't use Expo so I would love to see reports of success from anyone that tries it

I'll do a quick scan to make sure you're not a rogue bitcoin miner abusing Github free services (I'm joking) then approve the CI run so it generates the patches

barthap:
@mikehardy Thanks, it helped a bit :) In the _Test Plan_ section I prepared an instruction for anyone who wants to test it and prepared ready-to-go patched packages. Regarding other CI tests - i had to add `plugin/build` directories to eslint ignore, that test should pass now. Other tests passed (besides the title).

@GertSallaerts yes I considered that. Honestly, most functionality of the `@rnfb/app` plugin is also built in `expo-cli`, and partially in `expo-firebase-core`. And both approaches have some pros and cons. I decided to duplicate the functionality, because that makes RNFB more flexible and independent from potential expo-cli changes. Of course, I don't mind relying on the built-in plugin, but in that case the `app` plugin has little purpose to exist. cc @EvanBacon what do you think?

packages/app/plugin/src/index.ts
```diff
@@ -0,0 +1,63 @@
...
+      withCopyAndroidGoogleServices,
+      {
+        relativePath: resolvedAndroidServicesPath,
+      },
+    ],
+  ]);
+};
+
+const pak = require('@react-native-firebase/app/package.json');
+export default createRunOncePlugin(withRnFirebaseApp, pak.name, pak.version);
```
barthap, line 63: Should I make a versioned plugin? The config-plugins guide isn't very clear about it.

packages/app/plugin/src/android/constants.ts
```diff
@@ -0,0 +1,7 @@
+export const DEFAULT_TARGET_PATH = 'app/google-services.json';
+
+// TODO: Is there a better place to store these values?
+// Maybe package.json?
+export const googleServicesClassPath = 'com.google.gms:google-services';
+export const googleServicesPlugin = 'com.google.gms.google-services';
+export const googleServicesVersion = '4.3.8';
```
barthap, line 7: I saw some custom sections in `package.json` like `sdkVersions` - maybe there?

packages/app/plugin/src/index.ts
EvanBacon: Versioning is good, this also helps prevent duplication in case another library applies the plugin.

packages/app/plugin/src/ios/appDelegate.ts
```diff
@@ -0,0 +1,48 @@
...
+
+const methodInvocationBlock = `[FIRApp configure];`;
+
+export function modifyObjcAppDelegate(contents: string): string {
+  // Add import
+  if (!contents.includes('@import Firebase;')) {
+    contents = contents.replace(
+      /#import "AppDelegate.h"/g,
+      `#import "AppDelegate.h"
+@import Firebase;`,
```
EvanBacon, line 12: Maybe wrap this in an optional in case someone failed to link the native module. The JS interface would produce a better error in JS then the Xcode build would.

packages/app/plugin/tsconfig.json
```diff
@@ -0,0 +1,9 @@
+{
+  "extends": "@tsconfig/node10/tsconfig.json",
```
EvanBacon, line 2: Node 10 is out of LTS, Expo CLI min is now 12

packages/crashlytics/plugin/tsconfig.json
```diff
@@ -0,0 +1,9 @@
+{
+  "extends": "@tsconfig/node10/tsconfig.json",
```
EvanBacon, line 2: Node 12

packages/app/plugin/src/index.ts
```diff
@@ -0,0 +1,63 @@
...
+
+const DEFAULT_ANDROID_GOOGLE_SERVICES_PATH = './google-services.json';
+const DEFAULT_IOS_GOOGLE_SERVICES_PATH = './GoogleService-Info.plist';
+
+interface PluginProps {
+  /**
+   * Custom location of `google-services.json`,
+   * relative to project root
+   */
+  androidGoogleServicesPath?: string;
```
EvanBacon, line 18: For now, maybe just rely on the Expo CLI built-in values since they could be used across Google Sign-in and other packages. This will reduce initial complexity in the config plugin as well.

packages/app/plugin/src/android/constants.ts
Salakar: https://github.com/invertase/react-native-firebase/blob/master/packages/app/package.json#L63 would probably be the best place for versions, versions there are also used in docs generation

packages/app/package.json
```diff
@@ -70,6 +73,9 @@
       "compileSdk": 30,
       "buildTools": "30.0.2",
       "firebase": "28.2.1",
+      "firebaseCrashlytics": "2.7.1",
+      "firebasePerf": "1.4.0",
+      "gmsGoogleServices": "4.3.8",
```
mikehardy, line 78: With apologies, as this is a nitpick, but these strings are ambiguous to me, firebase crashlytics may refer to multiple things (as might the other three) so the name could use disambiguation - this impact all code referencing them

```suggestion
      "firebaseCrashlyticsPlugin": "2.7.1",
      "firebasePerfPlugin": "1.4.0",
      "gmsGoogleServicesPlugin": "4.3.8",
```

packages/app/plugin/__tests__/iosPlugin.test.ts
```diff
@@ -0,0 +1,14 @@
...
+
+import { modifyObjcAppDelegate } from '../src/ios/appDelegate';
+
+describe('Config Plugin iOS Tests', function () {
+  it('tests changes made to AppDelegate.m', async function () {
+    const appDelegate = await fs.readFile(path.join(__dirname, './fixtures/AppDelegate.m'), {
+      encoding: 'utf8',
+    });
+    const result = modifyObjcAppDelegate(appDelegate);
+    expect(result).toMatchSnapshot();
```
mikehardy, line 12: We don't use snapshots anywhere else in our testing infrastructure, so it may be that the next person that has to figure out why this failed may have no experience with them (like me)

Could you leave a comment here with the steps you used to construct the test? It may seem obvious, but something like "fetch current template AppDelegate from (?), put it ?, do this thing to it and put it here" etc when I'm trying to keep this thing running in spare time a year from now and it's e.g. a swift file in the template now so it all breaks

packages/app/plugin/tsconfig.json
```diff
@@ -0,0 +1,9 @@
+{
+  "extends": "@tsconfig/node12/tsconfig.json",
```
mikehardy, line 2: Out of curiosity, why this specific tsconfig node12 bit? Smells like maintainer-burden for the future node n+1 release etc

mikehardy (commented):
In general seems okay as far as I understand the Expo config system
My point of view for review is focused on answering these questions to my satisfaction:

As react-native and node etc advance over time and the files this is operating on change, how am I going to successfully maintain this, and how much time will it take

My comments reflect that in general - anything at all that will make this either version independent or explain the process of updating things to keep it moving is welcome, as our maintainership of the package is already time-limited given the sheer surface area of Firebase APIs

packages/app/package.json
barthap: These versions directly correspond to gradle dependencies from RNFB installation docs: 
- https://rnfirebase.io/#configure-firebase-with-android-credentials - `gmsGoogleServices`
- https://rnfirebase.io/crashlytics/android-setup#2-add-the-firebase-crashlytics-plugin-dependency - crashlytics
- https://rnfirebase.io/perf/usage#add-the-performance-monitoring-plugin - perf

Instead of `Plugin` suffix, I'd prefer more accurate suffix - `Gradle`, e.g. `firebaseCrashlyticsGradle`, unless you have any better idea. As @Salakar mentioned, these versions can be used for automatic docs generation.

packages/app/plugin/__tests__/iosPlugin.test.ts
barthap: Yeah, I'll leave some doc comments there.

Basically, what the tests do is loading `build.gradle` or `AppDelegate.m` test fixtures (they're taken from [bare minimum Expo template](https://github.com/expo/expo/tree/master/templates/expo-template-bare-minimum)). Then it applies the plugin changes on them and compares changed files with snapshots. If they're equal - the tests pass.
If you intentionally change anything in plugin code (e.g. gradle dependency versions), you have to update the snapshots by running `yarn tests:jest -u` (or interactively from _watch_ mode by pressing the `i` key). Anyway, I'll make a small `README` in the `__tests__` directory.

packages/app/plugin/tsconfig.json
barthap: I added this to avoid adding much heavier and not-really-necessary [`expo-module-scripts`](https://www.npmjs.com/package/expo-module-scripts) dependency.

My `tsconfig` should extend the [`expo-module-scripts/tsconfig.plugin.json`](https://github.com/expo/expo/blob/master/packages/expo-module-scripts/tsconfig.plugin.json), but when we see that tsconfig, it only extends the `node12` tsconfig, so why not use it directly 😄

barthap:
@mikehardy 
> As react-native and node etc advance over time and the files this is operating on change, how am I going to successfully maintain this, and how much time will it take
> 
> My comments reflect that in general - anything at all that will make this either version independent or explain the process of updating things to keep it moving is welcome, as our maintainership of the package is already time-limited given the sheer surface area of Firebase APIs

Generally, the plugin code shouldn't be difficult to maintain, because it is only related to installation instructions. And they're rather rarely changing. The most common subject to updates are the dependencies versions from `package.json` - but the only thing to do in that case is to bump versions there and update test snapshots.

I'd have more concerns about the functionality I have NOT implemented - e.g. [the FCM iOS installation](https://rnfirebase.io/messaging/usage/ios-setup), because users may eventually want that. I've implemented only `app`, `perf`, and `crashlytics` to give further contributors a foundation to implement the remaining RNFB plugins.

Also, my plan is to make a docs PR with updated installation instructions for Expo users.

packages/app/package.json
mikehardy: Yeah - the versions are ace, it's just disambiguating the variable name as it will escape it's gradle-local context here now that it's in package.json. I like your idea of a gradle suffix, certainly satisfies the issue I had in mind

packages/app/plugin/tsconfig.json
mikehardy: Interesting - I just spidered out and examined the whole `@tsconfig` thing. Interesting idea. This will be a maintenance burden syncing/updating over time, that's unavoidable as there must be some future state where ES2022 is the minimum required javascript target or similar, but it appears it may be a good maintenance burden in that it actually abstracts away other pain points in keeping tsconfigs in sync + valid. May adopt it in other areas. Resolved, in my mind, thanks

mikehardy:
I'll review this again with the README but I was thinking abstractly, separately about this theme:

> Generally, the plugin code shouldn't be difficult to maintain, because it is only related to installation instructions. And they're rather rarely changing. The most common subject to updates are the dependencies versions from package.json - but the only thing to do in that case is to bump versions there and update test snapshots.

I don't disagree. I'm just very sensitive to very small values of "to maintain" is all. Optimizing for "requires no maintenance" is the goal as there are enough features that are not difficult to maintain in a system this large that even a tiny amount on all of them in the aggregate is a mountain, inertially speaking.

That's not a quibble with this PR at this point pending README review on test snap maintenance, just a philosophical position

mikehardy (approved):
The README on test snap updating is great, my future self - trying to update this later as quickly as possible after forgetting how this all worked, has reviewed it and told me he appreciates it a lot ;-). The gradle naming I think will also help this be obvious.

This looks good to me. You already addressed `@Salakar` comments. @EvanBacon I believe you had some comments? They are not marked resolved, if you could re-check I would appreciate it.

This is good to go in my opinion though, I think a lot of people will be excited about this. A really powerful addition

oscar-gallog:
I literally hold an app that I'm building just to be able to do use react-native-firebase with Expo, very excited about this!

Salakar (dismissed):

packages/app/plugin/src/ios/googleServicesPlist.ts
```diff
@@ -0,0 +1,60 @@
+import { ConfigPlugin, withXcodeProject, XcodeProject } from '@expo/config-plugins';
+import { getSourceRoot } from '@expo/config-plugins/build/ios/Paths';
```
EvanBacon, line 2: Can use `IOSConfig.Paths.getSourceRoot()` so you aren't reaching into the package internals.

packages/app/plugin/src/ios/googleServicesPlist.ts
```diff
@@ -0,0 +1,60 @@
...
+import {
+  addResourceFileToGroup,
+  getProjectName,
+} from '@expo/config-plugins/build/ios/utils/Xcodeproj';
```
EvanBacon, line 6: `IOSConfig.XcodeUtils.getProjectName`, etc.

EvanBacon (dismissed):
LGTM, just a few comments about imports

mikehardy (approved):
That's a lot better for the review + fixup, thanks @EvanBacon @barthap 

I want to be clear that while I am certainly Expo-curious, and I am really excited that RNFB will now be more easily integrated into Expo, I will need help with user support and fixups in the area.

All of my support is based on knowledge of react-native built from `npx react-native init` and going from there :sweat_smile: . So, I will really need help in this area with any documentation or user support you can provide when the inevitable user questions arise :pray: :pray:

That actually makes me a bit nervous, but "fear of more users with legitimate questions" is no reason to hold up a PR in my opinion, so I'll merge this and let's go!

barthap:
> I will need help with user support and fixups in the area.

In such cases, just ping me or somebody at Expo, we'll try to help 😄

Salakar:
Thanks for this @barthap - would you also be able to update https://github.com/invertase/react-native-firebase/blob/master/docs/index.md#expo - it's quite out of date 😅

barthap:
Yeah, that's my plan 😄  I'll open a PR probably on Monday

Salakar:
> Yeah, that's my plan 😄 I'll open a PR probably on Monday

Thanks!

Ehesp:
Thanks for the help here @barthap @EvanBacon 👍

richardwu:
Thanks all for pushing this upstream. 

For other packages like `remote-config`, it seems like there are no additional steps on the native side, so AFAIK a config plugin is unnecessary. To confirm: should this be as simple as `yarn add @react-native-firebase/remote-config` if I am using EAS Build?

richardwu:
> Thanks all for pushing this upstream.
> 
> For other packages like `remote-config`, it seems like there are no additional steps on the native side, so AFAIK a config plugin is unnecessary. To confirm: should this be as simple as `yarn add @react-native-firebase/remote-config` if I am using EAS Build?

Oops, I just saw the line that packages without native code linking need not a plugin.

On a side note: is it necessary to include `com.google.firebase:firebase-iid:17.0.2` in the config plugin for `@react-native-firebase/app`? I'm running into this issue when using `expo-notifications` and a commentor mentioned that including that dependency fixed the issue: https://forums.expo.dev/t/cant-get-expotoken-in-expo-notification-due-to-failed-resolution-of-lcom-google-firebase-iid-firebaseinstanceid/53265/4

barthap:
I haven't dug into the issue itself, but if this dependency is needed for FCM / Notifications, the proper place for it would be the [`expo-notifications` plugin](https://github.com/expo/expo/blob/master/packages/expo-notifications/plugin/src/withNotificationsAndroid.ts) or `@react-native-firebase/fcm` _(plugin doesn't exist yet)_.

If you really need to include that dependency for notifications to work, the easiest workaround would be to write your own plugin (I replied in the forum thread).

mikehardy:
Worth noting, `iid` is a dead module for firebase, it should not be included as a direct dependency anywhere with current versions, as far as I know

techpanga:
Do we have config plugin for messaging?

barthap:
No, but if you mean to customize notification icons, please take look at this discussion: https://github.com/barthap/with-rn-firebase/issues/6 - you can "steal" `expo-notifications` plugin for that 😉

mikehardy:
Unrelated, looks like the dependency (expo-plugin) has a semver major to adopt. Is there anything we need to do here? Will it transitively cause a semver major here? Hadn't considered that this dependency might cause us to do majors, hopefully avoidable but if not so be it, I have another breaking change in the wings if so

barthap:
honestly I don't know, cc @EvanBacon

mikehardy:
when I tried to look (it came out a couple weeks ago) I have to say as an "expo-plugins" consumer, I was completely unable to find any CHANGELOG or commit diff or anything on it (I'm not even sure where the code is? I couldn't find it in the repo?) so it wasn't a smooth-n-easy consumer situation. If the package.json for it could actually point to the real location of expo-plugins (vs the general expo/expo repo) and there was a real changelog, I could have just done the migration...

brentvatne:
the changelog for config-plugins lives inside of the expo-cli repo (where config-plugins lives also): https://github.com/expo/expo-cli/blob/master/CHANGELOG.md

the major version bump was due to the ["extend base mods"](https://github.com/expo/expo-cli/issues/3852) feature landing, which adds to the public api. we bumped the major version because this shifted around the internals in a meaningful way and we are not entirely sure on the extent to which developers may have been depending on undocumented apis.

mikehardy:
Sweet! Thanks for the CHANGELOG pointer, that will help. Thanks also for the change insight, I don't think we were doing anything fancy at all in here, so this is likely safe for us to ingest then.

techpanga:
> No, but if you mean to customize notification icons, please take look at this discussion: [barthap/with-rn-firebase#6](https://github.com/barthap/with-rn-firebase/issues/6) - you can "steal" `expo-notifications` plugin for that 😉

@barthap , I am looking for imageUrl in notifications...

https://firebase.google.com/docs/cloud-messaging/ios/send-image
https://firebase.google.com/docs/cloud-messaging/android/send-image

Can this plugin be extended to include the image in notifications. (not for the icon)

nandorojo:
One thing I'm a bit unclear on from the docs:

> Not all React Native Firebase packages have Expo config plugins provided yet. You may see if a module is supported by checking if it contains the app.plugin.js file in its package directory.

`@react-native-firebase/auth` does not have a config plugin. I've only added `@react-native-firebase/app`. Does this mean that auth won't work, or should it work anyway? I ask because it isn't working for me, and I'm wondering if it's due to a missing config plugin, or if I messed something up in the config step.

Thanks!

PR #5521, opened 2021-07-15
test(ios): arm64 simulator is a valid target now

mikehardy:
### Description

Running the tests project on Apple Silicon results in an Xcode project file update that seems valid
Cement the seemingly valid setting via a Podfile post_install tweak so it is persistent after `pod install` regardless of current platform CPU architecture

### Test Plan

This already ran successfully on Apple Silicon
If it also runs in the CI environment (x86-64) then it's good to go

---


mikehardy:
Note that in testing the .pbxproj is now dirty on intel macs - so this may not be stable.
I tried altering post_install hook settings in the tests/Podfile but could not find one that worked correctly
I fear there may be a podspec messing with the user xcconfig settings or similar, but then further testing with any post_install build_settings in the Podfile resulted in no changes so I'm not sure what's going on.

TL;DR: this may be unstable and cause a dirty-local-workspace git issue in the short term while I investigate further

PR #5540, opened 2021-07-22
build(vercel): moved master branch script to website directory

dackers86:
Further research shows the build issues may be based on the location of the bash script.

The master branch is based on the `website` directory in Vercel.

This fix looks to move the script here so the `next` branch does not fail it's deployments when running master deployments.


mikehardy (approved):
3rd time's the charm? :crossed_fingers:

PR #5541, opened 2021-07-22
build: added website_modular build script

dackers86:
The next branch uses `website-modular` as the main entry point for the website.

Adding this in master will allow that deployment to see this configuration when running a `master` branch deployment


mikehardy (approved):
Do any of the other locations for this script now need removal? :thinking:

PR #5547, opened 2021-07-26
perf(messaging, ios): Improve time to delivery of background messages on iOS

brianathere:
Convert didReceiveRemoteNotification to wait until the RN bridge initialized based on the completion of the call to setBackgroundMessageHandler by the client

### Checklist

  - [x] Yes
  - [ ] `Android`
  - [x] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

Tested delivery using Firebase messaging, messages with data, in the foreground and background on an actual device. Confirmed foreground notification delivery is still immediate. Confirmed background notification delivery occurs when the application is running but hidden, as well as when the application has been force quit.

As this requires an actual device and no existing test automation existed for these scenarios all of the testing was manual.

---

### DIFF
--- a/packages/messaging/ios/RNFBMessaging/RNFBMessaging+AppDelegate.m
+++ b/packages/messaging/ios/RNFBMessaging/RNFBMessaging+AppDelegate.m
@@ -33,6 +33,8 @@
   __strong static RNFBMessagingAppDelegate *sharedInstance;
   dispatch_once(&once, ^{
     sharedInstance = [[RNFBMessagingAppDelegate alloc] init];
+    sharedInstance.conditionBackgroundMessageHandlerSet = [[NSCondition alloc] init];
+    sharedInstance.backgroundMessageHandlerSet = NO;
   });
   return sharedInstance;
 }
@@ -67,10 +69,13 @@
 }
 
 // used to temporarily store a promise instance to resolve calls to `registerForRemoteNotifications`
-- (void)setBackgroundMessageHandlerSet {
-  [self.conditionBackgroundMessageHandlerSet lock];
-  self.backgroundMessageHandlerSet = true;
-  [self.conditionBackgroundMessageHandlerSet unlock];
+- (void)signalBackgroundMessageHandlerSet {
+  RNFBMessagingAppDelegate *sharedInstance = [RNFBMessagingAppDelegate sharedInstance];
+  [sharedInstance.conditionBackgroundMessageHandlerSet lock];
+  NSLog(@"signalBackgroundMessageHandlerSet sharedInstance.backgroundMessageHandlerSet was %@", sharedInstance.backgroundMessageHandlerSet ? @"YES" : @"NO");
+  sharedInstance.backgroundMessageHandlerSet = YES;
+  [sharedInstance.conditionBackgroundMessageHandlerSet broadcast];
+  [sharedInstance.conditionBackgroundMessageHandlerSet unlock];  
 }
 
 // used to temporarily store a promise instance to resolve calls to `registerForRemoteNotifications`
@@ -109,6 +114,7 @@
 - (void)application:(UIApplication *)application didReceiveRemoteNotification:(NSDictionary *)userInfo fetchCompletionHandler:(void (^)(UIBackgroundFetchResult result))completionHandler {
   #if __has_include(<FirebaseAuth/FirebaseAuth.h>)
   if ([[FIRAuth auth] canHandleNotification:userInfo]) {
+    NSLog(@"didReceiveRemoteNotification Firebase Auth handeled the notification");
     completionHandler(UIBackgroundFetchResultNoData);
     return;
   }
@@ -117,6 +123,8 @@
   [[NSNotificationCenter defaultCenter] postNotificationName:@"RNFBMessagingDidReceiveRemoteNotification" object:userInfo];
     
   if (userInfo[@"gcm.message_id"]) {
+    NSLog(@"didReceiveRemoteNotification gcm.message_id was present %@", userInfo);
+
     if ([UIApplication sharedApplication].applicationState == UIApplicationStateBackground) {
       // If app is in background state, register background task to guarantee async queues aren't frozen.
       UIBackgroundTaskIdentifier __block backgroundTaskId = [application beginBackgroundTaskWithExpirationHandler:^{
@@ -136,14 +144,37 @@
         }
       });
 
-      [self.conditionBackgroundMessageHandlerSet lock];
-        while (!self.backgroundMessageHandlerSet)
-          [self.conditionBackgroundMessageHandlerSet wait];
-        dispatch_async(dispatch_get_main_queue(), ^{
-          [[RNFBRCTEventEmitter shared] sendEventWithName:@"messaging_message_received_background" body:[RNFBMessagingSerializer remoteMessageUserInfoToDict:userInfo]];
-        });
-      [self.conditionBackgroundMessageHandlerSet unlock];
+      RNFBMessagingAppDelegate *sharedInstance = [RNFBMessagingAppDelegate sharedInstance];
+      [sharedInstance.conditionBackgroundMessageHandlerSet lock];
+        NSLog(@"didReceiveRemoteNotification sharedInstance.backgroundMessageHandlerSet = %@", sharedInstance.backgroundMessageHandlerSet ? @"YES" : @"NO");
+        if (sharedInstance.backgroundMessageHandlerSet) {
+          // Normal path, backgroundMessageHandlerSet has already been set, queue the notification for immediate delivery
+          dispatch_async(dispatch_get_main_queue(), ^{
+            [[RNFBRCTEventEmitter shared] sendEventWithName:@"messaging_message_received_background" body:[RNFBMessagingSerializer remoteMessageUserInfoToDict:userInfo]];
+          });
+          NSLog(@"didReceiveRemoteNotification without waiting for backgroundMessageHandlerSet to be set");
+        } else {
+          // This spin needs to be on a background/concurrent queue to await the setup of backgroundMessageHandlerSet and not block the main thread
+          dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
+            // Reaquire the lock in this new closure
+            [sharedInstance.conditionBackgroundMessageHandlerSet lock];
+
+            // Spin/wait until backgroundMessageHandlerSet
+            // NB it is possible while this closure was being scheduled that backgroundMessageHandlerSet is already set and this loop is skipped
+            while (!sharedInstance.backgroundMessageHandlerSet) {
+              NSLog(@"didReceiveRemoteNotification waiting for sharedInstance.backgroundMessageHandlerSet %@", sharedInstance.backgroundMessageHandlerSet ? @"YES" : @"NO");
+              [sharedInstance.conditionBackgroundMessageHandlerSet wait];
+            }
+            dispatch_async(dispatch_get_main_queue(), ^{
+              [[RNFBRCTEventEmitter shared] sendEventWithName:@"messaging_message_received_background" body:[RNFBMessagingSerializer remoteMessageUserInfoToDict:userInfo]];
+            });
+            NSLog(@"didReceiveRemoteNotification after waiting for backgroundMessageHandlerSet");
+            [sharedInstance.conditionBackgroundMessageHandlerSet unlock];
+          });
+        }
+      [sharedInstance.conditionBackgroundMessageHandlerSet unlock];
   } else {
+      NSLog(@"didReceiveRemoteNotification while app was in foreground");
       [[RNFBRCTEventEmitter shared] sendEventWithName:@"messaging_message_received" body:[RNFBMessagingSerializer remoteMessageUserInfoToDict:userInfo]];
       completionHandler(UIBackgroundFetchResultNoData);
     }
--- a/packages/messaging/ios/RNFBMessaging/RNFBMessaging+AppDelegate.m
+++ b/packages/messaging/ios/RNFBMessaging/RNFBMessaging+AppDelegate.m
@@ -67,6 +67,13 @@
 }
 
 // used to temporarily store a promise instance to resolve calls to `registerForRemoteNotifications`
+- (void)setBackgroundMessageHandlerSet {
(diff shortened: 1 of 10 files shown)


mikehardy (changes requested):
Very interesting! I love this idea, where I will summarize the idea as "solve the iOS 'when is the RN bridge up' detection issue by using the 'setBackgroundMessageHandler' call as a positive signal to uncork messages'

Did I get that correct, as the idea?

If so, it seems like a nice fix and I'd like to collaborate with you to get this merged and released.

I see these things as necessary to get it in shape though

- nothing in the `tests/` directory should be included in this PR, they are not relevant to the fix and clutter the PR - please remove all those items
- the e2e test you changed is not executed on ios at all, and this is an ios fix, that change should  be reverted
- this is currently not backwards compatible, and as the `setBackgroundMessageHandler` API is marked as "android" it's quite possible that people are not calling it on iOS *but still receiving messages*. If I understand the code change proposed here, iOS will now hang indefinitely on message delivery if `setBackgroundMessageHandler` is not called. The previous time-delay/uncork logic needs to remain as a fallback I think, leaving your idea of the positive signal given by setBackgroundMessageHandler as a very nice optimization but without potentially causing problems for existing users.

What do you think?

brianathere:
> Very interesting! I love this idea, where I will summarize the idea as "solve the iOS 'when is the RN bridge up' detection issue by using the 'setBackgroundMessageHandler' call as a positive signal to uncork messages'
> 
> Did I get that correct, as the idea?
> 
> If so, it seems like a nice fix and I'd like to collaborate with you to get this merged and released.
> 
> I see these things as necessary to get it in shape though
> 
> * nothing in the `tests/` directory should be included in this PR, they are not relevant to the fix and clutter the PR - please remove all those items
> * the e2e test you changed is not executed on ios at all, and this is an ios fix, that change should  be reverted
> * this is currently not backwards compatible, and as the `setBackgroundMessageHandler` API is marked as "android" it's quite possible that people are not calling it on iOS _but still receiving messages_. If I understand the code change proposed here, iOS will now hang indefinitely on message delivery if `setBackgroundMessageHandler` is not called. The previous time-delay/uncork logic needs to remain as a fallback I think, leaving your idea of the positive signal given by setBackgroundMessageHandler as a very nice optimization but without potentially causing problems for existing users.
> 
> What do you think?

Excellent feedback. I accidentally sent this up for review, well before I got started, so let me start again here in an hour with a working solution, and then let's go from there. 

I wasn't aware setBackgroundMessageHandler was considered optional on iOS, and further saw it marked android which didn't make complete sense since it is needed on iOS, so yeah, I'll need to give how to mark this as backward breaking or not a think based on feedback.

mikehardy:
Hmm - I was keying that comment *mostly* off the fact the API call was marked `android` and not taken into account before. Perhaps I am fearing for backwards compatibility too much and/or incorrectly.

> I wasn't aware setBackgroundMessageHandler was considered optional on iOS

On a re-think, perhaps for this to matter at all, you would need a handler set? Stated differently: the only messages we are delaying here would be sent through the background message handler?

If so, then my assertion it is optional and needs management here is incorrect. I regret I'm not familiar enough with this exact area of code to know off-hand but it should be easy to verify the assertion, and I may be wrong...

brianathere:
> Hmm - I was keying that comment _mostly_ off the fact the API call was marked `android` and not taken into account before. Perhaps I am fearing for backwards compatibility too much and/or incorrectly.
> 
> > I wasn't aware setBackgroundMessageHandler was considered optional on iOS
> 
> On a re-think, perhaps for this to matter at all, you would need a handler set? Stated differently: the only messages we are delaying here would be sent through the background message handler?
> 
> If so, then my assertion it is optional and needs management here is incorrect. I regret I'm not familiar enough with this exact area of code to know off-hand but it should be easy to verify the assertion, and I may be wrong...



> Hmm - I was keying that comment _mostly_ off the fact the API call was marked `android` and not taken into account before. Perhaps I am fearing for backwards compatibility too much and/or incorrectly.
> 
> > I wasn't aware setBackgroundMessageHandler was considered optional on iOS
> 
> On a re-think, perhaps for this to matter at all, you would need a handler set? Stated differently: the only messages we are delaying here would be sent through the background message handler?
> 
> If so, then my assertion it is optional and needs management here is incorrect. I regret I'm not familiar enough with this exact area of code to know off-hand but it should be easy to verify the assertion, and I may be wrong...

My latest updates are what I was intending to share.

I believe nothing has changed in the case of foreground delivery with my change, it just gets sent to the app.

In the background delivery will be held on a dispatch queue until the client code calls setBackgroundMessageHandler and provides a callback. If the client code never calls, the delivery request will wait indefinitely.

Since setBackgroundMessageHandler is used on iOS and Android I removed the platform flag from that method.

I looked into writing automation for testing this change, however since it requires a real device to test I tested in my application instead. I'v only tested on an iPhone 12 at this point, but will go back to an iPhone 7 before merging.

As I've not written ObjC in 25 years, I'd very much appreciate eyes on that code.

packages/messaging/ios/RNFBMessaging/RNFBMessaging+AppDelegate.m
```diff
@@ -66,6 +68,16 @@ - (void)observe {
   });
 }
 
+// used to signal that a javascript handler for background messages is set
+- (void)signalBackgroundMessageHandlerSet {
+  RNFBMessagingAppDelegate *sharedInstance = [RNFBMessagingAppDelegate sharedInstance];
+  [sharedInstance.conditionBackgroundMessageHandlerSet lock];
+  NSLog(@"signalBackgroundMessageHandlerSet sharedInstance.backgroundMessageHandlerSet was %@", sharedInstance.backgroundMessageHandlerSet ? @"YES" : @"NO");
```
mikehardy, line 75: I'm afraid our logging is all over the place in the module, partly my fault.
But I believe the correct style is to use this shared utility:

> app/ios/RNFBApp/RNFBSharedUtils.h:#define DLog(fmt, ...) NSLog((@"%s [Line %d] " fmt), __PRETTY_FUNCTION__, __LINE__, ##__VA_ARGS__);

packages/messaging/ios/RNFBMessaging/RNFBMessagingModule.m
```diff
@@ -89,6 +89,15 @@ - (NSDictionary *)constantsToExport {
   return resolve([NSNull null]);
 }
 
+RCT_EXPORT_METHOD(signalBackgroundMessageHandlerSet) {
+  NSLog(@"signalBackgroundMessageHandlerSet called");
+  @try {
+    [[RNFBMessagingAppDelegate sharedInstance] signalBackgroundMessageHandlerSet];
+  } @catch (NSException *exception) {
+    NSLog(@"signalBackgroundMessageHandlerSet failed");
```
mikehardy, line 97: This one would use a similar log macro but it's `ELog` in this case

packages/messaging/lib/index.js
```diff
@@ -288,7 +288,8 @@ class FirebaseMessagingModule extends FirebaseModule {
   }
 
   /**
-   * @platform android
+   * Called when a message is received while the app is in the background. An app is considered to be in the background if no active window is displayed.
```
mikehardy, line 291:
```suggestion
   * Set a handler that will be called when a message is received while the app is in the background. 
   * Should be called before the app is registered in `AppRegistry`, for example in `index.js`. 
   * An app is considered to be in the background if no active window is displayed.
```

packages/messaging/lib/index.js
```diff
@@ -288,7 +288,8 @@ class FirebaseMessagingModule extends FirebaseModule {
...
+   * @param {*} handler called with message payload that contains the notification payload that is represented with firebase.messaging.NotificationPayload and the data payload that contains an arbitrary number of key-value pairs sent by developers through the Send API
```
mikehardy, line 292: Lifted from here https://rnfirebase.io/reference/messaging#setBackgroundMessageHandler

```suggestion
   * @param handler called with an argument of type messaging.RemoteMessage that must be async and return a Promise
```

mikehardy (changes requested):
I have some comments on logging and documentation but I don't have the Obj-C skills to know if I'm missing something. This is something that will affect the Notifee library a lot if there is an issue though, perhaps I could get a look from @helenaford ? For context if you have time Helena - the underlying idea seems sound to me, and the implementation in general even, to me - at this point I think we're just trying to suss out if if the code will actually spin-then-uncork messages correctly, with a side question (on my part) on what exactly will happen if no background handler is ever registered? (e.g., will it use 100% cpu, or ?)

packages/messaging/ios/RNFBMessaging/RNFBMessaging+AppDelegate.m
brianathere: Awesome. Do you know, will this end up in the system console log? I needed for debugging in development to correlate the log lines with the other iOS system logs to match up when notifications arrived to the phone with when the app responded. If ELog/DLog only write to the react native log, not the device logs, these statements are far less interesting and I will just remove them.

mikehardy: Yes - they are wrapping NSLog so should be going to the same place.
If they were RCTLog calls they'd go to the react-native console but these should be in Console.app or Xcode native device output

brianathere: I see it wraps NSLog, perfect.

mikehardy (dismissed):
This looks good to me know - as mentioned I myself am hoping for another set of eyes on the objective-c but I don't have any hangups personally and if no one else can take a look I'm biased towards merge.

Please note that this generates a patch set from the patch CI action (https://github.com/invertase/react-native-firebase/actions/runs/1069327151) - it generates the patch set as the full difference between current release (12.3.0 as of this typing) and current dev branch + this change so it contains more than just this PR, but stated differently it contains "everything approved for release plus the PR" so it's still worth trying to make sure it works for you in production via patch-package

In the absence of qualified objective-c review that's how I'll test integrate it in my app for verification as well, would love a report if you try it

brianathere:
> I have some comments on logging and documentation but I don't have the Obj-C skills to know if I'm missing something. This is something that will affect the Notifee library a lot if there is an issue though, perhaps I could get a look from @helenaford ? For context if you have time Helena - the underlying idea seems sound to me, and the implementation in general even, to me - at this point I think we're just trying to suss out if if the code will actually spin-then-uncork messages correctly, with a side question (on my part) on what exactly will happen if no background handler is ever registered? (e.g., will it use 100% cpu, or ?)

When didReceiveRemoteNotification is called and the app is in the background and the javascript client has not yet set a background message handler, a task is queued on the GCD thread pool, and the task waits for the client to set a handler. The task shouldn't use CPU, but it will consume a thread from the thread pool. This could lead to resource (address space, allowed threads?) exhaustion, is a large number (millions? - I don't know enough about modern iOS to know what the limits will be) of didReceiveRemoteNotification events are delivered. To mitigate this I added a change to the wait to give up after 25 seconds, aligned with the timeout for the beginBackgroundTaskWithExpirationHandler block. I tested a bit, seems to work well.

brianathere:
> This looks good to me know - as mentioned I myself am hoping for another set of eyes on the objective-c but I don't have any hangups personally and if no one else can take a look I'm biased towards merge.
> 
> Please note that this generates a patch set from the patch CI action (https://github.com/invertase/react-native-firebase/actions/runs/1069327151) - it generates the patch set as the full difference between current release (12.3.0 as of this typing) and current dev branch + this change so it contains more than just this PR, but stated differently it contains "everything approved for release plus the PR" so it's still worth trying to make sure it works for you in production via patch-package
> 
> In the absence of qualified objective-c review that's how I'll test integrate it in my app for verification as well, would love a report if you try it

Happy to try this, but I don't understand how to apply this patch. Can you point me to the documentation?

mikehardy:
@brianathere oh sure! The main to know is that https://github.com/ds300/patch-package exists (and it's pretty easy to install), then having learned about patch-package, the patches.zip we generate makes sense - it has a few patches you drop in your `./patches/` directory and whenever you `yarn` (or `npm i` or whatever) they apply and that's that. A great way to carry local changes without forking a repo, or in our case here since we are a monorepo about the only way for people to easily integrate PRs here into their actual project

mikehardy:
Unrelated: I just noticed the CLA wasn't signed yet either. Once things are reviewed I merge and release pretty quickly, so follow the details link on that pending check here and it'll help step through it

brianathere:
> Unrelated: I just noticed the CLA wasn't signed yet either. Once things are reviewed I merge and release pretty quickly, so follow the details link on that pending check here and it'll help step through it

I've attempted signing the CLA multiple times, it keeps timing out. Will try again. Definitely would like to get this into our users hands this week.

brianathere:
> > Unrelated: I just noticed the CLA wasn't signed yet either. Once things are reviewed I merge and release pretty quickly, so follow the details link on that pending check here and it'll help step through it
> 
> I've attempted signing the CLA multiple times, it keeps timing out. Will try again. Definitely would like to get this into our users hands this week.

This is what I get when I attempt to sign. Looks like I already have, but it still shows up as un-signed in the CI.

image

brianathere:
> @brianathere oh sure! The main to know is that https://github.com/ds300/patch-package exists (and it's pretty easy to install), then having learned about patch-package, the patches.zip we generate makes sense - it has a few patches you drop in your `./patches/` directory and whenever you `yarn` (or `npm i` or whatever) they apply and that's that. A great way to carry local changes without forking a repo, or in our case here since we are a monorepo about the only way for people to easily integrate PRs here into their actual project

I've tested on an iPhone 7 on iOS 12.1, and on a 12 Max running 14.7, works well in my application in both locations. I see there is still a block on the CLA being signed, do I need to do something more to get this cleared and this change merged?

mikehardy:
Sorry - just sweeping through notifications now (I do one big react-native-firebase pass a day, this is it). I do not believe there are any further actions required on your part.
I've asked for another set of Objective-C eyes on this just in case but I'm biased towards merge+release and will do so without review if I don't get a second look by this afternoon my time (that's a few hours from now).

mikehardy (approved):
Okay, just re-re-read this after the latest changes and it still LGTM. On a countdown for merge on my side, though it is possible we get a second review and they notice something I haven't

:+1:

mikehardy:
I suppose the doc you just made better for #5545 is now no longer up to date, since you rooted out the problem entirely!

Issue #5570, opened 2021-08-04
[🐛] FCM token does not refresh after deleting token via `deleteToken()` on iOS


helenaford:
Prior to the Firebase Messaging version before Installations, the FCM token would delete and refresh on iOS. Since installations, `deleteToken` has no effect. Same in FlutterFire. 

## Issue

From debugging, after calling `[FIRInstallations installations] deleteWithCompletion` once, `deleteToken` will start to refresh the token

```
  [[FIRInstallations installations] deleteWithCompletion:^(NSError * _Nullable error) {
    if (error) {
      [RNFBSharedUtils rejectPromiseWithNSError:reject error:error];
    } else {
      resolve([NSNull null]);
    }
  }];
```

### Javascript

<details><summary>Click To Expand</summary>
<p>

#### `package.json`:

```json
# N/A
```

#### `firebase.json` for react-native-firebase v6:

```json
# N/A
```

</details>

### iOS

<details><summary>Click To Expand</summary>
<p>

#### `ios/Podfile`:

- [ ] I'm not using Pods
- [x] I'm using Pods and my Podfile looks like:

```ruby
# N/A
```

#### `AppDelegate.m`:

```objc
// N/A
```

</p>
</details>

---

### Android

<details><summary>Click To Expand</summary>
<p>

#### Have you converted to AndroidX?

- [ ] my application is an AndroidX application?
- [ ] I am using `android/gradle.settings` `jetifier=true` for Android compatibility?
- [ ] I am using the NPM package `jetifier` for react-native compatibility?

#### `android/build.gradle`:

```groovy
// N/A
```

#### `android/app/build.gradle`:

```groovy
// N/A
```

#### `android/settings.gradle`:

```groovy
// N/A
```

#### `MainApplication.java`:

```java
// N/A
```

#### `AndroidManifest.xml`:

```xml
<!-- N/A -->
```

</p>
</details>

---

## Environment

<details><summary>Click To Expand</summary>
<p>

**`react-native info` output:**

```
 OUTPUT GOES HERE
```

- **Platform that you're experiencing the issue on**:
  - [ ] iOS
  - [ ] Android
  - [ ] **iOS** but have not tested behavior on Android
  - [ ] **Android** but have not tested behavior on iOS
  - [ ] Both
- **`react-native-firebase` version you're using that has this issue:**
  - `e.g. 5.4.3`
- **`Firebase` module(s) you're using that has the issue:**
  - `e.g. Instance ID`
[...]

PR #5571, opened 2021-08-04
fix(messaging): ensure token is refreshed when calling `deleteToken` on iOS

helenaford:
### Description

This will always refresh the fcm token, but it's unclear to me if this is the correct approach or not based on the documentation and changelog

I've opened up a ticket on https://github.com/firebase/firebase-ios-sdk/issues/8491 for guidance from the iOS team.

I don't believe there are any side-effects of always deleting the installations id, but it would be good to get confirmation as this approach always will delete it when refreshing the fcm token.

### Related issues
Fixes https://github.com/invertase/react-native-firebase/issues/5570 

### Checklist

  - [x] Yes
  - [ ] `Android`
  - [x] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [ ] No

### Test Plan

---

### DIFF
--- a/packages/messaging/ios/RNFBMessaging/RNFBMessagingModule.m
+++ b/packages/messaging/ios/RNFBMessaging/RNFBMessagingModule.m
@@ -125,7 +125,7 @@ RCT_EXPORT_METHOD(deleteToken:
   (RCTPromiseResolveBlock) resolve
     :(RCTPromiseRejectBlock) reject
 ) {
-  [[FIRMessaging messaging] deleteTokenWithCompletion:^(NSError *_Nullable error) {
+  [[FIRInstallations installations] deleteWithCompletion:^(NSError * _Nullable error) {
     if (error) {
       [RNFBSharedUtils rejectPromiseWithNSError:reject error:error];
     } else {


mikehardy:
Great - I just subscribed to the upstream issue and it looks like Paul B over there has already put it in a work queue, very curious to hear the result

mikehardy:
I'm not sure we ever received the upstream guidance we were looking for, specifically if we should delete the installation token, or if simply deleting the messaging token is good enough?

They do appear to have a fix coming for messaging token delete though and it is tagged for firebase-ios-sdk 8.6.0

mikehardy:
This appears to be blessed by upstream but they also indicate they have fixed the underlying problem. I'd like to see if underlying sdk 8.6.0 resolves this before using an API from another module but will merge this if 860 has no effect

mikehardy (approved):

helenaford:
@mikehardy thanks Mike. Yeah, thought that too - they seem to be happy with this and confirmed there are no side-effects to using it 😄 But agreed, probably don't need it if `delete token` is fixed.

mikehardy:
8.6.0 is rolling out now, already on cocoapods if you want to do a version override to test it?

The performance I'm getting from Installations.delete followed by getToken as I'm implementing that module is...wow `169099ms` on Android is one of the faster ones. iOS is very fast but android is slow slow slow, so hopefully messaging tokens work...

PR #5581, opened 2021-08-08
feat(app-check): add AppCheck implementation

mikehardy:
### Description

Initial implementation of AppCheck module

What works:
 - `@react-native-firebase/app-check` package namespace to match firebase-js-sdk
 - `appCheck` javascript namespace to match firebase-js-sdk
 - multi-app support
 - debug provider on Android if app is debuggable
 - `setTokenAutoRefreshEnabled` method, and parameter on `activate`
 - `getToken` in e2e testing works!
 - docs

What needs implementing:
- firebase.json entry that toggles `FirebaseAppCheckTokenAutoRefreshEnabled` plist entry
- token refresh listeners on Android
- some mechanism to configure providers prior to `[FIRApp configure]` in AppDelegate.m. Maybe we need a startup hook?

I propose that plist entry is deferred as it can be toggled already by early adopters and it's an easy follow-on PR

I prose that refresh listeners on Android are deferred indefinitely, as they don't exist on iOS so would not be cross-platform anyway.

### Related issues

#5346 

https://github.com/wix/Detox/issues/2933

### Release Summary

Each commit is separate and works on it's own with a conventional commit message and comment

### Checklist

  - [x] Yes
  - [x] `Android`
  - [x] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

- check automaticDataCollection == false to make sure token refresh is disabled
- see if setting automaticDataCollection == false then setting token refresh programmatically works? (the GDPR scenario)

---


docs/sidebar.yaml
```diff
@@ -23,6 +23,10 @@
     - - Building an Analytics Funnel
       - 'https://blog.theodo.com/2018/01/building-google-analytics-funnel-firebase-react-native'
   - '//static.invertase.io/assets/firebase/analytics.svg'
+- - AppCheck
+  - - - Usage
+      - '/app-check/usage'
+  - '//static.invertase.io/assets/social/firebase-logo.png'
```
mikehardy, line 29: I checked, there is no AppCheck official logo that I can find.

mikehardy, line 29: I patterned the sidebar stuff off the rest of them, but it is simply not linking in :sob:

docs/app-check/usage/usage.md
```diff
@@ -0,0 +1,93 @@
+---
```
Salakar, line 1: This should be located at `docs/app-check/usage.md` instead of `docs/app-check/usage/usage.md` or alternatively it could be `docs/app-check/usage/index.md` - otherwise it 404s: https://react-native-firebase-git-mikehardy-appcheck-invertase.vercel.app/app-check/usage

Salakar: Other docs are in the format `docs/<module>/usage/index.md` - so this makes sense to keep it the same

nomi9995:
@mikehardy I think, we should use JSI for `getToken` method for reduce getToken time

mikehardy:
@nomi9995 you think we should require a native build step (that is: NDK for java platforms) and use - for the first time - C/C++ code + makefiles etc + in this module for this API call? With the side effect that since the JSI interface is not stable so we will then have a new per-react-native-version API fragility in the module? Or am I misunderstanding how JSI would be used here? This sounds like a real maintenance headache that I have personally struggled with working with the react-native-reanimated repository and the react-native releases crew - it is a large additional maintenance burden for a module

That said, I'm always open to look at PRs and I do consider performance a feature. Just know that I consider maintainability a feature as well and it is my understanding that use of JSI right now is difficult to maintain

nomi9995:
@mikehardy yes, you are right, JSI is a real maintenance headache and made headache many times in reanimated

mikehardy:
I'm still open to a PR if you are strongly motivated, react-native-vision-camera uses it without quite as many ill effects, but they ship source vs pre-built binaries and that pushes build time expansion down to consumers.

Honestly I think JSI will be a future direction for the module but it will be used to best effect in Firestore for the document serialization/deserialization where performance really is critical.

For AppCheck, we're just waiting on the SDK to deliver results, I would guess - could be wrong but I would guess - removing JS/native bridge latency from that is an inconsequential modification

Issue #4428, opened 2020-10-21
[🐛] firebase.json not respecting "messaging_ios_auto_register_for_remote_messages" property on iOS


todesignandconquer:
## Issue
This is a very straightforward vanilla implementation, including a `firebase.json` as follows in the project root, as described here: https://rnfirebase.io/messaging/usage#auto-registration-ios

Regardless, I still get the following warning:

```json
Usage of "messaging().registerDeviceForRemoteMessages()" is not required. You only need to register if auto-registration is disabled in your 'firebase.json' configuration file via the 'messaging_ios_auto_register_for_remote_messages' property.
```

---

### Javascript

<details><summary>Click To Expand</summary>
<p>

#### `package.json`:

```json
    "@react-native-firebase/analytics": "^7.6.2",
    "@react-native-firebase/app": "^8.4.2",
    "@react-native-firebase/dynamic-links": "^7.5.3",
    "@react-native-firebase/iid": "^7.4.3",
    "@react-native-firebase/in-app-messaging": "^7.5.1",
    "@react-native-firebase/messaging": "^7.8.6",
    ...
    "react-native": "0.63.2",
```

#### `firebase.json` for react-native-firebase v6:

```json
{
  "react-native": {
    "messaging_ios_auto_register_for_remote_messages": false
  }
}
```

Redux implementation:

```javascript
...
import Messaging from '@react-native-firebase/messaging'
...
const refreshNotifications = () => async(dispatch, getState) => {
     ...
     await Messaging().registerDeviceForRemoteMessages()
     const token = await Messaging().getToken()
     ...
}
```

</details>

### iOS
<details><summary>Click To Expand</summary>
<p>

#### `ios/Podfile`:

- [ ] I'm not using Pods
- [x] I'm using Pods and my Podfile looks like:

```ruby
require_relative '../node_modules/react-native/scripts/react_native_pods'
require_relative '../node_modules/@react-native-community/cli-platform-ios/native_modules'

platform :ios, '11.0'

workspace 'App'

target 'App' do
  config = use_native_modules!

  use_react_native!(:path => config["reactNativePath"])

  permissions_path = '../node_modules/react-native-permissions/ios'

  pod 'Permission-Camera', :path => "#{permissions_path}/Camera.podspec"
  pod 'Permission-FaceID', :path => "#{permissions_path}/FaceID.podspec"
  pod 'Permission-LocationWhenInUse', :path => "#{permissions_path}/LocationWhenInUse.podspec"
[...]

Issue #4241, opened 2020-09-10
Remove use of configureWithApp from iOS init sequence


mikehardy:
Tracking planned upstream method removal: https://github.com/firebase/firebase-ios-sdk/issues/2933

It looks like dynamic links has a very visible example of what the change means - that configureWithApp should go away, and that 'componentsToRegister' becomes the new natural home for the logic, with a new flag to initialize quickly

https://github.com/firebase/firebase-ios-sdk/pull/4137/files/71f589926b74e69ea76d0a25ae24e58efee6836b#diff-61367186bfcedbaaa5b0554cd03f8821R126-R162

What I don't know though is where to put that flag that says we want init eagerly (for the default app since crashlytics does not do multi-app) - e.g. this https://github.com/firebase/firebase-ios-sdk/pull/4137/files/71f589926b74e69ea76d0a25ae24e58efee6836b#diff-f33c561b2a3430b947e6da076758c6d9R1887

For RNFB Crashlytics we need to figure out how to set whether crashlytics should be enabled or not during startup, in the new style

_Originally posted by @mikehardy in https://github.com/invertase/react-native-firebase/pull/4236#r486507138_

ryanwilson:
Thanks for picking this up! I'm curious how this affects this SDK - is this going to be a large refactor?

This is more of a code-cleanup thing and the breaking change isn't a hard deadline for this to be removed. It could also be deprecated for the breaking change and removed next time if needed.

Re: eager init, it goes in the `componentsToRegister:` bit which it looks like isn't populated at the moment. Perhaps we could sync up at some point in the near future about the needs of the RN SDK and make sure we don't do anything that breaks it.

mikehardy:
Hey @ryanwilson - this seemed like it was a 95% "understand the change and current code" / 1% implementation / 4% testing type change to me. I didn't fully understand the current code here in RNFB so I have work to do there if I implement it, and this issue was mostly me filing my notes for when it did happen. My quick take was that the proposed change upstream had an end result that offered everything we needed with regard to init sequence, we just needed to bend to match. 

With Invertase focused heavily on FlutterFire right now we (mostly "I" at the moment) are unfortunately really light on resources so I'm happy to hear there is not a hard deadline, though in practical terms it may not receive attention here until there *is* a hard deadline to motivate it, given we're just barely keeping up with breaking changes at the moment as a resources vs priorities practical matter, and are lagging on tracking new APIs

ryanwilson:
[...]

PR #5597, opened 2021-08-15
fix: sweep of firebase.json type config elements and related infrastructure

mikehardy:
### Description

Includes a lot of related work, commit by commit, with test support

- testing some of the data collection init settings more thoroughly (e.g. #4428)
- implementing in-app-messaging data-collection disable flag from json -> native configs
- implementing + testing app check's token auto refresh disable from json -> native config, important for iOS GDPR
- fixing up test platform inclusion styles (`ios.only` does not work, `if (device.getPlatform() === 'ios')` works)

Having read through the current state of ios/android native Info.plist/AndroidManifest.xml data collection settings, we're a fair bit behind with regard to available toggles - I'll try to get the rest in a follow-on PR

### Related issues

#4428 asserted that iOS auto registration wasn't correctly handled, but I was unable to reproduce

### Release Summary

All conventional commits --> rebase merge

### Checklist

  - [x] Yes
  - [x] `Android`
  - [x] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

---


packages/messaging/ios/RNFBMessaging/RNFBMessagingModule.m
```diff
@@ -247,6 +247,7 @@ - (NSDictionary *)constantsToExport {
   if (@available(iOS 10.0, *)) {
     if ([UIApplication sharedApplication].isRegisteredForRemoteNotifications == YES) {
       resolve(@([RCTConvert BOOL:@(YES)]));
+      return;
```
ourhut, line 250: @mikehardy This return is logically inconsistent with the comment 5 lines later 
```
    // Apple docs recommend that registerForRemoteNotifications is always called on app start
    // regardless of current status
```

I don't know if the comment is accurate and I wasn't able to find any documentation from apple stating this when I just looked, but it is unclear to me now which cases we should call registerForRemoteNotifications.

mikehardy: This was not intended to be any sort of a change that users should have to *alter* their code for. Whatever you have been doing, you should keep doing.

There have been other cases in this code base (and others I've been on) where resolving and then not returning has led to unintended side-effects as a resolve without return means code keeps executing.

In this case I judged the hygiene risk of that scenario much higher than attempting to register again, since by the conditional we are already registered.

That said, if you determine any case at all where your app does not seem to be registered, please let me know. That would be entirely unexpected (given the conditional being true in this case....)

ourhut: reading a bit further, it looks like the full/complete implementation of this Apple API would be implementing didRegisterForRemoteNotificationsWithDeviceToken and didFailToRegisterForRemoteNotificationsWithError and the resolving the promise with the result of them being called.

https://developer.apple.com/documentation/appkit/nsapplicationdelegate/1428766-application

I've found nothing yet to indicate if the app should always call registerForRemoteNotifications or if it can count on isRegisteredForRemoteNotifications and only call it registerForRemoteNotifications when it is false.

ourhut: Just noticed I'm using my personal account, this is all from brian@slashtalk.com, my mistake on not checking the browser tab.

mikehardy: > reading a bit further, it looks like the full/complete implementation of this Apple API would be implementing didRegisterForRemoteNotificationsWithDeviceToken and didFailToRegisterForRemoteNotificationsWithError and the resolving the promise with the result of them being called.
> 
> https://developer.apple.com/documentation/appkit/nsapplicationdelegate/1428766-application

That's a macOS API (`NSApplication` root).

I think  you'll like that when we register we store the promise resolver and rejecter on our AppDelegate object, and we've swizzled ourselves into the response chain for when certain items are invoked, and we do what you mention, but on the `UIApplication` iOS APis:

https://github.com/invertase/react-native-firebase/blob/8765a993f137964f11f2dffbb027d65cda897204/packages/messaging/ios/RNFBMessaging/RNFBMessaging%2BAppDelegate.m#L90-L115

> 
> I've found nothing yet to indicate if the app should always call registerForRemoteNotifications or if it can count on isRegisteredForRemoteNotifications and only call it registerForRemoteNotifications when it is false.

I have not either. It's actually called immediately if you're set to auto register for them 

https://github.com/invertase/react-native-firebase/blob/5daa5ff7f22aed5de338d1d1ca5bb430cc93f8c5/packages/app/firebase-schema.json#L81-L83

https://github.com/invertase/react-native-firebase/blob/8765a993f137964f11f2dffbb027d65cda897204/packages/messaging/ios/RNFBMessaging/RNFBMessaging%2BNSNotificationCenter.m#L91-L93

I'm unaware of who you are as @ourhut or `brian@slashtalk.com`, but if there is something wrong here, I can certainly correct it

brianathere: Thank you for pointing out the place the location the promise resolver and rejecter are referenced, I was looking around and hadn't found it yet. The makes perfect sense. 

I also now see why what you've done is correct, and I don't have any cause for concern other than what I got from reading the comment I mentioned.

A style comment that I think would make this clearer, move call to registerForRemoteNotifications into the else block after registering the resolvers, and remove the early return and old comment. I feel like that makes it clear the there are two distinct paths, one that returns TRUE, and one that registers the resolvers and dispatches a request that will resolve/reject them later. Logically the same as what you have now. 

Thanks for quickly responding, and I apologize for the noise. This library is amazing, we're getting a lot of value from it and I'm happy to help with improving it as I can.

mikehardy: Not sure if I'll have time to get to that suggestion right now, but I do agree things could be tidied up there.
Right now I'm still doggedly pursuing other initialization / config-setting changes that cross-cut the whole set of modules, there's quite a few things going on there and some of them have higher priority ecosystem impacts all the way back to firebase-ios-sdk like #4241 which I need to maintain focus on

mikehardy: Oh, and no worries about noise - I was moving literally as quickly as possible through things over the weekend and no one else has had time to review, so even if the result is the null hypothesis is upheld (no further change necessary) simply having anyone look at it, and even potentially forcing a defense of change, is hugely helpful. I appreciate it

Baterka:
Is this fixed already? Or is there some workaround?

brianathere:
> Is this fixed already? Or is there some workaround?

What issue are you running into?

Baterka:
> > Is this fixed already? Or is there some workaround?
> 
> What issue are you running into?

Even I disabled every auto-init:
```
{
  "react-native": {
    "crashlytics_debug_enabled": true,
    "messaging_ios_auto_register_for_remote_messages": false,
    "messaging_auto_init_enabled": false,
    "analytics_auto_collection_enabled": false
  }
}
```

I am still getting "Access to Local Network" permission popup on iOS which means that messaging is still auto-initialized and calling for additional permissions.

mikehardy:
Hi there! I'm not sure the assertion that the "access to local network" popup is connected to this module is correct. I don't believe this module does a local network scan and I thought that was the cause for that popup, if it popped up? Can you point me to some documentation on the apple developer site that shows me what the conditions are for that network popup?

As a separate item, you might like this toggle:

https://github.com/invertase/react-native-firebase/blob/79918977979e099721da634157ec18cb9699e746/tests/firebase.json#L5

Baterka:
> Hi there! I'm not sure the assertion that the "access to local network" popup is connected to this module is correct. I don't believe this module does a local network scan and I thought that was the cause for that popup, if it popped up? Can you point me to some documentation on the apple developer site that shows me what the conditions are for that network popup?
> 
> As a separate item, you might like this toggle:
> 
> https://github.com/invertase/react-native-firebase/blob/79918977979e099721da634157ec18cb9699e746/tests/firebase.json#L5

I thought that this: https://rnfirebase.io/reference/messaging#registerDeviceForRemoteMessages is causing permission popup appearance. Maybe I am wrong... 

Sadly I found just this about the permission popping up on my device when I initialize Firebase: https://www.howtogeek.com/692528/why-iphone-apps-ask-for-devices-on-your-local-network/

mikehardy:
@Baterka I will assert that the "local network permission" request from your app is unrelated to react-native-firebase (or firebase-ios-sdk, our underlying SDK) completely. I use the messaging module. I do not even have the local network permission request plist key entry in my app, and I do not see the popup, and my app works fine.

This is the permission https://support.apple.com/en-us/HT211870
Here is some stackoverflow https://stackoverflow.com/questions/65952902/how-to-check-local-network-permission-in-ios-14
The permissions library I use doesn't even support it! https://github.com/zoontek/react-native-permissions/issues/509

Near as I can tell, you've got something else going on in your app.

PR #5641, opened 2021-08-24
fix(messaging, ios): revert "fix(messaging, ios): only call onMessage handler if message is data-only or undelivered"

helenaford:
Reverts invertase/react-native-firebase#5604

From additional testing, fixes the duplication issue but causes another issue where apple throttle `willPresentNotification` with the presence of `mutableContent` in the payload

I will be pursuing other solutions as the duplicate notification bug causes a big issue in apps


mikehardy (approved):
Filed under: no good deed unpunished

PR #5667, opened 2021-08-31
fix(app-distribution, ios): avoid crash when releaseNotes is nil

YukiKitagata:
### Description

On iOS device crashes when called `firebase.app().appDistribution().checkForUpdate()`.

image

This happens when `release.releaseNotes` is nil. Because the current implementation is assigned to an NSDictionary even if `release.releaseNotes` is nil.

This PR has been changed to return NSNull if it is nil.

https://github.com/invertase/react-native-firebase/blob/600dee0547423e3a9e1cce2451aad5f95c5e28dd/packages/app-distribution/ios/RNFBAppDistribution/RNFBAppDistributionModule.m#L96-L102

In addition, it fixes the issue of `resolve` being called twice.

### Related issues

N/A

### Checklist

  - [x] Yes
  - [ ] `Android`
  - [x] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

---

### DIFF
--- a/packages/app-distribution/ios/RNFBAppDistribution/RNFBAppDistributionModule.m
+++ b/packages/app-distribution/ios/RNFBAppDistribution/RNFBAppDistributionModule.m
@@ -96,12 +96,10 @@ RCT_EXPORT_METHOD(checkForUpdate
     resolve(@{
       @"displayVersion" : release.displayVersion,
       @"buildVersion" : release.buildVersion,
-      @"releaseNotes" : release.releaseNotes,
+      @"releaseNotes" : release.releaseNotes == nil ? [NSNull null] : release.releaseNotes,
       @"isExpired" : [NSNumber numberWithBool:release.isExpired],
       @"downloadURL" : release.downloadURL
     });
-
-    resolve([NSNull null]);  // FIXME read the result into a map and return it!
   }];
 }
 


mikehardy (approved):
Thank you for this! I coded it to docs (it's supposed to always be there?) but didn't have a specific project to test on and you know docs and reality are always a little different. I'll do a release as soon as this is merged etc

mikehardy:
Definitely need the cla signed pre merge then should be quick to merge

YukiKitagata:
@mikehardy 
Thanks for the Approve!
I've signed the CLA.

mikehardy:
Great - merged and I have a separate crash fix for dynamic links queued up as well, will release both shortly.

It's worth saying that you may be the first person using App Distribution via react-native-firebase, which explains why you saw something immediately (this crash) but also leads me to ask in advance to please post anything else up you see since there may be more issues lurking  - I want it to be a good experience but it's impossible to test really in an E2E context. I couldn't think of a way anyway.

There aren't many APIs and they are all really really simple at least, so hopefully it all works great from this point :crossed_fingers:

mikehardy:
app distribution is actually beta upstream as well (it really is brand new...) and the related issue I filed upstream was marked for a documentation change on the nullability of the property, so this is helpful throughout the firebase ecosystem https://github.com/firebase/firebase-ios-sdk/issues/8602

mikehardy:
Android does not have a public implementation yet, so there won't be problems after you're done integrating iOS (android just resolves via a reject saying it doesn't work on the platform...) but they are implementing it internally pre-release and I verified that it is marked correctly as nullable there

https://github.com/firebase/firebase-android-sdk/blob/82b02af331d4674682bfb90c195da45543b808c3/firebase-app-distribution/src/main/java/com/google/firebase/appdistribution/AppDistributionRelease.java#L46-L48

I think we're all set.

PR #5694, opened 2021-09-04
fix(auth, android): linkWithCredential will not attempt to upgrade from anon user (matches iOS)

Albocal:
### Description
When doing "linkWithCredential" in Android and the users exists, it makes relogin with this new user instead of launch exception:
Error: [auth/credential-already-in-use] This credential is already associated with a different user account.

### Checklist

  - [x] Yes
  - [x] `Android`
  - [ ] `iOS`
- My change includes tests;
- This is a breaking change;
  - [x] Yes --> Only for Android
  - [ ] No

### Test Plan

---


mikehardy (approved):
Thanks for posting this! I went to the relevant PR #4487 and PR #4552 and per a comment there:

https://github.com/invertase/react-native-firebase/pull/4487#issuecomment-742100755

the original author has actually disavowed this style as well, with an indication it was no longer working anyway, so this looks like a good PR. Shame the style doesn't work - it would be a nicer DX if it did work and we could do the same thing in iOS but it is best if the platforms are consistent, and if this isn't working it should be removed

Thanks for the PR!

PR #5697, opened 2021-09-05
build(e2e): centralize build commands to use all yarn run scripts

mikehardy:
### Description

Just noticed that the detox / e2e build commands were still sometimes doing custom commands instead of using the yarn run scripts. This makes it hard to change/improve things as there are multiple command styles in use

With this change the e2e scripts are now all using yarn run scripts for everything I think

### Test Plan

If CI passes, it's good

---


mikehardy:
The affected parts of e2e have already run successfully, good to go

PR #5734, opened 2021-09-22
test(e2e): forward port to Detox test framework v18

mikehardy:
### Description

This is in prep for support of iOS15 in Detox - currently Detox breaks on iOS15 but support (when it comes) will certainly rely on a new version, so it's finally time to forward port here

1- two tests needed to be disabled
-- one test exposed a Detox (I think) crash in their network sync code in analytics
-- one test was simply a failure in crashlytics restart detection - very odd

Both are marked with TODO (or FIXME) and detox

2- may be a little flaky on iOS, I noticed some crashes on iOS but they were not reproducible, only time + experience will tell - I have noted that here https://github.com/invertase/react-native-firebase/issues/4058#issuecomment-930644198 and logged it here https://github.com/wix/Detox/issues/3000

3- because the app will never launch as part of the detox init now, the jet hot-patch to init that waited for node to be connected to the app needed to be removed. launchApp is now called separately and it was already hot-patched by jet to wait for connection so it is sufficient

4- sadly I continue to suffer from flakiness in the firestore pre-compiled framework inclusion so I had to comment it out again

5- because iOS 15 is not supported by Detox yet (https://github.com/wix/Detox/issues/2895) I have opted to pin our iOS e2e workflow on Xcode 12.5.1 via `xcode-select` github action, as that guarantees we will get iOS 14.x simulators. For local execution of the tests you must have a side-by-side install of Xcode-12.5.1 (or Xcode-12.4 I guess) and use `sudo xcode-select -s /Applications/Xcode-12.5.1.app` to use it or iOS e2e will fail for you.. Here is a reasonable article explaining how to do it https://medium.com/@YogevSitton/install-xcode-7-and-xcode-8-side-by-side-5bf40ea8f5ac though you need to go to your Apple Account's downloads / all downloads section and select something beside Xcode 7 obviously

6- I took the opportunity to update all the other javascript dependencies and de-flake CI a little bit since this PR was forcing a  bunch of compile / test cycles anyway and it was easy to qualify them

### Test Plan

If E2E works in CI, this works

---


mikehardy:
Okay, confirmed iOS 13.7 is not a valid OS specifier in CI. I dislike the OS specifier anyway, so this will either wait on Detox official iOS15 support or I'll hack in the same and post it for them. Probably wait - here's the URL to upstream issue https://github.com/wix/Detox/issues/2895

mikehardy:
also needs testing on an M1 mac, there was a note that Detox v18 may not work there

PR #5796, opened 2021-10-16
fix(app, expo): Update iOS `AppDelegate` plugin to work with Expo SDK 43

### LINKED ISSUE #5788: App / Expo plugin - iOS might break with Expo SDK 43
> Not sure where to post it, it's not a problem yet, feel free to move/edit this issue.

Because of [this commit in expo](https://github.com/expo/expo/commit/9781212eb3ec89a0b184da712ec7ca4e5ecc725b#diff-bff087577a5c9701fc698b349742e47dd252ee849307704985df6799de62f405) (scroll to `AppDelegate.m`),
The `@rnfb/app` plugin will likely stop working, because of [this regex](https://github.com/invertase/react-native-firebase/blob/master/packages/app/plugin/src/ios/appDelegate.ts#L19-L23)

We need to find a way to make that regex work more flexible - to support both old and new `AppDelegate` template.

barthap:
### Description

Resolves problem described in #5788 

Followed solution from the [`react-native-maps` plugin](https://github.com/expo/expo-cli/blob/4477bdde3c768e4ab822fb43224e94f85453683b/packages/config-plugins/src/ios/Maps.ts#L12) as suggested in the conversation: https://github.com/invertase/react-native-firebase/issues/5788#issuecomment-943832962 

I did NOT update other places to use the new syntax (`mergeContents()`), but I can if asked to do so.

To avoid possible problems with duplicate `[FIRApp configure]` call after updating, the new plugin skips if the call is found. The `mergeContents` has this functionality already, but compares only `@expo generated headers` which are introduced in this PR, so it wouldn't work with previous plugin versions.
These problems themselves could be easily solved by running `expo prebuild --clean`, but IMO it's better to have an extra check instead of users creating issues _"expo plugin stopped working. Firebase default app instance already initialized"_ etc 😉 

### Related issues

Resolves #5788

### Release Summary

- Updated iOS Expo config plugin for `@react-native-firebase/app` to work with Expo SDK 43.

### Checklist

  - [x] Yes
  - [ ] `Android`
  - [x] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

- Added test fixture for new `AppDelegate.m`
- Now we have two test cases: for current/old pre-SDK 43 AppDelegate as well as for SDK 43+ (and non-Expo) AppDelegate.
- Looking at the `.snap` snapshot files, we can see that the plugin works perfectly in both cases 😁 
   > I'm glad that we've decided to use snapshot testing technique here - it's really helpful in such cases

### DIFF
--- a/packages/app/plugin/src/ios/appDelegate.ts
+++ b/packages/app/plugin/src/ios/appDelegate.ts
@@ -17,6 +17,11 @@ export function modifyObjcAppDelegate(contents: string): string {
     );
   }
 
+  // To avoid potential issues with existing changes from older plugin versions
+  if (contents.includes(methodInvocationBlock)) {
+    return contents;
+  }
+
   // Add invocation
   return mergeContents({
     tag: '@react-native-firebase/app-didFinishLaunchingWithOptions',


mikehardy:
Thanks for this!
E2E test failures are obviously unrelated (two known flakes, alas) - I restarted them but they are not a gate here

When you say: 

> I did NOT update other places to use the new syntax (mergeContents()), but I can if asked to do so.

Does that imply perf and crashlytics will have similar issues? If that understanding is correct, any effort to bring them in line with expo-future-state would be really appreciated. This repo sees enough new users / day that if everything isn't nailed down (code + tests + docs) you can see how quickly issue traffic / response eats up all available time then no forward progress happens. The alternative (everything nailed down) is an actual time-saver then

PR #5840, opened 2021-11-07
fix(firestore)!: fix Long/Double conversion issues #3004

### LINKED ISSUE #3004: Cloud Firestore .where() in operator doesn't work for numbers
## Issue

When using `where()` and `in` for queries, numbers does't work.
```js
// does not work
const querySnapshot = await roomsRef.current.where('status', 'in', [1, 2]);

// after changing the field 'status' to string in Firebase console, it does work
const querySnapshot = await roomsRef.current.where('status', 'in', ['1', '2']);
```
In order to find out if this is `react-native-firebase`'s bug or Firestore simply does not support `in` numbers, I have used the JS SDK for web to perform the same query and it worked. So I came to the conclusion that this is a bug in `react-native-firebase`.

---

### Javascript

<p>

#### `package.json`:

```json
{
  "name": "test",
  "version": "0.0.1",
  "private": true,
  "scripts": {
    "android": "react-native run-android",
    "ios": "react-native run-ios",
    "start": "react-native start",
    "test": "jest",
    "lint": "eslint ."
  },
  "dependencies": {
    "@expo/react-native-action-sheet": "^3.4.1",
    "@invertase/react-native-apple-authentication": "^0.1.1",
    "@react-native-community/async-storage": "^1.7.1",
    "@react-native-community/google-signin": "^3.0.3",
    "@react-native-firebase/app": "^6.2.0",
    "@react-native-firebase/auth": "^6.2.0",
    "@react-native-firebase/firestore": "^6.2.0",
    "@react-native-firebase/functions": "^6.2.0",
    "@react-native-firebase/storage": "^6.2.0",
    "appcenter": "^2.6.0",
    "appcenter-analytics": "^2.6.0",
    "appcenter-crashes": "^2.6.0",
    "moment": "^2.24.0",
    "react": "16.9.0",
    "react-native": "0.61.5",
    "react-native-code-push": "^6.0.0",
    "react-native-fbsdk": "^1.1.1",
    "react-native-gesture-handler": "^1.5.2",
    "react-native-image-crop-picker": "^0.26.1",
    "react-native-keyboard-aware-scroll-view": "^0.9.1",
    "react-native-linear-gradient": "^2.5.6",
    "react-native-reanimated": "^1.4.0",
    "react-native-screens": "^1.0.0-alpha.23",
    "react-native-vector-icons": "^6.6.0",
    "react-navigation": "^4.0.10",
    "react-navigation-header-buttons": "^3.0.4",
    "react-navigation-stack": "^1.10.3"
  },
  "devDependencies": {
    "@babel/core": "^7.6.2",
    "@babel/runtime": "^7.6.2",
    "@react-native-community/eslint-config": "^0.0.5",
    "babel-jest": "^24.9.0",
    "eslint": "^6.5.1",
    "flow-bin": "^0.113.0",
    "jest": "^24.9.0",
    "metro-react-native-babel-preset": "^0.56.0",
    "react-test-renderer": "16.9.0"
  },
  "jest": {
    "preset": "react-native"
  }
}
```

```json
{
  "functions": {
    "predeploy": [
      "npm --prefix \"$RESOURCE_DIR\" run lint",
      "npm --prefix \"$RESOURCE_DIR\" run build"
    ]
  }
}
```

### iOS

<p>

#### `ios/Podfile`:

- [ ] I'm not using Pods

```ruby
platform :ios, '9.0'

target 'Test' do
  # Pods for Test
  pod 'FBLazyVector', :path => "../node_modules/react-native/Libraries/FBLazyVector"
  pod 'FBReactNativeSpec', :path => "../node_modules/react-native/Libraries/FBReactNativeSpec"
  pod 'RCTRequired', :path => "../node_modules/react-native/Libraries/RCTRequired"
  pod 'RCTTypeSafety', :path => "../node_modules/react-native/Libraries/TypeSafety"
  pod 'React', :path => '../node_modules/react-native/'
  pod 'React-Core', :path => '../node_modules/react-native/'
  pod 'React-CoreModules', :path => '../node_modules/react-native/React/CoreModules'
  pod 'React-Core/DevSupport', :path => '../node_modules/react-native/'
  pod 'React-RCTActionSheet', :path => '../node_modules/react-native/Libraries/ActionSheetIOS'
  pod 'React-RCTAnimation', :path => '../node_modules/react-native/Libraries/NativeAnimation'
  pod 'React-RCTBlob', :path => '../node_modules/react-native/Libraries/Blob'
  pod 'React-RCTImage', :path => '../node_modules/react-native/Libraries/Image'
  pod 'React-RCTLinking', :path => '../node_modules/react-native/Libraries/LinkingIOS'
  pod 'React-RCTNetwork', :path => '../node_modules/react-native/Libraries/Network'
  pod 'React-RCTSettings', :path => '../node_modules/react-native/Libraries/Settings'
  pod 'React-RCTText', :path => '../node_modules/react-native/Libraries/Text'
  pod 'React-RCTVibration', :path => '../node_modules/react-native/Libraries/Vibration'
  pod 'React-Core/RCTWebSocket', :path => '../node_modules/react-native/'

  pod 'React-cxxreact', :path => '../node_modules/react-native/ReactCommon/cxxreact'
  pod 'React-jsi', :path => '../node_modules/react-native/ReactCommon/jsi'
  pod 'React-jsiexecutor', :path => '../node_modules/react-native/ReactCommon/jsiexecutor'
  pod 'React-jsinspector', :path => '../node_modules/react-native/ReactCommon/jsinspector'
  pod 'ReactCommon/jscallinvoker', :path => "../node_modules/react-native/ReactCommon"
  pod 'ReactCommon/turbomodule/core', :path => "../node_modules/react-native/ReactCommon"
  pod 'Yoga', :path => '../node_modules/react-native/ReactCommon/yoga'

  # Facebook SDK
  pod 'FBSDKCoreKit'
  pod 'FBSDKLoginKit'

  # Google Signin SDL
  pod 'GoogleSignIn', '~> 5.0.2'

  # react-native-image-crop-picker
  pod 'QBImagePickerController', :path => '../node_modules/react-native-image-crop-picker/ios/QBImagePicker/QBImagePickerController.podspec'

  target 'TestTests' do
    inherit! :search_paths
    # Pods for testing
  end

  use_native_modules!
end

target 'Test-tvOS' do
  # Pods for Test-tvOS

  target 'HeroRN-tvOSTests' do
    inherit! :search_paths
    # Pods for testing
  end

end
```

#### `AppDelegate.m`:

```objc
/**
 * Copyright (c) Facebook, Inc. and its affiliates.
 *
 * This source code is licensed under the MIT license found in the
 * LICENSE file in the root directory of this source tree.
 */

@import Firebase;
#import "AppDelegate.h"

#import <React/RCTBridge.h>
#import <React/RCTRootView.h>
#import <FBSDKCoreKit/FBSDKCoreKit.h>
#import <RNGoogleSignin/RNGoogleSignin.h>
#import <AppCenterReactNative.h>
#import <AppCenterReactNativeAnalytics.h>
#import <AppCenterReactNativeCrashes.h>
#import <CodePush/CodePush.h>

@implementation AppDelegate

{
  RCTBridge *bridge = [[RCTBridge alloc] initWithDelegate:self launchOptions:launchOptions];
  RCTRootView *rootView = [[RCTRootView alloc] initWithBridge:bridge
                                                   moduleName:@"Test"
                                            initialProperties:nil];

  rootView.backgroundColor = [[UIColor alloc] initWithRed:1.0f green:1.0f blue:1.0f alpha:1];

  self.window = [[UIWindow alloc] initWithFrame:[UIScreen mainScreen].bounds];
  UIViewController *rootViewController = [UIViewController new];
  rootViewController.view = rootView;
  self.window.rootViewController = rootViewController;
  [self.window makeKeyAndVisible];
  
  if ([FIRApp defaultApp] == nil) {
    [FIRApp configure];
  }
  [[FBSDKApplicationDelegate sharedInstance] application:application
  didFinishLaunchingWithOptions:launchOptions];
  [AppCenterReactNative register];
  [AppCenterReactNativeAnalytics registerWithInitiallyEnabled:true];
  [AppCenterReactNativeCrashes registerWithAutomaticProcessing];
  return YES;
}

- (BOOL)application:(UIApplication *)application
            openURL:(NSURL *)url
            options:(NSDictionary<UIApplicationOpenURLOptionsKey,id> *)options {

  BOOL handled =  [[FBSDKApplicationDelegate sharedInstance] application:application openURL:url options:options] || [RNGoogleSignin application:application openURL:url options:options];
  // Add any custom logic here.
  return handled;
}

{
#if DEBUG
#else
  return [CodePush bundleURL];
#endif
}

@end
```

</p>
</details>

---

### Android

<p>

#### `android/build.gradle`:

```groovy
// Top-level build file where you can add configuration options common to all sub-projects/modules.

buildscript {
    ext {
        buildToolsVersion = "28.0.3"
        minSdkVersion = 16
        compileSdkVersion = 28
        targetSdkVersion = 28
        googlePlayServicesAuthVersion = "16.0.1"
    }
    repositories {
        google()
        jcenter()
    }
    dependencies {
        classpath("com.android.tools.build:gradle:3.4.2")
        classpath("com.google.gms:google-services:4.3.3")

        // NOTE: Do not place your application dependencies here; they belong
        // in the individual module build.gradle files
    }
}

allprojects {
    repositories {
        mavenLocal()
        maven {
            // All of React Native (JS, Obj-C sources, Android binaries) is installed from npm
            url("$rootDir/../node_modules/react-native/android")
        }
        maven {
            // Android JSC is installed from npm
            url("$rootDir/../node_modules/jsc-android/dist")
        }

        google()
        jcenter()
        maven { url 'https://jitpack.io' }
    }
}

```

```groovy
apply plugin: "com.android.application"

import com.android.build.OutputFile

/**
 * The react.gradle file registers a task for each build variant (e.g. bundleDebugJsAndAssets
 * and bundleReleaseJsAndAssets).
 * These basically call `react-native bundle` with the correct arguments during the Android build
 * cycle. By default, bundleDebugJsAndAssets is skipped, as in debug/dev mode we prefer to load the
 * bundle directly from the development server. Below you can see all the possible configurations
 * and their defaults. If you decide to add a configuration block, make sure to add it before the
 * `apply from: "../../node_modules/react-native/react.gradle"` line.
 *
 * project.ext.react = [
 *   // the name of the generated asset file containing your JS bundle
 *   bundleAssetName: "index.android.bundle",
 *
 *   // the entry file for bundle generation
 *   entryFile: "index.android.js",
 *
 *   // https://facebook.github.io/react-native/docs/performance#enable-the-ram-format
 *   bundleCommand: "ram-bundle",
 *
 *   // whether to bundle JS and assets in debug mode
 *   bundleInDebug: false,
 *
 *   // whether to bundle JS and assets in release mode
 *   bundleInRelease: true,
 *
 *   // whether to bundle JS and assets in another build variant (if configured).
 *   // See http://tools.android.com/tech-docs/new-build-system/user-guide#TOC-Build-Variants
 *   // The configuration property can be in the following formats
 *   //         'bundleIn${productFlavor}${buildType}'
 *   //         'bundleIn${buildType}'
 *   // bundleInFreeDebug: true,
 *   // bundleInPaidRelease: true,
 *   // bundleInBeta: true,
 *
 *   // whether to disable dev mode in custom build variants (by default only disabled in release)
 *   // for example: to disable dev mode in the staging build type (if configured)
 *   devDisabledInStaging: true,
 *   // The configuration property can be in the following formats
 *   //         'devDisabledIn${productFlavor}${buildType}'
 *   //         'devDisabledIn${buildType}'
 *
 *   // the root of your project, i.e. where "package.json" lives
 *   root: "../../",
 *
 *   // where to put the JS bundle asset in debug mode
 *   jsBundleDirDebug: "$buildDir/intermediates/assets/debug",
 *
 *   // where to put the JS bundle asset in release mode
 *   jsBundleDirRelease: "$buildDir/intermediates/assets/release",
 *
 *   // where to put drawable resources / React Native assets, e.g. the ones you use via
 *   // require('./image.png')), in debug mode
 *   resourcesDirDebug: "$buildDir/intermediates/res/merged/debug",
 *
 *   // where to put drawable resources / React Native assets, e.g. the ones you use via
 *   // require('./image.png')), in release mode
 *   resourcesDirRelease: "$buildDir/intermediates/res/merged/release",
 *
 *   // by default the gradle tasks are skipped if none of the JS files or assets change; this means
 *   // that we don't look at files in android/ or ios/ to determine whether the tasks are up to
 *   // date; if you have any other folders that you want to ignore for performance reasons (gradle
 *   // indexes the entire tree), add them here. Alternatively, if you have JS files in android/
 *   // for example, you might want to remove it from here.
 *   inputExcludes: ["android/**", "ios/**"],
 *
 *   // override which node gets called and with what additional arguments
 *   nodeExecutableAndArgs: ["node"],
 *
 *   // supply additional arguments to the packager
 *   extraPackagerArgs: []
 * ]
 */

project.ext.react = [
    entryFile: "index.js",
    enableHermes: false,  // clean and rebuild if changing
]

project.ext.vectoricons = [
    iconFontNames: ['Feather.ttf']
]

apply from: "../../node_modules/react-native/react.gradle"
apply from: "../../node_modules/react-native-code-push/android/codepush.gradle"
apply from: "../../node_modules/react-native-vector-icons/fonts.gradle"

/**
 * Set this to true to create two separate APKs instead of one:
 *   - An APK that only works on ARM devices
 *   - An APK that only works on x86 devices
 * The advantage is the size of the APK is reduced by about 4MB.
 * Upload all the APKs to the Play Store and people will download
 * the correct one based on the CPU architecture of their device.
 */
def enableSeparateBuildPerCPUArchitecture = false

/**
 * Run Proguard to shrink the Java bytecode in release builds.
 */
def enableProguardInReleaseBuilds = false

/**
 * The preferred build flavor of JavaScriptCore.
 *
 * For example, to use the international variant, you can use:
 * `def jscFlavor = 'org.webkit:android-jsc-intl:+'`
 *
 * The international variant includes ICU i18n library and necessary data
 * allowing to use e.g. `Date.toLocaleString` and `String.localeCompare` that
 * give correct results when using with locales other than en-US.  Note that
 * this variant is about 6MiB larger per architecture than default.
 */
def jscFlavor = 'org.webkit:android-jsc:+'

/**
 * Whether to enable the Hermes VM.
 *
 * This should be set on project.ext.react and mirrored here.  If it is not set
 * on project.ext.react, JavaScript will not be compiled to Hermes Bytecode
 * and the benefits of using Hermes will therefore be sharply reduced.
 */
def enableHermes = project.ext.react.get("enableHermes", false);

android {
    compileSdkVersion rootProject.ext.compileSdkVersion

    compileOptions {
        sourceCompatibility JavaVersion.VERSION_1_8
        targetCompatibility JavaVersion.VERSION_1_8
    }

    dexOptions {
        javaMaxHeapSize "4g"
    }

    defaultConfig {
        applicationId "com.test.test"
        minSdkVersion rootProject.ext.minSdkVersion
        targetSdkVersion rootProject.ext.targetSdkVersion
        versionCode 1
        versionName "1.0"
        multiDexEnabled true
    }
    splits {
        abi {
            reset()
            enable enableSeparateBuildPerCPUArchitecture
            universalApk false  // If true, also generate a universal APK
            include "armeabi-v7a", "x86", "arm64-v8a", "x86_64"
        }
    }
    signingConfigs {
        debug {
            storeFile file('debug.keystore')
            storePassword 'android'
            keyAlias 'androiddebugkey'
            keyPassword 'android'
        }
        release {
            if (project.hasProperty('MYAPP_UPLOAD_STORE_FILE')) {
                storeFile file(MYAPP_UPLOAD_STORE_FILE)
                storePassword MYAPP_UPLOAD_STORE_PASSWORD
                keyAlias MYAPP_UPLOAD_KEY_ALIAS
                keyPassword MYAPP_UPLOAD_KEY_PASSWORD
            }
        }
    }
    buildTypes {
        debug {
            signingConfig signingConfigs.debug
        }
        release {
            // Caution! In production, you need to generate your own keystore file.
            // see https://facebook.github.io/react-native/docs/signed-apk-android.
            signingConfig signingConfigs.release
            minifyEnabled enableProguardInReleaseBuilds
            proguardFiles getDefaultProguardFile("proguard-android.txt"), "proguard-rules.pro"
        }
    }
    // applicationVariants are e.g. debug, release
    applicationVariants.all { variant ->
        variant.outputs.each { output ->
            // For each separate APK per architecture, set a unique version code as described here:
            // https://developer.android.com/studio/build/configure-apk-splits.html
            def versionCodes = ["armeabi-v7a": 1, "x86": 2, "arm64-v8a": 3, "x86_64": 4]
            def abi = output.getFilter(OutputFile.ABI)
            if (abi != null) {  // null for the universal-debug, universal-release variants
                output.versionCodeOverride =
                        versionCodes.get(abi) * 1048576 + defaultConfig.versionCode
            }

        }
    }
}

dependencies {
    implementation fileTree(dir: "libs", include: ["*.jar"])
    implementation "com.facebook.react:react-native:+"  // From node_modules
    implementation 'androidx.appcompat:appcompat:1.1.0-rc01'
    implementation 'androidx.swiperefreshlayout:swiperefreshlayout:1.1.0-alpha02'
    implementation 'androidx.multidex:multidex:2.0.1'

    if (enableHermes) {
        def hermesPath = "../../node_modules/hermes-engine/android/";
        debugImplementation files(hermesPath + "hermes-debug.aar")
        releaseImplementation files(hermesPath + "hermes-release.aar")
    } else {
        implementation jscFlavor
    }
}

// Run this once to be able to run the application with BUCK
// puts all compile dependencies into folder libs for BUCK to use
task copyDownloadableDepsToLibs(type: Copy) {
    from configurations.compile
    into 'libs'
}

apply from: file("../../node_modules/@react-native-community/cli-platform-android/native_modules.gradle"); applyNativeModulesAppBuildGradle(project)
apply plugin: 'com.google.gms.google-services'
```

```groovy
rootProject.name = 'Test'
apply from: file("../node_modules/@react-native-community/cli-platform-android/native_modules.gradle"); applyNativeModulesSettingsGradle(settings)
include ':app'

```

#### `MainApplication.java`:

```java
package com.test.test;

import androidx.multidex.MultiDexApplication;
import android.content.Context;
import com.facebook.react.PackageList;
import com.facebook.react.ReactApplication;
import com.facebook.react.ReactNativeHost;
import com.facebook.react.ReactPackage;
import com.facebook.soloader.SoLoader;
import java.lang.reflect.InvocationTargetException;
import java.util.List;
import com.microsoft.codepush.react.CodePush;

public class MainApplication extends MultiDexApplication implements ReactApplication {

  private final ReactNativeHost mReactNativeHost =
      new ReactNativeHost(this) {
        @Override
        public boolean getUseDeveloperSupport() {
          return BuildConfig.DEBUG;
        }

        @Override
        protected String getJSBundleFile() {
            return CodePush.getJSBundleFile();
        }

        @Override
        protected List<ReactPackage> getPackages() {
          @SuppressWarnings("UnnecessaryLocalVariable")
          List<ReactPackage> packages = new PackageList(this).getPackages();
          // Packages that cannot be autolinked yet can be added manually here, for example:
          // packages.add(new MyReactNativePackage());
          return packages;
        }

        @Override
        protected String getJSMainModuleName() {
          return "index";
        }
      };

  @Override
  public ReactNativeHost getReactNativeHost() {
    return mReactNativeHost;
  }

  @Override
  public void onCreate() {
    super.onCreate();
    SoLoader.init(this, /* native exopackage */ false);
    initializeFlipper(this); // Remove this line if you don't want Flipper enabled
  }

  /**
   * Loads Flipper in React Native templates.
   *
   * @param context
   */
  private static void initializeFlipper(Context context) {
    if (BuildConfig.DEBUG) {
      try {
        /*
         We use reflection here to pick up the class that initializes Flipper,
        since Flipper library is not available in release mode
        */
        Class<?> aClass = Class.forName("com.facebook.flipper.ReactNativeFlipper");
        aClass.getMethod("initializeFlipper", Context.class).invoke(null, context);
      } catch (ClassNotFoundException e) {
        e.printStackTrace();
      } catch (NoSuchMethodException e) {
        e.printStackTrace();
      } catch (IllegalAccessException e) {
        e.printStackTrace();
      } catch (InvocationTargetException e) {
        e.printStackTrace();
      }
    }
  }
}

```

#### `AndroidManifest.xml`:

```xml
<manifest xmlns:android="http://schemas.android.com/apk/res/android"
  package="com.test.test">

    <uses-permission android:name="android.permission.INTERNET" />
    <uses-permission android:name="android.permission.CAMERA" />
    <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE"/>

    <application
      android:name=".MainApplication"
      android:label="@string/app_name"
      android:icon="@mipmap/ic_launcher"
      android:roundIcon="@mipmap/ic_launcher_round"
      android:allowBackup="false"
      android:theme="@style/AppTheme">
      <activity
        android:name=".MainActivity"
        android:label="@string/app_name"
        android:configChanges="keyboard|keyboardHidden|orientation|screenSize"
        android:windowSoftInputMode="adjustPan">
        <intent-filter>
            <action android:name="android.intent.action.MAIN" />
            <category android:name="android.intent.category.LAUNCHER" />
        </intent-filter>
      </activity>
      <activity android:name="com.facebook.react.devsupport.DevSettingsActivity" />
      <meta-data android:name="com.facebook.sdk.ApplicationId" android:value="@string/facebook_app_id" />
    </application>

</manifest>

```

</p>
</details>

---

## Environment

<p>

```
System:
    OS: macOS 10.15.2
    CPU: (12) x64 Intel(R) Core(TM) i7-8850H CPU @ 2.60GHz
    Memory: 112.74 MB / 16.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 13.3.0 - /usr/local/bin/node
    Yarn: 1.21.1 - /usr/local/bin/yarn
    npm: 6.13.2 - /usr/local/bin/npm
  SDKs:
    iOS SDK:
      Platforms: iOS 13.2, DriverKit 19.0, macOS 10.15, tvOS 13.2, watchOS 6.1
    Android SDK:
      API Levels: 23, 26, 27, 28
      Build Tools: 23.0.1, 26.0.2, 27.0.3, 28.0.1, 28.0.3
      System Images: android-28 | Google APIs Intel x86 Atom, android-28 | Google Play Intel x86 Atom
  IDEs:
    Android Studio: 3.5 AI-191.8026.42.35.5977832
    Xcode: 11.3/11C29 - /usr/bin/xcodebuild
  npmPackages:
    react: 16.9.0 => 16.9.0 
    react-native: 0.61.5 => 0.61.5
```

  - [ ] iOS
  - [ ] Android
  - [x] Both
  - `6.2.0`
  - `Firestore`
  - `N`
  
</p>
</details>

---

Think `react-native-firebase` is great? Please consider supporting all of the project maintainers and contributors by donating via our [Open Collective](https://opencollective.com/react-native-firebase/donate) where all contributors can submit expenses. [[Learn More]](https://invertase.io/oss/react-native-firebase/contributing/donations-expenses)

Yonom:
Fixes #3004.

## Test Plan

- call `firestore.collection('myCollection').where('myField', 'in', [0])` and confirm that correct results are being returned

### DIFF
--- a/packages/firestore/e2e/issues.e2e.js
+++ b/packages/firestore/e2e/issues.e2e.js
@@ -16,6 +16,23 @@
  */
 
 const COLLECTION = 'firestore';
+const { getE2eEmulatorHost } = require('@react-native-firebase/app/e2e/helpers');
+const jsFirebase = require('firebase/compat/app');
+const { wipe } = require('./helpers');
+require('firebase/compat/firestore');
+
+const testNumbers = {
+  zero: 0, // int
+  negativeZero: -0, // double
+  half: 0.5, // double
+  unsafeInt: Number.MAX_SAFE_INTEGER + 1, // double
+  nagativeUnsafeInt: Number.MIN_SAFE_INTEGER - 1, // double
+  safeInt: Number.MAX_SAFE_INTEGER, // int
+  nagativeSafeInt: Number.MIN_SAFE_INTEGER, // int
+  inf: Infinity, // double
+  negativeInf: -Infinity, // double
+  // nan: NaN, // double -- where-in queries on NaN do not work
+};
 
 describe('firestore()', function () {
   describe('issues', function () {
@@ -90,6 +107,53 @@ describe('firestore()', function () {
           doc.id.should.equal('mr7MdAygvuheF6AUtWma');
         }
       });
+    });
+  });
+
+  describe('number type consistency', function () {
+    before(async function () {
+      jsFirebase.initializeApp(FirebaseHelpers.app.config());
+      jsFirebase.firestore().useEmulator(getE2eEmulatorHost(), 8080);
+
+      // Put one example of each number in our collection using JS SDK
+      await Promise.all(
+        Object.entries(testNumbers).map(([testName, testValue]) => {
+          return jsFirebase
+            .firestore()
+            .doc(`${COLLECTION}/numberTestsJS/cases/${testName}`)
+            .set({ number: testValue });
+        }),
+      );
+
+      // Put one example of each number in our collection using Native SDK
+      await Promise.all(
+        Object.entries(testNumbers).map(([testName, testValue]) => {
+          return firebase
+            .firestore()
+            .doc(`${COLLECTION}/numberTestsNative/cases/${testName}`)
+            .set({ number: testValue });
+        }),
+      );
+    });
+
+    it('types inserted by JS may be queried by native with filters', async function () {
+      const testValues = Object.values(testNumbers);
+      const ref = firebase
+        .firestore()
+        .collection(`${COLLECTION}/numberTestsJS/cases`)
+        .where('number', 'in', testValues);
+      typesSnap = await ref.get();
+      should.deepEqual(typesSnap.docs.map(d => d.id).sort(), Object.keys(testNumbers).sort());
+    });
+
+    it('types inserted by native may be queried by JS with filters', async function () {
+      const testValues = Object.values(testNumbers);
+      const ref = jsFirebase
+        .firestore()
+        .collection(`${COLLECTION}/numberTestsNative/cases`)
+        .where('number', 'in', testValues);
+      typesSnap = await ref.get();
+      should.deepEqual(typesSnap.docs.map(d => d.id).sort(), Object.keys(testNumbers).sort());
     });
   });
 });
--- a/packages/firestore/e2e/issues.e2e.js
+++ b/packages/firestore/e2e/issues.e2e.js
@@ -18,7 +18,6 @@
 const COLLECTION = 'firestore';
 const { getE2eEmulatorHost } = require('@react-native-firebase/app/e2e/helpers');
 const jsFirebase = require('firebase/compat/app');
-const { wipe } = require('./helpers');
 require('firebase/compat/firestore');
 
 const testNumbers = {
--- a/packages/firestore/android/src/reactnative/java/io/invertase/firebase/firestore/ReactNativeFirebaseFirestoreSerialize.java
+++ b/packages/firestore/android/src/reactnative/java/io/invertase/firebase/firestore/ReactNativeFirebaseFirestoreSerialize.java
@@ -71,6 +71,7 @@ public class ReactNativeFirebaseFirestoreSerialize {
   private static final int INT_FIELDVALUE = 15;
   private static final int INT_OBJECT = 16;
   private static final int INT_INTEGER = 17;
+  private static final int INT_NEGATIVE_ZERO = 18;
   private static final int INT_UNKNOWN = -999;
 
   // Keys
@@ -462,6 +463,8 @@ public class ReactNativeFirebaseFirestoreSerialize {
         return true;
       case INT_BOOLEAN_FALSE:
         return false;
+      case INT_NEGATIVE_ZERO:
+        return -0.0;
       case INT_INTEGER:
         return (long) typeArray.getDouble(1);
       case INT_DOUBLE:
--- a/packages/firestore/android/src/reactnative/java/io/invertase/firebase/firestore/ReactNativeFirebaseFirestoreSerialize.java
+++ b/packages/firestore/android/src/reactnative/java/io/invertase/firebase/firestore/ReactNativeFirebaseFirestoreSerialize.java
@@ -60,7 +60,7 @@ public class ReactNativeFirebaseFirestoreSerialize {
   private static final int INT_DOCUMENTID = 4;
   private static final int INT_BOOLEAN_TRUE = 5;
   private static final int INT_BOOLEAN_FALSE = 6;
-  private static final int INT_NUMBER = 7;
+  private static final int INT_DOUBLE = 7;
   private static final int INT_STRING = 8;
   private static final int INT_STRING_EMPTY = 9;
   private static final int INT_ARRAY = 10;
@@ -70,6 +70,7 @@ public class ReactNativeFirebaseFirestoreSerialize {
   private static final int INT_BLOB = 14;
   private static final int INT_FIELDVALUE = 15;
   private static final int INT_OBJECT = 16;
+  private static final int INT_INTEGER = 17;
   private static final int INT_UNKNOWN = -999;
 
   // Keys
@@ -300,7 +301,7 @@ public class ReactNativeFirebaseFirestoreSerialize {
     }
 
     if (value instanceof Integer) {
-      typeArray.pushInt(INT_NUMBER);
+      typeArray.pushInt(INT_DOUBLE);
       typeArray.pushDouble(((Integer) value).doubleValue());
       return typeArray;
     }
@@ -325,19 +326,19 @@ public class ReactNativeFirebaseFirestoreSerialize {
         return typeArray;
       }
 
-      typeArray.pushInt(INT_NUMBER);
+      typeArray.pushInt(INT_DOUBLE);
       typeArray.pushDouble(doubleValue);
       return typeArray;
     }
 
     if (value instanceof Float) {
-      typeArray.pushInt(INT_NUMBER);
+      typeArray.pushInt(INT_DOUBLE);
       typeArray.pushDouble(((Float) value).doubleValue());
       return typeArray;
     }
 
     if (value instanceof Long) {
-      typeArray.pushInt(INT_NUMBER);
+      typeArray.pushInt(INT_DOUBLE);
       typeArray.pushDouble(((Long) value).doubleValue());
       return typeArray;
     }
@@ -461,14 +462,10 @@ public class ReactNativeFirebaseFirestoreSerialize {
         return true;
       case INT_BOOLEAN_FALSE:
         return false;
-      case INT_NUMBER:
-        // https://github.com/invertase/react-native-firebase/issues/3004
-        // Number values come from JS as Strings on Android so we can check for floating points
-        String numberStringValue = Objects.requireNonNull(typeArray.getString(1));
-        if (numberStringValue.contains(".")) {
-          return Double.valueOf(numberStringValue);
-        }
-        return Long.valueOf(numberStringValue, 10);
+      case INT_INTEGER:
+        return (long) typeArray.getDouble(1);
+      case INT_DOUBLE:
+        return typeArray.getDouble(1);
       case INT_STRING:
         return typeArray.getString(1);
       case INT_STRING_EMPTY:
--- a/packages/firestore/ios/RNFBFirestore/RNFBFirestoreSerialize.m
+++ b/packages/firestore/ios/RNFBFirestore/RNFBFirestoreSerialize.m
@@ -46,7 +46,7 @@ enum TYPE_MAP {
   INT_DOCUMENTID,
   INT_BOOLEAN_TRUE,
   INT_BOOLEAN_FALSE,
-  INT_NUMBER,
+  INT_DOUBLE,
   INT_STRING,
   INT_STRING_EMPTY,
   INT_ARRAY,
@@ -56,6 +56,7 @@ enum TYPE_MAP {
   INT_BLOB,
   INT_FIELDVALUE,
   INT_OBJECT,
+  INT_INTEGER,
   INT_UNKNOWN = -999,
 };
 
@@ -336,7 +337,7 @@ enum TYPE_MAP {
     }
 
     // Number
-    typeArray[0] = @(INT_NUMBER);
+    typeArray[0] = @(INT_DOUBLE);
     typeArray[1] = value;
     return typeArray;
   }
@@ -400,7 +401,9 @@ enum TYPE_MAP {
       return @(YES);
     case INT_BOOLEAN_FALSE:
       return @(NO);
-    case INT_NUMBER:
+    case INT_INTEGER:
+      return @([typeMap[1] longLongValue]);
+    case INT_DOUBLE:
       return @([typeMap[1] doubleValue]);
     case INT_STRING:
       return typeMap[1];
--- a/packages/firestore/ios/RNFBFirestore/RNFBFirestoreSerialize.m
+++ b/packages/firestore/ios/RNFBFirestore/RNFBFirestoreSerialize.m
@@ -57,6 +57,7 @@ enum TYPE_MAP {
   INT_FIELDVALUE,
   INT_OBJECT,
   INT_INTEGER,
+  INT_NEGATIVE_ZERO,
   INT_UNKNOWN = -999,
 };
 
@@ -401,6 +402,8 @@ enum TYPE_MAP {
       return @(YES);
     case INT_BOOLEAN_FALSE:
       return @(NO);
+    case INT_NEGATIVE_ZERO:
+      return @(-0.0);
     case INT_INTEGER:
       return @([typeMap[1] longLongValue]);
     case INT_DOUBLE:
--- a/packages/firestore/lib/utils/serialize.js
+++ b/packages/firestore/lib/utils/serialize.js
@@ -140,8 +140,10 @@ export function generateNativeData(value, ignoreUndefined) {
   if (isNumber(value)) {
     // mirror the JS SDK's integer detection algorithm
     // https://github.com/firebase/firebase-js-sdk/blob/086df7c7e0299cedd9f3cff9080f46ca25cab7cd/packages/firestore/src/remote/number_serializer.ts#L56
-    const isNegativeZero = value === 0 && 1 / value === -Infinity;
-    if (Number.isSafeInteger(value) && !isNegativeZero) {
+    if (value === 0 && 1 / value === -Infinity) {
+      return getTypeMapInt('negativeZero');
+    }
+    if (Number.isSafeInteger(value)) {
       return getTypeMapInt('integer', value);
     }
     return getTypeMapInt('double', value);
@@ -258,6 +260,7 @@ export function parseNativeData(firestore, nativeArray) {
       return false;
     case 'double':
     case 'integer':
+    case 'negativeZero':
     case 'string':
       return value;
     case 'stringEmpty':
--- a/packages/firestore/lib/utils/serialize.js
+++ b/packages/firestore/lib/utils/serialize.js
@@ -16,7 +16,6 @@
  */
 
 import {
-  isAndroid,
   isArray,
   isBoolean,
   isDate,
@@ -139,10 +138,13 @@ export function generateNativeData(value, ignoreUndefined) {
   }
 
   if (isNumber(value)) {
-    if (isAndroid) {
-      return getTypeMapInt('number', value.toString());
+    // mirror the JS SDK's integer detection algorithm
+    // https://github.com/firebase/firebase-js-sdk/blob/086df7c7e0299cedd9f3cff9080f46ca25cab7cd/packages/firestore/src/remote/number_serializer.ts#L56
+    const isNegativeZero = value === 0 && 1 / value === -Infinity;
+    if (Number.isSafeInteger(value) && !isNegativeZero) {
+      return getTypeMapInt('integer', value);
     }
-    return getTypeMapInt('number', value);
+    return getTypeMapInt('double', value);
   }
 
   if (isString(value)) {
@@ -254,7 +256,8 @@ export function parseNativeData(firestore, nativeArray) {
       return true;
     case 'booleanFalse':
       return false;
-    case 'number':
+    case 'double':
+    case 'integer':
     case 'string':
       return value;
     case 'stringEmpty':
--- a/packages/firestore/lib/utils/typemap.js
+++ b/packages/firestore/lib/utils/typemap.js
@@ -25,7 +25,7 @@ const MAP = {
   documentid: 4, // to native only
   booleanTrue: 5,
   booleanFalse: 6,
-  number: 7,
+  double: 7,
   string: 8,
   stringEmpty: 9,
   array: 10,
@@ -35,6 +35,7 @@ const MAP = {
   blob: 14,
   fieldvalue: 15,
   object: 16,
+  integer: 17,
   unknown: -999,
 };
 
--- a/packages/firestore/lib/utils/typemap.js
+++ b/packages/firestore/lib/utils/typemap.js
@@ -36,6 +36,7 @@ const MAP = {
   fieldvalue: 15,
   object: 16,
   integer: 17,
+  negativeZero: 18,
   unknown: -999,
 };
 


mikehardy (approved):
Oh my - that's one of those fixes that's so obvious when viewed as a diff in a PR.
There is a mystery embedded in the issues/PRs/tests where the question is: with #3895 we are testing numbers in where clauses, and it's working for us in our e2e rig, but obviously it is failing in practice. Why? I believe I have the answer, our e2e rig is a bit special and it's running the bundle in the e2e app via the javascript debugging interface on a node instance, on linux. Stated differently, the "isAndroid" checks are querying the node VM not the react-native app container, and since those run on unix (Darwin or Ubuntu) and android is also linux, they're answering "yes" to "isAndroid" in the javascript check.

So my hypothesis is that our test was always doing a string conversion. Remains to be proven that is happening, and if so how the backend is handling a string query on number contents (maybe it's auto-converting?) such that it works, but that's my hypothesis for why the test we already have is false-negative.

I'm preliminarily approving this (and really really appreciate it!) but I need to poke at it a little from the testing side to see if I can get tests to fail before merging.

That shouldn't be a blocker though, we release very frequently so the diffs are small, and we generate a patch-package patch set as one of our CI actions, so you may trivially integrate this fix in a reliable (via patch-package) way and not have to maintain a fork or worry about updates etc.

I won't delay long on merge+release either way

Yonom:
@mikehardy 

> So my hypothesis is that our test was always doing a string conversion. Remains to be proven that is happening, and if so how the backend is handling a string query on number contents (maybe it's auto-converting?) such that it works, but that's my hypothesis for why the test we already have is false-negative.

The (firebase) backend never received strings on either platform. Before this PR, the old code `return @([typeMap[1] doubleValue]);` took care of converting strings to doubles if they were passed as such. (Hence my note on the fact that this changelog is backwards compatible with existing native library versions).

In my tests, the firebase backend is stricter and does not accept neither a string nor a double as equal to an integer value due to the different type. So that shouldn't be a reason for tests not failing.

Idk why your tests aren't failing, but here are some thoughts:
- This only happens if the object fields are integers. Does the iOS SDK save integer object fields as long? The Python and JS SDKs certainly do.
- It silently fails and returns an empty array. 

---

To be honest, I consider this a bug in the Firebase iOS SDK though. The JS/Python SDKs must do some automatic integer detection to save `0.0` as an integer instead of double. This automatic conversion is left as an exercise for the user (in this case, react-native-firebase) for the iOS SDK.

I don't care enough about the iOS SDK to file a bug though, so this is an exercise for the reader. 😊

Yonom:
Oh look your tests are failing! 😄

---

Jokes aside; I've made up my mind on why your test's aren't failing on the mentioned issue though:

`await ref.add({ status: 1 });`

This saves 1 (double) as the value for `status`. Therefore, a query for double succeeds.

How to verify:

1. set rules for a firestore object to allow ALL: `allow read, write: if true;`
2. write to the object to firestore `await ref.add({ status: 1 });`
3. go to https://firestore.googleapis.com/v1/projects/$PORJECT_ID/databases/(default)/documents/$COLLECTION_ID/$DOCUMENT_ID

View the output:

```json
{
  "name": "projects/<redacted>/databases/(default)/documents/config/test",
  "fields": {
    "status": {
      "doubleValue": 1
    }
  },
  "createTime": "2021-10-14T13:15:10.763946Z",
  "updateTime": "2021-11-07T19:25:48.364106Z"
}
```

mikehardy:
Ah, so if I understand correctly / just restating: our tests are passing because setting a double then checking a double of course works. But there is no ability in the current react-native-firebase to set an *int* then check an int - but if we had (for instance) a test where there was already an int in there (not set via our android implementation as it exists, since it will never set an int) then it would fail. And that makes sense and would explain it completely.

Our tests are quite flaky at the moment unfortunately - android is a rarer flake but iOS has a serious problem right now with Detox that unfortunately causes the app to crash quickly. I restarted both those test runs though and I'm pretty sure they'll roll through if the flakiness doesn't halt them

Yonom:
No worries!

Some thing to keep in mind: 

If you have a case where there was already an **integer saved as double** (set by the iOS implementation as it currently exists) then the same object will no longer be found via `.where` after this patch is applied.

This makes this change a minor breaking change 😱. Mitigation would be to go through your whole DB and load and re-save every object (I've been there, done that)

mikehardy:
I'm okay with major version releases, as a consumer I actually prefer them to be conservative (implying this would be a major version) just with full details on exactly how to know if it affects you and what to do if it does. This will likely be one of those, now that you mention it. I handle that when I do the merge via special commit message tokens for conventional commits to work through

https://firebase.google.com/docs/firestore/manage-data/data-types indicates there is "floating point number" and "integer". I'm still getting the work bench set up to really look at this, and I'll inspect it from that perspective. Based on the related android PR I think this handles all the branches but I'll look at android and ios and have a think about it

I really appreciate all the thought you've put into it and the details you've commented on, they are very helpful

Yonom:
One more aspect that I thought about - there are multiple ways to detect double vs integer:

- toString() and check for the existence of "." (current approach)
- pass double and check for `n % 1 !== 0`

I chose the first because the android version used the same approach and it makes sense to apply the same design over to iOS. However, I was worried that the conversion to string and back to double would cause bits to get lost over the wire, so I wrote this quick and dirty test:

```js
for (let i = 0; i < 100000000; i++) {
  const n = Number.EPSILON * i;
  console.assert(n === parseFloat(n.toString()));
}
// no logs encountered
```

So I assume it's safe. Just documenting it here for your review, in case I'm wrong.

mikehardy:
Sorry for the delay on this! My work queue unexpectedly got very deep, very quickly. Mentally the delay is mostly about how to handle + message the breaking change, the change itself looks good to me still. For interested people this is the exact reason we generate patch-package patch-sets in CI though - if you need this, please grab it, and if you do that and use it please report any success or failure.

Thanks, and thanks for your patience

Yonom:
There are still several cases that are handled _incorrectly_ (differently than the JS SDK) by react-native-firebase:

| Value                       | JS SDK      | react-native-firebase (not tested, just my guess) |
|-----------------------------|-------------|-----------------------|
| -0                          | doubleValue | integerValue          |
| Number.MAX_SAFE_INTEGER + 2 | doubleValue | integerValue          |
| Number.MIN_SAFE_INTEGER - 2 | doubleValue | integerValue          |
| NaN                         | doubleValue | ??                     |
| Infinity                    | doubleValue | ??                    |
| -Infinity                   | doubleValue | ??                    |


Here are the relevant code lines in the JS SDK:
https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore/src/remote/number_serializer.ts#L56
https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore/src/util/types.ts#L44

---

**_Off topic:_**
I just realized that the nodejs sdk does not handle negative zero correctly either: https://github.com/googleapis/nodejs-firestore/blob/main/dev/src/serializer.ts#L121
I'll create create a bug report for that...

mikehardy:
Github does not offer me the :fearful: emoji as a reaction on your comment but it's appropriate. Really hard to get things right, but also it's math with an example implementation - there *is* a right answer - and we should get it right. Any further commits on this PR happily accepted, or a separate PR if you have time. Can stage them all as one "Hey everyone! We're getting numbers typed correctly in firebase now..." breaking release

Yonom:
Yeah I'll add it to this PR (probably get to doing it on Monday)

mikehardy:
This literally just bit me on a project, and now I'm using this patch-package too :see_no_evil: - I need to do the needful on the release here and get it out. I would have been stumped if I hadn't known about the issue and known about this PR.

mikehardy:
That didn't quite go as planned, it looks like the item is not actually a string as it pops out of the type map for some reason:
Screen Region 2021-11-24 at 02 00 38

mikehardy:
Really strange, in our e2e test harness I do not reproduce the crash but I've gone back and forth on my work project with the patch twice and it reproduces with the patch each time. I'm not certain what is wrong but still suspect something about my integration via the patch s incorrect, just haven't found it yet.

I did successfully probe the failure though, so we finally have a test case that correctly breaks with our existing broken implementation 

https://github.com/invertase/react-native-firebase/commit/f9ce428de0c584499642d1020c0a1b17ab15c6d5#diff-b7d5d6ae9890f87025278e8b804153bd92e1a0c6a6c9ab78acdf9ac2a8015855R178-R184

mikehardy:
It was my patch integration, the patch checks out :100: in a test integration with a real project.
The local problem I had is that I'm testing in a distributed cluster of emulators and browsers and simulators with them all pointing to bundler on one machine over a ssh port forwards. My native build (on a different machine) was out of sync with the bundle build (on my local machine) with regard to whether the patch was integrated or not, so the types were going in to native from JS without your change to string from number on the JS side. Totally my fault.

Okay, this is still on deck to integrate, and now it's making a contract project of mine work correctly ;-). Many thanks.

Yonom:
You can expect my fixes for the edge cases within the next 12 hours. 

I have decided to remove the toString conversion and pass double on both platforms. On the native side I check for bounds around MAX_SAFE_NUMBER and if the double value is an integer (analogous to how the JS SDK does it)

I have yet to implement negative zero handling but I can confirm that negative zero crosses the bridge without issues 👍

mikehardy:
Fantastic!

Yonom:
Okay, I've moved the int / double detection to the JS side and added a separate data type for double / integer.

This is what I'm using for my tests:
```js
await firebase.firestore().collection('test').doc('test').set({
  zero: 0, // int
  negativeZero: -0, // double
  half: 0.5, // double
  one: 1, // int
  negativeOne: -1, // int
  unsafeInt: Number.MAX_SAFE_INTEGER + 1, // double
  nagativeUnsafeInt: Number.MIN_SAFE_INTEGER - 1, // double
  safeInt: Number.MAX_SAFE_INTEGER, // int
  nagativeSafeInt: Number.MIN_SAFE_INTEGER, // int
  nan: NaN, // double
  inf: Infinity, // double
  negativeInf: -Infinity, // double
});
```

mikehardy (commented):
This looks good - do you think you could build on the test I put together here as a follow-on commit? This really needs automated verification I think and I actually have working code that may be applied 

Then instead of a comment telling me how you are testing, we can just see the test in the code and know it ran :-) https://github.com/invertase/react-native-firebase/commit/f9ce428de0c584499642d1020c0a1b17ab15c6d5

Taking that wholesale and dropping it + mutating is fine, I'm not concerned about attribution or provenance or anything, just want more working test coverage :)

chiendv:
@Yonom could you tell me how to test your fix?

I did changed package.json like bellow but not sure it correct or not because I still get return 0 in querySnapshot
```
    "@react-native-firebase/app": "88a82599b37d8a6cb65945671c2429c94f96286b",
    "@react-native-firebase/auth": "88a82599b37d8a6cb65945671c2429c94f96286b",
    "@react-native-firebase/firestore": "88a82599b37d8a6cb65945671c2429c94f96286b",
```

the issue I getting is return 0 when use object to query in array field, it is ok when I change to string.
But it only happen on iO, it's working fire on Android.

```
   .where('my_field', 'array-contains-any', [
      { 'id': '1', 'type': 'a' }, // work
      { 'id': 1, 'type': 'a' }, //  not work
  ])
```
thank you

mikehardy:
@chiendv I used this added to the e2e tests in order to test it: https://github.com/invertase/react-native-firebase/commit/f9ce428de0c584499642d1020c0a1b17ab15c6d5#diff-b7d5d6ae9890f87025278e8b804153bd92e1a0c6a6c9ab78acdf9ac2a8015855R178-R184

That test definitely fails with existing released code, but was fixed by the original version of this patch. I haven't really exercised the current version of the patch but I do have it integrated in a work project now (unreleased) and was hoping to test it shortly.

If you are not sure you integrated it correctly via git reference, you can do like I do and try using patch-package (https://github.com/ds300/patch-package) and integrating the patch set including all our unreleased changes (they are minor, and safe - just waiting for release) along with this change: https://github.com/invertase/react-native-firebase/actions/runs/1503114495

chiendv:
@mikehardy thank your help, I got it work on my project now.

mikehardy:
Fantastic, and I'm happy to hear the success report (though I'm not surprised: @Yonom appears to have done great work here and I really need to get it released!). Good luck with your project

Yonom:
Finally got around to writing tests! I caught two issues.

## -0 does not cross the bridge safely in the test environment (fixed)

The native side was receiving 0 instead of -0 on both Android and iOS, which resulted in the following being stored:
```     
"number": {
  "doubleValue": 0.0
}
```

It works on my other app, I can't figure out why... JSI? TurboModules? Hermes? 🤷
I assume the bridge payload is getting stringified somewhere, since: `-0.toString() === '0'` and `JSON.stringify(-0) === '0'`

To be sure that it works everywhere, I added an extra data type `INT_NEGATIVE_ZERO`, similar to the other edge cases (NaN, Inf, -Inf).

## can't query where-in NaN (wontfix)

Querying `.where('x', 'in', [NaN])` does not return any results
It doesn't return anything in the JS SDK either, so we are on parity with the official client, good enough IMO!

packages/firestore/e2e/issues.e2e.js
```diff
@@ -16,9 +16,26 @@
...
+  safeInt: Number.MAX_SAFE_INTEGER, // int
+  nagativeSafeInt: Number.MIN_SAFE_INTEGER, // int
+  inf: Infinity, // double
+  negativeInf: -Infinity, // double
+  // nan: NaN, // double -- where-in queries on NaN do not work
+};
 
 describe('firestore()', function () {
-  describe(COLLECTION, function () {
+  describe('pagination', function () {
```
mikehardy, line 38:
```suggestion
  describe(COLLECTION, function () {
```

I don't understand this change, it seems unrelated, can it be left as is?

mikehardy (changes requested):
Your experience matches my experience with regard to testing: there is nothing so cleansing as an automated test...the corner cases discovered, the surprise "enjoyed" by what is found ;-)

This looks fantastic. One tiny comment

mikehardy (dismissed):
It's a thing of beauty. I imagine you are now the master of all-thing-number-related when it comes to how a) Javascript deals with them and b) how the RN bridge deals with them and c) their native representations. Believe it or not that'll probably come in handy for life. I knew much of this stuff but not all of it, and learned quite a bit as you made this PR, thanks!

packages/firestore/e2e/issues.e2e.js
Yonom: Oh right, I based my changes on https://github.com/invertase/react-native-firebase/commit/f9ce428de0c584499642d1020c0a1b17ab15c6d5#diff-b7d5d6ae9890f87025278e8b804153bd92e1a0c6a6c9ab78acdf9ac2a8015855R178-R184 and that's where this change came from. I have reverted it now :)

mikehardy:
My work stack is finally unwinding! Preparing to merge this. Looking for guidance on the changelog breaking change note. Currently thought:

BREAKING CHANGE: Previous versions had an issue with number type handling during save on iOS. Integers would be saved as doubles, incorrectly. Previously you had to save numbers as strings if you wanted `.where` queries to work cross-platform. Number types are now handled correctly. However, If you have integers saved (incorrectly!) as double (from previous versions) and you use `where / in` style queries on numbers, then the same document will no longer be found via `.where`. Mitigation could be to go through your whole DB and load and re-save the integers correctly, or alter queries. Please test your `where / in` queries that use number types.

mikehardy:
I think I've mitigated iOS E2E CI flakiness to a large degree with #5932 - I'm going to re-base and repush this branch to hopefully get CI green pre-merge

mikehardy (approved):
Just wrestling with CI at this point, my hoped-for iOS E2E flakiness mitigation is unfortunately not the fix I was hoping for. Anyway, this is on deck, RNFB's getting a release with a bunch of pent-up fixes shortly and this will be in

Yonom:
Awesome! :)

PR #5898, opened 2021-12-01
feat(firestore): stricter DocumentData type

Yonom:
### Description

Beginning with react-native-firebase 12, setting `undefined` in a firestore object throws instead of setting it to `null`. However, typescript does not warn you if you do something like:

```ts
await ref.set({
  test: undefined,
});
```

This PR aims to fix that.

### Release Summary

feat(firestore): stricter DocumentData type

### Checklist

  - [x] Yes
  - [x] `Android`
  - [x] `iOS`
- My change includes tests;
- This is a breaking change;
  - [x] Yes (but only typescript types)
  - [ ] No

### Test Plan

Set an object value to `undefined` and ensure that typescript shows an error:

image

### DIFF
--- a/packages/firestore/lib/index.d.ts
+++ b/packages/firestore/lib/index.d.ts
@@ -90,7 +90,7 @@ export namespace FirebaseFirestoreTypes {
    * A `DocumentData` object represents the data in a document.
    */
   export interface DocumentData {
-    [key: string]: any;
+    [key: string]: DocumentFieldType;
   }
 
   /**


Yonom:
I just applied this to a ~ 10k LOC project of mine where it caught some potential bugs involving undefined values. 

I was worried about the way Typescript would deal with optional properties without [`exactOptionalPropertyTypes`](https://devblogs.microsoft.com/typescript/announcing-typescript-4-4/#exact-optional-property-types), but it works as expected.

```ts
type GoodObject = {
  someValue?: number;
};
export const setGoodObject = async (obj: GoodObject): Promise<void> => {
  await firestore.collection('test').doc('test').set(obj); // OK
};

type BadObject = {
  someValue: number | undefined;
};
export const setBadObject = async (obj: BadObject): Promise<void> => {
  await firestore.collection('test').doc('test').set(obj); // TS2345: Argument of type 'BadObject' is not assignable to parameter of type 'DocumentData'.

};
```

Yonom:
Applying it to another project of mine, I am getting warnings about `new Date()`
I guess Date is a valid input type but it gets converted to Timestamp when you read from the DB? (to be investigated...)

I also wonder if they are other types that are supported but only as input...

Yonom:
A different but related issue is that when you use typescript to define collection types, this gets in the way of using `Date`s when setting values in firebase (this is also the case in `main` and not caused by this PR)

Example:

```ts
type TestObject = {
  someValue: Timestamp;
};

export const test = async (): Promise<void> => {
  await firestore.collection<TestObject>('test').doc('test').set({
    someValue: new Date(), // Type 'Date' is missing the following properties from type 'Timestamp', [...]
  });
};
```

I thought of a fix for both this issue and the issue from my previous comment: https://github.com/invertase/react-native-firebase/commit/6815b601bad6c8e51989de98dcbf0302bcd51dec
I'm not sure if its too clever, so I need your input @mikehardy 

Basically, for whatever type `T` you give the collection, I replace all instances of `Timestamp` with `Timestamp | Date` for the `.set` and `.update` functions.

mikehardy:
Hmm - I like where this is going, but my hesitation here is that, you *can* set it to allow undefined. It's okay if you do, that may be your preference, we just set the default to match firebase-js-sdk (if I remember that correctly...) but it is a toggle. How does firebase-js-sdk handle it when people allow undefined? You are right about the Date vs Timestamp transformation in vs out though that trips me up as I have one canonical set of object model definitions in my project(s) but I use TypeORM in one of them and I had to "educate" it that both types were possible, yes. A bit fussy

Yonom:
# Timestamp -> Timestamp | Date conversion

It might be a good idea to allow Dates in-place of Timestamp. 
I'll move the change to a separate PR because it seems less problematic (and smaller in scope) than the undefined case.

# undefined in set/update

I wasn't aware that there is an option for that, ugh!

Three solutions:
- do not disallow undefined (the majority of users who do not have the setting enabled will not profit from the benefits)
- disallow undefined by default but allow opt-out
- allow undefined by default but add opt-in

The API for either of the last two could look something like this:

```ts
// variant 1 (requires us to pass the type all the way to the doc)
firebase.firebase<true /* allow undefined */>().collection('test').doc('test').set({ undefined });
// variant 2 (less code in rn-firebase, my preference)
firebase.firebase().collection('test').doc('test').set<true /* allow undefined */>({ undefined });
```

I'm not happy with either variant tbh, it doesn't look pretty

Yonom:
Closing for inactivity

PR #5899, opened 2021-12-02
feat(sdks): firestore-ios-sdk 8.10.0 / firestore-android-sdk 29.0.1

mikehardy:
Standard dependency updates PR really, nothing major happening in the SDKs but I like to keep them fresh :sparkles:


mikehardy:
Passed iOS for me locally, modulo the flake that CI is having as well with DetoxSync crash

PR #5939, opened 2021-12-15
test(e2e, ios): use correct env var name to workaround iOS crashes

asafkorem:
### Description

Fix typo from `SIMCTL_CHILD_NSZombiesEnabled` to `SIMCTL_CHILD_NSZombieEnabled`.

See https://developer.apple.com/documentation/foundation/nszombieenabled.

### Related issues

https://github.com/invertase/react-native-firebase/pull/5932

@mikehardy


mikehardy (approved):
Perhaps I have watched too many movies, but who has only one zombie? Of course it would be plural 🤦  - thanks a bunch for continuing the investigation here and finding this error then even making the PR! Really appreciated

mikehardy:
💥 clean iOS run first time through, that's really refreshing. Thanks again

PR #5979, opened 2021-12-31
fix(app-check): getToken returns `{token: string}` not `string` matching firebase-js-sdk

quentin-fox:
### Description

The return type of the `getToken` method is not of type `{ token: string }`, but rather of type `string`.

On the Android side, the task result here is the `AppCheckTokenResult` class.

https://github.com/invertase/react-native-firebase/blob/master/packages/app-check/android/src/main/java/io/invertase/firebase/appcheck/ReactNativeFirebaseAppCheckModule.java#L85

The `AppCheckTokenResult` class itself has a `getToken` method, which returns a `String`. 

https://github.com/firebase/firebase-android-sdk/blob/6da8617afe903126afac46ef71b1d857dcfe7ef9/appcheck/firebase-appcheck/src/main/java/com/google/firebase/appcheck/internal/DefaultAppCheckTokenResult.java#L55

Since the string is what is resolved in the Promise, the return type of the javascript method bound to this native code should also be a string.

On the iOS side, the promise is also passed the `token.token`, which should be a string, instead of the `token`.

https://github.com/invertase/react-native-firebase/blob/master/packages/app-check/ios/RNFBAppCheck/RNFBAppCheckModule.m#L101

In our own app, we found that at runtime the `getToken` call was returning a string, so we had to use typescript assertions to prevent compiler errors due to the incorrect type.

### Checklist

  - [x] Yes
  - [x] `Android`
  - [x] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [ ] No
  - [x] Somewhat? (it may break existing typescript builds, but they would have been typed incorrectly anyways)

### Test Plan

I added a few tests to assert that `getToken` returns a string. It should have already be returning this type, as in the tests it is passed into `jwt.decode`, which accepts a string.

---

### DIFF
--- a/packages/app-check/android/src/main/java/io/invertase/firebase/appcheck/ReactNativeFirebaseAppCheckModule.java
+++ b/packages/app-check/android/src/main/java/io/invertase/firebase/appcheck/ReactNativeFirebaseAppCheckModule.java
@@ -82,10 +82,8 @@ public class ReactNativeFirebaseAppCheckModule extends ReactNativeFirebaseModule
             getExecutor(),
             (task) -> {
               if (task.isSuccessful()) {
-                String token = task.getResult().getToken();
-
                 WritableMap tokenResultMap = Arguments.createMap();
-                tokenResultMap.putString("token", tokne);
+                tokenResultMap.putString("token", task.getResult().getToken());
                 promise.resolve(tokenResultMap);
               } else {
                 Log.e(
--- a/packages/app-check/android/src/main/java/io/invertase/firebase/appcheck/ReactNativeFirebaseAppCheckModule.java
+++ b/packages/app-check/android/src/main/java/io/invertase/firebase/appcheck/ReactNativeFirebaseAppCheckModule.java
@@ -82,7 +82,11 @@ public class ReactNativeFirebaseAppCheckModule extends ReactNativeFirebaseModule
             getExecutor(),
             (task) -> {
               if (task.isSuccessful()) {
-                promise.resolve(task.getResult().getToken());
+                String token = task.getResult().getToken();
+
+                WritableMap tokenResultMap = Arguments.createMap();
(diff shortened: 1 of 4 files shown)


mikehardy (changes requested):
Mmm - first of all, thanks for noticing this, second :raising_hand: I made this mistake, sorry about that, but third - I think we need to go other direction and return (or somehow simulate and return) an AppCheckTokenResult for real. https://firebase.google.com/docs/reference/js/app-check.md#gettoken - we're trying to mirror the firebase-js-sdk where ever possible and though the AppCheck APIs diverge a bit I think we can get there with this one. Do you think it's possible to actually return AppCheckTokenResult, and alter the return values to be like that? It won't be a breaking change since AppCheck is in beta but I will make certain the release notes warn people in big bold terms so they're paying attention, if it's possible

quentin-fox:
It should be simple enough to do! I'll quickly rejig the PR so that it confirms to the firebase-js-sdk.

quentin-fox:
I couldn't for the life of me get the android project in `packages/app-check/android` to Gradle sync in my version of android studio, so there might be errors. Not at all experienced writing code on the bridge, so please tear it apart if it's not done correctly.

mikehardy:
Ah. sorry - the general development style is to actually load the tests/android directory into gradle. It has all of the modules hooked into it correctly for live editing while using the tests app as the "module host", and allowing quick re-runs in an e2e test harness context --> https://github.com/invertase/react-native-firebase/blob/main/tests/README.md

Regardless, I'll look through it - thanks!

mikehardy (dismissed):
This looks correct - originally on review I was thinking we needed to have expireTimeMillis in there as well but when I examined the firebase-js-sdk reference, it really is just an object with nothing but a token: string property in it, so this seems exactly right. I'll approve the CI runs and we'll see how it goes :-)

PR #6078, opened 2022-02-16
fix(analytics, ios): handle RNFirebaseAnalyticsWithoutAdIdSupport == false Podfile case correctly

OskarJuszczyk:
### Description

Right know it looks only if it is defined and not if it set as true, same error as in older appsflyer-react-native-plugin xD

### Release Summary

Fix RNFirebaseAnalyticsWithoutAdIdSupport flag in Podfile

### Checklist

  - [ ] Yes
  - [ ] `Android`
  - [x] `iOS`
- My change includes tests;
- This is a breaking change; (Hard to decide I would say yes because it will change someones dependencies after upgrade)
  - [X] Yes
  - [ ] No

### DIFF
--- a/packages/analytics/RNFBAnalytics.podspec
+++ b/packages/analytics/RNFBAnalytics.podspec
@@ -34,7 +34,7 @@ Pod::Spec.new do |s|
   end
 
   # Firebase dependencies
-  if defined?($RNFirebaseAnalyticsWithoutAdIdSupport)
+  if defined?($RNFirebaseAnalyticsWithoutAdIdSupport) && ($RNFirebaseAnalyticsWithoutAdIdSupport == true)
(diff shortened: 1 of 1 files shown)


mikehardy (approved):
Oh my, that looks like a :facepalm: error / incorrect assumption.
Of course most that define this will *only* define it if they want to set it to true, but, also (of course) you can define it false if you like.
Thanks for the catch + PR!

mikehardy:
The PR Title failure is okay, it's more of a warning for maintainers on merge since github handles single-commits in a surprising way. If the rest are green, we're :heavy_check_mark:

mikehardy:
Sorry for the delay publishing this one! Oversight on my part I did not realize we had this fix unpublished. I just kicked off the release process now

PR #6122, opened 2022-03-11
fix(android): firebase-android-sdk 30.0.1

### LINKED ISSUE #6158: Android firestore:24.0.1 has bug preventing reconnection to firestore after being backgrounded (workaround: use bom 29.2.1)
Per the title the Android firebase-firestore SDK v24.0.1 has a bug that will cause it to fail to reconnect to firestore after an app has been backgrounded and come back to the foreground. 

This is actually a bug in grpc-java that was fixed in v1.44.1 https://github.com/grpc/grpc-java/releases/tag/v1.44.1

firebase-firestore did not upgrade to grpc-java 1.44.1 until version firebase-firestore:24.0.2 but the current version of react-native-firebase uses firebase-firestore:24.0.1 which still has the bug

https://github.com/firebase/firebase-android-sdk/issues/3302

I resolved this locally by overriding the firebase sdk version on Android so that 24.0.2 is used like so in `build.gradle`

```
ext {
    set('react-native', [
                versions: [
                  // Overriding Library SDK Versions
                  firebase: [
                    // Override Firebase SDK Version
                    bom           : "29.2.1"
                  ]
                ],
              ])
}
```

mikehardy:
Update the native SDKs and javascript dependencies, pretty normal, except...

~~This is not working because of https://github.com/firebase/firebase-android-sdk/issues/3526~~

I expect a 29.2.1 or similar from firebase-android-sdk as soon as they determine root cause + fix

Fixes #6158

### DIFF
--- a/packages/app-check/android/build.gradle
+++ b/packages/app-check/android/build.gradle
@@ -93,9 +93,8 @@ repositories {
 dependencies {
   api appProject
   implementation platform("com.google.firebase:firebase-bom:${ReactNative.ext.getVersion('firebase', 'bom')}")
-  // These are beta, so they are not in the BoM yet. We have to specify the versions explicitly.
-  implementation 'com.google.firebase:firebase-appcheck-safetynet:16.0.0-beta04'
-  implementation 'com.google.firebase:firebase-appcheck-debug:16.0.0-beta04'
+  implementation "com.google.firebase:firebase-appcheck-safetynet"
+  implementation "com.google.firebase:firebase-appcheck-debug"
 }
 
 ReactNative.shared.applyPackageVersion()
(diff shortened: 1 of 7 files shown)


mikehardy:
https://github.com/firebase/firebase-android-sdk/issues/3526 is fixed in 29.2.1 of firebase-android-sdk

Currently blocked on inability of firebase-android-sdk auth to connect to auth emulator for some reason, causing failure of auth e2e chunk of test suite: https://github.com/firebase/firebase-android-sdk/issues/3557 - fails in 29.2.1, works in 29.1.0 with same code

mikehardy:
Also if you upgrade firebase-tools from 10.2.2 to 10.4.0 then firebase-ios-sdk has a problem with one of the storage upload tests. Not possible to say which of react-native-firebase/storage or firebase-tools or firebase-ios-sdk is fault location for that one yet.

mikehardy:
https://github.com/firebase/firebase-android-sdk/issues/3557 has resolution waiting on firebase-android-sdk > 29.3.2 as it will hopefully contain fix

mikehardy:
This is finally good to go! Resolves firestore connection issue, ingests new SDK and auth emulator is working.
At long last...

PR #6192, opened 2022-04-08
Code_Of_Conduct.md not Working

Jagroop2000:
##  In the Readme.md Code_Of_Conduct.md not working below i add the screenshot for your reference :
Screenshot from 2022-04-08 20-33-09

This is because the path given is incorrect . So i have two options : 
## Change Link to  https://github.com/invertase/react-native-firebase/blob/main/CODE_OF_CONDUCT.md 
But this will connect to below screenshot as listed below : 
Screenshot from 2022-04-08 20-39-07
## Secondly directly link to https://github.com/invertase/meta/blob/master/CODE_OF_CONDUCT.md
which connects to correct file as shown in below screenshot : 
Screenshot from 2022-04-08 20-40-37

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@
 
 **React Native Firebase** is a collection of official React Native modules connecting you to Firebase services; each module is a light-weight JavaScript layer connecting you to the native Firebase SDKs for both iOS and Android.
 
-React Native Firebase is built with four key principals in mind;
+React Native Firebase is built with four key principles in mind;
 
 - 🧪 **Well tested**
   - every module is extensively tested to >95% coverage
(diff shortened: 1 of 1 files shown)


mikehardy (commented):
I've asked internally that the meta repo switch default branch name to main, then the link will work correctly and have inclusive language, so it will be win / win

Jagroop2000:
@mikehardy  Okay, thank you for letting me know. You can close this.

mikehardy:
Thanks for bringing this up though, I hate 404s in docs, but I hadn't noticed it! This would be the right fix if we couldn't move it, but I just verified - our meta repo has default branch main now - so this is no longer a 404. Thanks again

PR #6205, opened 2022-04-19
chore: Set permissions for GitHub actions

naveensrinivasan:
Restrict the GitHub token permissions only to the required ones; this way, even if the attackers will succeed in compromising your workflow, they won’t be able to do much.

- Included permissions for the action. https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

[Keeping your GitHub Actions and workflows secure Part 1: Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)

Signed-off-by: naveensrinivasan <172697+naveensrinivasan@users.noreply.github.com>

### DIFF
--- a/.github/workflows/pr_title.yml
+++ b/.github/workflows/pr_title.yml
@@ -6,8 +6,13 @@ on:
       - edited
       - synchronize
 
+permissions:
+  contents: read
+
 jobs:
   main:
+    permissions:
+      contents: none
     runs-on: ubuntu-latest
     env:
       GITHUB_TOKEN: ${{ secrets.GH_TOKEN }}


mikehardy (commented):
Interesting - thanks for checking in to our security here, always good to have more eyes on this stuff.
I looked into this recently for a different repo and I think this is the important part of the linked document:

> TL;DR: Combining pull_request_target workflow trigger with an explicit checkout of an untrusted PR is a dangerous practice that may lead to repository compromise.

Since the conventional PR title check does not actually check out any code, I'm not sure if we have an opening in our attack surface area here? I can't think of how I'd use this workflow as an attack vector with no code checkout. So I'm not sure any change is necessary

One could argue that restricting tokens is always a best practice, and you'd be correct, but I want to make sure I understand it before just changing it

What do you think?

naveensrinivasan:
> Since the conventional PR title check does not actually check out any code, I'm not sure if we have an opening in our attack surface area here? I can't think of how I'd use this workflow as an attack vector with no code checkout. So I'm not sure any change is necessary

I agree with that case. 

> 
> One could argue that restricting tokens is always a best practice, and you'd be correct, but I want to make sure I understand it before just changing it

You are right, it is the best practice, and if there are future code changes to that workflow this will make sure it is in place. 

> 
> What do you think?

I'll let you decide.

mikehardy (approved):
Okay - I feel like I understand it (thanks for checking) and it is a best practice, let's put it in and hopefully the next PR the check continues to work :-)

Thanks again

PR #6255, opened 2022-05-17
docs(crashlytics): Add link to iOS symbol upload docs

lukebars:
### Description

Currently docs are missing needed steps for iOS debug symbol upload, which you really want to have on crashlytics. I've added a link to firebase official documentation.

### Checklist

  - [x] Yes
  - [ ] `Android`
  - [x] `iOS`
- This is a breaking change;
  - [ ] Yes
  - [x] No

### DIFF
--- a/docs/crashlytics/usage/index.md
+++ b/docs/crashlytics/usage/index.md
@@ -26,6 +26,10 @@ Once installed, you must complete the following additional setup steps for Andro
 
 - [Android Additional Setup](/crashlytics/android-setup).
 
+To get readable iOS crash reports in the Crashlytics dashboard, you must setup debug symbol upload:
+
+- [iOS Additional Setup](https://firebase.google.com/docs/crashlytics/get-deobfuscated-reports?platform=ios).
+
 > If you're using Expo, make sure to add the `@react-native-firebase/crashlytics` config plugin to your `app.json` or `app.config.js`. It handles the Android installation steps for you. For instructions on how to do that, view the [Expo](/#expo) installation section.
 
 If you're using an older version of React Native without autolinking support, or wish to integrate into an existing project,


mikehardy (changes requested):
Hmm - this should be set up by default, there is no need to take extra steps I think. I get symbol uploads on build normally.  However, it has a deeper problem: it is also not useful information, it leads to a false sense that things are working. More useful would be a discussion of why symbols uploaded from the local build are useless except for debug builds you are testing right then, and how to get useful symbols - I wrote it up here: https://stackoverflow.com/a/55796619/9910298 

What do you think?

lukebars:
> Hmm - this should be set up by default, there is no need to take extra steps I think. I get symbol uploads on build normally. However, it has a deeper problem: it is also not useful information, it leads to a false sense that things are working. More useful would be a discussion of why symbols uploaded from the local build are useless except for debug builds you are testing right then, and how to get useful symbols - I wrote it up here: https://stackoverflow.com/a/55796619/9910298
> 
> What do you think?

Reading your writeup, it all makes sense now. Maybe we should include this somewhere in crashlytics setup?

mikehardy:
Hmmm - this is definitely a documentation issue at least, we can improve this

As a user, do you *want* these uploaded by default? Or do you just want the option?
I'm hesitant to upload them by default because the files are large and it takes time to process them etc.

But I like the idea of having the option, we could add a firebase.json option and parse that in the podspec, then on pod install we could refer to it in the podspec and install this script as thing to do after the build

> /Pods/FirebaseCrashlytics/upload-symbols -gsp

Then regardless, a link to the writeup about the need to download bitcode-reprocessed dSYMs is probably useful

PR #6270, opened 2022-05-25
build(deps): update javascript dependencies

mikehardy:
Standard javascript dependency updates, plus a test fix for a flaky storage test fixture creation


mikehardy:
lerna v5 requires node v16+ (or a really update v14) so initial vercel deployment failed
I have access to config there so I changed the node engine in use, it is retrying.

PR #6319, opened 2022-06-19
fix(storage, android)!: android now updates customMetadata as a group

mikehardy:
### Description

StorageMetadata.customMetadata is an interesting beast. This PR resolves a number of problems with it so that we behave the same way as web, despite android and ios platforms differing in their behavior.

The basic idea is that customMetadata is typed as `Map<String, String>` on web, meaning you may never send a null in there, which implies that you may not use a null value to indicate a single key/value pair should be removed.

Rather, you update the customMetadata all at once as a single unit, passing in the whole map of customMetadata you want to keep as part of the update. Any keys you omit will be removed, and sending in null for customMetadata removes the whole map.

iOS used to allow per-key null to remove keys, but with v9 they are enforcing non-null with Swift so that doesn't work.
android allows per-key null to remove it, but that's not the same as web or firebase-ios-sdk v9

So here we converge all platforms so that all entries coming in are String/String, null deletes the customMetadata entirely, and any keys missing from user-supplied customMetadata are removed

Additionally, setting/deletion of all the other settable keys has been verified and behaves the same cross-platform.

### Related issues

https://github.com/firebase/firebase-ios-sdk/issues/9849

### Release Summary

Each commit is conventional, in particular since this is a breaking change, the actual breaking change has a special commit message which should provide direct advice on affected users.

### Checklist

  - [x] Yes
  - [x] `Android`
  - [] `iOS`
- My change includes tests;
- This is a breaking change;
  - [x] Yes
  - [ ] No

### Test Plan

So much testing, the e2e test was the primary driver of this change to begin with, and I used it extensively to validate the changes

---

### DIFF
--- a/packages/messaging/android/src/main/java/io/invertase/firebase/messaging/ReactNativeFirebaseMessagingModule.java
+++ b/packages/messaging/android/src/main/java/io/invertase/firebase/messaging/ReactNativeFirebaseMessagingModule.java
@@ -122,7 +122,8 @@ public class ReactNativeFirebaseMessagingModule extends ReactNativeFirebaseModul
 
   @ReactMethod
   public void getToken(String appName, String senderId, Promise promise) {
-    FirebaseMessaging messagingInstance = FirebaseApp.getInstance(appName).get(FirebaseMessaging.class);
+    FirebaseMessaging messagingInstance =
+        FirebaseApp.getInstance(appName).get(FirebaseMessaging.class);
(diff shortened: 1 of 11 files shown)


mikehardy:
@Salakar / @russellwheatley this was ...illuminating and very difficult to work through. It came up as I was porting to firebase-ios-sdk v9 and exposed an underlying platform difference that I/we had not noticed between web/ios/android. I think the same issue will likely apply for FlutterFire thus pining in case you want to have some :eyes:  on it

PR #6529, opened 2022-09-13
feat(messaging): add support for setDeliveryMetricsExportToBigQuery

pavan168:
### Description

We have had very specific requirements to send out push notifications using FCM and have implemented an in-house solution using the Firebase Admin SDK. However, we don't get access to the message delivery information in the Firebase Console anymore and want to use BigQuery to analyse this information.

Reading through [the documentation](https://firebase.google.com/docs/cloud-messaging/understand-delivery?platform=android#enable-message-delivery-data-export) we found that the message delivery data export needs to be enabled on the device (after asking for consent of course) and both [ios](https://firebase.google.com/docs/reference/ios/firebasemessaging/api/reference/Classes/FIRMessagingExtensionHelper#-exportdeliverymetricstobigquerywithmessageinfo:) and [android](https://firebase.google.com/docs/reference/android/com/google/firebase/messaging/FirebaseMessaging#public-method-summary) native methods are not exposed as per the RNFirebase documentation. 

With this PR, we can use `setDeliveryMetricsExportToBigQuery` method to pass down the status (a boolean value) and have used `deliveryMetricsExportToBigQueryEnabled()` method that exist already on android to return the updated status whilst iOS does not have a similar method, `isDeliveryMetricsExportToBigQueryEnabled` is introduced that holds the most up-to-date status that can be used as per your requirement.

### Related issues
n/a

### Release Summary
n/a

### Checklist

  - [x] Yes :fire:
  - [x] `Android`
  - [x] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

1. iOS test cases
ios

2. Android test cases
android

---

### DIFF
--- a/packages/messaging/android/src/main/java/io/invertase/firebase/messaging/ReactNativeFirebaseMessagingModule.java
+++ b/packages/messaging/android/src/main/java/io/invertase/firebase/messaging/ReactNativeFirebaseMessagingModule.java
@@ -221,10 +221,29 @@ public class ReactNativeFirebaseMessagingModule extends ReactNativeFirebaseModul
             });
   }
 
+  @ReactMethod
+  public void setDeliveryMetricsExportToBigQuery(Boolean enabled, Promise promise) {
+    Tasks.call(
+            getExecutor(),
+            () -> {
+              FirebaseMessaging.getInstance().setDeliveryMetricsExportToBigQuery(enabled);
+              return null;
+            })
+        .addOnCompleteListener(
+            task -> {
+              if (task.isSuccessful()) {
+                promise.resolve(FirebaseMessaging.getInstance().deliveryMetricsExportToBigQueryEnabled());
+              } else {
+                rejectPromiseWithExceptionMap(promise, task.getException());
+              }
+            });
+  }
+
   @Override
   public Map<String, Object> getConstants() {
     final Map<String, Object> constants = new HashMap<>();
     constants.put("isAutoInitEnabled", FirebaseMessaging.getInstance().isAutoInitEnabled());
+    constants.put("isDeliveryMetricsExportToBigQueryEnabled", FirebaseMessaging.getInstance().deliveryMetricsExportToBigQueryEnabled());
     return constants;
   }
 
--- a/packages/messaging/e2e/messaging.e2e.js
+++ b/packages/messaging/e2e/messaging.e2e.js
@@ -320,4 +320,30 @@ describe('messaging()', function () {
       }
     });
   });
+
+  describe('setDeliveryMetricsExportToBigQuery()', function () {
+    afterEach(async function () {
+      await firebase.messaging().setDeliveryMetricsExportToBigQuery(false);
+    });
+
+    it('throws if enabled is not a boolean', function () {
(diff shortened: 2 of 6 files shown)


mikehardy:
Hey! Thanks for your patience on this - I'm still catching up (as fast as I can, but never fast enough...) after summer travels. Queued for review

packages/messaging/android/src/main/java/io/invertase/firebase/messaging/ReactNativeFirebaseMessagingModule.java
```diff
@@ -221,10 +221,29 @@ public void unsubscribeFromTopic(String topic, Promise promise) {
...
+                rejectPromiseWithExceptionMap(promise, task.getException());
+              }
+            });
+  }
+
   @Override
   public Map<String, Object> getConstants() {
     final Map<String, Object> constants = new HashMap<>();
     constants.put("isAutoInitEnabled", FirebaseMessaging.getInstance().isAutoInitEnabled());
+    constants.put("isDeliveryMetricsExportToBigQueryEnabled", FirebaseMessaging.getInstance().deliveryMetricsExportToBigQueryEnabled());
```
mikehardy, line 246: Interesting - I was under the impression that constants were fetched once and cached such that setting new values into constants would not be reflected in future checks, but your test probes this, so, :+1:

mikehardy (approved):
This looks great. For a little bit I was confused why the iOS implementation wasn't actually *doing* anything, then I remembered our previous conversation where it came out that on the iOS side you can only send things per-message + on receipt. At which point the property is perfect as it may be queried by an extension

Great work here, all around

mikehardy:
v15.6.0 just went out with this feature - good luck with your project!

packages/messaging/android/src/main/java/io/invertase/firebase/messaging/ReactNativeFirebaseMessagingModule.java
pavan168: I had the same thought at first. Gotta say that the tests were super helpful for a starter 👍

pavan168: Thank you very much @mikehardy Hoping to contribute more in future!

mattThousand:
@mikehardy @pavan168 in order for an iOS app to export message data to BigQuery, I will still need to make my own extension, is that correct? But on Android, all I would need to do is call `setDeliveryMetricsExportToBigQuery(true)` from the javascript layer?

Or do the changes introduced in this PR mean that we no longer need to make our own Notification Service Extension? 

I suppose I am a bit confused as to what the iOS changes introduced in this PR do, exactly 🤔

mikehardy:
I don't believe there is a way to avoid the extension on iOS bec you have to send it every notification. The PR made the status of user acceptance very accessible, but it still requires an extension. If this isn't clear in the docs you can hit the edit button at the top of any page and drop a note in to help future intrepid explorers 😁

Issue #3371, opened 2020-03-30
Firebase Performance Monitoring: Not able to get screen trace.


tusharthinksys:
I am getting only MainActivity screen traces but how can i get screen traces per screen in my React native app. Please help me.

mikehardy:
#2994 #3226 

Everything in react-native runs in one activity, it's the way the whole platform works

You need to bridge the logical screen changes in your app (typically you will have some sort of hook that is called from your navigation library) into firebase (via setCurrentScreen or similar) such that navigation events results in firebase-level screen changes

comvenger-brandon:
Using setCurrentScreen does not fix this problem. The performance library only reports "Mainactivity" for any screen.
"setCurrentScreen" is only for Analytics.

mikehardy:
@comvenger-brandon this is a react-native-general problem. Has nothing to do with the specific module in use. It is literally fundamental to the design of react native to use a single activity and navigate within the bundle in javascript stack frames. If you have a workaround for the performance library as well, that would be interesting.

MCervenka:
monitoring screen rendering would be really useful. Anyone found some way to implement it?

sampsonjoliver:
Based on [this codelab](https://firebase.google.com/codelabs/measure-android-view-performance#7) from the Firebase team for running screen traces on Fragments, it seems like there is a way to start and stop screen traces programmatically, with custom names. At least for Android.

With that ScreenTrace util bundled for RN, it seems you start it on RN screen component mounts, and end it on screen component unmount. This might miss a some of the render lifecycle (between the activity create and the react mount), but you'll capture component updates well enough.

Or, for `react-navigation` users, and based on my understanding on `react-native-screens` [using Fragments under-the-hood](https://github.com/software-mansion/react-native-screens#using-createnativestacknavigator-with-react-navigation) if the Screens library could accept custom Fragment definitions then you'd just need to extend their [`ScreenFragment`](https://github.com/software-mansion/react-native-screens/blob/2918b44255dd85f00e74cc00b474b0cd7f84d4a6/android/src/main/java/com/swmansion/rnscreens/ScreenFragment.kt) and as per the codelab, start and stop a named trace on Fragment attach and detach.

shamilovtim:
I've implemented per screen FPS profiling here (based on ad hoc start / stop via React lifecycle): #6588 

[...]

PR #6588, opened 2022-10-04
feat(perf): add custom screen rendering traces for android

shamilovtim:
### Description

The Firebase Android SDK automatically records screen traces ( janky / slow / frozen frames) for activities and fragments. However, these traces do not include information about what Javascript side screen the user is on. Users are left trying to triage which screen introduced a slowdown into their application. In large applications this is nearly impossible. This PR adds support for custom screen rendering traces. They can be used ad hoc however the user wishes. We plan to start them on screen mount in react native and end them on screen cleanup. 

For the most part I've followed the existing patterns in this area of the project. I'm not a big fan of ES6 classes but I implemented using an ES6 class just to stay consistent. 

Performance considerations: because there is no data being returned to the Javascript, there is almost no serialization happening across the bridge. This is why I am comfortable and happy with this Javascript side solution.

brown m&m: 🔥

### Related issues

#3371

### Checklist

  - [x] Yes
  - [x] `Android`
  - [ ] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

The initial implementation (Circa Oct 6) has been used by our team in prod to great success. It has helped us find out where in the javascript logic our react native app has jank.

---

### DIFF
--- a/docs/perf/usage/index.md
+++ b/docs/perf/usage/index.md
@@ -101,6 +101,7 @@ async function screenTrace() {
     await trace.stop();
   } catch (e) {
     // rejects if iOS or (Android == 8 || Android == 8.1)
+    // or if hardware acceleration is off
   }
 }
 ```
--- a/docs/perf/usage/index.md
+++ b/docs/perf/usage/index.md
@@ -95,10 +95,13 @@ import perf from '@react-native-firebase/perf';
 
 async function screenTrace() {
   // Define & start a screen trace
-  const trace = await perf().startScrenTrace('FooScreen');
-
-  // Stop the trace
-  await trace.stop();
+  try {
+    const trace = await perf().startScrenTrace('FooScreen');
+    // Stop the trace
+    await trace.stop();
+  } catch (e) {
+    // rejects if iOS or < Android 9.0
+  }
 }
 ```
 
--- a/docs/perf/usage/index.md
+++ b/docs/perf/usage/index.md
@@ -100,7 +100,7 @@ async function screenTrace() {
     // Stop the trace
     await trace.stop();
   } catch (e) {
-    // rejects if iOS or < Android 9.0
+    // rejects if iOS or (Android == 8 || Android == 8.1)
   }
 }
 ```
--- a/docs/perf/usage/index.md
+++ b/docs/perf/usage/index.md
@@ -86,6 +86,22 @@ async function customTrace() {
 }
 ```
 
+## Custom screen traces
+
+Record a custom screen rendering trace (slow frames / frozen frames)
+
+```jsx
+import perf from '@react-native-firebase/perf';
+
+async function screenTrace() {
+  // Define & start a screen trace
+  const trace = await perf().startScrenTrace('FooScreen');
+
+  // Stop the trace
+  await trace.stop();
+}
+```
+
 ## HTTP Request Tracing
 
 Below illustrates you would measure the latency of a HTTP request.
--- a/packages/perf/android/src/main/java/io/invertase/firebase/perf/ScreenTrace.java
+++ b/packages/perf/android/src/main/java/io/invertase/firebase/perf/ScreenTrace.java
@@ -132,11 +132,6 @@ public class ScreenTrace {
       }
     }
 
-    if (totalFrames == 0 && slowFrames == 0 && frozenFrames == 0) {
-      // All metrics are zero, no need to send screen trace.
-      // return;
-    }
-
     // Only incrementMetric if corresponding metric is non-zero.
     if (totalFrames > 0) {
       perfScreenTrace.putMetric(Constants.CounterNames.FRAMES_TOTAL.toString(), totalFrames);
--- a/packages/perf/android/src/main/java/io/invertase/firebase/perf/ScreenTrace.java
+++ b/packages/perf/android/src/main/java/io/invertase/firebase/perf/ScreenTrace.java
@@ -17,6 +17,7 @@ package io.invertase.firebase.perf;
  */
 
 import android.app.Activity;
+import android.os.Build;
 import android.util.Log;
 import android.util.SparseIntArray;
 import android.view.WindowManager;
@@ -163,13 +164,16 @@ public class ScreenTrace {
   // region Helper Functions
 
   private static boolean checkScreenTraceSupport(Activity activity) {
-    boolean hasFrameMetricsAggregatorClass = hasFrameMetricsAggregatorClass();
+    boolean isValidSDKVersion = checkSDKVersion();
+    boolean hasFrameMetricsAggregatorClass = checkFrameMetricsAggregatorClass();
     boolean isActivityHardwareAccelerated = activity.getWindow() != null
       && ((activity.getWindow().getAttributes().flags & WindowManager.LayoutParams.FLAG_HARDWARE_ACCELERATED) != 0);
 
-    boolean supported = hasFrameMetricsAggregatorClass && isActivityHardwareAccelerated;
+
+    boolean supported = isValidSDKVersion && hasFrameMetricsAggregatorClass && isActivityHardwareAccelerated;
 
     Log.d(TAG, new StringBuilder()
+      .append("isValidSDKVersion: ").append(isValidSDKVersion)
       .append("isScreenTraceSupported(").append(activity).append("): ").append(supported)
       .append(" [hasFrameMetricsAggregatorClass: ").append(hasFrameMetricsAggregatorClass)
       .append(", isActivityHardwareAccelerated: ").append(isActivityHardwareAccelerated).append("]").toString());
@@ -177,10 +181,18 @@ public class ScreenTrace {
     return supported;
   }
 
+  private static boolean checkSDKVersion() {
+    if (Build.VERSION.SDK_INT == 26 || Build.VERSION.SDK_INT == 27) {
+      return false;
+    }
+
+    return true;
+  }
+
   /**
    * Inspired by fireperf source.
    */
-  private static boolean hasFrameMetricsAggregatorClass() {
+  private static boolean checkFrameMetricsAggregatorClass() {
     try {
       Class<?> initializerClass = Class.forName(FRAME_METRICS_AGGREGATOR_CLASSNAME);
       return true;
--- a/packages/perf/android/src/main/java/io/invertase/firebase/perf/ScreenTrace.java
+++ b/packages/perf/android/src/main/java/io/invertase/firebase/perf/ScreenTrace.java
@@ -75,7 +75,7 @@ public class ScreenTrace {
     boolean isScreenTraceSupported = checkScreenTraceSupport(activity);
 
     if (!isScreenTraceSupported) {
-      throw new IllegalStateException("Device does not support screen traces. Hardware acceleration must be enabled and Android must be > 8.1.");
+      throw new IllegalStateException("Device does not support screen traces. Hardware acceleration must be enabled and Android must not be 8.0 or 8.1.");
     }
 
     frameMetricsAggregator = new FrameMetricsAggregator();
--- a/packages/perf/android/src/main/java/io/invertase/firebase/perf/ScreenTrace.java
+++ b/packages/perf/android/src/main/java/io/invertase/firebase/perf/ScreenTrace.java
@@ -29,7 +29,7 @@ import com.google.firebase.perf.util.Constants;
 
 /**
  * Utility class to capture Screen rendering information (Slow/Frozen frames) for the
- * {@code Activity} passed to the constructor {@link ScreenTrace#ScreenTrace(Activity, String)}.
+ * {@code Activity} passed to the constructor {@link io.invertase.firebase.perf.ScreenTrace#ScreenTrace(Activity, String)}.
  * <p>
  * Learn more at https://firebase.google.com/docs/perf-mon/screen-traces?platform=android.
  * <p>
@@ -42,7 +42,7 @@ import com.google.firebase.perf.util.Constants;
  * - https://www.youtube.com/playlist?list=PLOU2XLYxmsIKEOXh5TwZEv89aofHzNCiu (Android Performance Patterns)
  * <p>
  * References:
- * - Fireperf Source Code: https://bityl.co/5v2O
+ * - Fireperf Source Code
  */
 public class ScreenTrace {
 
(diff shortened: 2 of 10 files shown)


packages/perf/android/src/main/java/io/invertase/firebase/perf/ScreenTrace.java
```diff
@@ -0,0 +1,227 @@
...
+
+    if (isScreenTraceSupported) {
+      frameMetricsAggregator = new FrameMetricsAggregator();
+    }
+  }
+
+  // region Public APIs
+
+  /**
+   * Force enable Hardware acceleration to support screen traces as we can't observe frame
```
mikehardy, line 87: Ooo! This is dangerous. I have some deep scar tissue from android hardware acceleration. This side effect needs to be clearly documented (https://github.com/ankidroid/Anki-Android/issues/7369#issuecomment-794410275 / https://github.com/ankidroid/Anki-Android/issues/8160).

I am okay with it going in only with a warning that doing a perf trace requires hardware acceleration, which has caused problems in the past, may cause them in the future, and because of known incompatibilities with Android 8/8.1 plus WebViews it will never be enabled on Android 8/8.1 devices

There are no known issues with any other devices + versions but that's a big enough unexpected problem that I think the only safe way to do this is to check for SDK_INT 26/27 and refuse to enable hardware acceleration.

Later on in the flow this will reject the promise when the trace start is attempted since isScreenTraceSupported will return false

packages/perf/android/src/main/java/io/invertase/firebase/perf/ScreenTrace.java
```diff
@@ -0,0 +1,227 @@
+package io.invertase.firebase.perf;
+
+/**
+ * Copyright 2021 Google Inc. All Rights Reserved.
```
mikehardy, line 4: where did this code come from? You must sign the CLA here which asserts that it is your code, but the copyright is Google :thinking: ?

packages/perf/android/src/main/java/io/invertase/firebase/perf/ScreenTrace.java
```diff
@@ -0,0 +1,227 @@
...
+ * A slow screen rendering often leads to a UI Jank which creates a bad user experience. Below are
+ * some tips and references to understand and fix common UI Jank issues:
+ * - https://developer.android.com/topic/performance/vitals/render.html#fixing_jank
+ * - https://youtu.be/CaMTIgxCSqU (Why 60fps?)
+ * - https://youtu.be/HXQhu6qfTVU (Rendering Performance)
+ * - https://youtu.be/1iaHxmfZGGc (Understanding VSYNC)
+ * - https://www.youtube.com/playlist?list=PLOU2XLYxmsIKEOXh5TwZEv89aofHzNCiu (Android Performance Patterns)
+ * <p>
+ * References:
+ * - Fireperf Source Code: https://bityl.co/5v2O
```
mikehardy, line 45:
```suggestion
 * - Fireperf Source Code: https://bityl.co/5v2O
```

This link is `bityl` instead of bitly, but in general it's best to put a real code reference in there. Branch names change, code lines move around, but a permalink to the actual code is best I think as opposed to any shortening. When you can see the actual link structure it is usually possible to find the current link even if linkrot takes hold

Same applies to all similar links through this file

packages/perf/android/src/main/java/io/invertase/firebase/perf/ScreenTrace.java
```diff
@@ -0,0 +1,227 @@
...
+            // Slow frames are anything above 16ms (i.e. 60 frames/second)
+            slowFrames += numFrames;
+          }
+        }
+      }
+    }
+
+    if (totalFrames == 0 && slowFrames == 0 && frozenFrames == 0) {
+      // All metrics are zero, no need to send screen trace.
+      // return;
```
mikehardy, line 160: interesting decision to a) comment out the code versus simply remove it - committing commented out code is usually a bad sign or b) alternatively not document why it is commented out but included

Stated differently: why is this return commented out, contrary to the stated intent of the original comment?

packages/perf/lib/ScreenTrace.js
```diff
@@ -0,0 +1,46 @@
...
+    if (this._started) {
+      return Promise.resolve(null);
+    }
+    this._started = true;
+
+    return this.native.startScreenTrace(this._id, this._identifier);
+  }
+
+  stop() {
+    if (this._stopped) {
```
mikehardy, line 39: maybe 
```suggestion
    if (!this._started || this._stopped) {
```

...that way it is safe to call stop() if start never worked, this makes stop safe even if on iOS where start won't work, or on unsupported platforms where start doesn't work

packages/perf/lib/ScreenTrace.js
```diff
@@ -0,0 +1,46 @@
...
+    this._stopped = false;
+  }
+
+  start() {
+    if (this._started) {
+      return Promise.resolve(null);
+    }
+    this._started = true;
+
+    return this.native.startScreenTrace(this._id, this._identifier);
```
mikehardy, line 35: I did not see an iOS implementation, so this will crash on iOS unless there is a platform check here.
I think it's best to platform-check and then Promise.reject "not supported on iOS", document the same in the index.d.ts
This will get people in the habit of checking for rejections, which should steer them into implementations that handle Android 8/8.1 rejections correctly as well

packages/perf/e2e/perf.e2e.js
```diff
@@ -42,6 +42,17 @@ describe('perf()', function () {
       trace._identifier.should.equal('invertase');
       trace._started.should.equal(true);
       await trace.stop();
+      trace._stopped.should.equal(true);
+    });
+  });
+
+  describe('startScreenTrace()', function () {
+    it('resolves a started instance of a ScreenTrace', async function () {
+      const screenTrace = await firebase.perf().startScreenTrace('FooScreen');
```
mikehardy, line 51: I love that there is a test, but for iOS it should expect an exception? Our test runs are all higher than Android 8/8.1 right now but it could use a note in here about that as well in case we do an SDK matrix later that includes it

docs/perf/usage/index.md
```diff
@@ -86,6 +86,22 @@ async function customTrace() {
...
+## Custom screen traces
+
+Record a custom screen rendering trace (slow frames / frozen frames)
+
+```jsx
+import perf from '@react-native-firebase/perf';
+
+async function screenTrace() {
+  // Define & start a screen trace
+  const trace = await perf().startScrenTrace('FooScreen');
```
mikehardy, line 98: make the example do a try/catch style with mention that iOS rejects, and Android 8/8.1 rejects, as discussed in other comments

mikehardy (changes requested):
Very interesting - this is a nice way around the mis-match between react-native view construction and Android underlying objects + firebase performance expectations on view construction

No problem with including something like this but I left a big batch of comments that I think all center around two main ideas 1- hardware acceleration is a powerful + error-prone side effect so needs a cutout+documentation and 2- rejection on trace start should be socialized as normal and occur on unsupported platform + places where hardware acceleration is too dangerous to enable

Thoughts?

packages/perf/android/src/main/java/io/invertase/firebase/perf/ScreenTrace.java
shamilovtim: 99.99% of this code is Google's which they offered to the community. I modified it slightly for our use cases here.

packages/perf/lib/ScreenTrace.js
shamilovtim: Sounds good to me.

packages/perf/android/src/main/java/io/invertase/firebase/perf/ScreenTrace.java
shamilovtim: I am happy to completely refactor this code if that's necessary for licensing. It looks like we're making plenty of modifications to the file anyways.

mikehardy: I think the Apache license is fine, I was just curious - this won't block

docs/perf/usage/index.md
shamilovtim: fixed

packages/perf/e2e/perf.e2e.js
shamilovtim: fixed

packages/perf/lib/ScreenTrace.js
shamilovtim: fixed

packages/perf/lib/ScreenTrace.js
shamilovtim: fixed

packages/perf/lib/ScreenTrace.js
shamilovtim: @mikehardy I am not seeing the @platform tag being pulled in from the JSDoc comments onto the site: https://react-native-firebase-43jdz2ws9-invertase.vercel.app/reference/perf/screentrace. s there something bugged with the docs website?

mikehardy: I haven't seen that go through either, I usually put it in the method text like so https://rnfirebase.io/reference/app-distribution#checkForUpdate - works for me ?

docs/perf/usage/index.md
```diff
@@ -86,6 +86,25 @@ async function customTrace() {
...
+import perf from '@react-native-firebase/perf';
+
+async function screenTrace() {
+  // Define & start a screen trace
+  try {
+    const trace = await perf().startScrenTrace('FooScreen');
+    // Stop the trace
+    await trace.stop();
+  } catch (e) {
+    // rejects if iOS or < Android 9.0
```
mikehardy, line 103:
```suggestion
    // rejects if iOS or (Android == 8 || Android == 8.1)
```
It's actually just API26 and API27, Android 5, 6, 7 etc, totally fine.

packages/perf/lib/ScreenTrace.js
shamilovtim: Oh I see

docs/perf/usage/index.md
shamilovtim: Question on this. ScreenTrace is currently used in the auto instrumentation. Would we be bifurcating `Android == 8 || Android == 8.1` support if we reject these custom screen traces on Android 8 but allow the automatic screen tracing on Android 8?

mikehardy: Maybe? I'm not sure what the performance module is doing under the covers for Android 8/8.1 all I know is what I have personally had to solve, and the solution I came up with, on an app with 2 million users and active use of 15mins / day (that is: a lot of users, enough I'm pretty confident in our work there). 

It doesn't look like the underlying code takes a position

seems to note if it cannot do it https://github.com/firebase/firebase-android-sdk/blob/fa4adcf92b354898396cac614b4e2149bb29c797/firebase-perf/src/main/java/com/google/firebase/perf/application/FrameMetricsRecorder.java#L113

but does not seem to throw if you create a trace without hardware acceleration and you only find out on stop?

Below firebase-android-sdk is https://developer.android.com/reference/androidx/core/app/FrameMetricsAggregator#add(android.app.Activity) which does not seem to do anything about it.

Perhaps instead of forcing it on here, except for API26/27 we should just document that it requires hardware acceleration, and throw if it is not on, with a side note indicating that it is on in almost all cases but there are valid reasons to turn it off in some cases e.g. API26/27+WebView.

I like that better as the idea of toggling a new code pathway on in view hierarchy seems like a very unexpected side effect, to me, of a measurement tool

shamilovtim: Good call, I'll make those changes.

shamilovtim:
I'll be getting this merged as soon as I'm back from vacation. Thanks!

docs/perf/usage/index.md
shamilovtim: changes made

packages/perf/android/src/main/java/io/invertase/firebase/perf/ScreenTrace.java
shamilovtim: fixed

packages/perf/android/src/main/java/io/invertase/firebase/perf/ScreenTrace.java
shamilovtim: fixed

packages/perf/android/src/main/java/io/invertase/firebase/perf/ScreenTrace.java
shamilovtim: fixed

shamilovtim:
all changes made

shamilovtim:
Not stale

mikehardy (dismissed):
I'm really sorry this sat in the review queue so long, it appears you nailed everything feedback-related quite some time ago and it went stale obviously through no fault of your own.

It looks fantastic now

packages/perf/android/src/main/java/io/invertase/firebase/perf/ScreenTrace.java
mikehardy: this one may still not be resolved, unsure, but there are no major side effects that will impact users and if the metrics are off, we'll have to see folks using it in practice submit fixes I suppose

packages/perf/lib/index.js
```diff
@@ -80,6 +81,21 @@ class FirebasePerfModule extends FirebaseModule {
...
+    if (!isString(identifier) || identifier.length > 100) {
+      throw new Error(
+        "firebase.perf().newScreenTrace(*) 'identifier' must be a string with a maximum length of 100 characters.",
+      );
+    }
+
+    return new ScreenTrace(this.native, identifier);
+  }
+
+  startScreenTrace(identifier) {
```
mikehardy, line 94: Note for @russellwheatley - this PR is from before the modular PR work and was in flight while you have been working. It adds two native-only APIs that need to go in the #6771 API after this is merged (should merge as soon as CI goes green) so this PR is no longer orphaned and modular/future API doesn't drop it in the shuffle

mikehardy (approved):
pulled PR locally, rebased against main (since main has moved quite a bit...), reran e2e tests - everything looks good, when CI goes green here I'll merge and release this one before more things land

mikehardy:
Just released this, it's live out on npmjs.com!

PR #6592, opened 2022-10-06
docs(messaging,ios): Added Schema to Disabling Auto Registration (iOS)

Bilal-Abdeen:
Added the following to the change in the  firebase.json file, which is referenced in the section titled "Auto Registration (iOS)." 
"$schema": "./node_modules/@react-native-firebase/app/firebase-schema.json",

### Related issues
#4428 

### Checklist

  - [ x] Yes
  - [ ] `Android`
  - [ x] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [ x] No

### Test Plan

---

:fire:

### DIFF
--- a/docs/messaging/usage/index.md
+++ b/docs/messaging/usage/index.md
@@ -369,6 +369,7 @@ to manually control registration you can disable this via the `firebase.json` fi
 ```json
 // <projectRoot>/firebase.json
 {
+  "$schema": "./node_modules/@react-native-firebase/app/firebase-schema.json",
   "react-native": {
     "messaging_ios_auto_register_for_remote_messages": false
   }


mikehardy (changes requested):
This seems like a reasonable change as it shows how to use the schema - however:
- it's not really a "messaging" change, the schema is part of Core/App so should be documented there
- I don't believe for the issue you linked this was actually the problem, I would need to hear from you that you have done a positive *and* a negative test where:
-- you were able to see the error message described in the issue without the schema (the baseline)
-- you added the schema and no longer see the error (the positive test)
-- you removed the schema again and now see the error again (the negative test)

I assert - without proof - that the claim of success in the related issue was more related to the steps taken or not taken, that is that pod install was not run every time prior to build. But the positive and negative test above should determine that

If my assertion is correct, then this is not related to messaging at all, it's just a generic example of how to put the schema into the JSON, so should be in the docs next to the JSON docs themselves

If I'm wrong, then...I don't understand the mechanism for the linked issue nor do I understand JSON schemas but that won't surprise me *too* much, it'll be an opportunity to learn something...

Bilal-Abdeen:
This change is not actually required. I misunderstood the problem. I am sorry for wasting your time.

mikehardy:
No worries! I appreciate the attempt and appreciate the help

PR #6633, opened 2022-10-26
feat(analytics): add getSessionId

dlackty:
### Description

Firerbase Analytics SDKs provide an getSessionId method, and getting this value can help app developers to do cross environment trackings (e.g. Posting events using Google's measurement protocol).

So I implemented a React Native binding for this method.

### Release Summary

analytics: implement getSessionId

### Checklist

  - [x] Yes
  - [x] `Android`
  - [x] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

I made an demo app to test on devices:

Simulator Screen Shot - iPhone 14 - 2022-10-26 at 04 12 59

---

### DIFF
--- a/packages/analytics/android/src/main/java/io/invertase/firebase/analytics/UniversalFirebaseAnalyticsModule.java
+++ b/packages/analytics/android/src/main/java/io/invertase/firebase/analytics/UniversalFirebaseAnalyticsModule.java
@@ -60,6 +60,10 @@ public class UniversalFirebaseAnalyticsModule extends UniversalFirebaseModule {
     return FirebaseAnalytics.getInstance(getContext()).getAppInstanceId();
   }
 
+  Task<Long> getSessionId() {
+    return FirebaseAnalytics.getInstance(getContext()).getSessionId();
+  }
+
   Task<Void> setUserId(String id) {
     return Tasks.call(
         () -> {
--- a/packages/analytics/android/src/reactnative/java/io/invertase/firebase/analytics/ReactNativeFirebaseAnalyticsModule.java
+++ b/packages/analytics/android/src/reactnative/java/io/invertase/firebase/analytics/ReactNativeFirebaseAnalyticsModule.java
@@ -94,6 +94,21 @@ public class ReactNativeFirebaseAnalyticsModule extends ReactNativeFirebaseModul
   }
 
   @ReactMethod
+  public void getSessionId(Promise promise) {
+    module
+        .getSessionId()
+        .addOnCompleteListener(
+            task -> {
+              if (task.isSuccessful()) {
+                Long result = task.getResult();
+                promise.resolve(result != null ? result.doubleValue() : null);
+              } else {
+                rejectPromiseWithExceptionMap(promise, task.getException());
+              }
+            });
+  }
+
+  @ReactMethod
   public void setUserId(String id, Promise promise) {
     module
         .setUserId(id)
--- a/packages/analytics/e2e/analytics.e2e.js
+++ b/packages/analytics/e2e/analytics.e2e.js
@@ -49,6 +49,40 @@ describe('analytics()', function () {
     });
   });
 
+  describe('getSessionId()', function () {
+    it('calls native fn without error', async function () {
+      await firebase.analytics().getSessionId();
+    });
+
+    it('returns a non empty session ID', async function () {
+      let sessionId = await firebase.analytics().getSessionId();
+      // On iOS it can take ~ 3 minutes for the session ID to be generated
+      // Otherwise, `Analytics uninitialized` error will be thrown
+      const retries = 240;
+      while (!sessionId && retries > 0) {
+        await Utils.sleep(1000);
+        sessionId = await firebase.analytics().getSessionId();
+      }
+
+      if (!sessionId) {
+        return Promise.reject(
+          new Error('Firebase SDK did not return a session ID after 4 minutes'),
+        );
+      }
+
+      sessionId.should.not.equal(0);
+    });
+
+    it('returns a null value if session expires', async function () {
+      // Set session duration to 1 millisecond
+      firebase.analytics().setSessionTimeoutDuration(1);
+      // Wait 100 millisecond to ensure session expires
+      await Utils.sleep(100);
+      const sessionId = await firebase.analytics().getSessionId();
+      should.equal(sessionId, null);
+    });
+  });
+
   describe('setUserId()', function () {
     it('allows a null values to be set', async function () {
       await firebase.analytics().setUserId(null);
--- a/packages/analytics/e2e/analytics.modular.e2e.js
+++ b/packages/analytics/e2e/analytics.modular.e2e.js
@@ -50,6 +50,40 @@ describe('analytics() modular', function () {
       });
     });
 
+    describe('getSessionId()', function () {
+      it('calls native fn without error', async function () {
+        await firebase.analytics().getSessionId();
+      });
+
+      it('returns a non empty session ID', async function () {
+        let sessionId = await firebase.analytics().getSessionId();
+        // On iOS it can take ~ 3 minutes for the session ID to be generated
+        // Otherwise, `Analytics uninitialized` error will be thrown
+        const retries = 240;
+        while (!sessionId && retries > 0) {
+          await Utils.sleep(1000);
+          sessionId = await firebase.analytics().getSessionId();
+        }
+
+        if (!sessionId) {
+          return Promise.reject(
+            new Error('Firebase SDK did not return a session ID after 4 minutes'),
+          );
+        }
+
+        sessionId.should.not.equal(0);
+      });
+
+      it('returns a null value if session expires', async function () {
+        // Set session duration to 1 millisecond
+        firebase.analytics().setSessionTimeoutDuration(1);
+        // Wait 100 millisecond to ensure session expires
+        await Utils.sleep(100);
+        const sessionId = await firebase.analytics().getSessionId();
+        should.equal(sessionId, null);
+      });
+    });
+
     describe('setUserId()', function () {
       it('allows a null values to be set', async function () {
         await firebase.analytics().setUserId(null);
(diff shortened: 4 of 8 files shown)


dlackty:
From what I read from [Firebase' official doc](https://firebase.google.com/docs/reference/android/com/google/firebase/analytics/FirebaseAnalytics#public-void-setconsent-mapfirebaseanalytics.consenttype,-firebaseanalytics.consentstatus-consentsettings), there's only a way to set consent status but no way to get consent status.

For case that the API returns null because of consent status, Firebase analytics' built-in verbose logging already shows warnings.

Last but not least, I also spend time thinking about the choice of data type. The original Firebase SDK uses long as return value type, but React Native doesn't support it, and thus I chose string instead.

mikehardy:
Interesting! You are right there does not appear to be a getConsent, it all appears to be just a one-way black box - you start with defaults (default true or optionally disabled in native level in AndroidManifest/iOS plist), you overlay with setConsent calls (not implemented / exposed here, yet) and that's it!

I logged a feature request here to at least expose the native default settings in app configuration, and to expose the APIs to set: https://github.com/invertase/react-native-firebase/discussions/6636

For this feature then, we may at least assert that it *should* work as we may set things how we like in testing, given that the firebase.json for the test app does not deactivate collection entirely:

https://github.com/invertase/react-native-firebase/blob/b2670ea2537a35ae653cad210b27cac276036a4e/tests/firebase.json#L23-L25

On iOS the e2e test is currently failing though, which is odd:

https://github.com/invertase/react-native-firebase/actions/runs/3326085899/jobs/5505977736#step:23:133

>      NativeFirebaseError: [analytics/unknown] Analytics uninitialized

1) I think the `analytics/unknown` (at least: the unknown part) is odd, as this is an error message we should be able to map to something, ideally to the same code that web would throw, so people on all 3 platforms may match the same error.

2) why is analytics uninitialized for the iOS e2e test app, at this point in the testing? Is there something fundamentally wrong with the iOS e2e test config, or is this test making an assumption about analytics collection state (that may be altered by other tests) and needs to set the test up more thoroughly?


For the number typing - this reminded me of a couple things at once. 

1- isn't that actual session id suspiciously like the unix epoch time? https://www.unixtimestamp.com/ - which is a long normally
2- both the underlying APIs return numbers not strings, and while the firebase-js-sdk does not seem to expose session id for some reason (I could not find it in their types nor in an issue or PR in their repo or docs) but I think if firebase-js-sdk implements this, they will return number too

I propose returning it as a number at the JS level and you can follow the style from the firestore count implementation (where [firebase-js-sdk matches the numeric style](https://firebase.google.com/docs/reference/js/firestore_.md#getcountfromserver) from native SDK). It appears to work well (and if not, whatever fix is needed will be applicable here too as this will follow same style):

https://github.com/invertase/react-native-firebase/blob/b2670ea2537a35ae653cad210b27cac276036a4e/packages/firestore/android/src/reactnative/java/io/invertase/firebase/firestore/ReactNativeFirebaseFirestoreCollectionModule.java#L173

I think the e2e test really should have an some assertions based on:
- session id being non-null number when analytics collection is enabled
- session id being null when session timeout set to a very small number and a Utils.wait has elapsed for longer than that number

There are no methods yet for setConsent (unless you want to expose those 😅 ) so e2e cannot assert on null/non-null return dependent on that status yet

What do you think?

mikehardy (changes requested):
as commented above

mikehardy:
I always feel like a huge comment with all sorts of criticism (constructive or otherwise) feels bad, so I want to reiterate this:

> This looks fantastic! Thank you for posting it

I'm excited to see this go in - it's not far from going really :-)

dlackty:
While implementing tests on session id, I found that:

> session id being null when session timeout set to a very small number and a Utils.wait has elapsed for longer than that number

This might not work on iOS because Firebase iOS SDK doesn't provide session configuration method. Do you think it's okay to test this behavior only on Android? I'll still test session id != null on iOS though.

mikehardy:
That's odd 🤔 - I see the API here https://firebase.google.com/docs/reference/ios/firebaseanalytics/api/reference/Classes/FIRAnalytics#+setsessiontimeoutinterval:

and it appears to be implemented with a call through to native on both platforms?

https://github.com/invertase/react-native-firebase/blob/97a4ea573dd0795b157e73299e705dbd7bb7e3d4/packages/analytics/lib/index.js#L118-L132

Am I examining the wrong API?

dlackty:
Ah I got it, so it has different naming across platforms (`setSessionTimeoutDuration` / `setSessionTimeoutInterval`)

I was also confused by this:

https://github.com/invertase/react-native-firebase/blob/a9423dbe363ea67d3794d65d27714b3d24c7c928/packages/analytics/ios/RNFBAnalytics/RNFBAnalyticsModule.m#L114-L120

So maybe let me send a PR to fix this first. I come up with few possible options:

1. Just implement the native binding `setSessionTimeoutDuration` to call `setSessionTimeoutInterval` on iOS
2. Like above, but also add a `setSessionTimeoutInterval` method in JS as an alias
3. Add `setSessionTimeoutInterval` in JS and make it only calls `setSessionTimeoutInterval` on iOS, i.e. different method for different platforms

What do you think? I prefer 1. but 2. sounds good too.

mikehardy:
Ha! This one is more subtle + involved than it first seemed. Surprising

firebase-js-sdk does not support it so we did not have naming coherence across the different SDKs (https://firebase.google.com/docs/reference/js/v8/firebase.analytics.Analytics) and we just picked one then map it down the right thing underneath in the SDK...


...except as you see we apparently do not! Just no-oping on iOS, I did not think to look at the SDK, we usually do platform cutouts at the javascript layer.

Implementing it by just calling the appropriate firebase-ios-sdk method, no change to javascript layer seems perfect. Can be a commit on this PR to keep the whole thing together + easier to manage, suggest `feat(analytics, ios): implement setSessionTimeoutDuration for apple platform` as commit for so it's beautiful in the changelog

mikehardy:
Hey @dlackty :wave: - no pressure, just curious on your plans for this, it came to mind today because someone requested the feature in a discussion, and I linked this PR as an in-progress reference

rameshwaranb:
Any updates on this? This feature would be very useful for me.

mikehardy (dismissed):
This looks very very nice, CI kicking off now

mikehardy:
Hmm :thinking: neither of the e2e runs completed successfully, each failing for slightly different reasons?

android
```
  1) analytics()
       getSessionId()
         returns a non empty session ID:
     DetoxRuntimeError: expect() argument is invalid, expected a native or web matcher, but got number
      at AndroidExpect.expect (node_modules/detox/src/android/AndroidExpect.js:51:11)
      at Context.<anonymous> (/Users/runner/work/react-native-firebase/react-native-firebase/packages/analytics/e2e/analytics.e2e.js:58:7)
      at runMicrotasks (<anonymous>)
```

ios
```

  1) analytics()
       getSessionId()
         calls native fn without error:
     NativeFirebaseError: [analytics/unknown] Analytics uninitialized
      at FirebaseAnalyticsModule.getSessionId (http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false&modulesOnly=false&runModule=true&app=io.invertase.testing&sourceMapURL=true:6573:28)
      at Context.<anonymous> (/Users/runner/work/react-native-firebase/react-native-firebase/packages/analytics/e2e/analytics.e2e.js:54:34)

```

mikehardy:
> I did some research online and found that [people advise](https://stackoverflow.com/a/60691387/605165) against matching JS raw values in Detox E2E tests (should match UI elements instead). A quick search in current react-native-firebase code bases also show no expect usage in any e2e tests.

That doesn't really apply here. Our e2e app is not a regular app, it's really just an on-device engine for running code, where we can and do compare raw JS values with great success. So I do not think that's the problem, and either way we're definitely converting it to UI elements in this PR

There is no usage of expect but there is plenty of usage mocha matchers etc

Use a matching style more like this I think

https://github.com/invertase/react-native-firebase/blob/3c8e6ba2eb9ce5decf8f72fe9d9192ee65b2b3f9/packages/auth/e2e/auth.e2e.js#L43-L48

The other one where analytics is not initialized, I'm unsure on

dlackty:
Thanks for the direction! Indeed I overlooked `should` usage and I just updated the tests.
I changed the test direction a little bit to test "session expired" thrown correctly because I couldn't find a way to get session ID again after expiration.

dlackty:
Sorry about the mess. I'm only able to run iOS e2e test locally, and that initialization exception seems only happening on fresh setup. 

Let me set up GitHub Actions on my fork and once I got it passed, I'll ping you for reviewing. Sorry again for wasting your time.

mikehardy:
:thinking: any technical reason in-repo that the android tests don't work? Just curious
Also, not sure if it's too painful or not but you can reset the simulator data for local iOS testing, and/or alter the test device definitions to have it pop up a different (maybe fresh) simulator so you don't have to wait for CI on the fork - because that sounds *really* painful

dlackty:
For Android test setting, that's my dev environment and not repo-related. I was stuck at setting up `TestAVD' in Android Studio, but fortunately it now works after I re-installed Android Studio & SDKs.

For iOS testing, I've tried:

1. Clear app sandbox
2. Uninstall app
3. Erase all keychain data
4. Erase iOS content & settings
5. Delete the simulator from Xcode and recreate one

I was also surprised that altering the test device definitions (say, iPhone 14 -> iPhone 14 Pro) doesn't work. Not sure where's the data being cached. Anyway, at least I could try to reproduce the issue on GitHub Actions in my fork. Hope it could fix that in few tries 😅

There's an API design issue would like to hear your feedback.

`getSessionId` has different behaviors across iOS / Android:

- iOS throws exception (e.g. not initialized / session expired)
- Android just returns null when there's problem getting one

I think we should try to avoid inconsistency and provide unified interface for Read Native Firebase users. Though I would prefer throwing exceptions in both platforms and let API callers know the reason, Android just returns null and we can only guess the reason.

Could you please let me know your preferences? I'll implement accordingly. Thanks!

mikehardy:
Hmm - good question on API shape. I agree it should be consistent.
I actually prefer not to throw exceptions, but also a "polymorphic return" (that is, "real result | null") is not great either. 
We don't have any guidance from the JS SDK so it is really a tossup.
I'll go with implementor's choice then, and you are busy in there in the code, if you want an exception for the "no session id" case, that works for me. Just needs to be documented in the index.d.ts

giovaniortolani:
Hello! Any updates on this? Thanks for implementing it.

kwolniak:
I just leave it here:
```
FA | my.package_id | W | getSessionId has been disabled.
```

mikehardy:
@kwolniak - I don't understand  your comment - is the information you posted an answer to some question in the comment chain above? Or ?

giovaniortolani:
@dlackty, thanks for the implementation :handshake:. It will be very useful in my organization. We are looking forward to being able to use this feature.

Do you intend to finish this task?

mikehardy:
reopening to see if we can get this one in, it has a lot of user interest and I may have time to get it in shape

dlackty:
I also received some messages asking this PR, so I made a second attempt last weekend. The code is complete, but there is an "Analytics uninitialized" error on iOS tests that I have been unable to reproduce on my local device. It appears to only happen on fresh new VMs. I will continue to try to find a workaround on my forked repo to test with GitHub Actions and once I passed the CI, I'll ping @mikehardy to have a code review. Thanks!

dlackty:
@mikehardy Hi Mike, I've made improvements to the test by adding a retrying logic. Could you please review it and give me your feedback? I believe this could be beneficial for other teams as well, so I'm hoping it can be merged.

mikehardy (approved):
Sorry for the delay! This looks clean + correct, and your worry about the long e2e test is perhaps justified, but let's start with a default stance of "testing + working reliably" and then optimize as needed after. So I think it's fine as a start and I *really* like that it is tested + working reliably.

Nice sneaking in the iOS session duration set as well :-)

mikehardy:
I'm rebasing this to current main and will push it out, CI should run again and go green (or I'll do little fixups as needed to shepherd it through) then I think we can merge+release

mikehardy:
I noticed it was missing an implementation for the modular / firebase-js-sdk v9+ API style (that was added after this PR took flight I think!) so I added that and e2e support for it. It all passes testing + linting locally

CI should go green, then can merge

Fantastic, in my opinion @dlackty - thank you!

mikehardy (approved):

mikehardy:
Passed iOS CI full + Android past the analytics part (flaky test in auth, after analytics) stopped it. Good to go

PR #6683, opened 2022-11-10
fix(auth): Typescript support for Multi-factor Auth

### LINKED ISSUE #6661: react-native-firebase Typescript support for Multi-factor Auth
### Discussed in https://github.com/invertase/react-native-firebase/discussions/6656

<div type='discussions-op-text'>

<sup>Originally posted by **SO2OS1EN45QGF81** November  3, 2022</sup>
In the documentation it is mentioned to use
```
import auth from '@react-native-firebase/auth';
const multiFactorUser = await auth.multiFactor(auth());

...

const session = await multiFactorUser.getSession();
 and 
const verificationId = await auth().verifyPhoneNumberForMultiFactor(phoneOptions);
```
But it seems at least some of these functions are not available in Typescript ( verifyPhoneNumberForMultiFactor for example )
I am using v16.4.1
</div>

### LINKED ISSUE #6672: react-native-firebase Typescript support for Multi-factor Auth
I think something is missing in index.d.ts around line 317
Something like :

    /**
     * A multiFactor interface.
     */
    multiFactor(auth: FirebaseAuthTypes.Module): Promise<MultiFactorUser>;

BasBouw:
### Description

A Typescript function was missing, this adds the missing function

### Related issues

Fixes #6661  the final comment 
Fixes #6672 

### Checklist

  - [x] Yes
  - [x] `Android`
  - [x] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

---

### DIFF
--- a/packages/auth/lib/index.d.ts
+++ b/packages/auth/lib/index.d.ts
@@ -313,7 +313,12 @@ export namespace FirebaseAuthTypes {
      * A PhoneMultiFactorGenerator interface.
      */
     PhoneMultiFactorGenerator: PhoneMultiFactorGenerator;
-  }
+  
+    /**
+     * A multiFactor interface.
+     */
+    multiFactor(auth: FirebaseAuthTypes.Module): Promise<MultiFactorUser>;
+}
 
   /**
    * A structure containing additional user information from a federated identity provider via {@link auth.UserCredential}.


packages/auth/lib/index.d.ts
```diff
@@ -313,7 +313,12 @@ export namespace FirebaseAuthTypes {
      * A PhoneMultiFactorGenerator interface.
      */
     PhoneMultiFactorGenerator: PhoneMultiFactorGenerator;
-  }
+  
+    /**
+     * A multiFactor interface.
+     */
+    multiFactor(auth: FirebaseAuthTypes.Module): Promise<MultiFactorUser>;
```
mikehardy, line 320: Hmm :thinking: - this deviates in a significant way from the javascript typing, where the auth instance is *not* passed:

https://firebase.google.com/docs/reference/js/auth#multifactor

```typescript
export declare function multiFactor(user: User): MultiFactorUser;
```

It needs to take a `User` type, not the auth module, in it's exposed API.

I think this may need a structural change in javascript where these things happen:

1- the `multifactor` method in `packages/auth/lib/multiFactor.js` instead of being just stuck on statics in `packages/auth/lib/index.js` should be wrapped in an API that takes the user (to match the firebase-js-sdk), and that new method can call the current multiFactor.js constructor with itself as the module etc - this implies a change to `packages/auth/e2e/helpers.js` to create the user with the new method instead of just static access
2- the existing multifactor in statics should be declared deprecated and slated for removal in v17 here
3- the type in index.d.ts should now be altered to look like it does in the firebase-js-sdk reference

What do you think? And I'm curious what @fzuellich thinks, as multi-factor was their baby and it *is* working, but this is a bit of a rearrangement...

mikehardy (changes requested):

packages/auth/lib/index.d.ts
BasBouw: Mike, I am just the reporter, in no way do I pretend to have a final solution. In my project I can use the code as proposed. 
The example shows to use : 

`const multiFactorUser = await auth.multiFactor(auth());
`
And that call is missing in the TypeScript code.

The example I see you make, is based on version 9, while the react-native-auth seems to be based on version 8

mikehardy: API compatibility with the JS SDK is a vital part of react-native-firebase so I do want that, and we will have v9/modular API support here shortly so being close to that / ready to adapt to it is important even if we are currently modeled after v8 JS SDK API

@fzuellich  So we really do want to swap the argument out from the auth instance to a user instance. If there is a mismatch in capability (only one user may be handled native vs multiple on JS) we can just do a comparison on the user's userID vs auth.currentUser and throw an error indicating the same: "on mobile platforms only auth.currentUser is valid for multi-factor" or similar. Thanks for the cross-check that this doesn't seem terrible, assuming we handle the single user / multi-user difference you mention

@BasBouw no worries on only being the reporter, this is a real issue you've brought to our attention and it is important to fix. I'm going to close this PR but detail our thoughts in a separate issue + open a PR for the full fix.

Thanks!

mikehardy:
Will be superceded with #6689

PR #6689, opened 2022-11-12
fix(auth, multifactor): put multiFactor and getMultiFactorResolver on auth()

### LINKED ISSUE #6672: react-native-firebase Typescript support for Multi-factor Auth

mikehardy:
### Description

The original implementation of multifactor auth worked beautifully, but during review I should have noticed the methods were statics when they should have been off the auth() module

Granted, firebase-js-sdk does not support MFA in it's v8/chained API, however this lines the API up with the rest of the APIs on modules such that they will transform v9/modular API

This also fills in a missing type as noticed + requested in #6672

### Related issues

Fixes #6672

### Release Summary

THERE IS ONLY CONVENTIONAL COMMITS :laughing: 

### Checklist

  - [x] Yes
  - [x] `Android`
  - [x] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

Dropping the deprecated version will be a semver major later though

### Test Plan

I checked it incrementally while making changes by using the fabulous e2e tests provided by @fzuellich - trivial changes really, it worked every time as I altered API surface

---

### DIFF
--- a/docs/auth/multi-factor-auth.md
+++ b/docs/auth/multi-factor-auth.md
@@ -22,7 +22,7 @@ operations:
 
 ```js
 import auth from '@react-native-firebase/auth';
-const multiFactorUser = await auth.multiFactor(auth());
+const multiFactorUser = await auth().multiFactor(auth().currentUser);
 ```
 
 Request the session identifier and use the phone number obtained from the user
@@ -81,7 +81,7 @@ Using the error object you can obtain a
 continue the flow:
 
 ```js
-const resolver = auth.getMultiFactorResolver(auth(), error);
+const resolver = auth().getMultiFactorResolver(error);
 ```
 
 The resolver object has all the required information to prompt the user for a
@@ -143,7 +143,7 @@ authInstance
     const { code } = error;
     // Make sure to check if multi factor authentication is required
     if (code !== 'auth/multi-factor-auth-required') {
-      const resolver = auth.getMultiFactorResolver(authInstance, error);
+      const resolver = auth.getMultiFactorResolver(error);
 
       if (resolver.hints.length > 1) {
         // Use resolver.hints to display a list of second factors to the user
--- a/packages/auth/e2e/helpers.js
+++ b/packages/auth/e2e/helpers.js
@@ -212,7 +212,7 @@ exports.createUserWithMultiFactor = async function createUserWithMultiFactor() {
   const email = 'verified@example.com';
   const password = 'test123';
   await createVerifiedUser(email, password);
-  const multiFactorUser = await firebase.auth.multiFactor(firebase.auth());
+  const multiFactorUser = await firebase.auth().multiFactor(firebase.auth().currentUser);
   const session = await multiFactorUser.getSession();
   const phoneNumber = getRandomPhoneNumber();
   const verificationId = await firebase
@@ -242,7 +242,7 @@ exports.signInUserWithMultiFactor = async function signInUserWithMultiFactor(
     e.message.should.equal(
       '[auth/multi-factor-auth-required] Please complete a second factor challenge to finish signing into this account.',
     );
-    const resolver = firebase.auth.getMultiFactorResolver(firebase.auth(), e);
+    const resolver = firebase.auth().getMultiFactorResolver(e);
     let verificationId = await firebase
       .auth()
       .verifyPhoneNumberWithMultiFactorInfo(resolver.hints[0], resolver.session);
--- a/packages/auth/lib/index.d.ts
+++ b/packages/auth/lib/index.d.ts
@@ -1871,6 +1871,19 @@ export namespace FirebaseAuthTypes {
      * @param url: emulator URL, must have host and port (eg, 'http://localhost:9099')
      */
     useEmulator(url: string): void;
+    /**
+     * Provides a MultiFactorResolver suitable for completion of a multi-factor flow.
+     *
+     * @param error: The MultiFactorError raised during a sign-in, or reauthentication operation.
+     */
+    getMultiFactorResolver(error: MultiFactorError): MultiFactorResolver;
+    /**
+     * The MultiFactorUser corresponding to the user.
+     *
+     * This is used to access all multi-factor properties and operations related to the user.
+     * @param user The user.
+     */
+    multiFactor(user: User): MultiFactorUser;
   }
 }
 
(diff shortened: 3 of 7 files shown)


docs/auth/multi-factor-auth.md
```diff
@@ -81,7 +81,7 @@ Using the error object you can obtain a
 continue the flow:
 
 ```js
-const resolver = auth.getMultiFactorResolver(auth(), error);
+const resolver = auth().getMultiFactorResolver(error);
```
fzuellich, line 84: I believe this deviates from the Web-API, correct? https://firebase.google.com/docs/auth/web/multi-factor#signing_users_in_with_a_second_factor

```javascript
.catch(function (error) {
        if (error.code == 'auth/multi-factor-auth-required') {
            // The user is a multi-factor user. Second factor challenge is required.
            resolver = getMultiFactorResolver(auth, error);
            // ...
        } else if (error.code == 'auth/wrong-password') {
            // Handle other errors such as wrong password.
        }
```

I remember I tried to make the import work as well, when implementing MFA, but I couldn't make it work:

`import { getAuth, getMultiFactorResolver} from "firebase/auth";`

packages/auth/e2e/helpers.js
```diff
@@ -212,7 +212,7 @@ exports.createUserWithMultiFactor = async function createUserWithMultiFactor() {
   const email = 'verified@example.com';
   const password = 'test123';
   await createVerifiedUser(email, password);
-  const multiFactorUser = await firebase.auth.multiFactor(firebase.auth());
+  const multiFactorUser = await firebase.auth().multiFactor(firebase.auth().currentUser);
```
fzuellich, line 215: If we want to follow the API closely, then the `multiFactor` function should be available as import and not be derived from an Auth instance. Looking at the implementation in `firebase-js-sdk` we could probably copy that (packages/auth/src/mfa/mfa_user.ts): Instantiate the user by copying the data from the user object.

packages/auth/lib/index.d.ts
```diff
@@ -1871,6 +1871,19 @@ export namespace FirebaseAuthTypes {
      * @param url: emulator URL, must have host and port (eg, 'http://localhost:9099')
      */
     useEmulator(url: string): void;
+    /**
+     * Provides a MultiFactorResolver suitable for completion of a multi-factor flow.
+     *
+     * @param error: The MultiFactorError raised during a sign-in, or reauthentication operation.
+     */
+    getMultiFactorResolver(error: MultiFactorError): MultiFactorResolver;
```
fzuellich, line 1879: I'm curious about your opinion. Should this be typed as `unknown` just the way Typescript types exceptions in the catch-clause, or do we want to make the user check that the exception fits a `MultiFactorError` when calling?

packages/auth/lib/multiFactor.js
```diff
@@ -2,14 +2,19 @@
  * Return a MultiFactorUser instance the gateway to multi-factor operations.
  */
 export function multiFactor(auth) {
+  // eslint-disable-next-line no-console
+  console.warn('This method is deprecated. Please auth().multiFactor(user) instead');
```
fzuellich, line 6: Missing a use?

```suggestion
  console.warn('This method is deprecated. Please use auth().multiFactor(user) instead');
```

fzuellich (commented):
Hey Mike, thanks for taking care of the "broken" API! I thought I'd chime in without being asked. ;) I'm uncertain if I understand the reason for putting the functions on auth(). Please ignore my remarks about deviating from the version 9 Web API, if that's not your intention.

docs/auth/multi-factor-auth.md
mikehardy: The problem with these specific APIs is that they are not on v8 at all so we do not have a model.
However, we do have a model for APIs that are on both, and the typical part is:

- v8-style: `auth().method(params)`
- v9-style: `method(getAuth(), params)`

Since these ones follow the typical v9 style of being a method in auth package that takes an auth instance, I back-APId them to the typical "method on the auth instance" style vs the static export.

The goal is that they feel like the rest of the chained v8 APIs and then when we do the big v8->v9 API port (which may actually start next week!) it's the same methodical step for these APIs as the rest 

Does that make sense? What do you think, looking at it from that lense - the back-port-style API perspective?

packages/auth/e2e/helpers.js
mikehardy: again I think this is just an artifact of me taking a v9-only JS API and faking a "v8 style" for it. This should resolve exactly as you say in the coming v9 style - it will be an export and you'll be sending result of `getAuth()` into it

packages/auth/lib/index.d.ts
mikehardy: I want it to be typed exactly the same as it is in firebase-js-sdk (allowing for the v8/v9 style difference from previous comments) - and this was copy/pasted from their docs, so, I think this is the right way to go?

packages/auth/lib/multiFactor.js
mikehardy: :facepalm: nice catch thank you

mikehardy:
@fzuellich 

>  I thought I'd chime in without being asked

On the contrary - I specifically asked on the linked issue and I appreciate the review a lot, it'll make the final product better

docs/auth/multi-factor-auth.md
fzuellich: I think it makes sense. 👍

packages/auth/e2e/helpers.js
fzuellich: 👍

packages/auth/lib/index.d.ts
fzuellich: Looked it up. This is the exact typing `firebase-js-sdk` uses. 👍

fzuellich (dismissed):
Looking good. Existing code shouldn't be too hard to update.

docs/auth/multi-factor-auth.md
mikehardy: Cool - and I temporarily have a bunch of the Invertase crew besides myself crawling in the repo starting this week doing the v9 API, which is very exciting for me (I'm usually solo here in this repo, and v9 finally!). No deadlines because open source...but we should have v9 API support soon, alongside v8, so people can start migrating.

mikehardy:
Just re-pushed with a squash of commits 2+3, no functional change but I'll wait for CI to clear

PR #6694, opened 2022-11-14
feat(app): migrate `app` module to a modular API

Lyokone:
### Checklist

  - [ ] Yes
  - [ ] `Android`
  - [ ] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [ ] No

### Test Plan

---


Lyokone:
Yeah didn't knew what to do about those, but I can definitely add those 👍

mikehardy:
I think it depends on the goal, and I'm not sure we have a defined goal there which is why I did an @ for Elliot.
If our goal is for people to not have to do a bunch of `if (Platform.....)` then at least stubbing them might be nice. I personally think this would be nice.

But maybe being explicit that they are not supported is better, and making it explicit by not implementing them, so people have to Platform is "safest" ? I personally don't think that's useful in practice but again I could be missing something.

So my opinions are weakly held on stubbing the non-native APIs but I lean towards the idea stubbing them and eliminating all need for Platform checks would be cool

Lyokone:
The coverage bot is saying that some lines are not tested even thought the test was properly added.

packages/app/e2e/app.v9.e2e.js
```diff
@@ -0,0 +1,316 @@
...
+      const newApp = await initializeApp(platformAppConfig, name);
+
+      newApp.name.should.equal(name);
+      newApp.toString().should.equal(name);
+      newApp.options.apiKey.should.equal(platformAppConfig.apiKey);
+
+      await deleteApp(newApp);
+      try {
+        await deleteApp(newApp);
+      } catch (e) {
```
mikehardy, line 274: this one and the next five that should throw either need to have a fail after the line that throws (like a couple cases above) to make *sure* it threw, or they should be some special assertion that expects an exception.

I think we tried the method that was like `expectException` or whatever they called it and it wasn't working for some unimportant reason so we typically just throw something like 
```javascript
        throw new Error('Should have rejected incorrect initializeApp input');
````
after the lines that should throw

Lyokone: Oh okay, I see why we need to do that, thanks :) 
I've added this to the following cases :)

mikehardy (approved):
LGTM - I don't want to delay any longer sorry it sat this long!

PR #6720, opened 2022-11-23
feat(firestore): make nested collection as generic

bglgwyng:
…nts.

### Description

Currently, top level `collection` is generic so you can type its documents by `firestore().collection<User>()`.
However, nested collection that looks like `firestore().collection<User>().doc(userId).collection()` is not generic.
Then it gives its document data naive `DocumentData` type which helps developers little.
I made a very small change that changed nested `collection` into generic also, not only top level one.

### Checklist

  - [x] Yes
  - [ ] `Android`
  - [ ] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

---

### DIFF
--- a/packages/firestore/lib/index.d.ts
+++ b/packages/firestore/lib/index.d.ts
@@ -237,7 +237,9 @@ export namespace FirebaseFirestoreTypes {
      *
      * @param collectionPath A slash-separated path to a collection.
      */
-    collection(collectionPath: string): CollectionReference;
+    collection<T extends DocumentData = DocumentData>(
+      collectionPath: string,
+    ): CollectionReference<T>;
 
     /**
      * Deletes the document referred to by this DocumentReference.
@@ -2084,7 +2086,7 @@ export namespace FirebaseFirestoreTypes {
      *
      * #### Example
      *
-     * ```js
+     * ```jsc
      * const collectionGroup = firebase.firestore().collectionGroup('orders');
      * ```
      *


packages/firestore/lib/index.d.ts
```diff
@@ -2084,7 +2086,7 @@ export namespace FirebaseFirestoreTypes {
      *
      * #### Example
      *
-     * ```js
+     * ```jsc
```
mikehardy, line 2089: I don't understand this change, it does not seem relevant to the PR, and it does not seem to be in the [languages.yaml](https://github.com/github/linguist/blob/master/lib/linguist/languages.yml) file anywhere I would expect for javascript highlighting in [the package github uses](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-and-highlighting-code-blocks#syntax-highlighting) ?

packages/firestore/lib/index.d.ts
```diff
@@ -237,7 +237,9 @@ export namespace FirebaseFirestoreTypes {
      *
      * @param collectionPath A slash-separated path to a collection.
      */
-    collection(collectionPath: string): CollectionReference;
+    collection<T extends DocumentData = DocumentData>(
+      collectionPath: string,
+    ): CollectionReference<T>;
```
mikehardy, line 242: Normally we try to match the types from firebase-js-sdk, I am consulting those and I don't see the generics passed in but I welcome a pointer to a better set of reference lines, I can definitely be missing something here:

https://github.com/firebase/firebase-js-sdk/blob/e650f649854f3c39737fe4bade43f9eedc3e611f/packages/firestore-types/index.d.ts#L252

They have generics in a bunch of other places, I wonder why not here? Until their types move I am cautious on moving ours, as I consider their types "upstream" with ours derivative.

Have you brought this up with firebase-js-sdk?

mikehardy (changes requested):
This seems like it might work, but I left a note about the upstream firebase-js-sdk types - we need them to match and we typically can't move first

packages/firestore/lib/index.d.ts
bglgwyng: Sorry, I missed this change. This isn't intended one(maybe from linter?). I'll remove this.

packages/firestore/lib/index.d.ts
bglgwyng: No. I didn't know that typings here were from firebase-js-sdk. 
I think that I should make a pull request on firebase-js-sdk repository first.

PR #6771, opened 2022-12-07
feat(perf): Expose modular API that matches the Firebase web JS SDK v9 API

russellwheatley:
### Description

 Firebase web JS SDK v9 mapping. We've tried our best to make it as JS centric as possible to keep it as close to the web API as we can. If you want more information, I'd suggest reading the feedback to follow reasoning.

### Checklist

  - [X] Yes
  - [X] `Android`
  - [X] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [X] No

### Test Plan

---


Lyokone:
```it would have to be async so we could call the native to update dataCollectionEnabled```

I thought React Native had no constraints on calling the native part 🤔 Still learning 😅 

LGTM on first read, but would love to get @mikehardy thoughts

**Comment by russellwheatley** [comment]
@mikehardy I've updated to include the `initializePerformance()`. I don't think we can get rid of `setPerformanceCollectionEnabled()` or `isPerformanceCollectionEnabled()`. It would really diminish flexibility for the user. But I'll defer to your opinion 😄. If you have any other suggestions, let me know!

**Comment by mikehardy** [comment]
I am still mulling over this one, I'm struggling with the fundamental design goal vs this proposed mapping.

My goal is to hopefully have the modular API renewal effectively handle the work for react-native-web compatibility as well (or at least as much as possible) which means that (as much as possible...) we need 100% of the firebase-js-sdk's to map to their analogues (even if named differently) in ios/android so the code needs zero changes.

I'd set that as a firm design goal of any modular API mapping really:

1) *all* firebase-js-sdk code should work out of the box without modification on the react-native-firebase modular APIs, doing what a reasonable person expects it to do.

2) So a follow-on firm design goal should be that the only non-firebase-js-sdk APIs we have are native-only APIs. *Not* APIs that are just named differently or were accessed differently in the past, but actual novel capabilities only present in the native SDKs.

In that sense. the `dataCollectionEnabled` and `instrumentationEnabled` properties are not novel to the native SDK I think. The only allowance we need to make for native vs firebase-js-sdk is that it needs to be async as you mention (because of react-native bridge asynchronicity)

We need to support the `initializePerformance` API with the optional settings option, but it is documented as "only callable one time", so we can expect JS code use that API but also to access the `Performance.dataCollectionEnabled` and `Performance.instrumentationEnabled` properties for setting things as well and I think we need to support that or we'll have a fidelity issue with our JS API mapping / idiomatic usage - it won't feel the same using the APIs here if those properties don't work or if those properties don't behave as expected.

That's all theory / design discussion (and feel free to disagree, push back, whatever!)

But based on that, this statement:

>  don't think we can get rid of setPerformanceCollectionEnabled() or isPerformanceCollectionEnabled(). It would really diminish flexibility for the user.

...is not correct I think. *If* you grant that a settable `Performance.dataCollectionEnabled` property (either used in `initializePerformance` when called with properties, *or* set directly) is equivalent to the native idea of "PerformanceCollectionEnabled", *then* we are not removing flexibility for the user *assuming* we have that property and it works.

For the other property (`instrumentationEnabled`) I think it should always be boolean true since instrumentation happens at compile time in the native SDKs (if I understand correctly?) and setting it can just be documented as a no-op, but not error (so JS code will interoperate well)

What do you think?

File: packages/perf/lib/index.d.ts [hunk 1]
```diff
@@ -384,7 +384,22 @@ export namespace FirebasePerformanceTypes {
...
+     *
+     * #### Example
+     *
+     * ```js
+     * const isEnabled = firebase.perf().isDataCollectionEnabled;
+     * console.log('Performance collection enabled: ', isEnabled);
+     * ```
+     */
 
+    isDataCollectionEnabled: boolean;
```
mikehardy, line 402: These deviate from the firebase-js-sdk API where they are properties that are settable, but are not getters/setters - that (javascript properties) is something we do not do very much in react-native-firebase but we do it in a couple places, and this is one where we have to in order to be firebase-js-sdk compatible

https://github.com/firebase/firebase-js-sdk/blob/49ee786f2b022e65aef45693e1a8b546d889ec10/packages/performance-types/index.d.ts#L30-L34

russellwheatley: @mikehardy  How will that work with existing API?

Currently:
```js

  get isPerformanceCollectionEnabled() {
    return this._isPerformanceCollectionEnabled;
  }

  get isInstrumentationEnabled() {
    return this._isInstrumentationEnabled;
  }

  isDataCollectionEnabled() {
    return this.isPerformanceCollectionEnabled;
  }

  setDataCollectionEnabled(enabled) {
    return this.setPerformanceCollectionEnabled(enabled);
  }

  setPerformanceCollectionEnabled(enabled) {
    if (!isBoolean(enabled)) {
      throw new Error(
        "firebase.perf().setPerformanceCollectionEnabled(*) 'enabled' must be a boolean.",
      );
    }
```

Just so I understand. Wouldn't I have to remove existing methods `setPerformanceCollectionEnabled` & `isPerformanceCollectionEnabled` if we're moving to a simple `isDataCollectionEnabled` property without getter/setter? This would mean we wouldn't have the type checking for `enabled` here:

```js
  setPerformanceCollectionEnabled(enabled) {
    if (!isBoolean(enabled)) {
      throw new Error(
        "firebase.perf().setPerformanceCollectionEnabled(*) 'enabled' must be a boolean.",
      );
    }
```

packages/perf/lib/modular/index.js
```diff
@@ -0,0 +1,70 @@
...
+
+  return perf;
+}
+
+/**
+ * Determines whether performance monitoring is enabled or disabled.
+ * @param perf - Performance instance
+ * @returns {Boolean}
+ */
+export function isPerformanceCollectionEnabled(perf) {
```
russellwheatley, line 38: @mikehardy I guess we remove this as it isn't part of Firebase JS web API?

packages/perf/lib/modular/index.js
```diff
@@ -0,0 +1,70 @@
...
+  return perf.isPerformanceCollectionEnabled;
+}
+
+/**
+ * Returns nothing.
+ * @param perf - Performance instance
+ * @param enabled - A Boolean for setting Performance collection
+ * @returns {void}
+ */
+export function setPerformanceCollectionEnabled(perf, enabled) {
```
russellwheatley, line 48: and this

packages/perf/lib/modular/index.js
```diff
@@ -0,0 +1,70 @@
...
+  return perf.newTrace(identifier);
+}
+
+/**
+ * Returns a HttpMetric instance.
+ * @param perf - Performance instance
+ * @param identifier - A String to identify the HttpMetric instance
+ * @returns {HttpMetric}
+ */
+export function httpMetric(perf, identifier, httpMethod) {
```
russellwheatley, line 68: and this can be removed as well. It isn't a web API

packages/perf/lib/index.d.ts
```diff
@@ -384,7 +384,22 @@ export namespace FirebasePerformanceTypes {
...
+    /**
+     * Determines whether performance monitoring is enabled or disabled.
+     *
+     * #### Example
+     *
+     * ```js
+     * const isEnabled = firebase.perf().dataCollectionEnabled;
+     * console.log('Performance collection enabled: ', isEnabled);
+     * ```
+     */
```
mikehardy, line 401: will this doc comment associate correctly with the blank line in between comment and property? Probably? But it's different then the rest

```suggestion
```

packages/perf/lib/modular/index.js
mikehardy: good point - the JS web API for modular exposes the javascript properties and the initialize method but not this specific getter or a setter - should remove, I think you are right

packages/perf/lib/modular/index.js
mikehardy: good point - the JS web API for modular exposes the javascript properties and the initialize method but not this specific getter or a setter - should remove, I think you are right

packages/perf/lib/index.js
```diff
@@ -47,12 +56,35 @@ class FirebasePerfModule extends FirebaseModule {
...
+
+  get dataCollectionEnabled() {
+    return this.isPerformanceCollectionEnabled;
+  }
+
+  set dataCollectionEnabled(enabled) {
+    if (!isBoolean(enabled)) {
+      throw new Error("firebase.perf().dataCollectionEnabled = 'enabled' must be a boolean.");
+    }
+    this._isPerformanceCollectionEnabled = enabled;
```
mikehardy, line 85: I think right here should be the current implementation of `setPerformanceCollectionEnabled` (code just below this comment) so that it actually turns on or off the native logging of perf events. Right now this doesn't actually do anything, but `this.native.setPerformanceCollectionEnabled(enabled)` is the effect it *should* have I think?

Then setPerformanceCollectionEnabled should delegate to this in it's implementation, and in the future state where it is gone completely this method (with the native call...) is all that's left + and it works

packages/perf/lib/index.js
```diff
@@ -47,12 +56,35 @@ class FirebasePerfModule extends FirebaseModule {
...
 
+  get instrumentationEnabled() {
+    return this._instrumentationEnabled;
+  }
+
+  set instrumentationEnabled(enabled) {
+    if (!isBoolean(enabled)) {
+      throw new Error("firebase.perf().instrumentationEnabled = 'enabled' must be a boolean.");
+    }
+    this._instrumentationEnabled = enabled;
```
mikehardy, line 74: don't look now, but iOS actually has a native API for this. it may not be implemented in native yet but is likely trivial (should mirror the setPerformanceCollectionEnabled API...) but should be implemented and should be called in this setter

packages/perf/lib/index.js
```diff
@@ -47,12 +56,35 @@ class FirebasePerfModule extends FirebaseModule {
...
     this._isPerformanceCollectionEnabled = this.native.isPerformanceCollectionEnabled;
+    this._instrumentationEnabled = true;
   }
 
   get isPerformanceCollectionEnabled() {
     return this._isPerformanceCollectionEnabled;
   }
 
+  get instrumentationEnabled() {
+    return this._instrumentationEnabled;
```
mikehardy, line 67: on iOS should return the value of this property https://firebase.google.com/docs/reference/ios/firebaseperformance/api/reference/Classes/FIRPerformance#instrumentationenabled

on android I'm not sure if there is any way to tell, should just return true I guess, with an implementation note saying the same in docs?

packages/perf/lib/modular/index.js
mikehardy: I think this is one where we need to leave it. Native is a superset of web in general and this is a function that is on android and ios and is useful. It does not represent an incompatibility because if you were on web you don't care if there is an extra feature, but if you need it on native you will want it exposed.

Then for a web implementation (in future) we would stub it instead of calling in to JS API

mikehardy (changes requested):
Really close now - the properties seem to line up well
- questions/thoughts as commented about how to actually call in to the native collection enable/disable toggles
- not commented except here: the codecov checks are still unhappy after the experiment with removing tests, not sure that will work? 🤔 

Dangerous / false precision to assign a percentage to it, but I'd say this is 90% done, in that all the real thought behind it is complete really. And doing an iOS native API as commented - in this module at least / for the one contemplated - really is trivial (not always the case with obj-c as I'm sure you are very aware haha)

packages/perf/lib/index.js
russellwheatley: I think you're right. It won't be await'ed (i.e. it's not a promise) but I guess it doesn't matter because it only takes effect the next time the app is run 🤔

russellwheatley: If I understand you correctly, I don't think `setPerformanceCollectionEnabled` can delegate to this as it is a setter. You cannot return a setter.

mikehardy: You cannot return a setter, but maybe you can call it internally? Just thinking how to de-duplicate the code (DRY principle) but that's an implementation detail. As long as the backwards-compatible method and the setter are actually *doing* the necessary thing, it's okay

packages/perf/ios/RNFBPerf/RNFBPerfModule.m
```diff
@@ -73,7 +73,6 @@ + (BOOL)requiresMainQueueSetup {
                   : (RCTPromiseResolveBlock)resolve rejecter
                   : (RCTPromiseRejectBlock)reject) {
   [FIRPerformance sharedInstance].dataCollectionEnabled = (BOOL)enabled;
-  [FIRPerformance sharedInstance].instrumentationEnabled = (BOOL)enabled;
```
mikehardy, line 76: Removing this here seems correct, but I wonder if we need to also export instrumentationEnabled in constantsToExport  so we can use it in the javascript layer

packages/perf/lib/index.js
```diff
@@ -47,12 +50,41 @@ class FirebasePerfModule extends FirebaseModule {
   constructor(...args) {
     super(...args);
     this._isPerformanceCollectionEnabled = this.native.isPerformanceCollectionEnabled;
+    this._instrumentationEnabled = true;
```
mikehardy, line 53: If platform is iOS this has to call through to the "I think we should export it" `this.native.isInstrumentationEnabled` ?

packages/perf/lib/index.js
mikehardy: Something about the local / javascript-level constants bothers me - in gray-box situations (where people have native code that might perturb the state) this could be out of sync with the actual native state of data collection enabled (both platforms) and instrumentation enabled (ios at least).

I think it is most correct to get rid of the javascript state holders and just always get the native state - and it would simplify things really. If Platform == android just always return true for instrumentation enabled but otherwise just always reach in and set them or reach in and get them from native

mikehardy (commented):
I'm sorry I don't think of everything at once, but I just had a correctness thought for mixed (react-native / native) developments and had a thought about those javascript level variables...I *think* it will actually be simplifying ?

packages/perf/lib/index.js
russellwheatley: I think the best we can do is:

```js
 set dataCollectionEnabled(enabled) {
    if (!isBoolean(enabled)) {
      throw new Error("firebase.perf().dataCollectionEnabled = 'enabled' must be a boolean.");
    }
    this.setPerformanceCollectionEnabled(enabled);
  }
```

No way of de-duplicating the Error I'm afraid, but it will at least throw the correct error message for the correct API.

mikehardy: That works for me!

russellwheatley:
@mikehardy not sure what to do about the code coverage as it is covered in the e2e tests 🤔

mikehardy:
> @mikehardy not sure what to do about the code coverage as it is covered in the e2e tests 🤔

Looks like something around 54% is our actual baseline *if* the Java coverage files upload, it's around 72% if the Java coverage files do not upload:

- with java coverage: https://app.codecov.io/gh/invertase/react-native-firebase/commit/97a5523bac4570ba7b47fbfdbb22348f38246d11
- without java coverage: https://app.codecov.io/gh/invertase/react-native-firebase/commit/da82c1036051f0518da0401de24cef24c7ac091f

So the codecov/project coverage @ 54ish % is actually expected and shouldn't block the PR, it just represents a failure on previous main/HEAD commit to successfully run+upload java coverage leading codecov to temporarily think we should be at 70+%

That's a codecov config / flake type issue, not related to this PR

packages/perf/lib/index.js
```diff
@@ -47,12 +50,40 @@ class FirebasePerfModule extends FirebaseModule {
...
+    return this._isPerformanceCollectionEnabled;
+  }
+
+  set dataCollectionEnabled(enabled) {
+    if (!isBoolean(enabled)) {
+      throw new Error("firebase.perf().dataCollectionEnabled = 'enabled' must be a boolean.");
+    }
+    this.setPerformanceCollectionEnabled(enabled);
+  }
+
```
mikehardy, line 86: for 100% backwards compatibility, people that currently call setPerformanceCollectionEnabled on iOS were historically getting both the effect of what is now `this.native.setPerformanceCollectionEnabled` and `this.native.setInstrumentationEnabled` so this method `setPerformanceCollectionEnabled` just below (if you expand the diff) - which is non-JS-SDK / historical should also toggle isInstrumentationEnabled / call through to same on native (for iOS at least?) and in index.d.ts should have a note I think?

perfect JS compatibility (or as close as we can get) + backwards compatibility is super fun

mikehardy: I think the setPerformanceCollectionEnabled method should also be marked deprecated ?

packages/perf/lib/index.js
```diff
@@ -47,12 +50,40 @@ class FirebasePerfModule extends FirebaseModule {
...
+  get dataCollectionEnabled() {
+    return this._isPerformanceCollectionEnabled;
+  }
+
+  set dataCollectionEnabled(enabled) {
+    if (!isBoolean(enabled)) {
+      throw new Error("firebase.perf().dataCollectionEnabled = 'enabled' must be a boolean.");
+    }
+    this.setPerformanceCollectionEnabled(enabled);
```
mikehardy, line 84: And now, irritatingly because I think I was trying to de-duplicate and *asked* for this to delegate, I don't think this can call through because the non-JS-API `this.setPerformanceCollectionEnabled` needs to toggle both to match historical behavior

I was tempted to be a bit cynical and question if anyone was even using this but it is actually a vital part of a correct GDPR opt-in flow and I am motivated to respect user privacy so...no cynicism, just effort 😆

mikehardy (approved):
I looked again, I can't see anything wrong. Fantastic @russellwheatley

mikehardy (changes requested):
Oh snap! We need a modular version of the new screen trace method from #6588 

Should be easy, it's just the addition of the transformed API ?

packages/perf/lib/index.d.ts
```diff
@@ -486,7 +486,7 @@ export namespace FirebasePerformanceTypes {
 
     /**
      * Creates a ScreenTrace instance with the given identifier.
-     * Throws if hardware acceleration is diabled or if Android is 9.0 or 9.1.
+     * Throws if hardware acceleration is disabled or if Android is 9.0 or 9.1.
```
mikehardy, line 489: Nice one

mikehardy (approved):
nailed it

PR #6801, opened 2022-12-25
docs(expo): update Expo managed integration

ssysm:
### Description

Add relevant information for pinning Firebase SDK version number to iOS Podfile after prebuild. Workaround for mismatch version number between RNF and Expo's module. 

### Related issues

N/A

### Checklist

  - [x] Yes
- This is a breaking change;
  - [ ] Yes
  - [X] No

---

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -323,6 +323,14 @@ iOS only, please use [expo-build-properties](https://docs.expo.dev/versions/late
 
 Next, you need to use the `expo prebuild --clean` command as described in the ["Adding custom native code"](https://docs.expo.io/workflow/customizing/) guide to rebuild your app with the plugin changes. If this command isn't run, you'll encounter connection errors to Firebase. Note, this is not required if you build your project with EAS (Expo Application Services).
 
+iOS only, after prebuild is done, manually pin the Firebase SDK version in `Podfile` inside the `ios` folder. Add this at the beginnig of the file, this is a workaround for version mismatch for Expo's module:
+
+```ruby
+$FirebaseSDKVersion='10.3.0' # The SDK used in the React Native Firebase Module.
+```
+
+The speific version number can be found in `node_modules/@react-native-firebase/app/package.json` under `sdkVersions.ios.firebase`
+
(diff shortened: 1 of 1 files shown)


mikehardy (changes requested):
Hi there! Interesting. People have been combining Expo and react-native-firebase for a while now and I have not heard anyone mention this specific problem, and your instructions do not give the reader any ability to determine: 1) if they have this problem (is there some error message that would indicate they are suffering from whatever this would fix?) 2) what the actual problem is (why is it happening? SDK versions are pretty serious things to change or pin, it should not be done lightly and should really be solved by recommending compatible react-native module versions, not pinning underlying native SDK versions) 3) whether this is a good idea or not and 4) when they can remove it (if you add this, you're now responsible for managing the underlying native SDK version and you will be in a race with me - who works on this module 10-20hrs/week - to keep it up to date, is that something you want to sign up for?)

So in general, I'm really not sure about this one as a recommendation other than as an emergency measure with a lot of information on why/what/how/when to do it + remove it accompanied by some upstream PR link where the real problem in the other react-native modules is being fixed so this is not necessary

ssysm:
Hello! Thanks for the response, I'll try to replicate this issue again in the next few days and get back to you.

ssysm:
Hi there, sorry for the late response, I have tried to recreate this issue using a brand new expo app and I was not able to reproduce this issue. Seems like my expo based RN app is a bit old, but I was able to reproduce the issue in my old project with the following error without pinning the pod version:

```bash
[!] CocoaPods could not find compatible versions for pod "Firebase/Core":
  In snapshot (Podfile.lock):
    Firebase/Core (= 10.3.0)

  In Podfile:
    EXFirebaseAnalytics (from `../node_modules/expo-firebase-analytics/ios`) was resolved to 7.0.0, which depends on
      Firebase/Core (= 8.14.0)


You have either:
 * out-of-date source repos which you can update with `pod repo update` or with `pod install --repo-update`.
 * changed the constraints of dependency `Firebase/Core` inside your development pod `EXFirebaseAnalytics`.
   You should run `pod update Firebase/Core` to apply changes you've made.
```

I'm using Expo version 45, and the module in question `expo-firebase-*` is no longer available in version 48. I don't know how my project got mixed up with RNF and expo's firebase module, but it does not seems like a library problem. I'll close this PR for now and leave my finding here in case anyone else is stuck on an older expo version.

PR #6806, opened 2022-12-28
feat(messaging): Expose modular API that matches the Firebase web JS SDK v9 API

russellwheatley:
## Action Items summary (details under the "API review" header)
1. What to do about the `GetTokenOptions.serviceWorkerRegistration` value. We could pull in this Type: https://www.npmjs.com/package/@types/serviceworker and implement in the `index.d.ts` file?
2. What to do about the `isSupported()` API? I created a new function in existing code but it perhaps it shouldn't be there?
	1. If it is supposed to be there, we have to pass in `messaging` as the first argument which isn't the same as the Firebase Web JS SDK v9. 
3. Should we put `icon`, `image` & `title` on their own separate `web` object like `android` & `ios` for `Notification` payload?
4. I've also modularised a bunch of API that have no corresponding API on the Firebase Web JS SDK v9. I assume they are fine as they simply wrap around existing.

### Firebase Web JS SDK v9
```js
export declare function getMessagingInWindow(app?: FirebaseApp): Messaging;
```

### RNFB modular v9

```dart
export function getMessaging(app) {
if (app) {
  return firebase.app(app.name).messaging();
}

return firebase.app().messaging();
}
```

### Compare API
The same.

### Firebase Web JS SDK v9
```js
export declare function deleteToken(messaging: Messaging): Promise<boolean>;
```

### RNFB modular v9

```js
export function deleteToken(messaging, tokenOptions) {
  if (tokenOptions != null) {
    return messaging.deleteToken();
  }

  return messaging.deleteToken(tokenOptions);
}
```

### Existing API

```js
deleteToken({ appName, senderId } = {}) {
    if (!isUndefined(appName) && !isString(appName)) {
      throw new Error("firebase.messaging().deleteToken(*) 'appName' expected a string.");
    }

    if (!isUndefined(senderId) && !isString(senderId)) {
      throw new Error("firebase.messaging().deleteToken(*) 'senderId' expected a string.");
    }

    return this.native.deleteToken(
      appName || this.app.name,
      senderId || this.app.options.messagingSenderId,
    );
  }
```

### Compare API
1. We allow passing `tokenOptions` as a second argument as it is needed for native SDKs. 
2. I have updated `projectId` to `appName` string in `Error()` thrown.

### Firebase Web JS SDK v9
```js
export declare function getToken(messaging: Messaging, options?: GetTokenOptions): Promise<string>;
```

### RNFB modular v9

```js
export function getToken(messaging, options) {
  if (options != null) {
    return messaging.getToken();
  }

  return messaging.getToken(options);
}
```

### Existing API

```js
  getToken({ appName, senderId } = {}) {
    if (!isUndefined(appName) && !isString(appName)) {
      throw new Error("firebase.messaging().getToken(*) 'appName' expected a string.");
    }

    if (!isUndefined(senderId) && !isString(senderId)) {
      throw new Error("firebase.messaging().getToken(*) 'senderId' expected a string.");
    }

    return this.native.getToken(
      appName || this.app.name,
      senderId || this.app.options.messagingSenderId,
    );
  }
```

### Compare API
1. I have updated `projectId` to `appName` string in `Error()` thrown.
2. I've removed the type `TokenOptions` in RNFB. Introducing `GetTokenOptions` & `DeleteTokenOptions`. 
3. `GetTokenOptions` is the same as there is no corresponding object in the Firebase Web JS SDK
5. [[Action item]] What to do about `serviceWorkerRegistration` type?

`GetTokenOptions`

```ts
  /**
   * Options for `getToken()`
   */
  export interface GetTokenOptions {
    /**
     * The app name of the FirebaseApp instance.
     *
     * @platform android Android
     */
    appName?: string;

    /**
     * The senderID for a particular Firebase project.
     *
     * @platform ios iOS
     */
    senderId?: string;
  }
```

`DeleteTokenOptions`

```ts
/**
   * Options for `deleteToken()`
   */
  export interface DeleteTokenOptions {
    /**
     * The app name of the FirebaseApp instance.
     *
     * @platform android Android
     */
    appName?: string;

    /**
     * The senderID for a particular Firebase project.
     *
     * @platform ios iOS
     */
    senderId?: string;

    /**
     * The VAPID key used to authenticate the push subscribers
     *  to receive push messages only from sending servers
     * that hold the corresponding private key.
     *
     * @platform web
     */
    vapidKey?: string;

    /**
     * The VAPID key used to authenticate the push subscribers
     * to receive push messages only from sending servers
     * that hold the corresponding private key.
     *
     * @platform web
     */

    // TODO - A typed service worker. Could use this https://www.npmjs.com/package/@types/serviceworker ??
    serviceWorkerRegistration;
  }
```

### Firebase Web JS SDK v9
```js
export declare function onMessage(messaging: Messaging, nextOrObserver: NextFn<MessagePayload> | Observer<MessagePayload>): Unsubscribe;
```

### RNFB modular v9

```js
export function onMessage(messaging, nextOrObserver) {
  return messaging.onMessage(nextOrObserver);
}
```

### Existing API

```js
 onMessage(listener) {
    if (!isFunction(listener)) {
      throw new Error("firebase.messaging().onMessage(*) 'listener' expected a function.");
    }

    const subscription = this.emitter.addListener('messaging_message_received', listener);
    return () => subscription.remove();
  }
```

### Compare API
The same.

### Firebase Web JS SDK v9
```js
export declare function isWindowSupported(): Promise<boolean>;
```

### RNFB modular v9

```js
export function isSupported(messaging) {
  // No-op as web platform isn't implemented
  return messaging.isSupported();
}
```

### New API

```js
  async isSupported() {
    // No-op as web platform isn't implemented
    return true;
  }
```

### Compare API
1. We have to pass in `messaging` but it isn't necessary for the Firebase Web JS SDK v9. [[Action item]].

### Firebase Web JS SDK v9
```js
export interface MessagePayload {
  /**
   * {@inheritdoc NotificationPayload}
   */
  notification?: NotificationPayload;

  /**
   * Arbitrary key/value payload.
   */
  data?: { [key: string]: string };

  /**
   * {@inheritdoc FcmOptions}
   */
  fcmOptions?: FcmOptions;

  /**
   * The sender of this message.
   */
  from: string;

  /**
   * The collapse key of the message. See
   * {@link https://firebase.google.com/docs/cloud-messaging/concept-options#collapsible_and_non-collapsible_messages | Non-collapsible and collapsible messages}
   */
  collapseKey: string;

  /**
   * The message id of a message.
   */
  messageId: string;
}
```

### New API

```js
/**
   * The `RemoteMessage` interface describes an outgoing & incoming message from the remote FCM server.
   */
  export interface RemoteMessage {
    /**
     * The collapse key a message was sent with. Used to override existing messages with the same
     * key.
     */
    collapseKey?: string;

    /**
     * A unique ID assigned to every message.
     *
     * If not provided, a random unique ID is generated.
     */
    messageId?: string;

    /**
     * The message type of the message.
     */
    messageType?: string;
    /**
     * The topic name or message identifier.
     */
    from?: string;
    /**
     * The address for the message.
     */
    to?: string;

    /**
     * The time to live for the message in seconds.
     *
     * Defaults to 3600.
     */
    ttl?: number;

    /**
     * The time the message was sent, in milliseconds since the start of unix epoch
     */
    sentTime?: number;

    /**
     * Any additional data sent with the message.
     */
    data?: { [key: string]: string };

    /**
     * Additional Notification data sent with the message
     */
    notification?: Notification;

    /**
     * Whether the iOS APNs message was configured as a background update notification.
     *
     * @platform ios iOS
     */
    contentAvailable?: boolean;

    /**
     * Whether the iOS APNs `mutable-content` property on the message was set
     * allowing the app to modify the notification via app extensions.
     *
     * @platform ios iOS
     */
    mutableContent?: boolean;

    /**
     * The iOS category this notification is assigned to.
     *
     * @platform ios iOS
     */
    category?: string;

    /**
     * An iOS app specific identifier used for notification grouping.
     */
    threadId?: string;
	/**
     * Options for features provided by the FCM SDK for Web.
     */
	fcmOptions: FcmOptions;
  }

export interface FcmOptions {
  /**
   * The link to open when the user clicks on the notification.
   */
  link?: string;

  /**
   * The label associated with the message's analytics data.
   */
  analyticsLabel?: string;
}
```

### Compare API
1. I've added the missing property `fcmOptions` in the `RemoteMessage` in the "index.d.ts" file.

### Firebase Web JS SDK v9
```js
export interface Notification {
    /**
     * The notification title.
     */
    title?: string;

    /**
     * The native localization key for the notification title.
     */
    titleLocKey?: string;

    /**
     * Any arguments that should be formatted into the resource specified by titleLocKey.
     */
    titleLocArgs?: string[];

    /**
     * The notification body content.
     */
    body?: string;

    /**
     * Web only. The URL to use for the notification's icon. If you don't send this key in the request,
     * FCM displays the launcher icon specified in your app manifest.
     */
    icon?: string;

    /**
     * Web only. The URL of an image that is downloaded on the device and displayed in the notification.
     */
    image?: string;

    /**
     * Web only. The notification's title.
     */
    title?: string;

    /**
     * The native localization key for the notification body content.
     */
    bodyLocKey?: string;

    /**
     * Any arguments that should be formatted into the resource specified by bodyLocKey.
     */
    bodyLocArgs?: string[];

    ios?: {
      /**
       * The notification's subtitle.
       */
      subtitle?: string;

      /**
       * The native localization key for the notification's subtitle.
       */
      subtitleLocKey?: string;

      /**
       * Any arguments that should be formatted into the resource specified by subtitleLocKey.
       */
      subtitleLocArgs?: string[];

      /**
       * The value of the badge on the home screen app icon.
       * If not specified, the badge is not changed.
       * If set to 0, the badge has been removed.
       */
      badge?: string;

      /**
       * The sound played when the notification was delivered on the device (if permissions permit).
       */
      sound?: string | NotificationIOSCriticalSound;
    };

    /**
     * Additional Android specific properties set on the notification.
     */
    android?: {
      /**
       * The sound played when the notification was delivered on the device (channel settings permitted).
       *
       * Set as "default" if the default device notification sound was used.
       */
      sound?: string;

      /**
       * The channel ID set on the notification. If not set, the notification uses the default
       * "Miscellaneous" channel set by FCM.
       */
      channelId?: string;

      /**
       * The custom color used to tint the notification content.
       */
      color?: string;

      /**
       * The custom small icon used to display on the notification. If not set, uses the default
       * application icon defined in the AndroidManifest file.
       */
      smallIcon?: string;

      /**
       * The custom image was provided and displayed in the notification body.
       */
      imageUrl?: string;

      /**
       * Deep link URL provided to the notification.
       */
      link?: string;

      /**
       * The current unread notification count for this application, managed by the device.
       */
      count?: number;

      /**
       * Name of the click action set on the notification.
       */
      clickAction?: string;

      /**
       * The notification priority.
       *
       * Note; on devices which have channel support (Android 8.0 (API level 26) +),
       * this value will be ignored. Instead, the channel "importance" level is used.
       */
      priority?:
        | NotificationAndroidPriority.PRIORITY_MIN
        | NotificationAndroidPriority.PRIORITY_LOW
        | NotificationAndroidPriority.PRIORITY_DEFAULT
        | NotificationAndroidPriority.PRIORITY_HIGH
        | NotificationAndroidPriority.PRIORITY_MAX;

      /**
       * Ticker text set on the notification.
       *
       * Ticker text is used for devices with accessibility enabled (e.g. to read out the message).
       */
      ticker?: string;

      /**
       * The visibility of a notification. This value determines how the notification is shown on the users
       * devices (e.g. on the lock-screen).
       */
      visibility?:
        | NotificationAndroidVisibility.VISIBILITY_SECRET
        | NotificationAndroidVisibility.VISIBILITY_PRIVATE
        | NotificationAndroidVisibility.VISIBILITY_PUBLIC;
    };
  }
```

### New API

```js
export interface Notification {
    /**
     * The notification title.
     */
    title?: string;

    /**
     * The native localization key for the notification title.
     */
    titleLocKey?: string;

    /**
     * Any arguments that should be formatted into the resource specified by titleLocKey.
     */
    titleLocArgs?: string[];

    /**
     * The notification body content.
     */
    body?: string;

    /**
     * Web only. The URL to use for the notification's icon. If you don't send this key in the request,
     * FCM displays the launcher icon specified in your app manifest.
     */
    icon?: string;

    /**
     * Web only. The URL of an image that is downloaded on the device and displayed in the notification.
     */
    image?: string;

    /**
     * Web only. The notification's title.
     */
    title?: string;

    /**
     * The native localization key for the notification body content.
     */
    bodyLocKey?: string;

    /**
     * Any arguments that should be formatted into the resource specified by bodyLocKey.
     */
    bodyLocArgs?: string[];

    ios?: {
      /**
       * The notification's subtitle.
       */
      subtitle?: string;

      /**
       * The native localization key for the notification's subtitle.
       */
      subtitleLocKey?: string;

      /**
       * Any arguments that should be formatted into the resource specified by subtitleLocKey.
       */
      subtitleLocArgs?: string[];

      /**
       * The value of the badge on the home screen app icon.
       * If not specified, the badge is not changed.
       * If set to 0, the badge has been removed.
       */
      badge?: string;

      /**
       * The sound played when the notification was delivered on the device (if permissions permit).
       */
      sound?: string | NotificationIOSCriticalSound;
    };

    /**
     * Additional Android specific properties set on the notification.
     */
    android?: {
      /**
       * The sound played when the notification was delivered on the device (channel settings permitted).
       *
       * Set as "default" if the default device notification sound was used.
       */
      sound?: string;

      /**
       * The channel ID set on the notification. If not set, the notification uses the default
       * "Miscellaneous" channel set by FCM.
       */
      channelId?: string;

      /**
       * The custom color used to tint the notification content.
       */
      color?: string;

      /**
       * The custom small icon used to display on the notification. If not set, uses the default
       * application icon defined in the AndroidManifest file.
       */
      smallIcon?: string;

      /**
       * The custom image was provided and displayed in the notification body.
       */
      imageUrl?: string;

      /**
       * Deep link URL provided to the notification.
       */
      link?: string;

      /**
       * The current unread notification count for this application, managed by the device.
       */
      count?: number;

      /**
       * Name of the click action set on the notification.
       */
      clickAction?: string;

      /**
       * The notification priority.
       *
       * Note; on devices which have channel support (Android 8.0 (API level 26) +),
       * this value will be ignored. Instead, the channel "importance" level is used.
       */
      priority?:
        | NotificationAndroidPriority.PRIORITY_MIN
        | NotificationAndroidPriority.PRIORITY_LOW
        | NotificationAndroidPriority.PRIORITY_DEFAULT
        | NotificationAndroidPriority.PRIORITY_HIGH
        | NotificationAndroidPriority.PRIORITY_MAX;

      /**
       * Ticker text set on the notification.
       *
       * Ticker text is used for devices with accessibility enabled (e.g. to read out the message).
       */
      ticker?: string;

      /**
       * The visibility of a notification. This value determines how the notification is shown on the users
       * devices (e.g. on the lock-screen).
       */
      visibility?:
        | NotificationAndroidVisibility.VISIBILITY_SECRET
        | NotificationAndroidVisibility.VISIBILITY_PRIVATE
        | NotificationAndroidVisibility.VISIBILITY_PUBLIC;
    };
  }
```

### Compare API
1. I've added the missing properties to `Notification` found on the Firebase Web JS SDK v9. They are `icon`, `image` & `title`. Maybe we should create our own `web` object like `android` & `iOS`? [[Action item]]

### Checklist

  - [ ] Yes
  - [ ] `Android`
  - [ ] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [X] No

### Test Plan

---

### DIFF
--- a/packages/messaging/lib/index.d.ts
+++ b/packages/messaging/lib/index.d.ts
@@ -139,14 +139,35 @@ export namespace FirebaseMessagingTypes {
 
     /**
      * An iOS app specific identifier used for notification grouping.
-     */
     threadId?: string;
+    */
+    threadId?: string;
+
+    /**
+     * Options for features provided by the FCM SDK for Web.
+     */
+    fcmOptions: FcmOptions;
   }
 
   /**
-   * Options for `getToken()`, `deleteToken()`
+   * Options for features provided by the FCM SDK for Web.
    */
-  export interface TokenOptions {
+  export interface FcmOptions {
+    /**
+     * The link to open when the user clicks on the notification.
+     */
+    link?: string;
+
+    /**
+     * The label associated with the message's analytics data.
+     */
+    analyticsLabel?: string;
+  }
+
+  /**
+   * Options for `getToken()`
+   */
+  export interface GetTokenOptions {
     /**
      * The app name of the FirebaseApp instance.
      *
@@ -160,6 +181,45 @@ export namespace FirebaseMessagingTypes {
      * @platform ios iOS
      */
     senderId?: string;
+  }
+
+  /**
+   * Options for `deleteToken()`
+   */
+  export interface DeleteTokenOptions {
+    /**
+     * The app name of the FirebaseApp instance.
+     *
+     * @platform android Android
+     */
+    appName?: string;
+
+    /**
+     * The senderID for a particular Firebase project.
+     *
+     * @platform ios iOS
+     */
+    senderId?: string;
+
+    /**
+     * The VAPID key used to authenticate the push subscribers
+     *  to receive push messages only from sending servers
+     * that hold the corresponding private key.
+     *
+     * @platform web
+     */
+    vapidKey?: string;
+
+    /**
+     * The service worker registration for receiving push messaging.
+     * If the registration is not provided explicitly, you need to
+     * have a firebase-messaging-sw.js at your root location.
+     *
+     * @platform web
+     */
+
+    // TODO - A typed service worker. Could use this https://www.npmjs.com/package/@types/serviceworker ??
+    serviceWorkerRegistration;
   }
 
   export interface Notification {
@@ -182,6 +242,22 @@ export namespace FirebaseMessagingTypes {
      * The notification body content.
      */
     body?: string;
+
+    /**
+     * Web only. The URL to use for the notification's icon. If you don't send this key in the request,
+     * FCM displays the launcher icon specified in your app manifest.
+     */
+    icon?: string;
+
+    /**
+     * Web only. The URL of an image that is downloaded on the device and displayed in the notification.
+     */
+    image?: string;
+
+    /**
+     * Web only. The notification's title.
+     */
+    title?: string;
 
     /**
      * The native localization key for the notification body content.
@@ -630,7 +706,7 @@ export namespace FirebaseMessagingTypes {
      *
      * @param options Options to override senderId (iOS) and projectId (Android).
      */
-    getToken(options?: TokenOptions): Promise<string>;
+    getToken(options?: GetTokenOptions): Promise<string>;
 
     /**
      * Returns whether the root view is headless or not
@@ -653,7 +729,7 @@ export namespace FirebaseMessagingTypes {
      *
      * @param options Options to override senderId (iOS) and projectId (Android).
      */
-    deleteToken(options?: TokenOptions): Promise<void>;
+    deleteToken(options?: DeleteTokenOptions): Promise<void>;
 
     /**
      * When any FCM payload is received, the listener callback is called with a `RemoteMessage`.
@@ -1012,6 +1088,12 @@ export namespace FirebaseMessagingTypes {
      * @param enabled A boolean value to enable or disable exporting of message delivery metrics to BigQuery.
      */
     setDeliveryMetricsExportToBigQuery(enabled: boolean): Promise<void>;
+    /**
+     * Checks if all required APIs exist in the browser.
+     *
+     * @web
+     */
+    isSupported(): Promise<boolean>;
   }
 }
 
--- a/packages/messaging/lib/index.d.ts
+++ b/packages/messaging/lib/index.d.ts
@@ -111,9 +111,9 @@ export namespace FirebaseMessagingTypes {
     data?: { [key: string]: string };
 
     /**
-     * Additional Notification data sent with the message
+     * Additional NotificationPayload data sent with the message
      */
-    notification?: Notification;
+    notification?: NotificationPayload;
 
     /**
      * Whether the iOS APNs message was configured as a background update notification.
@@ -165,9 +165,9 @@ export namespace FirebaseMessagingTypes {
   }
 
   /**
-   * Options for `getToken()`
+   * Options for `getToken()` and `deleteToken()`
    */
-  export interface GetTokenOptions {
+  export interface NativeTokenOptions {
     /**
      * The app name of the FirebaseApp instance.
      *
@@ -184,23 +184,9 @@ export namespace FirebaseMessagingTypes {
   }
 
   /**
-   * Options for `deleteToken()`
+   * Options for `getToken()`
    */
-  export interface DeleteTokenOptions {
-    /**
-     * The app name of the FirebaseApp instance.
-     *
-     * @platform android Android
-     */
-    appName?: string;
-
-    /**
-     * The senderID for a particular Firebase project.
-     *
-     * @platform ios iOS
-     */
-    senderId?: string;
-
+  export interface GetTokenOptions {
     /**
      * The VAPID key used to authenticate the push subscribers
      *  to receive push messages only from sending servers
@@ -217,10 +203,14 @@ export namespace FirebaseMessagingTypes {
      *
      * @platform web
      */
-
-    // TODO - A typed service worker. Could use this https://www.npmjs.com/package/@types/serviceworker ??
-    serviceWorkerRegistration;
+    serviceWorkerRegistration?: ServiceWorker;
   }
+
+  /**
+   * NotificationPayload is an alias for Notification. This is to keep it the same as
+   * Firebase Web JS SDK v9 and to make it backwards compatible.
+   */
+  type NotificationPayload = Notification;
 
   export interface Notification {
     /**
@@ -704,9 +694,9 @@ export namespace FirebaseMessagingTypes {
      *   });
      * ```
      *
-     * @param options Options to override senderId (iOS) and projectId (Android).
+     * @param options Options can be either `GetTokenOptions` or `NativeTokenOptions`
      */
-    getToken(options?: GetTokenOptions): Promise<string>;
+    getToken(options?: GetTokenOptions | NativeTokenOptions): Promise<string>;
 
     /**
      * Returns whether the root view is headless or not
@@ -727,9 +717,9 @@ export namespace FirebaseMessagingTypes {
      * await firebase.messaging().deleteToken();
      * ```
      *
-     * @param options Options to override senderId (iOS) and projectId (Android).
+     * @param options Options to override senderId (iOS) and appName (android)
      */
-    deleteToken(options?: DeleteTokenOptions): Promise<void>;
+    deleteToken(options?: NativeTokenOptions): Promise<void>;
 
     /**
      * When any FCM payload is received, the listener callback is called with a `RemoteMessage`.
@@ -1094,6 +1084,15 @@ export namespace FirebaseMessagingTypes {
      * @web
      */
     isSupported(): Promise<boolean>;
+
+    /**
+     * Enables or disables Firebase Cloud Messaging message delivery metrics export to BigQuery. By
+     * default, message delivery metrics are not exported to BigQuery. Use this method to enable or
+     * disable the export at runtime.
+     *
+     * @web
+     */
+    experimentalSetDeliveryMetricsExportedToBigQueryEnabled(): void;
   }
 }
 
--- a/packages/messaging/lib/index.d.ts
+++ b/packages/messaging/lib/index.d.ts
@@ -203,7 +203,7 @@ export namespace FirebaseMessagingTypes {
      *
      * @platform web
      */
-    serviceWorkerRegistration?: ServiceWorker;
+    serviceWorkerRegistration?: ServiceWorkerRegistration;
   }
 
   /**
@@ -693,10 +693,9 @@ export namespace FirebaseMessagingTypes {
      *     fcmTokens: firebase.firestore.FieldValues.arrayUnion(fcmToken),
      *   });
      * ```
-     *
-     * @param options Options can be either `GetTokenOptions` or `NativeTokenOptions`
+     * @param options Options composite type with all members of `GetTokenOptions` and `NativeTokenOptions`
      */
-    getToken(options?: GetTokenOptions | NativeTokenOptions): Promise<string>;
+    getToken(options?: GetTokenOptions & NativeTokenOptions): Promise<string>;
 
     /**
      * Returns whether the root view is headless or not
--- a/packages/messaging/lib/index.js
+++ b/packages/messaging/lib/index.js
@@ -482,7 +482,8 @@ class FirebaseMessagingModule extends FirebaseModule {
 
   async isSupported() {
     if (Platform.isAndroid) {
-      return this.native.isSupported();
+      playServicesAvailability = firebase.utils().playServicesAvailability;
+      return playServicesAvailability.isAvailable;
     }
     // Always return "true" for iOS. Web will be implemented when it is supported
     return true;
--- a/packages/messaging/lib/index.js
+++ b/packages/messaging/lib/index.js
@@ -30,7 +30,7 @@ import {
   FirebaseModule,
   getFirebaseRoot,
 } from '@react-native-firebase/app/lib/internal';
-import { AppRegistry } from 'react-native';
+import { AppRegistry, Platform } from 'react-native';
 import remoteMessageOptions from './remoteMessageOptions';
 import version from './version';
 
@@ -62,6 +62,7 @@ export {
   unsubscribeFromTopic,
   setDeliveryMetricsExportToBigQuery,
   isDeliveryMetricsExportToBigQueryEnabled,
+  isSupported,
 } from '../modular/index';
 
 const statics = {
@@ -480,7 +481,10 @@ class FirebaseMessagingModule extends FirebaseModule {
   }
 
   async isSupported() {
-    // No-op as web platform isn't implemented
+    if (Platform.isAndroid) {
+      return this.native.isSupported();
+    }
+    // Always return "true" for iOS. Web will be implemented when it is supported
     return true;
   }
 
--- a/packages/messaging/lib/index.js
+++ b/packages/messaging/lib/index.js
@@ -196,7 +196,7 @@ class FirebaseMessagingModule extends FirebaseModule {
 
   getToken({ appName, senderId } = {}) {
     if (!isUndefined(appName) && !isString(appName)) {
-      throw new Error("firebase.messaging().getToken(*) 'projectId' expected a string.");
+      throw new Error("firebase.messaging().getToken(*) 'appName' expected a string.");
     }
 
     if (!isUndefined(senderId) && !isString(senderId)) {
@@ -211,7 +211,7 @@ class FirebaseMessagingModule extends FirebaseModule {
 
   deleteToken({ appName, senderId } = {}) {
     if (!isUndefined(appName) && !isString(appName)) {
-      throw new Error("firebase.messaging().deleteToken(*) 'projectId' expected a string.");
+      throw new Error("firebase.messaging().deleteToken(*) 'appName' expected a string.");
     }
 
     if (!isUndefined(senderId) && !isString(senderId)) {
@@ -477,6 +477,11 @@ class FirebaseMessagingModule extends FirebaseModule {
 
     this._isDeliveryMetricsExportToBigQueryEnabled = enabled;
     return this.native.setDeliveryMetricsExportToBigQuery(enabled);
+  }
+
+  async isSupported() {
+    // No-op as web platform isn't implemented
+    return true;
   }
 }
 
--- a/packages/messaging/lib/index.js
+++ b/packages/messaging/lib/index.js
@@ -61,7 +61,8 @@ export {
   subscribeToTopic,
   unsubscribeFromTopic,
   setDeliveryMetricsExportToBigQuery,
-} from './modular/index';
+  isDeliveryMetricsExportToBigQueryEnabled,
+} from '../modular/index';
 
 const statics = {
   AuthorizationStatus: {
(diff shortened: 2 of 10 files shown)


mikehardy:
I'll be a better collaborator on these after Jan 3rd I promise :-), I tuned down a lot of my other work commitments at the start of December so I'll have proper time to dedicate to modular effort

mikehardy:
I want to spend more time looking at the mappings but result of my preliminary check on the TokenOptions object passed to getToken / deleteToken is that we may extend the firebase-js-sdk API here without much consequence (except for typescript issues for some people I suppose...) as firebase-js-sdk will just discard the parameter, whereas we strictly need it on native to support multi-project use cases (re-learned after reading the associated issue + PR where TokenOptions was added last summer...)

packages/messaging/lib/index.d.ts
```diff
@@ -630,7 +706,7 @@ export namespace FirebaseMessagingTypes {
      *
      * @param options Options to override senderId (iOS) and projectId (Android).
      */
-    getToken(options?: TokenOptions): Promise<string>;
+    getToken(options?: GetTokenOptions): Promise<string>;
```
mikehardy, line 709: I think a union type (assuming there was a new `NativeTokenOptions` type with senderId/appName members) could work like this ?

```suggestion
    getToken(options?: GetTokenOptions | NativeTokenOptions): Promise<string>;
```

packages/messaging/lib/index.d.ts
```diff
@@ -653,7 +729,7 @@ export namespace FirebaseMessagingTypes {
      *
      * @param options Options to override senderId (iOS) and projectId (Android).
      */
-    deleteToken(options?: TokenOptions): Promise<void>;
+    deleteToken(options?: DeleteTokenOptions): Promise<void>;
```
mikehardy, line 732: again assuming `NativeTokenOptions` existed, to give a home to the native-only appName/senderId stuff

```suggestion
    deleteToken(options?: NativeTokenOptions): Promise<void>;
```

packages/messaging/lib/index.d.ts
```diff
@@ -1012,6 +1088,12 @@ export namespace FirebaseMessagingTypes {
      * @param enabled A boolean value to enable or disable exporting of message delivery metrics to BigQuery.
      */
     setDeliveryMetricsExportToBigQuery(enabled: boolean): Promise<void>;
```
mikehardy, line 1090: I think we also need to export this same feature with the current "experimental" web method name?

https://github.com/firebase/firebase-js-sdk/blob/49ee786f2b022e65aef45693e1a8b546d889ec10/packages/messaging/src/api.ts#L184

packages/messaging/lib/index.d.ts
```diff
@@ -1012,6 +1088,12 @@ export namespace FirebaseMessagingTypes {
...
+    /**
+     * Checks if all required APIs exist in the browser.
+     *
+     * @web
+     */
+    isSupported(): Promise<boolean>;
```
mikehardy, line 1096: This appears a little complicated. On native, it is supported on iOS in general, and on Android *if* our Utils method says google play services exist, so there is a way to implement it.

firebase-js-sdk definitely implements it - but dynamically (!) - by defining it after the getMessagingInWindow / getMessagingInSw definitions like so --> https://github.com/firebase/firebase-js-sdk/blob/49ee786f2b022e65aef45693e1a8b546d889ec10/packages/messaging/src/api.ts#L53

it appears we need to have the symbol available to be firebase-js-sdk compatible

packages/messaging/lib/index.js
```diff
@@ -448,6 +478,11 @@ class FirebaseMessagingModule extends FirebaseModule {
     this._isDeliveryMetricsExportToBigQueryEnabled = enabled;
     return this.native.setDeliveryMetricsExportToBigQuery(enabled);
   }
+
+  async isSupported() {
+    // No-op as web platform isn't implemented
+    return true;
```
mikehardy, line 484: I think technically we can check on android if play services exists? FCM doesn't work on non-play-services devices as far as I know, and we have that native implementation (with JS wrapper) in app utils

mikehardy (commented):
getting really close I think! what do you think of these nitty-gritty typing comments?

packages/messaging/lib/index.d.ts
```diff
@@ -162,6 +183,45 @@ export namespace FirebaseMessagingTypes {
...
+    /**
+     * The service worker registration for receiving push messaging.
+     * If the registration is not provided explicitly, you need to
+     * have a firebase-messaging-sw.js at your root location.
+     *
+     * @platform web
+     */
+
+    // TODO - A typed service worker. Could use this https://www.npmjs.com/package/@types/serviceworker ??
+    serviceWorkerRegistration;
```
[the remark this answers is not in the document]
russellwheatley: woops, updated.

packages/messaging/lib/index.d.ts
russellwheatley: What API corresponds to this for iOS? For FlutterFire, we just return `true`: https://github.com/firebase/flutterfire/blob/master/packages/firebase_messaging/firebase_messaging_platform_interface/lib/src/method_channel/method_channel_messaging.dart#L148

As for the web, not sure what more can be done at this moment in time until web is an available platform?

packages/messaging/lib/index.js
russellwheatley: I'm not sure on this one. I think I'll see what the consensus is on checking play service availability for android before implementing 😄

packages/messaging/lib/index.js
```diff
@@ -483,6 +483,10 @@ class FirebaseMessagingModule extends FirebaseModule {
     // No-op as web platform isn't implemented
     return true;
   }
+
+  experimentalSetDeliveryMetricsExportedToBigQueryEnabled(enable) {
+    // No-op as web platform isn't implemented
```
mikehardy, line 488: 🤔  the idea of the modular API in react-native-firebase is not to no-op things that web doesn't implement or that we haven't implemented yet, rather it is to expose all available functionality in a way that is web-compatible.

In this case, we actually have an implementation available that works so I am not sure why the no-op here?

Should be similar to existing typing of 
https://github.com/invertase/react-native-firebase/blob/main/packages/messaging/lib/index.d.ts#L1014

with implementations callable like
https://github.com/invertase/react-native-firebase/blob/97a5523bac4570ba7b47fbfdbb22348f38246d11/packages/messaging/lib/index.js#L135

and 

https://github.com/invertase/react-native-firebase/blob/97a5523bac4570ba7b47fbfdbb22348f38246d11/packages/messaging/lib/index.js#L441

(that is, just delegating to the existing implementations which actually work, but using the firebase-js-sdk v9's `experimentalNNN` API symbol ? )

packages/messaging/lib/index.d.ts
```diff
@@ -162,6 +183,35 @@ export namespace FirebaseMessagingTypes {
...
+    vapidKey?: string;
+
+    /**
+     * The service worker registration for receiving push messaging.
+     * If the registration is not provided explicitly, you need to
+     * have a firebase-messaging-sw.js at your root location.
+     *
+     * @platform web
+     */
+    serviceWorkerRegistration?: ServiceWorker;
```
mikehardy, line 206: I think `ServiceWorkerRegistration` is the upstream type here, not `ServiceWorker` 🤔  ?

https://github.com/firebase/firebase-js-sdk/blob/49ee786f2b022e65aef45693e1a8b546d889ec10/packages/messaging/src/interfaces/public-types.ts#L134

```suggestion
    serviceWorkerRegistration?: ServiceWorkerRegistration;
```

tsconfig.json
```diff
@@ -18,7 +18,7 @@
     "experimentalDecorators": true,
     "emitDecoratorMetadata": true,
     "esModuleInterop": true,
-    "lib": ["es2015", "es2016", "esnext"]
+    "lib": ["es2015", "es2016", "esnext", "DOM"]
```
mikehardy, line 21: I have no idea if this is important or not, it probably is not, but they have it as lower-case `dom` ?
https://github.com/firebase/firebase-js-sdk/blob/49ee786f2b022e65aef45693e1a8b546d889ec10/packages/messaging/tsconfig.json#L7

packages/messaging/lib/index.d.ts
```diff
@@ -628,9 +694,9 @@ export namespace FirebaseMessagingTypes {
      *   });
      * ```
      *
-     * @param options Options to override senderId (iOS) and projectId (Android).
+     * @param options Options can be either `GetTokenOptions` or `NativeTokenOptions`
```
mikehardy, line 697: When I made the original suggestion I had to actually look up the typescript typing syntax and then I still steered you wrong because I misunderstood the typescript type syntax. Many apologies.

This did not seem correct to me though so I've just re-re-read the typescript docs and what we want here is an intersection type (symbol `&`) so we get a composite object with all properties, not just the intersecting properties (which would be empty in this case!)

```suggestion
     * @param options Options composite type with all members of `GetTokenOptions` and `NativeTokenOptions`
```

packages/messaging/lib/index.d.ts
```diff
@@ -628,9 +694,9 @@ export namespace FirebaseMessagingTypes {
...
      */
-    getToken(options?: TokenOptions): Promise<string>;
+    getToken(options?: GetTokenOptions | NativeTokenOptions): Promise<string>;
```
mikehardy, line 699: And this is the correct syntax to get the composite / all-members type, if I actually understand it now 

https://www.typescriptlang.org/docs/handbook/2/objects.html#intersection-types

```suggestion
    getToken(options?: GetTokenOptions & NativeTokenOptions): Promise<string>;
```

packages/messaging/lib/index.js
```diff
@@ -448,6 +479,18 @@ class FirebaseMessagingModule extends FirebaseModule {
     this._isDeliveryMetricsExportToBigQueryEnabled = enabled;
     return this.native.setDeliveryMetricsExportToBigQuery(enabled);
   }
+
+  async isSupported() {
+    if (Platform.isAndroid) {
+      return this.native.isSupported();
```
mikehardy, line 485: 🤔  we are guaranteed to have the app module loaded because it's a strict requirement for react-native-firebase, and it always loads the native module named `RNFBUtilsModule` which exports a constant here we could reference similar to how the Utils statics off app do it, meaning we could do away with the native implementation here and delegate to the existing on already here https://github.com/invertase/react-native-firebase/blob/94536b024b4db404ba7c6bb7f3735291a4a4642e/packages/app/android/src/reactnative/java/io/invertase/firebase/utils/ReactNativeFirebaseUtilsModule.java#L145

The status map has an `isAvailable` boolean on it, so I think this API could delegate like so?

https://github.com/invertase/react-native-firebase/blob/94536b024b4db404ba7c6bb7f3735291a4a4642e/packages/app/e2e/utils.e2e.js#L35-L37

```suggestion
      playServicesAvailability = firebase.utils().playServicesAvailability;
      return playServicesAvailability.isAvailable;
```

I *think* that will work anyway? There may be some sort of javascript import junk to handle, or perhaps a better way to gain access to the symbol since we should have internal access to it (that is: a different way to import and use it vs `firebase.utils().NNN`) ?

Sorry for beating this one to death - what I'm hoping for is that in the end it really is "that easy" - the native implementation is already done in app/utils so I was thinking it was just an easy reference to it

mikehardy (approved):
This looks great, a big step forward for the v9 effort

Issue #6853, opened 2023-01-23
[🐛] Build fails on macOS M1 with sample React Native 0.71.1 project


iskeld:
## Issue

Hello, I created a sample React Native project using the [official tutorial](https://reactnative.dev/docs/environment-setup) (macOS + iOS - NON Expo version) and added React-Native-Firebase using the [official tutorial](https://rnfirebase.io).

_(I didn't add any code there besides the parts described in the Firebase tutorial - it's a plain sample React project and a plain Firebase setup)_

Unfortunately trying to run the app with `npx react-native run-ios` fails with the following errors:

```
Undefined symbols for architecture arm64:
  "facebook::jsi::Value::Value(facebook::jsi::Runtime&, facebook::jsi::Value const&)", referenced from:
      facebook::jsi::detail::toValue(facebook::jsi::Runtime&, facebook::jsi::Value const&) in Instance.o
  "facebook::jsi::Value::~Value()", referenced from:
      facebook::react::handleJSError(facebook::jsi::Runtime&, facebook::jsi::JSError const&, bool) in Instance.o
      facebook::jsi::Value facebook::jsi::Function::call<facebook::jsi::Value const&>(facebook::jsi::Runtime&, facebook::jsi::Value const&) const in Instance.o
      facebook::jsi::Function::call(facebook::jsi::Runtime&, facebook::jsi::Value const*, unsigned long) const in Instance.o
  "facebook::jsi::JSError::JSError(facebook::jsi::Runtime&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >)", referenced from:
      facebook::react::handleJSError(facebook::jsi::Runtime&, facebook::jsi::JSError const&, bool) in Instance.o
  "facebook::jsi::JSError::~JSError()", referenced from:
      facebook::react::handleJSError(facebook::jsi::Runtime&, facebook::jsi::JSError const&, bool) in Instance.o
  "facebook::jsi::Object::getPropertyAsFunction(facebook::jsi::Runtime&, char const*) const", referenced from:
      facebook::react::handleJSError(facebook::jsi::Runtime&, facebook::jsi::JSError const&, bool) in Instance.o
  "facebook::jsi::Value::asObject(facebook::jsi::Runtime&) const &", referenced from:
[...]

PR #6856, opened 2023-01-23
docs(app): clarify iOS setup instructions, especially around use_frameworks

gosunman:
### Description

- I fixed the frequently asked issue (#6853) after looking `https://github.com/mikehardy/rnfbdemo/blob/main/make-demo.sh` by @mikehardy 
- according to `make-demo.sh`, I updated the documentation (iOS setup part in `getting started`)

### Related issues

#6853

### Checklist

  - [ ] Yes
  - [x] No (link should be changed with `../blob/main/CONTRIBUTING.md` instead of `../CONTRIBUTING.md`)
  - [ ] `Android`
  - [x] `iOS`
- My change includes tests;
  - [x] N/A
- [x] (N/A) I have updated TypeScript types that are affected by my change.
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

---

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -104,7 +104,7 @@ To allow Firebase on iOS to use the credentials, the Firebase iOS SDK must be co
 
 To do this, open your `/ios/{projectName}/AppDelegate.mm` file (or `AppDelegate.m` if on older react-native), and add the following:
 
-At the top of the file, import the Firebase SDK:
+At the top of the file, import the Firebase SDK right after `'#import "AppDelegate.h"'`:
 
 ```
 #import <Firebase.h>
@@ -125,15 +125,16 @@ Within your existing `didFinishLaunchingWithOptions` method, add the following t
 
 Beginning with firebase-ios-sdk v9+ (react-native-firebase v15+) you must tell CocoaPods to use frameworks.
 
-Open the file `./ios/Podfile` and add this line inside your targets:
-
-```
-use_frameworks!
-```
-
-To use Static Frameworks on iOS, you also need to manually enable this for the project with the following global to the top of your `/ios/Podfile` file:
+Open the file `./ios/Podfile` and add this line inside your targets (right after the line calling the react native Podfile function to get the native modules config):
 
 ```ruby
+use_frameworks! :linkage => :static
+```
+
+To use Static Frameworks on iOS, you also need to manually enable this for the project with the following global to your `/ios/Podfile` file:
+
+```ruby
+# right after `use_frameworks! :linkage => :static`
 $RNFirebaseAsStaticFramework = true
 ```
 


docs/index.md
```diff
@@ -125,18 +125,64 @@ Within your existing `didFinishLaunchingWithOptions` method, add the following t
 
 Beginning with firebase-ios-sdk v9+ (react-native-firebase v15+) you must tell CocoaPods to use frameworks.
 
-Open the file `./ios/Podfile` and add this line inside your targets:
+Open the file `./ios/Podfile` and add this line inside your targets (right after `'config = use_native_modules!'`):
```
mikehardy, line 128: This is unstable advice unfortunately, as it changes from react-native version to react-native version. Perhaps something a little less specific code and a little more general 🤔 

```suggestion
Open the file `./ios/Podfile` and add this line inside your targets (right after the line calling the react native Podfile function to get the native modules config):
```

docs/index.md
```diff
@@ -115,7 +115,7 @@ Within your existing `didFinishLaunchingWithOptions` method, add the following t
 ```
 - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions {
   // Add me --- \/
-  [FIRApp configure];
+  if ([FIRApp defaultApp] == nil) { [FIRApp configure]; }
```
mikehardy, line 118: This "check for null / only configure if defaultApp is null" has been the source of some debate, I'm not sure it is strictly required and I should probably remove it from my demo script.

Can you test if it works just fine as currently documented, that is just a straight `[FIRApp configure];` call?

docs/index.md
```diff
@@ -104,7 +104,7 @@ To allow Firebase on iOS to use the credentials, the Firebase iOS SDK must be co
 
 To do this, open your `/ios/{projectName}/AppDelegate.mm` file (or `AppDelegate.m` if on older react-native), and add the following:
 
-At the top of the file, import the Firebase SDK:
+At the top of the file, import the Firebase SDK right after `'#import "AppDelegate.h"'`:
```
mikehardy, line 107: This will likely help people who mistakenly put it inside a compiler IFDEF conditional, that happens frequently. Good suggestion

docs/index.md
```diff
@@ -125,18 +125,64 @@ Within your existing `didFinishLaunchingWithOptions` method, add the following t
 
 Beginning with firebase-ios-sdk v9+ (react-native-firebase v15+) you must tell CocoaPods to use frameworks.
 
-Open the file `./ios/Podfile` and add this line inside your targets:
+Open the file `./ios/Podfile` and add this line inside your targets (right after `'config = use_native_modules!'`):
 
-```
-use_frameworks!
+```ruby
+use_frameworks! :linkage => :static
```
mikehardy, line 131: Really good suggestion, linkage does need to be static for hermes to work

docs/index.md
```diff
@@ -125,18 +125,64 @@ Within your existing `didFinishLaunchingWithOptions` method, add the following t
...
 
-To use Static Frameworks on iOS, you also need to manually enable this for the project with the following global to the top of your `/ios/Podfile` file:
+To use Static Frameworks on iOS, you also need to manually enable this for the project with the following global to your `/ios/Podfile` file:
 
 ```ruby
+# right after `use_frameworks! :linkage => :static`
 ```
 
+For build setting, add below codes inside your targets (right after `'__apply_Xcode_12_5_M1_post_install_workaround(installer)'`)
```
mikehardy, line 141: This part is dangerous - that workaround will likely go away soon. It is already empty inside the react-native scripts, and scheduled for removal I think

docs/index.md
```diff
@@ -125,18 +125,64 @@ Within your existing `didFinishLaunchingWithOptions` method, add the following t
...
+
+```ruby
+  installer.pods_project.targets.each do |target|
+    target.build_configurations.each do |config|
+      config.build_settings["GCC_WARN_INHIBIT_ALL_WARNINGS"] = "YES"
+    end
+  end
```
mikehardy, line 148: This is a matter of taste and must not be suggested as a general configuration for anyone
In general warnings are useful, they just aren't in my build demo context

```suggestion
```

docs/index.md
```diff
@@ -125,18 +125,64 @@ Within your existing `didFinishLaunchingWithOptions` method, add the following t
...
+  
+  installer.pods_project.targets.each do |target|
+    target.build_configurations.each do |config|
+      config.build_settings["CC"] = "clang"
+      config.build_settings["LD"] = "clang"
+      config.build_settings["CXX"] = "clang++"
+      config.build_settings["LDPLUSPLUS"] = "clang++"
+    end
+  end
```
mikehardy, line 157: This is only required for one specific method of doing the performance optimization I worked on here: https://reactnative.dev/docs/build-speed#xcode-specific-setup

It cannot be recommended as a general configuration for people because a) it is only a performance optimization and those are matters of opinion, while b) it opens a security risk because the compiler is now referenced via relative path, leading to potential for impersonating programs to be executed unexpectedly. I personally use it and I need it for the build demo script so I can run builds quickly but it should not be in react-native-firebase docs

```suggestion
```

docs/index.md
```diff
@@ -125,18 +125,64 @@ Within your existing `didFinishLaunchingWithOptions` method, add the following t
...
+  
+  installer.aggregate_targets.each do |aggregate_target|
+    aggregate_target.user_project.native_targets.each do |target|
+      target.build_configurations.each do |config|
+        config.build_settings['ONLY_ACTIVE_ARCH'] = 'YES'
+        config.build_settings['EXCLUDED_ARCHS'] = 'i386'
+      end
+    end
+    aggregate_target.user_project.save
+  end
```
mikehardy, line 167: I'm not sure this can be a general suggestion. I think this should be in the tips-and-tricks section instead https://rnfirebase.io/faqs-and-tips

There may be cases where a developer wants to build a non-active arch, and this eliminates that capability.

docs/index.md
```diff
@@ -125,18 +125,64 @@ Within your existing `didFinishLaunchingWithOptions` method, add the following t
...
+
+  installer.pods_project.targets.each do |target|
+    if (target.name.eql?('FBReactNativeSpec'))
+      target.build_phases.each do |build_phase|
+        if (build_phase.respond_to?(:name) && build_phase.name.eql?('[CP-User] Generate Specs'))
+          target.build_phases.move(build_phase, 0)
+        end
+      end
+    end
+  end
```
mikehardy, line 177: Have you tried without this? I don't think it is required anymore - I have not had a chance to check yet:

https://github.com/facebook/react-native/issues/31149#issuecomment-1282465756

docs/index.md
```diff
@@ -125,18 +125,64 @@ Within your existing `didFinishLaunchingWithOptions` method, add the following t
...
+  
+  installer.pods_project.targets.each do |target|
+    target.build_configurations.each do |config|
+      config.build_settings["ENABLE_BITCODE"] = "NO"
+    end
+  end
```
mikehardy, line 183: This is a general issue, not a react-native-firebase issue. I don't believe this should be in our documentation - it should be (and is?I think?) documented elsewhere.

It could be a good tips-and-tricks addition though

docs/index.md
```diff
@@ -232,7 +278,8 @@ Open your projects `/ios/Podfile` and add any of the globals shown below to the
 
 ```ruby
 # Override Firebase SDK Version
-$FirebaseSDKVersion = '10.3.0'
+# Between `use_react_native!()` paragraph and `target 'RNFBTests' do ... end`
+$FirebaseSDKVersion = '10.4.0'
```
mikehardy, line 282: As mentioned in another comment we are not compatible with 10.4.0 yet. This is an advanced configuration method anyway and basically no one should use it, definitely no one should use it long-term, as react-native-firebase should control the dependency version.

```suggestion
$FirebaseSDKVersion = '10.3.0'
```

docs/index.md
```diff
@@ -232,7 +278,8 @@ Open your projects `/ios/Podfile` and add any of the globals shown below to the
 
 ```ruby
 # Override Firebase SDK Version
-$FirebaseSDKVersion = '10.3.0'
+# Between `use_react_native!()` paragraph and `target 'RNFBTests' do ... end`
```
mikehardy, line 281: It does not matter where it goes actually. It's just defining a variable in Ruby syntax, at the top level.

```suggestion
```

mikehardy (changes requested):
Quite a few thoughts on this one - i think a couple of the suggested changes could be really helpful, but many of the suggestions here are not valid for react-native-firebase specific install docs I don't think

What do you think?

docs/index.md
gosunman: @mikehardy 
it works with `[FIRApp configure];`
so I will change this as currently documented!

docs/index.md
gosunman: yeah I checked it also works with 10.3.0 too
thanks 👍

docs/index.md
gosunman: It works normally without modifying this part. I will delete the related content because it seems unnecessary.

docs/index.md
gosunman: you are right.. haha.. I will also delete it!

docs/index.md
gosunman: After testing, it turned out that this was not an important part... I will delete it..

gosunman:
@mikehardy 

I reviewed the comments you gave me one by one.
As a result, there are only a few lines left. lol
As you've already left a lot of comments on "Issue",
There was no fault in the document. Haha

My PR is a modification to the extent that it once again clarifies where the code should go in,
but thank you for your detailed review.

I can understand if you accept this PR or reject it.

Anyway, thank you again for the detailed review. :)

jckw:
Would just like to add that I was recently tripped up by exactly this. I worked out that I needed to make those changes, but I would've saved a few hours had the docs read like this.

🎉

mikehardy (approved):
Apologies this went to sleep! I think it is small now but still quite a useful change, thank you!

PR #6868, opened 2023-01-26
feat(remote-config): Expose modular API that matches the Firebase web JS SDK v9 API

russellwheatley:
### Feedback welcome on the following:

1. `setLogLevel()` API on web has no counterpart on native. At the moment, it just returns `debug`. [See here](https://github.com/invertase/react-native-firebase/compare/feat/modular-remote-config?expand=1#diff-85993e262f14abe7fdaf5ec03dd18a7a0b51546fd8cff2617c2538f28e240bb3R126-R129). 
2. Web JS SDK has a setter for [defaultConfig](https://github.com/firebase/firebase-js-sdk/blob/68369740aa8802ea8391432f10f9baa4abb84aa0/packages/remote-config-compat/src/remoteConfig.ts#L52-L54). We have a native counterpart that is async called [setDefaults()](https://github.com/invertase/react-native-firebase/blob/main/packages/remote-config/lib/index.js#L196-L202). What should we do about this?
3. Exact same situation for [settings](https://github.com/firebase/firebase-js-sdk/blob/68369740aa8802ea8391432f10f9baa4abb84aa0/packages/remote-config-compat/src/remoteConfig.ts#L68-L70). Native counterpart is [setConfigSettings()](https://github.com/invertase/react-native-firebase/blob/main/packages/remote-config/lib/index.js#L122-L155) which is async. What to do about this?

### Checklist

  - [ ] Yes
  - [ ] `Android`
  - [ ] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [ ] No

### Test Plan

---


mikehardy:
interesting - on the 3 questions listed currently in description -

1- to be firebase-js-sdk compatible we need to define the LogLevel type as well https://firebase.google.com/docs/reference/js/remote-config.md#loglevel and support the setLogLevel API, I think 'error' is probably the better choice for default return value as it will imply on native that you are not going to get any special non-error logging, and have a note that just says "setter is ignored on native"

2- we have to have a setter to be compatible, so it should do delegate to the existing implementation and just not await on it I guess. javascript property setters can delegate to other methods right? It's just the other direction (non-property implementations delegating to property setters) that doesn't work, as you learned+mentioned the other day?

3- same here we have to have the setter to be compatible. I just read an issue logged against the TC39 committee plus some stackoverflow and there is not and probably will never by async setters which - to me - just argues for never using javascript properties but firebase-js-sdk forces us to.

So both 2 and 3 - implement the setter so we are compatible, perhaps with a note that it is async on native platforms, and people should await on the native API or await on a short sleep if avoiding a race condition is critical 🤷

russellwheatley:
> interesting - on the 3 questions listed currently in description -
> 
> 1- to be firebase-js-sdk compatible we need to define the LogLevel type as well [firebase.google.com/docs/reference/js/remote-config.md#loglevel](https://firebase.google.com/docs/reference/js/remote-config.md#loglevel) and support the setLogLevel API, I think 'error' is probably the better choice for default return value as it will imply on native that you are not going to get any special non-error logging, and have a note that just says "setter is ignored on native"
> 
> 2- we have to have a setter to be compatible, so it should do delegate to the existing implementation and just not await on it I guess. javascript property setters can delegate to other methods right? It's just the other direction (non-property implementations delegating to property setters) that doesn't work, as you learned+mentioned the other day?
> 
> 3- same here we have to have the setter to be compatible. I just read an issue logged against the TC39 committee plus some stackoverflow and there is not and probably will never by async setters which - to me - just argues for never using javascript properties but firebase-js-sdk forces us to.
> 
> So both 2 and 3 - implement the setter so we are compatible, perhaps with a note that it is async on native platforms, and people should await on the native API or await on a short sleep if avoiding a race condition is critical 🤷

@mikehardy I've updated the PR to delegate the setters to the existing async API. The difference between perf and remote-config is; it didn't matter that we didn't have to await dataCollectionEnabled setter on perf because it would only take effect on the next app run, and we updated a simple boolean on the perf instance before getting to native. 

The two setters for remote-config update the settings and default config on the remote-config instance after sending them to native. I guess we should update them before they get to native? 🤔

mikehardy:
> The two setters for remote-config update the settings and default config on the remote-config instance after sending them to native. I guess we should update them before they get to native? 🤔

Updating the local instance before going to native seems best so that at least after the property setting returns it has an expected value if re-tested, even though native will be potentially still working async. This is a platform inconsistency we can't really bridge but I think a note in the docs "due to the nature of the react-native javascript/native bridge the actual native effect is asynchronous and there may be a slight delay at the native level before this takes effect. If that's important to your app, and you want to use the firebase-js-sdk compatible property setters, insert a small sleep after calling the setters" or similar ?

russellwheatley:
We have a bit of an issue with web. The current, native `setConfigSettings()` is designed for iOS and android which has different property names for settings and is also set in seconds 😓. 

[See here](https://github.com/invertase/react-native-firebase/blob/main/packages/remote-config/lib/index.js#L122-L154).
It also gets the returned values from settings with the native property names and converts them [here](https://github.com/invertase/react-native-firebase/blob/main/packages/remote-config/lib/index.js#L218-L228).

I've drafted [this commit](https://github.com/invertase/react-native-firebase/pull/6868/commits/a0570344a1993936303a3b5b3c19adac7f7f2044) to get your thoughts, @mikehardy? 

Will update tests and make note about non-async stuff for the setter but thought I'd see what you think of this first?

mikehardy:
@russellwheatley I think the native interfaces do not matter much, the only really important thing is that the JS (and Typescript) API is firebase-js-sdk compatible, and it appears we took some care to do that - firebase-js-sdk API is in millis and our API is in millis so we don't have a breaking change on tap here. No problem if our *internal* JS code is translating between millis and seconds (which it is doing now), we can just keep right on doing that.

And same with the property names - as long as we present a firebase-js-sdk compatible facade at the JS layer, native can be whatever it needs to be

Stated more practically: I think this may actually be close then - allowing for the "dangit, react-native is async..." stuff as a slight difference, which I think you've handled reasonably given constraints that javascript properties can't be async but react-native has to be over the bridge

russellwheatley:
@mikehardy, I changed the API for a couple of reasons. I wanted to make them cross platform and follow DRY principles.

The way to now think about `settings` & `setConfigSettings` is the former is sync and the latter is async. I've implemented them so they're cross platform. (i.e. I've checked what platform is running and updated the values accordingly). They can now be used on all 3 platforms.

This is the same for `defaultConfig` & `setDefaults`.

`_updateFromConstants` has been updated to accept values from all platforms to this end. [See here](https://github.com/invertase/react-native-firebase/pull/6868/commits/f7195682907dd11eb2ba0f66837373984dff92f5#diff-dd309df0215e2eea9d1d6b6aa5816654c61cf223e7843bd7af7e3eaddb644358R285-R306). You might notice, I've now wrapped [constants.lastFetchTime](https://github.com/invertase/react-native-firebase/pull/6868/commits/f7195682907dd11eb2ba0f66837373984dff92f5#diff-dd309df0215e2eea9d1d6b6aa5816654c61cf223e7843bd7af7e3eaddb644358R285-R306) and [constants.lastFetchStatus](https://github.com/invertase/react-native-firebase/pull/6868/commits/f7195682907dd11eb2ba0f66837373984dff92f5#diff-dd309df0215e2eea9d1d6b6aa5816654c61cf223e7843bd7af7e3eaddb644358R292-R294). This is because we have to call this [function here](https://github.com/invertase/react-native-firebase/pull/6868/commits/f7195682907dd11eb2ba0f66837373984dff92f5#diff-dd309df0215e2eea9d1d6b6aa5816654c61cf223e7843bd7af7e3eaddb644358R119) & [here](https://github.com/invertase/react-native-firebase/pull/6868/commits/f7195682907dd11eb2ba0f66837373984dff92f5#diff-dd309df0215e2eea9d1d6b6aa5816654c61cf223e7843bd7af7e3eaddb644358R106) as we need to update values on the instance before it gets to native for the sync setters. Those properties won't be available until the round trip from the platform is complete.

I have been a bit sneaky with the error messages so they correctly output the correct API call. I do this by using [call method](https://github.com/invertase/react-native-firebase/pull/6868/commits/f7195682907dd11eb2ba0f66837373984dff92f5#diff-dd309df0215e2eea9d1d6b6aa5816654c61cf223e7843bd7af7e3eaddb644358R120) to pass another argument that isn't part of the API. [Retrieve it here](https://github.com/invertase/react-native-firebase/pull/6868/commits/f7195682907dd11eb2ba0f66837373984dff92f5#diff-dd309df0215e2eea9d1d6b6aa5816654c61cf223e7843bd7af7e3eaddb644358R187) for correct error message. 

I'm sure you know, but simple demonstration of how that works:
```js
function func1(a) {
  console.log(arguments[0]);
  // Expected output: 1

  console.log(arguments[1]);
  // Expected output: 2

}

func1.call(this, 1, 2)
```

I hope you see this as an improvement? Can change back if it's a little too complicated, and you wanted to keep each API separate and dedicated to their respective platforms 😅

packages/remote-config/lib/index.d.ts
```diff
@@ -283,6 +286,11 @@ export namespace FirebaseRemoteConfigTypes {
     [key: string]: number | string | boolean;
   }
 
+  /**
+   * The status of the latest Remote RemoteConfig fetch action.
+   */
+  type LastFetchStatusType = 'success' | 'failure' | 'no_fetch_yet' | 'throttled';
```
mikehardy, line 292: nice - a real type is an improvement

packages/remote-config/lib/index.d.ts
```diff
@@ -306,15 +314,20 @@ export namespace FirebaseRemoteConfigTypes {
...
+    lastFetchStatus: LastFetchStatusType;
 
     /**
      * Provides an object which provides the properties `minimumFetchIntervalMillis` & `fetchTimeMillis` if they have been set
      * using setConfigSettings({ fetchTimeMillis: number, minimumFetchIntervalMillis: number }). A description of the properties
      * can be found above
      *
      */
-    settings: { fetchTimeMillis: number; minimumFetchIntervalMillis: number };
+    settings: ConfigSettings;
```
mikehardy, line 325: 😆 why wasn't it like this before !? nice improvement

packages/remote-config/lib/index.js
```diff
@@ -59,6 +83,41 @@ class FirebaseConfigModule extends FirebaseModule {
...
+
+    return updatedDefaultConfig;
+  }
+
+  set defaultConfig(defaults) {
+    if (!isObject(defaults)) {
+      throw new Error("firebase.remoteConfig().defaultConfig: 'defaults' must be an object.");
+    }
+    // To make Firebase web v9 API compatible, we update the config first so it immediately
+    // updates defaults on the instance. We then pass to web platform to update. We do this because
```
mikehardy, line 104: This is just a quibble on wording, but it's a mentality - if we work with apple/android/web we are not passing to web just because it's the modular API we present, we're wrapping any of the SDKs, so mental model is "we handle our business in the wrapping layer then hand off to SDK...", not just "web"

```suggestion
    // updates defaults on the instance. We then pass to underlying SDK to update. We do this because
```

packages/remote-config/lib/index.js
```diff
@@ -59,6 +83,41 @@ class FirebaseConfigModule extends FirebaseModule {
...
+    this.setDefaults.call(this, defaults, true);
+  }
+
+  get settings() {
+    return this._settings;
+  }
+
+  set settings(settings) {
+    // To make Firebase web v9 API compatible, we update the settings first so it immediately
+    // updates settings on the instance. We then pass to web platform to update. We do this because
```
mikehardy, line 116:
```suggestion
    // updates settings on the instance. We then pass to underlying SDK to update. We do this because
```

packages/remote-config/lib/index.js
mikehardy: I didn't perturb the PR with these FWIW but still worth thinking about w.r.t. mental model :-)

PR #6869, opened 2023-01-27
fix(firestore): support not equals null query

pafry7:
### Description

Retrieving all documents with field not equal to null doesn't work. Example:
```
firestore().collection("USERS").where("address", "!=", null);
```
Returns
```
firebase.firestore().collection().where(_, _, *) 'value' is invalid. You can only perform equals comparisons on null
```
I've looked into  `FirestoreQuery.js` and only equality comparison is allowed. 
```
if (isNull(value) && !this._modifiers.isEqualOperator(opStr)){...}
```
Is there any specific reason for it? Is there anything else I need to change, that I'm not aware of ? 

### Checklist

  - [x] Yes
  - [x] `Android`
  - [x] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

I've run tests locally and everything passes. I have not found a test related to this problem (maybe this is something that needs to be added).

🔥
---

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -102,7 +102,7 @@ cd react-native-firebase
 yarn
 yarn tests:ios:pod:install
 brew tap wix/brew
-brew install applesimutils xbeautify
+brew install applesimutils xcbeautify
 ```
 
 > Note that this project is a mono-repo, so you only need to install NPM dependencies once at the root of the project with `yarn`.
--- a/packages/firestore/lib/FirestoreQuery.js
+++ b/packages/firestore/lib/FirestoreQuery.js
@@ -437,7 +437,11 @@ export default class FirestoreQuery {
       );
     }
 
-    if (isNull(value) && !this._modifiers.isEqualOperator(opStr)) {
+    if (
+      isNull(value) &&
+      !this._modifiers.isEqualOperator(opStr) &&
+      !this._modifiers.isNotEqualOperator(opStr)
+    ) {
       throw new Error(
         "firebase.firestore().collection().where(_, _, *) 'value' is invalid. You can only perform equals comparisons on null",
       );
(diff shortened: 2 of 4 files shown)


CONTRIBUTING.md
```diff
@@ -102,7 +102,7 @@ cd react-native-firebase
 yarn
 yarn tests:ios:pod:install
 brew tap wix/brew
-brew install applesimutils xbeautify
+brew install applesimutils xcbeautify
```
mikehardy, line 105: 🤦 thanks :-)

packages/firestore/lib/FirestoreQuery.js
```diff
@@ -437,7 +437,11 @@ export default class FirestoreQuery {
       );
     }
 
-    if (isNull(value) && !this._modifiers.isEqualOperator(opStr)) {
+    if (
+      isNull(value) &&
+      !this._modifiers.isEqualOperator(opStr) &&
+      !this._modifiers.isNotEqualOperator(opStr)
```
mikehardy, line 443: This *does* in fact appear valid + correct. `!=` is supported in comparisons against null values, so this change looks good

mikehardy (changes requested):
This looks great - nice catch on a corner case missed in the implementation - you ask the question if there is any other place to change - and the answer is just in exercising it in the e2e harness to validate:

- an additional test like this one to verify that it is allowed now: https://github.com/invertase/react-native-firebase/blob/da82c1036051f0518da0401de24cef24c7ac091f/packages/firestore/e2e/Query/where.e2e.js#L87-L89

Other than that, I don't think more testing is necessary, it appears the `!=` operator is already exercised pretty well here https://github.com/invertase/react-native-firebase/blob/da82c1036051f0518da0401de24cef24c7ac091f/packages/firestore/e2e/Query/where.e2e.js#L532-L589

I think that's a really trivial addition so while I normally won't hang up on community contributions on testing, given you have installed xcbeautify (guessing from your diff line there ;-) ) adding this simple adaptation of an existing test seems like it isn't onerous?

That along with the CLA signed (check the related comment on the PR) - then good to go

What do you think?

pafry7:
@mikehardy Done 🔥

mikehardy (approved):
Fantastic - thanks of course just for noticing + fixing the error, and bonus thanks for a docs fix and a test addition. Great PR

PR #6910, opened 2023-02-15
feat(dynamic-links): Expose modular API that matches the Firebase web JS SDK v9 API

Lyokone:
### Checklist

  - [ ] Yes
  - [ ] `Android`
  - [ ] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [ ] No

### Test Plan

---


mikehardy:
I still think - even though it's native only - having the modular *style* for the API is useful, otherwise this module will feel pretty strange with regard to API flavor.

We'll have the same problem with app-distribution and crashlytics I think? Also maybe in-app-messaging, installations and ml. But having non-modular and modular just seems strange to me as the API style is so regular (sending the app in to get an instance, using it...) for v9 and then we'll have modules that you just use directly

mikehardy:
For clarity - this was discussed on slack, and the consensus was that all modules will move to the v9 / modular API so that the DX of using the various modules is consistent

exaby73:
Closed in favour of https://github.com/invertase/react-native-firebase/pull/7296

Issue #6705, opened 2022-11-17
☂️ Bring Modular V9 to React Native Firebase 


Lyokone:
This issue will track the progress and gather feedback for bringing the V9 API of Firebase to React Native Firebase.

The objective is to have a compatible API so people may adopt the V9 API if they wish, or if they were already using firebase-js-sdk v9 and find they need some native features provided by react-native-firebase, they may adopt react-native-firebase with zero code changes.

## Modules
- [x]  App: #6694
- [x] app-check https://github.com/invertase/react-native-firebase/pull/6912       
- [x] app-distribution https://github.com/invertase/react-native-firebase/pull/7307
- [x] analytics https://github.com/invertase/react-native-firebase/pull/6816      
- [x] auth  https://github.com/invertase/react-native-firebase/pull/7015            
- [x] crashlytics https://github.com/invertase/react-native-firebase/pull/7283
- [x] database https://github.com/invertase/react-native-firebase/pull/7136
- [x] dynamic-links https://github.com/invertase/react-native-firebase/pull/7296
- [x] firestore https://github.com/invertase/react-native-firebase/pull/7235
- [x] functions    https://github.com/invertase/react-native-firebase/pull/6764    
- [x] in-app-messaging https://github.com/invertase/react-native-firebase/pull/7306
- [x] installations   https://github.com/invertase/react-native-firebase/pull/7095 
- [x] messaging  https://github.com/invertase/react-native-firebase/pull/6806      
- [x] ml https://github.com/invertase/react-native-firebase/pull/7390
- [x] perf https://github.com/invertase/react-native-firebase/pull/6771    
- [x] remote-config #6868 
- [x] storage  #6958

jeveloper:
Github bot doesn't respect humans needing to enjoy some of that down time and self care. Tsk tsk.
Enjoy the holidays all and thank you for all the wonderful work over many years!

mikehardy:
The bot doesn't sleep no

adrianotadao:
hey @Lyokone @mikehardy, thank you for your hard work on it! #heroes

I'm quite interested in the **firestore** item, is this progress list up to date?

Thank you

Lyokone:
Hello @adrianotadao, Firestore hasn't yet been brought to modular v9 APIs. It will probably be dealt with in the coming weeks.

Slapbox:
If I'm not mistaken, at least one of these items and possibly several are released and need to be checked off in this issue.

[...]

PR #7015, opened 2023-03-31
feat(auth): Expose modular API that matches the Firebase web JS SDK v9 API

Lyokone:
### Related issues

https://github.com/invertase/react-native-firebase/issues/6705

### Checklist

  - [ ] Yes
  - [ ] `Android`
  - [ ] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [ ] No

### Test Plan

---

### DIFF
--- a/packages/auth/e2e/auth.e2e.js
+++ b/packages/auth/e2e/auth.e2e.js
@@ -23,923 +23,958 @@ const DISABLED_PASS = 'test1234';
 
 const { clearAllUsers, disableUser, getLastOob, resetPassword } = require('./helpers');
 
-describe('auth()', function () {
-  before(async function () {
-    await clearAllUsers();
-    await firebase.auth().createUserWithEmailAndPassword(TEST_EMAIL, TEST_PASS);
-    const disabledUserCredential = await firebase
-      .auth()
-      .createUserWithEmailAndPassword(DISABLED_EMAIL, DISABLED_PASS);
-    await disableUser(disabledUserCredential.user.uid);
-  });
-
-  beforeEach(async function () {
-    if (firebase.auth().currentUser) {
-      await firebase.auth().signOut();
-      await Utils.sleep(50);
-    }
-  });
-
-  describe('namespace', function () {
-    it('accessible from firebase.app()', function () {
-      const app = firebase.app();
-      should.exist(app.auth);
-      app.auth().app.should.equal(app);
-    });
-
-    // removing as pending if module.options.hasMultiAppSupport = true
-    it('supports multiple apps', async function () {
-      firebase.auth().app.name.should.equal('[DEFAULT]');
-
-      firebase
-        .auth(firebase.app('secondaryFromNative'))
-        .app.name.should.equal('secondaryFromNative');
-
-      firebase.app('secondaryFromNative').auth().app.name.should.equal('secondaryFromNative');
-    });
-  });
-
-  describe('applyActionCode()', function () {
-    // Needs a different setup to work against the auth emulator
-    xit('works as expected', async function () {
-      await firebase
+describe('auth() modular', function () {
+  describe('firebase v8 compatibility', function () {
+    before(async function () {
+      await clearAllUsers();
+      await firebase.auth().createUserWithEmailAndPassword(TEST_EMAIL, TEST_PASS);
+      const disabledUserCredential = await firebase
         .auth()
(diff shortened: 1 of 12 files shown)


mikehardy (commented):
just to note passes all tests locally, structurally looks right, I'm running through index.js right now with an eye towards how platform differences were handled (always a point of discussion, but already discussed I think...) and why github is telling me that there was no test coverage on the new modular APIs (when it looks like they were covered... 🤔 )

mikehardy (changes requested):
All seems to work, and the cause of coverage notices appears to be that the declaration lines are not covered but the implementation lines are? This seems odd and worth a (separate, not related to this PR) investigation but shouldn't be a blocker. Hypothesis based on looking at this: https://app.codecov.io/gh/invertase/react-native-firebase/pull/7015

I have a couple real questions though:

- can we have *only* the modular e2e tests, with their non-modular compat testing inside of them, and remove the non-modular tests? Really, just make the modular tests the new default, no need to have modular in the name
- the provider, rnReload, and multifactor tests do not seem to be modularized, if those are not modularized / exercising those features in the v9 / modular way I fear they will break in the future. Unless I am misunderstanding (I could be! please let me know) I think that's a blocker for merge

Lyokone:
Thanks for the review.
- For the first point, all the other migrated modules have kept the `.modular` for the new tests. I think we can remove the non-modular of all plugins at the same time but I don't see a reason to do it only for auth.
- For the second point, I missed that you could access multifactor with a top level function now, I'll add the 3 missing e2e tests.

packages/auth/e2e/rnReload.modular.e2e.js
```diff
@@ -0,0 +1,190 @@
...
+        currentUser.uid.should.be.a.String();
+        currentUser.toJSON().should.be.an.Object();
+        should.equal(currentUser.toJSON().email, null);
+        currentUser.isAnonymous.should.equal(true);
+        currentUser.providerId.should.equal('firebase');
+        currentUser.should.equal(defaultAuth.currentUser);
+
+        // test correct user is returned after signing
+        // in with a different user then reloading
+        await signOut(defaultAuth);
```
Lyokone, line 162: This signOut hangs out, I have no idea why @mikehardy

mikehardy: Huh - that is unexpected (of course!) - I'll check it locally and see what I can do about CI / investigating

mikehardy:
> For the first point, all the other migrated modules have kept the .modular for the new tests. I think we can remove the non-modular of all plugins at the same time but I don't see a reason to do it only for auth.


Okay - agreed there, I think we did one of them (the most recent?) with just convert-main-file-name-to-modular + compat-inside-modular and I'd like that going forward. You are correct that since there are quite a few in the old style (that this one is using) of old-style + modular-with-compat-inside we can do one PR that realigns all at once. No need to hold this up waiting for that

I'll check the new tests and see what I can find out about signOut hanging. Thanks @Lyokone !

russellwheatley (changes requested):
I think a PhoneAuthProvider test needs to be updated. And there's the lint failure that Mike mentioned. Other than that, LGTM 👌.

mikehardy:
I fixed the PhoneAuthProvider constructor to throw in a  backwards-compatible way and cleared the lint error, rebased to current main and re-pushed, we'll see what CI says, seems to run fine locally...

Lyokone:
@mikehardy Thanks for taking the on the rest of the PR 👍

mikehardy:
Note to self: need to add API for #7148 that came in while this was in-flight

mikehardy:
Please do not add merge commits unless necessary for some reason, they make it more difficult for me to reason about PRs
In this case I don't think any new commits were added but main/HEAD was merged anyway, which does not seem to me to add any value / serve any purpose but does make the commit string unclean

mikehardy:
I think everything is fixed now, pending CI

mikehardy:
Passed all the tests except lint on the last run, I fixed the lint one - merging!

aedocoding:
Hello, I was wondering if this PR being merged means that the next release of `react-native-firebase` would have the [Config interface](https://firebase.google.com/docs/reference/js/auth.config) accessible to users to make changes to properties like the `apiHost` or `tokenApiHost`? Apologies if this is the wrong place to ask this question by the way

PR #7045, opened 2023-04-14
feat(firestore): Firestore `Filter` instance. Use `Filter`, `Filter.or()` & `Filter.and()` in Firestore queries.

russellwheatley:
### Description

Latest `Filter` feature for RNFB. Integrated into the existing implementation. This allows chaining even with existing query capability:

```js
firebase.firestore().collection('some-collection').where(
        Filter.or(
          Filter.and(Filter('foo', 'array-contains', 'some-string'), Filter('bar', '==', 'baz')),
          Filter.and(Filter('foo', '==', 'not-this'), Filter('foo', '==', 'bar')),
        ),
      )
      .where('existing', '==', 'where')
      .where(Filter.and(Filter('another', '==', 'filter'), Filter('chain', '==', 'and')))
      .get();
```

It is largely based on recursion. 
1. We have recursion on the JS side for validation & generating the correct map for native parsing [here](https://github.com/invertase/react-native-firebase/pull/7045/files#diff-78d578c73633c0807e153034aafb7f9e899cc7fbfa2fa2fc4170680e8bf5fc3dR76). 
2. We have recursion on the native side for generating the filters. [android](https://github.com/invertase/react-native-firebase/pull/7045/files#diff-e44ab9a51100c8820d63c64443b468cf72dd5d4aba3faaf570e41b632266d1e2R188) & [iOS](https://github.com/invertase/react-native-firebase/pull/7045/files#diff-58931c2966f2a235c2d87ecb1c41cd5cace0aafe0071ecb3426f236a92018f88R131).

This PR also preserves the query order of chained filters that the user generates. 

We still send the same list of filters over to the native side except the list can have objects that are nested Filter queries:
```js
[
  {
    "operator": "OR",
    "queries": [
      {
        "operator": "AND",
        "queries": [
          {
            "fieldPath": {
              "_segments": [
                "foo"
              ]
            },
            "operator": "ARRAY_CONTAINS",
            "value": [
              8,
              "1681731303828"
            ]
          },
          {
            "fieldPath": {
              "_segments": [
                "bar"
              ]
            },
            "operator": "EQUAL",
            "value": [
              8,
              "baz"
            ]
          }
        ]
      },
      {
        "operator": "AND",
        "queries": [
          {
            "fieldPath": {
              "_segments": [
                "foo"
              ]
            },
            "operator": "EQUAL",
            "value": [
              8,
              "not-this"
            ]
          },
          {
            "fieldPath": {
              "_segments": [
                "bar"
              ]
            },
            "operator": "EQUAL",
            "value": [
              8,
              "baz"
            ]
          }
        ]
      }
    ]
  },
  {
    "fieldPath": {
      "_segments": [
        "existing"
      ]
    },
    "operator": "EQUAL",
    "value": [
      8,
      "where"
    ]
  },
  {
    "operator": "AND",
    "queries": [
      {
        "fieldPath": {
          "_segments": [
            "another"
          ]
        },
        "operator": "EQUAL",
        "value": [
          8,
          "filter"
        ]
      },
      {
        "fieldPath": {
          "_segments": [
            "chain"
          ]
        },
        "operator": "EQUAL",
        "value": [
          8,
          "and"
        ]
      }
    ]
  }
]
```

### Checklist

  - [x] Yes
  - [ ] `Android`
  - [x] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

---


docs/firestore/usage/index.md
```diff
@@ -239,6 +239,43 @@ firestore()
...
+```
+
+You can use the `Filter.or()` static method to make logical OR queries:
+
+```js
+const snapshot = await firestore()
+  .collection('Users')
+  .where(
+    Filter.or(
+      Filter.and(Filter('user', '==', 'Tim'), Filter('email', '==', 'tim@email.com')),
```
Ehesp, line 269: Can we add checks in to validate they can't do these types of things:

```
Filter.or(
  Filter.or(...),
```

```
Filter.or(
  Filter.and(Filter.or(...)),
```

Ehesp (dismissed):

docs/firestore/usage/index.md
```diff
@@ -239,6 +239,43 @@ firestore()
...
 [Firebase documentation](https://firebase.google.com/docs/firestore/query-data/queries).
 
+It is now possible to use the `Filter` instance to make queries. They can be used with the existing query API.
+For example, you could chain like so:
+
+```js
+const snapshot = await firestore()
+  .collection('Users')
+  .where(Filter('user', '==', 'Tim'))
+  .where('email', '==', 'tim@email.com')
```
mikehardy, line 249: Going to make a batch of suggestions + commit them that changes these to the RFC-prescribed example domain: https://datatracker.ietf.org/doc/html/rfc2606#section-3

```suggestion
  .where('email', '==', 'tim@example.com')
```

docs/firestore/usage/index.md
```diff
@@ -239,6 +239,43 @@ firestore()
...
+  .get();
+```
+
+You can use the `Filter.and()` static method to make logical AND queries:
+
+```js
+const snapshot = await firestore()
+  .collection('Users')
+  .where(Filter.and(Filter('user', '==', 'Tim'), Filter('email', '==', 'tim@email.com')))
```
mikehardy, line 258:
```suggestion
  .where(Filter.and(Filter('user', '==', 'Tim'), Filter('email', '==', 'tim@example.com')))
```

docs/firestore/usage/index.md
mikehardy, line 269:
```suggestion
      Filter.and(Filter('user', '==', 'Tim'), Filter('email', '==', 'tim@example.com')),
```

docs/firestore/usage/index.md
```diff
@@ -239,6 +239,43 @@ firestore()
...
+
+You can use the `Filter.or()` static method to make logical OR queries:
+
+```js
+const snapshot = await firestore()
+  .collection('Users')
+  .where(
+    Filter.or(
+      Filter.and(Filter('user', '==', 'Tim'), Filter('email', '==', 'tim@email.com')),
+      Filter.and(Filter('user', '==', 'Dave'), Filter('email', '==', 'dave@email.com')),
```
mikehardy, line 270:
```suggestion
      Filter.and(Filter('user', '==', 'Dave'), Filter('email', '==', 'dave@example.com')),
```

mikehardy (approved):
I was about to ask "where's the modular API?" then remembered firestore doesn't have it yet :-). Will commit the example.com changes and otherwise LGTM, this is an amazing feature

I did see some formatting moving on the java side and I think we've had false-negatives on google-java-format before so I'll check that locally and push a spacing commit as needed, but otherwise good to go

mikehardy (changes requested):
Oh wait - what about types? Should there be a typescript types update in https://github.com/invertase/react-native-firebase/blob/main/packages/firestore/lib/index.d.ts ? Seems like there has to be or this won't be usable in typescript projects

mikehardy:
Yep as suspected there was a false-negative on java formatting, I ran `lint android:format` locally until it was clean then commited+pushed the result.

Only pending issue for me is typescript types question now

packages/firestore/lib/index.d.ts
```diff
@@ -49,6 +49,39 @@ import { ReactNativeFirebase } from '@react-native-firebase/app';
...
+    /**
+     * The Filter.or() static function used to generate a logical OR query using multiple Filter instances.
+     * e.g. Filter.or(Filter('name', '==', 'Ada'), Filter('name', '==', 'Bob'))
+     */
+    or(...queries: Filter[]): Filter;
+    /**
+     * The Filter.and() static function used to generate a logical AND query using multiple Filter instances.
+     * e.g. Filter.and(Filter('name', '==', 'Ada'), Filter('name', '==', 'Bob'))
+     */
+    and(...queries: Filter[]): Filter;
```
mikehardy, line 78: @russellwheatley 

🤔  these types don't seem like a fit vs the firebase-js-sdk types, is that intentional?

Here's the `and()` from firebase-js-sdk https://firebase.google.com/docs/reference/js/firestore_.md#and

Has all sorts of QueryFilterCompositeConstraint stuff, takes QueryFilterConstraint[] etc types that aren't present here

Not sure how this will work as a drop-in replacement

russellwheatley: I've updated types to match naming: https://github.com/invertase/react-native-firebase/pull/7045/commits/20e9c8d7f200a9b00c461b54d9659b3afb2b6305. Let me know if this works.

Aleksandar-FFWD:
Great work! 👏 When can we expect this release? 👍

illlama:
Thanks for update! 
But I have a question.
Where do I have to import `Filter` ? 

I used like this and it occurs an error from where 
`Expected  3  arguments, but got  1 `
I think the way I import was wrong..


```
import firestore, { FirebaseFirestoreTypes } from '@react-native-firebase/firestore';
import Filter = FirebaseFirestoreTypes.Filter;
...
    .where(Filter('name', '==', 'test'))
```

my versions are like bellow
```
 "@react-native-firebase/app": "^17.5.0",
    "@react-native-firebase/auth": "^17.5.0",
    "@react-native-firebase/firestore": "^17.5.0",
    "@react-native-firebase/storage": "^17.5.0",
```

filipef101:
@russellwheatley @mikehardy Any reason for the 10 query operators [limit](https://github.com/invertase/react-native-firebase/blob/0103e12714de3c106128859eeadaf0fe07f9674c/packages/firestore/lib/FirestoreFilter.js)? I don't see any reason for that check to exist, if we hit any limit firebase api's should be the ones complaining right?

mikehardy:
@filipef101 in general we try to enforce things limitations that are clearly described in the documents before attempting to call the cloud APIs.

However in this case I believe the limits have changed from a strict "10" to a new kind-of-vague "no more than 30 disjunctions based on the query's disjunctive normal form"

https://cloud.google.com/firestore/docs/query-data/queries#query_limitations

It may make sense for you open an issue related to this, quoting the new limitations and lobbying for a removal of local validation here, to be replaced with server validation / error handling.

Before you do that, can you verify what *exactly* will happen (via direct manipulation of the code in node_modules to allow things through...) if you send two examples to probe the behavior:
- one that is 20 or filters (should pass?)
- one that is 31 or filters (should fail, but what exact error message is thrown?)

Issue #6785, opened 2022-12-15
[🐛🔥] @react-native-firebase/auth config plugin for phone auth + iOS does not recognize googleServicesFile if it is in Expo secrets vs filesystem


TrevPennington:
## Issue

Hello, I am trying to implement phone auth using Expo managed. I follow the instructions and add `@react-native-firebase/auth` to `app.json` plugins array. Running `eas build` for a custom dev client will work on Android but not iOS.

iOS output:
`Error: [ios.infoPlist]: withIosInfoPlistBaseMod: [@react-native-firebase/auth] Your app.json file is missing ios.googleServicesFile. Please add this field.`

I have created the secret and uploaded the google plist file to expo secrets. Everything works fine on Android. iOS does not recognize the file.

expo: 47.0.8
@react-native-firebase/auth: 16.4.6

app.config.js

```
export default ({ config }) => {
  return {
    ...config,

    ios: {
      ...config.ios,
      googleServicesFile: process.env.FIREBASE_IOS,
    },

    android: {
      ...config.android,
      googleServicesFile: process.env.FIREBASE_ANDROID,
    },
    plugins: [...config.plugins, "@react-native-firebase/auth"],
  };
};
```

---

### Javascript

<details><summary>Click To Expand</summary>
<p>

#### `package.json`:

```json
# N/A
```

#### `firebase.json` for react-native-firebase v6:

```json
# N/A
```

</details>

### iOS

<details><summary>Click To Expand</summary>
<p>

#### `ios/Podfile`:

- [ X] I'm not using Pods
- [] I'm using Pods and my Podfile looks like:

```ruby
# N/A
```

#### `AppDelegate.m`:

```objc
// N/A
```

</p>
</details>

---

### Android

<details><summary>Click To Expand</summary>
<p>

#### Have you converted to AndroidX?

- [ ] my application is an AndroidX application?
- [ ] I am using `android/gradle.settings` `jetifier=true` for Android compatibility?
- [ ] I am using the NPM package `jetifier` for react-native compatibility?

#### `android/build.gradle`:

```groovy
// N/A
```

#### `android/app/build.gradle`:

```groovy
// N/A
```

#### `android/settings.gradle`:

```groovy
// N/A
```

#### `MainApplication.java`:

```java
// N/A
```

#### `AndroidManifest.xml`:

```xml
<!-- N/A -->
```

</p>
</details>

---

## Environment

<details><summary>Click To Expand</summary>
<p>

[...]

PR #7067, opened 2023-04-23
fix(expo): Update @react-native-firebase/auth iOS plugin

kreshnikalidema:
### Description
Fixes firebase auth plugin for Expo iOS. The current plugin doesn't work if the file is uploaded to EAS secrets, so it requires the `GoogleService-Info.plist` file to be present locally, in the root of the project folder to be able to initiate the build. After analyzing the code I realized that `@react-native-firebase/auth` iOS plugin does nothing more than `IOSConfig.Google.withGoogle` which is built-in `@expo/config-plugins`, and for this reason I have replaced it, also this replacement fixes the issue with EAS secret files.

### Related issues
https://github.com/invertase/react-native-firebase/issues/6785


mikehardy:
Hey @kreshnikalidema 👋 !

This looks like a nice fix, lots of red/deleted lines and only a few green/added lines, my favorite type of PR if the functionality is maintained :-)

You state this:

>  that @react-native-firebase/auth iOS plugin does nothing more than IOSConfig.Google.withGoogle which is built-in @expo/config-plugins, and for this reason I have replaced it, also this replacement fixes the issue with EAS secret files.

Do you have a link ref to that `IOSConfig.Google.withGoogle` so I can make sure it still has the recaptcha URLs that the old style was adding? It should be linkable somewhere - it is difficult to review otherwise

Two other items:
- we need the CLA signed before we may accept any contributions from anyone, can you follow the instructions in the comment posted by the CLA bot above ☝️ ?
- you have not mentioned how you tested this, how did you prove to yourself that it was working after the change?

Wael-Ibn-Douma (approved):

mikehardy:
@kreshnikalidema hey there 👋  this PR is still blocked on the CLA and the comments above - do you have any plans to complete it? Cheers

kreshnikalidema:
@mikehardy  I apologize for the delayed reply. I have been very busy recently. I will deal with this today.

mikehardy:
Hmm. This still seems useful. Anyone up for fixing this up to be current (in case that is needed) and supplying at as a PR with the CLA signed?

Issue #7072, opened 2023-04-25
[📚] Clarifying React Native Linking for Older Versions


KiranAminPanjwani:
## Documentation Feedback

The paragraph in the [index.md](https://github.com/gosunman/react-native-firebase/blob/d78f93e6dcaf512d4b88bcbf5879854645404aba/docs/index.md) document may create ambiguity for users who are using React Native version below 0.60 and are not familiar with React Native versions and autolinking support. 
The paragraph reads: "Manual Linking: If you're using an older version of React Native without autolinking support, or wish to integrate into an existing project, you can follow the manual installation steps for iOS and Android."

To improve the clarity of the paragraph, it is suggested to specify the version of React Native that does not have autolinking support. The revised paragraph could read: "**Manual Linking:** If you're using an older version **(below 0.60)** of React Native without autolinking support, or wish to integrate into an existing project, you can follow the manual installation steps for iOS and Android."

<kbd>![image](https://user-images.githubusercontent.com/90326051/234265292-ef3d79b9-00e0-4565-8adf-31a642a889a0.png)</kbd>

@mikehardy Kindly review my issue. Thank you.😊

---

- 👉 Check out [`React Native Firebase`](https://twitter.com/rnfirebase) and [`Invertase`](https://twitter.com/invertaseio) on Twitter for updates on the library.

PR #7102, opened 2023-05-04
docs: Clarifying rn linking

mtalhabaig3:
fixes issue #7072 

Clarified React Native Linking for Older Versions in the manual linking section


mikehardy (changes requested):
Interesting - had not even considered the react-native < 0.60 case in a while.
A better solution would be a stronger statement of support following the guidelines here https://github.com/reactwg/react-native-releases#which-versions-are-currently-supported and then a complete deletion of any documentation related to manual linking

PR #7119, opened 2023-05-15
feat(remote-config): realtime config updates

Lyokone:
### Checklist

  - [x] Yes
  - [x] `Android`
  - [x] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

---

### DIFF
--- a/packages/remote-config/android/src/reactnative/java/io/invertase/firebase/config/ReactNativeFirebaseConfigModule.java
+++ b/packages/remote-config/android/src/reactnative/java/io/invertase/firebase/config/ReactNativeFirebaseConfigModule.java
@@ -19,19 +19,44 @@ package io.invertase.firebase.config;
 
 import com.facebook.react.bridge.*;
 import com.google.firebase.FirebaseApp;
-import com.google.firebase.remoteconfig.FirebaseRemoteConfigFetchThrottledException;
+import com.google.firebase.remoteconfig.*;
+import io.invertase.firebase.common.ReactNativeFirebaseEvent;
+import io.invertase.firebase.common.ReactNativeFirebaseEventEmitter;
 import io.invertase.firebase.common.ReactNativeFirebaseModule;
+import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.Iterator;
+import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import javax.annotation.Nullable;
+import org.jetbrains.annotations.NotNull;
 
 public class ReactNativeFirebaseConfigModule extends ReactNativeFirebaseModule {
   private static final String SERVICE_NAME = "Config";
   private final UniversalFirebaseConfigModule module;
 
+  private static HashMap<String, ConfigUpdateListenerRegistration> mConfigUpdateRegistrations =
+      new HashMap<>();
+
   ReactNativeFirebaseConfigModule(ReactApplicationContext reactContext) {
     super(reactContext, SERVICE_NAME);
     module = new UniversalFirebaseConfigModule(reactContext, SERVICE_NAME);
+  }
+
+  @Override
+  public void onCatalystInstanceDestroy() {
+    super.onCatalystInstanceDestroy();
+
+    Iterator<Map.Entry<String, ConfigUpdateListenerRegistration>> configRegistrationsIterator =
+        mConfigUpdateRegistrations.entrySet().iterator();
+
+    while (configRegistrationsIterator.hasNext()) {
+      Map.Entry<String, ConfigUpdateListenerRegistration> pair = configRegistrationsIterator.next();
+      ConfigUpdateListenerRegistration mConfigRegistration = pair.getValue();
+      mConfigRegistration.remove();
+      configRegistrationsIterator.remove();
+    }
   }
 
   @ReactMethod
@@ -149,6 +174,85 @@ public class ReactNativeFirebaseConfigModule extends ReactNativeFirebaseModule {
                 rejectPromiseWithExceptionMap(promise, task.getException());
               }
             });
+  }
+
+  @ReactMethod
+  public void onConfigUpdated(String appName) {
+    if (mConfigUpdateRegistrations.get(appName) == null) {
+      FirebaseApp firebaseApp = FirebaseApp.getInstance(appName);
+      ConfigUpdateListenerRegistration registration =
+          FirebaseRemoteConfig.getInstance(firebaseApp)
+              .addOnConfigUpdateListener(
+                  new ConfigUpdateListener() {
+                    @Override
+                    public void onUpdate(@NotNull ConfigUpdate configUpdate) {
+                      ReactNativeFirebaseEventEmitter emitter =
+                          ReactNativeFirebaseEventEmitter.getSharedInstance();
+
+                      Set<String> updatedKeys = configUpdate.getUpdatedKeys();
+                      List<String> updatedKeysList = new ArrayList<>(updatedKeys);
+
+                      Map<String, Object> results = new HashMap<>();
+                      results.put("appName", appName);
+                      results.put("resultType", "success");
+                      results.put("updatedKeys", updatedKeysList);
+                      ReactNativeFirebaseEvent event =
+                          new ReactNativeFirebaseEvent(
+                              "on_config_updated", Arguments.makeNativeMap(results), appName);
+                      emitter.sendEvent(event);
+                    }
+
+                    @Override
+                    public void onError(@NotNull FirebaseRemoteConfigException error) {
+                      ReactNativeFirebaseEventEmitter emitter =
+                          ReactNativeFirebaseEventEmitter.getSharedInstance();
+
+                      WritableMap userInfoMap = Arguments.createMap();
+                      userInfoMap.putString("resultType", "error");
+                      userInfoMap.putString("appName", appName);
+
+                      FirebaseRemoteConfigException.Code code = error.getCode();
+                      switch (code) {
+                        case CONFIG_UPDATE_STREAM_ERROR:
+                          userInfoMap.putString("code", "config_update_stream_error");
+                          break;
+                        case CONFIG_UPDATE_MESSAGE_INVALID:
+                          userInfoMap.putString("code", "config_update_message_invalid");
+                          break;
+                        case CONFIG_UPDATE_NOT_FETCHED:
+                          userInfoMap.putString("code", "config_update_not_fetched");
+                          break;
+                        case CONFIG_UPDATE_UNAVAILABLE:
+                          userInfoMap.putString("code", "config_update_unavailable");
+                          break;
+                        case UNKNOWN:
+                          userInfoMap.putString("code", "unknown");
+                          break;
+                        default:
+                          // contract violation internal to the RNFB / SDK boundary
+                          userInfoMap.putString("code", "internal");
+                      }
+
+                      userInfoMap.putString("message", error.getMessage());
+                      userInfoMap.putString("nativeErrorMessage", error.getMessage());
+                      ReactNativeFirebaseEvent event =
+                          new ReactNativeFirebaseEvent("on_config_updated", userInfoMap, appName);
+                      emitter.sendEvent(event);
+                    }
+                  });
+
+      mConfigUpdateRegistrations.put(appName, registration);
+    }
+  }
+
+  @ReactMethod
+  public void removeConfigUpdateRegistration(String appName) {
+    ConfigUpdateListenerRegistration mConfigRegistration = mConfigUpdateRegistrations.get(appName);
+
+    if (mConfigRegistration != null) {
+      mConfigRegistration.remove();
+      mConfigUpdateRegistrations.remove(appName);
+    }
   }
 
   private WritableMap resultWithConstants(Object result) {
--- a/packages/remote-config/ios/RNFBConfig/RNFBConfigModule.m
+++ b/packages/remote-config/ios/RNFBConfig/RNFBConfigModule.m
@@ -22,6 +22,10 @@
 #import "RNFBConfigModule.h"
 #import "RNFBSharedUtils.h"
 
+static NSString *const ON_CONFIG_UPDATED_EVENT = @"on_config_updated";
+
+static __strong NSMutableDictionary *configUpdateHandlers;
+
 @implementation RNFBConfigModule
 #pragma mark -
 #pragma mark Converters
@@ -64,6 +68,21 @@ NSString *convertFIRRemoteConfigSourceToNSString(FIRRemoteConfigSource value) {
   }
 }
 
+NSString *convertFIRRemoteConfigUpdateErrorToNSString(FIRRemoteConfigUpdateError value) {
+  switch (value) {
+    case FIRRemoteConfigUpdateErrorStreamError:
+      return @"config_update_stream_error";
+    case FIRRemoteConfigUpdateErrorMessageInvalid:
+      return @"config_update_message_invalid";
+    case FIRRemoteConfigUpdateErrorNotFetched:
+      return @"config_update_not_fetched";
+    case FIRRemoteConfigUpdateErrorUnavailable:
+      return @"config_update_unavailable";
+    default:
+      return @"internal";
+  }
+}
+
 NSDictionary *convertFIRRemoteConfigValueToNSDictionary(FIRRemoteConfigValue *value) {
   return @{
     @"value" : (id)value.stringValue ?: [NSNull null],
@@ -82,6 +101,28 @@ RCT_EXPORT_MODULE();
 
 + (BOOL)requiresMainQueueSetup {
   return YES;
+}
+
+- (id)init {
+  self = [super init];
+  static dispatch_once_t onceToken;
+  dispatch_once(&onceToken, ^{
+    configUpdateHandlers = [[NSMutableDictionary alloc] init];
+  });
+  return self;
+}
+
+- (void)dealloc {
+  [self invalidate];
+}
+
+- (void)invalidate {
+  for (NSString *key in configUpdateHandlers) {
+    FIRConfigUpdateListenerRegistration *registration = [configUpdateHandlers objectForKey:key];
+    [registration remove];
+  }
+
+  [configUpdateHandlers removeAllObjects];
 }
 
 #pragma mark -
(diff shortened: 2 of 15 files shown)


packages/remote-config/android/src/main/java/io/invertase/firebase/config/UniversalFirebaseConfigModule.java
```diff
@@ -26,13 +26,17 @@
...
 import com.google.firebase.remoteconfig.FirebaseRemoteConfigInfo;
 import com.google.firebase.remoteconfig.FirebaseRemoteConfigSettings;
 import com.google.firebase.remoteconfig.FirebaseRemoteConfigValue;
+import com.google.firebase.remoteconfig.FirebaseRemoteConfigException;
+import com.google.firebase.remoteconfig.FirebaseRemoteConfigFetchThrottledException;
 import io.invertase.firebase.common.UniversalFirebaseModule;
-import java.util.HashMap;
-import java.util.Map;
+import com.facebook.react.bridge.*;
+import java.util.*;
```
Lyokone, line 39: Shouldn't we import each Class individually?

packages/remote-config/android/src/main/java/io/invertase/firebase/config/UniversalFirebaseConfigModule.java
```diff
@@ -53,12 +57,12 @@ Task<Void> fetch(String appName, long expirationDuration) {
...
-          FirebaseRemoteConfig config = FirebaseRemoteConfig.getInstance(firebaseApp);
-          Tasks.await(expirationDuration == -1 ? config.fetch() : config.fetch(expirationDuration));
-          return null;
-        });
+      getExecutor(),
+      () -> {
+        FirebaseRemoteConfig config = FirebaseRemoteConfig.getInstance(firebaseApp);
+        Tasks.await(expirationDuration == -1 ? config.fetch() : config.fetch(expirationDuration));
+        return null;
+      });
```
Lyokone, line 65: Are you using the same formatter as the project?

packages/remote-config/e2e/config.modular.e2e.js
```diff
@@ -743,5 +749,14 @@ describe('remoteConfig() modular', function () {
         should(supported).equal(true);
       });
     });
+
+    describe('onConfigUpdated', function () {
+      it('onConfigUpdated can run without an issue', async function () {
+        const { getRemoteConfig, onConfigUpdated } = remoteConfigModular;
+        const remoteConfig = getRemoteConfig();
+
+        onConfigUpdated(remoteConfig, function (event, error) {});
```
Lyokone, line 758: I think this test need to be completed

packages/remote-config/ios/RNFBConfig/RNFBConfigModule.m
```diff
@@ -223,6 +223,31 @@ + (BOOL)requiresMainQueueSetup {
...
+        }
+
+        callback(@[
+            @{
+                @"updatedKeys" : [configUpdate.updatedKeys allObjects],
+            },
+            userInfo != nil ? userInfo : [NSNull null],
+        ]);
+    }];
+}
```
Lyokone, line 249: Is there any way to remove the listener once Javascript has stopped listening?

packages/remote-config/android/src/main/java/io/invertase/firebase/config/UniversalFirebaseConfigModule.java
```diff
@@ -143,9 +147,53 @@ Task<FirebaseRemoteConfigInfo> ensureInitialized(String appName) {
     return ensureInitializedTask;
   }
 
+  void onConfigUpdated(String appName, Callback callback) {
+    FirebaseApp firebaseApp = FirebaseApp.getInstance(appName);
+
+    FirebaseRemoteConfig.getInstance(firebaseApp).addOnConfigUpdateListener(new ConfigUpdateListener() {
```
russellwheatley, line 153: This event listener should probably be tracked like we do with auth [here](https://github.com/invertase/react-native-firebase/blob/main/packages/auth/android/src/main/java/io/invertase/firebase/auth/ReactNativeFirebaseAuthModule.java#L96) and removed using [this overridden method](https://github.com/invertase/react-native-firebase/blob/main/packages/auth/android/src/main/java/io/invertase/firebase/auth/ReactNativeFirebaseAuthModule.java#L117). This means you will have to move it to the `ReactNativeFirebaseConfigModule`.  Remote Config has multi-app support so you will need to track them on a per Firebase app basis so the implementation shouldn't be too far removed from how we implement the authStateListener [here](https://github.com/invertase/react-native-firebase/blob/main/packages/auth/android/src/main/java/io/invertase/firebase/auth/ReactNativeFirebaseAuthModule.java#L152-L181). i.e. use the app name as a key for the listener.

packages/remote-config/ios/RNFBConfig/RNFBConfigModule.m
```diff
@@ -223,6 +223,31 @@ + (BOOL)requiresMainQueueSetup {
   }
 }
 
+RCT_EXPORT_METHOD(onConfigUpdated
+                  : (FIRApp *)firebaseApp
+                  : (RCTResponseSenderBlock)callback) {
+    [[FIRRemoteConfig remoteConfigWithApp:firebaseApp] addOnConfigUpdateListener:^( FIRRemoteConfigUpdate * _Nonnull configUpdate, NSError * _Nullable error) {
```
russellwheatley, line 229: as mentioned above for android will also be required for iOS.

russellwheatley (changes requested):

packages/remote-config/e2e/config.modular.e2e.js
exaby73: I couldn't think of a test here. Followed tests/integration_test/firebase_remote_config/firebase_remote_config_e2e_test.dart in this case. In the comment there, you mentioned that it's not possible to change values on the fly to test the callback. Any ideas on how to improve this test?

packages/remote-config/android/src/main/java/io/invertase/firebase/config/UniversalFirebaseConfigModule.java
exaby73: I'm using IntelliJ, and it formatted the code like this (force of habit to hit the reformat code keybind). I'm not familiar with Java formatters. Could you point me to a command I can run to fix this formatting or should I manually revert the formatting changes?

packages/remote-config/android/src/main/java/io/invertase/firebase/config/UniversalFirebaseConfigModule.java
exaby73: Done :)

packages/remote-config/android/src/main/java/io/invertase/firebase/config/UniversalFirebaseConfigModule.java
russellwheatley: That's a good question actually, I can't see a script in the `package.json` for formatting android. What is the correct way @mikehardy?

packages/remote-config/android/src/reactnative/java/io/invertase/firebase/config/ReactNativeFirebaseConfigModule.java
```diff
@@ -17,23 +17,54 @@
...
 
+  @Override
+  public void onCatalystInstanceDestroy() {
+    super.onCatalystInstanceDestroy();
+
+    Iterator configRegistrationsIterator = mConfigUpdateRegistrations.entrySet().iterator();
+
+    while (configRegistrationsIterator.hasNext()) {
+      Map.Entry pair = (Map.Entry) configRegistrationsIterator.next();
+      String appName = (String) pair.getKey();
```
russellwheatley, line 60: this isn't used. Can remove.

packages/remote-config/ios/RNFBConfig/RNFBConfigModule.m
```diff
@@ -223,6 +247,41 @@ + (BOOL)requiresMainQueueSetup {
...
+            [userInfo setValue:@"unknown" forKey:@"code"];
+            [userInfo setValue:error.localizedDescription forKey:@"message"];
+            [userInfo setValue:@(error.code) forKey:@"nativeErrorCode"];
+            [userInfo setValue:error.localizedDescription forKey:@"nativeErrorMessage"];
+        }
+
+        callback(@[
+            @{
+                @"updatedKeys" : [configUpdate.updatedKeys allObjects],
+            },
```
russellwheatley, line 269: if there is an error, I'm not sure we should be returning the first value, it should probably be `[NSNull null]`.

packages/remote-config/android/src/reactnative/java/io/invertase/firebase/config/ReactNativeFirebaseConfigModule.java
exaby73: This code is to remove all listeners on destroy, right? This is how auth is doing it

packages/remote-config/ios/RNFBConfig/RNFBConfigModule.m
```diff
@@ -223,6 +247,41 @@ + (BOOL)requiresMainQueueSetup {
...
+                  : (FIRApp *)firebaseApp
+                  : (RCTResponseSenderBlock)callback) {
+    FIRConfigUpdateListenerRegistration *newRegistration = [[FIRRemoteConfig remoteConfigWithApp:firebaseApp] addOnConfigUpdateListener:^( FIRRemoteConfigUpdate * _Nonnull configUpdate, NSError * _Nullable error) {
+        NSMutableDictionary *userInfo;
+
+        if (error != nil) {
+            userInfo = [NSMutableDictionary dictionary];
+
+            [userInfo setValue:@(NO) forKey:@"fatal"];
+            [userInfo setValue:@"unknown" forKey:@"code"];
```
russellwheatley, line 260: We should probably try to propagate the original error code `error.code` before making it defacto `unknown`.

packages/remote-config/android/src/main/java/io/invertase/firebase/config/UniversalFirebaseConfigModule.java
```diff
@@ -143,9 +155,53 @@ Task<FirebaseRemoteConfigInfo> ensureInitialized(String appName) {
...
+        results.put("updatedKeys", updatedKeysList);
+        callback.invoke(Arguments.makeNativeMap(results), null);
+      }
+
+      @Override
+      public void onError(FirebaseRemoteConfigException error) {
+        WritableMap userInfoMap = Arguments.createMap();
+
+        if (error == null) {
+          userInfoMap.putString("code", "unknown");
```
russellwheatley, line 177: same deal here, try to find the error code first before making it `unknown`.

packages/remote-config/android/src/main/java/io/invertase/firebase/config/UniversalFirebaseConfigModule.java
```diff
@@ -143,9 +155,53 @@ Task<FirebaseRemoteConfigInfo> ensureInitialized(String appName) {
...
+        if (error.getCause() instanceof FirebaseRemoteConfigFetchThrottledException) {
+          userInfoMap.putString("code", "throttled");
+          userInfoMap.putString(
+            "message",
+            "fetch() operation cannot be completed successfully, due to throttling.");
+        } else {
+          userInfoMap.putString("code", "failure");
+          userInfoMap.putString(
+            "message",
+            "fetch() operation cannot be completed successfully.");
```
russellwheatley, line 194: use `error.getMessage()` here rather than hardcoded value like you do with iOS.

packages/remote-config/android/src/main/java/io/invertase/firebase/config/UniversalFirebaseConfigModule.java
```diff
@@ -143,9 +155,53 @@ Task<FirebaseRemoteConfigInfo> ensureInitialized(String appName) {
...
+        Map<String, Object> results = new HashMap<>();
+        results.put("updatedKeys", updatedKeysList);
+        callback.invoke(Arguments.makeNativeMap(results), null);
+      }
+
+      @Override
+      public void onError(FirebaseRemoteConfigException error) {
+        WritableMap userInfoMap = Arguments.createMap();
+
+        if (error == null) {
```
russellwheatley, line 176: I don't think this can be `null`. Remove.

packages/remote-config/android/src/main/java/io/invertase/firebase/config/UniversalFirebaseConfigModule.java
```diff
@@ -143,9 +155,53 @@ Task<FirebaseRemoteConfigInfo> ensureInitialized(String appName) {
...
+          userInfoMap.putString("code", "throttled");
+          userInfoMap.putString(
+            "message",
+            "fetch() operation cannot be completed successfully, due to throttling.");
+        } else {
+          userInfoMap.putString("code", "failure");
+          userInfoMap.putString(
+            "message",
+            "fetch() operation cannot be completed successfully.");
+        }
```
russellwheatley, line 195: I think remove all this and propagate error as it is. Here are the docs for `FirebaseRemoteConfigException`. https://firebase.google.com/docs/reference/kotlin/com/google/firebase/remoteconfig/FirebaseRemoteConfigException

You need to use `@NonNull` annotations on this parameter like this:
```
public void onError(@NonNull FirebaseRemoteConfigException error) {
```
Are you using android studio for android dev? It will flag when you need to use it.

packages/remote-config/ios/RNFBConfig/RNFBConfigModule.m
exaby73: Fixed this

russellwheatley (changes requested):
some updates required but looking good 😎

packages/remote-config/android/src/reactnative/java/io/invertase/firebase/config/ReactNativeFirebaseConfigModule.java
russellwheatley: where is appName used?

russellwheatley: If you use android studio, it will tell you when variables aren't used:
Screenshot 2023-05-18 at 11 14 44

exaby73: Ahh sorry didn't see the appName thingy. Removed it :)

packages/remote-config/android/src/main/java/io/invertase/firebase/config/UniversalFirebaseConfigModule.java
exaby73: Done

packages/remote-config/android/src/main/java/io/invertase/firebase/config/UniversalFirebaseConfigModule.java
mikehardy: We (that is, myself on behalf of Invertase) repackage https://github.com/google/google-java-format in https://www.npmjs.com/package/google-java-format so we have autoformatting, and it is available here: https://github.com/invertase/react-native-firebase/blob/a58ea4bfd72c903d43844da30d4f3bf5feb0f057/package.json#L11

If you run that command locally until it finally succeeds (it may take quite a few runs, it is a bit unreliable unfortunately but it will always eventually run to completion) then all your files will be formatted locally to our standard

packages/remote-config/e2e/config.modular.e2e.js
mikehardy: I think the tests are inadequate and I'm working on a test style that uses a cloud function I will deploy (I'm working on this currently, nearly ready).

The cloud function will have access to the admin SDK which allows it to read/write the remote-config template, and will have an API that lets you add a remote config template key, remove a remote config template key, and update a remote config template key value as a list of 0 or more operations that it will do at once.

So then in the e2e tests I expect this to happen:

- single listener addition/removal case:
  - we test updating a known remote-config template key value while there are no listeners, no listeners are called
  - we add a listener, do same update, listener should be called
  - we remove the listener, do same update, no listener should be called

- update type verification
  - make sure key addition is handled correctly
  - same with key removal
  - same with key update of all update types

- react-native reload verification
  - add a listener, verify it is called when you update the remote-config template
  - do a react-native hot reload, make sure the listener is removed down to native level

- multiple listener addition/removal accounting verification
  - add one listener, make sure it is called when template updates
  - add second listener, make sure both are called when template updates
  - remove first listener, make sure only second listener is called
  - add third/fourth listener, make sure 2nd/3rd/4th are called
  - remove 3rd listener, make sure 2nd/4th are called
  - remove all listeners, make sure none are called


All of this is predicated on the ability to trigger the listener, which is based on the assumption we'll have a function+admin-sdk deployed on the project as an enabler - once that's up, the ability to e2e test this thing should be pretty thorough

packages/remote-config/android/src/main/java/io/invertase/firebase/config/UniversalFirebaseConfigModule.java
exaby73: Following @mikehardy's advice, I ran the formatter. It did change a few files, but these I think are now in sync with our code style

packages/remote-config/lib/index.js
```diff
@@ -282,6 +289,45 @@ class FirebaseConfigModule extends FirebaseModule {
...
+          Promise.resolve().then(() => {
+            listener(event, undefined);
+          });
+          return;
+        }
+
+        Promise.resolve().then(() => {
+          listener(undefined, event);
+        });
+      },
```
russellwheatley, line 316: I think this can be removed now, this is specific to auth logic

TaylorDale:
Have been looking for this... thanks for the work 😎

tests/app.js
```diff
@@ -40,9 +40,9 @@ import * as remoteConfigModular from '@react-native-firebase/remote-config';
 import '@react-native-firebase/storage';
 import * as storageModular from '@react-native-firebase/storage';
 import jet from 'jet/platform/react-native';
-import React from 'react';
 import { AppRegistry, Button, NativeModules, Text, View } from 'react-native';
 import DeviceInfo from 'react-native-device-info';
+import React from 'react';
```
mikehardy, line 45: Please avoid unnecessary changes in PRs, this file would not be touched at all except for this change, which is not related to the PR, so it just adds noise to the diff / increases review burden

mikehardy (commented):
I've completed implementation of test scaffolding for this feature on branch https://github.com/invertase/react-native-firebase/tree/%40mikehardy/remote-config-realtime-updates

You can use the `updateTemplate` function from `helpers.js` to execute a batch of delete/update/add operations against the existing template, which will enable listener testing to verify this works.

We need to coordinate how to integrate the work, but in this PR I notice a few things:

- there is no update to the index.d.ts file so there are no typescript typings, what exactly is sent to the listener when the config is updated?
- I do not see complete support for multiple listeners? it appears there is a one-to-one limitation between app and listener, but I think you can have more than one listener per app? Is that an intentional constraint on the implementation to limit it to one listener per app?
- I do not see any documentation of the feature? (I have started some on my branch and was expecting a conflict here, but the docs are not updated here...)
- as noted above - I do not see testing

My branch has a sketch in the e2e file of the testing I expect to see to validate the feature

If you would like to continue working on it, you can perhaps rebase this branch on to my branch (just dropping the Podfile.lock commit, most likely...) and then start working through the e2e test file with the idea that your implementation should then be exercised

Follow that up with typescript typings and making sure the docs match the API you've chosen and we should be able to ship

What do you think?

tests/app.js
exaby73: Done

packages/remote-config/lib/index.js
exaby73: Done

mikehardy:
I unpicked the merge commits which made a hash of the commit trail, rebased main back into the branch (vs merge), squashed the remaining changes on this branch into one change (since it is one feature addition) and re-pushed.

Do *not* merge into this branch, rather re-fetch it - `git fetch origin && git reset --hard origin/feat/remote-config-onconfigupdated`

I'm working on now on listener changes and automating the testing of the listener using the cloud function implemented earlier, will re-push when those are stable locally.

mikehardy:
single listener test is more less stable now and definitely tests the back-and-forth of the listener triggering via use of the cloud function to update the config in realtime during the e2e test.

It is yielding some unexpected results that bear examination though - the full set of keys in the remote config are coming back as updated each time at the moment, which does not match with upstream docs

the branch has also been rebased to include #7180 so it is up to date with main

mikehardy:
single listener e2e test now definitely appears to be stable and types documentation is updated in an attempt to explain the strong need to activate the new config after the callback is called in order to get results that make sense

last step is to handle multiple listeners in an efficient manner by de-subscribing the native listener when no more javascript listeners exist. will have to get that tomorrow morning my time (~12 hrs from now) then we can shipit

mikehardy:
This works in e2e now with multiple listeners working and with optimized native subscribe/unsubscribe to make sure we don't open a persistent web socket unless necessary.

However, there is a problem locally for me where my android emulator is no longer receiving realtime updates at all and I need to determine why prior to cleaning up the commit chain and releasing this. Apologies for the delay but it's close!

mikehardy:
I'm getting unexpected behavior from the underlying firebase-ios-sdk with regard to removing listeners and have opened an issue related: https://github.com/firebase/firebase-ios-sdk/issues/11458

Still working through why android is completely failing now

mikehardy:
Need to clean the commit trail up but I believe the final form of the PR is present now, including a workaround for the firebase-ios-sdk item I uncovered

mikehardy:
firebase-ios-sdk appears to have something else going on where it receives a "template was updated" message on it's websocket but then is not successful at fetching the new remote config template. Android does not show this behavior so I have asked upstream for guidance: https://github.com/firebase/firebase-ios-sdk/issues/11462

The feature will appear pretty flaky on iOS until a reliable method to handle it is implemented, if our e2e CI is any indicator.

mikehardy:
commit stream cleaned up, added a trailing commit that turns on SDK logging while investigating ios template fetch failure

tests are clamped to just realtime remoteconfig tests until the above issue is sorted

ready to go other than that

PR #7123, opened 2023-05-18
docs: fix broken links

ashuvssut:


docs/auth/usage/index.md
```diff
@@ -130,7 +130,7 @@ Once successfully signed in, any [`onAuthStateChanged`](#listening-to-authentica
 with the [`User`](/reference/auth/user) details.
 
 In case any errors occur, the module provides support for identifying what specifically went wrong by attaching a code
-to the error. For a full list of error codes available, view the [Firebase documentation](https://firebase.google.com/docs/reference/js/firebase.auth.Auth#error-codes_9).
+to the error. For a full list of error codes available, view the [Firebase documentation](https://firebase.google.com/docs/reference/js/v8/firebase.auth.Auth#error-codes_9).
```
mikehardy, line 133: Prefer pointing to current URL vs the v8 one

```suggestion
to the error. For a full list of error codes available, view the [Firebase documentation](https://firebase.google.com/docs/reference/js/auth.md#autherrorcodes).
```

mikehardy (changes requested):
Thank you for catching this! We do need to repoint the URL but I'm going to make two code suggestions that I will commit where I point them to the new URL vs the v8 one which will eventually go away

mikehardy (approved):
ok - pointing to the new URL this works for me, thanks again for catching it!

PR #7124, opened 2023-05-18
fix(firestore, types): add types for Filter constraints on Queries

falleco:
### Description

When using the firestore document queries, typescript types seem to ignore the Filter object. Filter is already supported and well documented at https://rnfirebase.io/firestore/usage#filtering.

The documentation example does not compile when using typescript:

```ts
const snapshot = await firestore()
  .collection('Users')
  .where(Filter('user', '==', 'Tim'))
  .where('email', '==', 'tim@example.com')
  .get();
```

### Release Summary

Adding type support for Filter constraints on document queries.

### Checklist

  - [X] Yes
- This is a breaking change;
  - [ ] Yes
  - [X] No

### DIFF
--- a/packages/firestore/lib/index.d.ts
+++ b/packages/firestore/lib/index.d.ts
@@ -1395,6 +1395,24 @@ export namespace FirebaseFirestoreTypes {
      * @param value The comparison value.
      */
     where(fieldPath: keyof T | FieldPath, opStr: WhereFilterOp, value: any): Query<T>;
+
+    /**
+     * Creates and returns a new Query with the additional filter that documents must contain the specified field and
+     * the value should satisfy the relation constraint provided.
+     *
+     * #### Example
+     *
+     * ```js
+     * // Get all users who's age is 30 or above
+     * const querySnapshot = await firebase.firestore()
+     *   .collection('users')
+     *   .where(Filter('age', '>=', 30));
+     *   .get();
+     * ```
+     *
+     * @param filter The filter to apply to the query.
+     */
+    where(filter: FilterFunction): Query<T>;
   }
 
   /**
--- a/packages/firestore/lib/index.d.ts
+++ b/packages/firestore/lib/index.d.ts
@@ -1412,7 +1412,7 @@ export namespace FirebaseFirestoreTypes {
      *
      * @param filter The filter to apply to the query.
      */
-    where(filter: FilterFunction): Query<T>;
+    where(filter: QueryFilterConstraint | QueryCompositeFilterConstraint): Query<T>;
   }
 
   /**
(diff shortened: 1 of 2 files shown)


packages/firestore/lib/index.d.ts
```diff
@@ -2317,6 +2335,8 @@ export const firebase: ReactNativeFirebase.Module & {
   ): ReactNativeFirebase.FirebaseApp & { firestore(): FirebaseFirestoreTypes.Module };
 };
 
+export const Filter: FilterFunction;
```
pafry7, line 2338: I think it should be added inside `Statics` interface:
```
 export interface Statics {
    /**
     * Returns the `Blob` class.
     */
    Blob: typeof Blob;

    /**
     * Returns the `FieldPath` class.
     */
    FieldPath: typeof FieldPath;

    /**
     * Returns the `FieldValue` class.
     */
    FieldValue: typeof FieldValue;

    /**
     * Returns the `GeoPoint` class.
     */
    GeoPoint: typeof GeoPoint;

    /**
     * Returns the `Timestamp` class.
     */
    Timestamp: typeof Timestamp;

    /**
     * Returns the `Filter` function.
     */
    Filter: typeof Filter

    /**
     * Used to set the cache size to unlimited when passing to `cacheSizeBytes` in
     * `firebase.firestore().settings()`.
     */
    CACHE_SIZE_UNLIMITED: number;
 }
```

packages/firestore/lib/index.js
```diff
@@ -33,6 +33,7 @@ import FirestoreCollectionReference from './FirestoreCollectionReference';
 import FirestoreDocumentReference from './FirestoreDocumentReference';
 import FirestorePath from './FirestorePath';
 import FirestoreQuery from './FirestoreQuery';
+export { Filter } from './FirestoreFilter';
```
pafry7, line 36: Probably not needed, as it's already in `FirestoreStatics`

falleco: nice! I've removed the explicit import! Thanks!

packages/firestore/lib/index.d.ts
falleco: Added!

mikehardy (approved):
This looks great to me, I can't tell you how much I appreciate it / how much it helps that you (@falleco) proposed a change and then @pafry7 stepped in to help review. I probably would not have been able to get to it for a while in the absence of your efforts, but now I can just scan it + approve it + likely merge it in just a few minutes of effort. ❤️ 

I will let it sit for a couple hours just in case anyone from the linked documentation issue has a moment to scan the current state of the PR to make sure it's going to work as expected before merge

PR #7142, opened 2023-05-30
fix(firestore): Allow queries with combined in and array-contains-any

ibobo:
### Description

Since late april queries containing `in` and `array-contains-any` together are supported as queries with multiple `in` statements, given that the number of generated `or` queries in normal form is below 30. Since `@react-native-firebase/firestore` is currently blocking any of these queries, this PR aims at removing that block, leaving the check for the normal form limit to the underlying firestore engine.

This is the relevant firebase docs on these limitations: https://firebase.google.com/docs/firestore/query-data/queries#limitations_3

Please note: I do not have a history of the changes available, I know this change has been added with the `or` operator support.

### Related issues

None

### Release Summary

Allow the use of `in` and `array-contains-any` in the same query.

### Checklist

  - [x] Yes
  - [x] `Android`
  - [x] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

I ran the e2e tests, these are the results:

```
  1679 passing (3m)
  116 pending
```

Apart from running e2e tests I did use my fork in my own project using the aforementioned `in` and `array-contains-any` filters in some queries, and they work fine.

:fire:

### DIFF
--- a/packages/firestore/e2e/Query/where.and.filter.e2e.js
+++ b/packages/firestore/e2e/Query/where.and.filter.e2e.js
@@ -245,24 +245,6 @@ describe(' firestore().collection().where(AND Filters)', function () {
     }
   });
 
-  it('throws if query has in & array-contains-any filter', function () {
-    try {
-      firebase
-        .firestore()
-        .collection(COLLECTION)
-        .where(
-          Filter.and(Filter('foo.bar', 'in', [1]), Filter('foo.bar', 'array-contains-any', [2])),
-        );
-
-      return Promise.reject(new Error('Did not throw an Error.'));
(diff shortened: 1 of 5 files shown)


mikehardy (approved):
Fascinating - you are exactly right! This was changed recently and makes the query language much more powerful, *assuming* we let them through our filtering :-). Thanks for taking the time to fix this not just for yourself but for all users, much appreciated!

mikehardy:
The iOS CI failure is a timeout I'm fixing in #7163 and does not affect this PR - will merge this after that one passes CI so it doesn't fail on main as well post-merge

PR #7249, opened 2023-07-18
fix: Android RN 72 Support - Update build.gradle w/ Namespace

alexhernandez:
RN Team Update:
https://github.com/react-native-community/discussions-and-proposals/issues/671

Fix:
https://github.com/react-native-community/discussions-and-proposals/issues/671#issuecomment-1598892444

### DIFF
--- a/packages/app/android/build.gradle
+++ b/packages/app/android/build.gradle
@@ -59,6 +59,11 @@ if (rootProject.ext && rootProject.ext.firebaseJson) {
 }
 
 android {
+  def agpVersion = com.android.Version.ANDROID_GRADLE_PLUGIN_VERSION
+  if (agpVersion.tokenize('.')[0].toInteger() >= 7) {
+    namespace "io.invertase.firebase"
+  }
+  
   defaultConfig {
     multiDexEnabled true
     manifestPlaceholders = [


mikehardy (changes requested):
Oh yes! I've seen this a bit! This is also the current preferred solution I believe (it is what I've accepted in other repos I maintain anyway) - conditionally adding it to build.gradle so it is forward-compatible but *not* removing it from AndroidManifest.xml because that is sort of fiddly and makes it backwards-incompatible

Anyway - I think we have to do this in the build.gradle of every package though, not just app? That is, analytics, app-check, app-distribution, database etc etc - so there should be a large-ish batch of build.gradle diffs with the same change, one perpackage?

What do you think?

mikehardy:
This idea was taken up by @AndreiCalazans in #7323 and we've merged it and released it there. Thanks for raising this PR!

ivanignatiev:
For RN 0.73 for

- react-native-firebase_analytics
- react-native-firebase_messaging

You also need to add `android.buildFeatures.buildConfig true` otherwise it will fail:

```
FAILURE: Build failed with an exception.

* What went wrong:
A problem occurred configuring project ':react-native-firebase_analytics'.
> com.android.builder.errors.EvalIssueException: defaultConfig contains custom BuildConfig fields, but the feature is disabled.
  To enable the feature, add the following to your module-level build.gradle:
  `android.buildFeatures.buildConfig true`

BUILD FAILED in 4s
error Failed to install the app.
```

PR #7309, opened 2023-08-22
feat(app-check, android): Implement app check token change listener

MrLibya:
### Description

Implement app check token change listener only for android ( as the sdk don't support for iOS )

`onTokenChanged` function will be trigger with object contain `token, expireTimeMillis, appName`

### Checklist

  - [x] Yes
  - [x] `Android`
- This is a breaking change;
  - [x] No

### Test Plan

```js
  const onAppCheckTokenChanges = (result) => {
    console.log("onAppCheckTokenChanges: ",result)
  }

  React.useEffect(() => {
    const subscriber = appCheck().onTokenChanged(onAppCheckTokenChanges)

    return subscriber
  }, [])
```

### DIFF
--- a/packages/app-check/lib/index.d.ts
+++ b/packages/app-check/lib/index.d.ts
@@ -180,7 +180,7 @@ export namespace FirebaseAppCheckTypes {
   /**
    * The result return from `onTokenChanged`
    */
-  export type AppCheckListenerResult = AppCheckToken & { readonly appName: String }
+  export type AppCheckListenerResult = AppCheckToken & { readonly appName: string };
 
   // eslint-disable-next-line @typescript-eslint/no-empty-interface
   export interface Statics {
--- a/packages/app-check/lib/index.d.ts
+++ b/packages/app-check/lib/index.d.ts
@@ -88,6 +88,18 @@ export namespace FirebaseAppCheckTypes {
     isTokenAutoRefreshEnabled?: boolean;
   }
 
+  export type NextFn<T> = (value: T) => void;
+  export type ErrorFn = (error: Error) => void;
+  export type CompleteFn = () => void;
+
+  export interface Observer<T> {
+    next: NextFn<T>;
+    error: ErrorFn;
+    complete: CompleteFn;
+  }
+
+  export type PartialObserver<T> = Partial<Observer<T>>;
+
   export interface ReactNativeFirebaseAppCheckProviderOptions {
     /**
      * debug token to use, if any. Defaults to undefined, pre-configure tokens in firebase web console if needed
@@ -165,6 +177,10 @@ export namespace FirebaseAppCheckTypes {
      */
     readonly expireTimeMillis: number;
   }
+  /**
+   * The result return from `onTokenChanged`
+   */
+  export type AppCheckListenerResult = AppCheckToken & { readonly appName: String }
 
   // eslint-disable-next-line @typescript-eslint/no-empty-interface
   export interface Statics {
@@ -258,12 +274,9 @@ export namespace FirebaseAppCheckTypes {
      *
      * @returns A function that unsubscribes this listener.
      */
-    // TODO there is a great deal of Observer / PartialObserver typing to carry-in
-    // onTokenChanged(observer: PartialObserver<AppCheckTokenResult>): () => void;
+    onTokenChanged(observer: PartialObserver<AppCheckListenerResult>): () => void;
 
     /**
-     * TODO implement token listener for android.
-     *
      * Registers a listener to changes in the token state. There can be more
      * than one listener registered at the same time for one or more
      * App Check instances. The listeners call back on the UI thread whenever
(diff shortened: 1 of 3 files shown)


MrLibya:
@mikehardy @InvertaseBot Is an anything I need todo to get reviewed ?

packages/app-check/lib/index.d.ts
```diff
@@ -259,7 +261,7 @@ export namespace FirebaseAppCheckTypes {
      * @returns A function that unsubscribes this listener.
      */
     // TODO there is a great deal of Observer / PartialObserver typing to carry-in
-    // onTokenChanged(observer: PartialObserver<AppCheckTokenResult>): () => void;
+    onTokenChanged(observer: AppCheckListenerCallback): () => void;
```
mikehardy, line 264: 🤔 we must follow the upstream typings here, they are our source of truth for typing reference

https://github.com/firebase/firebase-js-sdk/blob/25cda8af6a6fc15e33f0ce5644dd29d996e4716f/packages/app-check-types/index.d.ts#L69

```typescript
   onTokenChanged(observer: PartialObserver<AppCheckTokenResult>): Unsubscribe;
```

packages/app-check/lib/index.js
```diff
@@ -40,6 +40,16 @@ const namespace = 'appCheck';
...
 
 class FirebaseAppCheckModule extends FirebaseModule {
+  constructor(...args) {
+    super(...args);
+
+    this.emitter.addListener(this.eventNameForApp('appCheck_token_changed'), event => {
+      this.emitter.emit(this.eventNameForApp('onAppCheckTokenChanged'), event);
+    });
+
+    this.native.addAppCheckListener();
```
mikehardy, line 50: Interesting - this adds a listener even if nothing is listening, but similar to remote-config's `onConfigUpdated` functionality, the upstream firebase-js-sdk docs indicate that an app may register zero or more listeners for a single app and they will all be called for the event

We can do that at the javascript level by allowing multiple listeners per app and just keeping track of how many listeners we have

So this should maintain a count of listeners at the javascript level, only subscribe to the native level if adding the first javascript level listener, and only remove the native listener from the native level if removing the last listener, similar to remote-config

https://github.com/invertase/react-native-firebase/blob/ea38a20988fbbb030d1a28d1ae83f627b47d615d/packages/remote-config/lib/index.js#L312-L331

mikehardy (changes requested):
Nice! The native implementation looks great - I especially like that you correctly handled the destruction of the catalyst instance ✅ 

At the javascript level it looks like we need a little bit of change here in order to comply with the firebase-js-sdk app-check contract, where they allow for multiple listeners for the same app, and their typings are a  bit different. Here in react-native-firebase we must be a drop-in replacement for firebase-js-sdk where ever possible and I think in this case we can meet their behavior (multiple listeners) and typings,. What do you think?

packages/app-check/lib/index.d.ts
MrLibya: Ok I'll update code to match the js sdk.

packages/app-check/lib/index.js
mikehardy: Hmm - well one of the wonderful things about mirroring firebase-js-sdk behavior is that we don't have to make judgement calls about whether something is useful or not. Our criteria is fidelity - we need to behave the *same* as firebase-js-sdk, regardless of opinion on the functionality.

So our job is really just to make sure we understand the functionality, then we implement it. Here is what I read in the related docs:

> Registers a listener to changes in the token state. There can be more than one listener registered at the same time for one or more App Check instances. The listeners call back on the UI thread whenever the current token associated with this App Check instance changes.

ref: https://firebase.google.com/docs/reference/js/app-check.md#ontokenchanged

I read that - if I reduce it slightly but I believe in keeping with the logical conditions: "there can be more than one listener registered at the same time for one App Check instances". I believe that is still logically consistent because one branch of an or clause may be eliminated without altering the meaning, if the remaining clauses return the same value ("one instance" === true).

So I believe we need to support multiple listeners for one app. Perhaps someone has a use case they identified while designing the API? It would be similar to the remote-config listener API they implemented, I believe they intend to support multiple listeners for the same app.

Then concretely, they specifically test that multiple listeners for the same app work:

https://github.com/firebase/firebase-js-sdk/blob/2d0a9f5fd921568bc76fcdee325b9ab5e6be8a58/packages/app-check/src/api.test.ts#L336-L337

So in order to have fidelity with the firebase-js-sdk API and continue our intended status as a drop-in replacement, we need to handle it as well

The remote-config onConfigUpdated implementation has the useful bits of code to implement it I believe and should be pretty easy to do here ?

mikehardy (dismissed):
This looks really good, thanks for handling the comments, and so quickly.
I think lint is going to be a bit annoyed at some of the formatting but I can auto-fix that and re-push to get this through + released, I'm doing a bunch of PRs today so it fits in my flow
Will merge as soon as CI goes green

🏆 !

mikehardy (approved):

PR #7362, opened 2023-09-21
feat: Adding support for Firebase Messaging via Expo config plugin.

MonchiLin:
### Description

This PR adds support for Firebase Messaging in Expo projects via a config plugin. This is mainly achieved by setting the appropriate keys (`com.google.firebase.messaging.default_notification_icon` and `com.google.firebase.messaging.default_notification_color`) in the AndroidManifest of the Android project during the build stage. The addition of this plugin will automatically enable Firebase Messaging support for developers using `@react-native-firebase/messaging` in their Expo projects which have the necessary notification configuration.

Displaying notifications correctly on Android requires setting a pure white icon. By default, the system uses the app's own icon as the notification icon, which usually isn't purely white. Therefore, it's necessary to specify a notification icon. If no notification configuration is detected from the Expo configuration, a friendly warning will be printed.

### Release Summary

Added support for Firebase Messaging in Expo projects via a config plugin.

### Checklist

  - [x] Yes
  - [x] `Android`
  - [ ] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

I created the repo to test the feature.

https://github.com/MonchiLin/firebase-messaging-test-app

### DIFF
--- a/packages/app/plugin/__tests__/androidPlugin.test.ts
+++ b/packages/app/plugin/__tests__/androidPlugin.test.ts
@@ -1,9 +1,14 @@
 import fs from 'fs/promises';
 import path from 'path';
-import { beforeAll, describe, expect, it } from '@jest/globals';
+import { beforeAll, describe, expect, it, jest } from '@jest/globals';
 
 import { applyPlugin } from '../src/android/applyPlugin';
 import { setBuildscriptDependency } from '../src/android/buildscriptDependency';
+import { setFireBaseMessagingAndroidManifest } from '../src/android/setupFirebaseNotifationIcon';
+import { ExpoConfig } from '@expo/config-types';
+import expoConfigExample from './fixtures/expo-config-example';
+import manifestApplicationExample from './fixtures/application-example';
+import { ManifestApplication } from '@expo/config-plugins/build/android/Manifest';
 
 describe('Config Plugin Android Tests', function () {
   let appBuildGradle: string;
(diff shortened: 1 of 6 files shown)


mikehardy (commented):
Thanks for posting this!

A couple questions - is it possible to go with a no-code solution wherein the standard Expo plugins for adding elements to AndroidManifest.xml are used? I assume those exist? Code here (especially Expo-related code which evolves rapidly) adds a non-trivial maintenance burden so I'd prefer a docs solution vs a code solution

If not, then 2) shouldn't this be in the messaging area? (as opposed to app)

MonchiLin:
> Thanks for posting this!
> 
> A couple questions - is it possible to go with a no-code solution wherein the standard Expo plugins for adding elements to AndroidManifest.xml are used? I assume those exist? Code here (especially Expo-related code which evolves rapidly) adds a non-trivial maintenance burden so I'd prefer a docs solution vs a code solution
> 
> If not, then 2) shouldn't this be in the messaging area? (as opposed to app)

I. To my knowledge, if you do not use [Expo bare](https://docs.expo.dev/bare/overview/), you can only modify the AndroidManifest.xml through Expo plugins.

I've checked [BuildProperties](https://docs.expo.dev/versions/latest/sdk/build-properties/), and it does not provide a way to modify AndroidManifest.xml.

II. You made a good point, and an idea I've considered. There are two reasons why I placed this part of the code in the Messaging module:

1. I noticed that the [Getting Started](https://rnfirebase.io/) section of the documentation is the only part that mentions how to use it with Expo. I think if I add it to the Messaging module, developers who don't notice the documentation on how Messaging works with Expo might miss this information. (Yes, I would also add a PR for Docs if the code could be merged).

2. The current way to use Expo plugins is as follows:

```json
{
  "expo": {
    "android": {
      "googleServicesFile": "./google-services.json"
    },
    "ios": {
      "googleServicesFile": "./GoogleService-Info.plist"
    },
    "plugins": [
      "@react-native-firebase/app",
      "@react-native-firebase/perf",
      "@react-native-firebase/crashlytics"
    ]
  }
}
```

If it were added to the Messaging module, it would become:

```json
{
  "expo": {
    "android": {
      "googleServicesFile": "./google-services.json"
    },
    "ios": {
      "googleServicesFile": "./GoogleService-Info.plist"
    },
    "plugins": [
      "@react-native-firebase/app",
      "@react-native-firebase/perf",
      "@react-native-firebase/crashlytics",
      "@react-native-firebase/messaging"
    ]
  }
}
```

I think this might increase the cognitive load somewhat, but I'm not sure if what I'm doing is correct. Since you've raised doubts, I think what I did may not be intuitive. Now I have moved the code to the Messaging module.

Following your advice, I've noticed that continuing to commit in this PR would pollute the commits, so I've created a new PR. [https://github.com/invertase/react-native-firebase/pull/7369]

Have a good day.

PR #7422, opened 2023-10-26
fix: Return Promise.reject instead of throw error

moonjava2005:
### Description
In order to make sure `crash free` with all even log method, we should change `throw Error` to `Promise.reject` because the below example:
In some case, developer will call log event in their sync method. This approach will help the log function not blocking the code running
` export const logFirebaseEvent = (eventName: string, properties?: object) => {
    firebaseAnalytic().logEvent(snakeCase(eventName), properties)  
}
`
Then
`......
logFirebaseEvent("GA-Event-Too-Long")
...Other logics`
But, in this case, the current code will throw Error, therefore, `other logics` won't be executed. This doesn't make sense
So, we will change from `throw Error` to `Promise.reject` in order to make sure the code pattern as above can run without any breaking logic

### Release Summary

- Fix crash error on analytic when calling synchronize with exception

### Checklist

  - [ X] Yes
  - [ X] `Android`
  - [ X] `iOS`
- My change includes tests;
  - [ X] `e2e` tests added or updated in `packages/\*\*/e2e`
  - [ X] `jest` tests added or updated in `packages/\*\*/__tests__`
- This is a breaking change;
  - [ ] Yes
  - [ X] No

### Test Plan

---

### DIFF
--- a/packages/analytics/lib/index.js
+++ b/packages/analytics/lib/index.js
@@ -133,36 +133,40 @@ const nativeModuleName = 'RNFBAnalyticsModule';
 class FirebaseAnalyticsModule extends FirebaseModule {
   logEvent(name, params = {}, options = {}) {
     if (!isString(name)) {
-      throw new Error("firebase.analytics().logEvent(*) 'name' expected a string value.");
+      return Promise.reject(new Error(
+        "firebase.analytics().logEvent(*) 'name' expected a string value."
+      ));
     }
 
     if (!isUndefined(params) && !isObject(params)) {
-      throw new Error("firebase.analytics().logEvent(_, *) 'params' expected an object value.");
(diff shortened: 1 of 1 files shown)


mikehardy:
Hi there! I understand your goal as stated ("in order to make sure crash free") however there is another goal which is also quite important: to never silently drop analytics events.

If you log an analytics event with parameters that are too long they will *silently* drop, which is a worse sort of failure than crashing which should be seen in development - this is based on our experience troubleshooting it in the past and why these throw errors.

The solution, if you are sending possibly-too-long event names despite the warnings of the limits of the API is to take the function you show above that wraps the logEvent, and add a try / catch to it.

Converting the API to something that could silently fail does not seem like a positive change though

Issue #7394, opened 2023-10-04
App Check Replay Protection


apetta:
Hi!
Would it be possible to use the app-check [replay protection](https://firebase.google.com/docs/app-check/ios/custom-resource#replay-protection) with this library?

mikehardy:
Interesting - I haven't contemplated adding support for it, but I would happily merge any reasonable PR that implemented the API and wrapped the underlying native calls

This is the API to model, as we use the firebase-js-sdk API shapes: https://firebase.google.com/docs/app-check/web/custom-resource#replay-protection

apetta:
> Interesting - I haven't contemplated adding support for it, but I would happily merge any reasonable PR that implemented the API and wrapped the underlying native calls
> 
> This is the API to model, as we use the firebase-js-sdk API shapes: https://firebase.google.com/docs/app-check/web/custom-resource#replay-protection

Just submitted a PR. Would be great if you could have a look :)

mikehardy:
Not stale and neither is the PR - definitely going to happen - sorry for the delay - thank you for your patience

JasonPan:
I seem to be getting an issue related to the work on the linked PR, is this feature still a work in progress?

I'm just trying to set up the baseline AppCheck functionality at the moment and came across this error. Led me to this issue / PR

<img width="817" alt="image" src="https://github.com/invertase/react-native-firebase/assets/6129538/ba0fc20e-a8de-4947-9263-a463dc1df184">

JasonPan:
(This is on the latest version 18.7.3, 18.6.1 seems to work okay)

apetta:
Hmm, I was able to get it up & running on the latest version. 

It might be worth opening a new issue with more details of your setup

mikehardy:
@JasonPan You have most likely overridden your firebase-ios-sdk version, which is something we intend to clearly note is *at your own risk* as we will adopt new features that require new versions of the SDK from time to time.

If you override your SDK and are not using the correct minimum version of firebase-ios-sdk as we direct in our app module package.json, you will see compilation errors like this

They are an indicator that you should take your firebase-ios-sdk version override out of your Podfile (preferred) or you must bump your firebase pod versions

mikehardy:
The feature has been merged though, and there's no further work planned. It appears to be working as intended.

JasonPan:
Nevermind, thanks! You're right, I did just have to make sure all my RN Firebase dependencies were upgraded to the latest version, working now :)

[...]

PR #7424, opened 2023-10-26
feat(app-check): Replay Protection

apetta:
### Description

This PR adds the Replay Protection feature for App Check, where it provides a consumable token for single-use server calls.

### Related issues

#7394 

### Release Summary

- Introduces Replay Protection feature for App Check
- `getLimitedUseToken` Provides single-use tokens for enhanced security on the server.

### Checklist

  - [x] Yes
  - [x] `Android`
  - [x] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

Tested in a private repo with the emulator.

---


mikehardy (changes requested):
Nice!
I think there should be entries in the index.d.ts files so that they are available in typescript?
Also, I think it should be possible to add an entry for this in the e2e test suite for app-check, that way we can see in CI that the methods really are working + not crashing (also, they will help you determine locally that they are working before pushing to the PR branch - best practice is to add an e2e test and put `.only` on it so when you run the suite locally it is super fast to run it + iterate on code changes

apetta:
Thanks! 

I remember I couldn't get the project installed locally last time I tried, so I had to skip the tests. Will try and look at this again soon

mikehardy:
When I took this one and attempted to add tests to it, it hung on iOS - need to investigate why but I suspect it has something to do with not being able to proxy to the currently configured provider from our custom provider, and calling the API directly on the FIRAppCheck.

It appears this new API isn't implemented on the Provider interfaces unlike getToken, for some reason, and I've got an API shape question out upstream to see if I can understand it better https://github.com/firebase/firebase-ios-sdk/issues/11284#issuecomment-1828389986

apetta:
Thank you so much!!

mikehardy (approved):

PR #7478, opened 2023-11-28
test(ci): do not pre-compress logs sent to upload-archive action

chipitsine:
GHA artifacts are already compressed using zip, no need to compress them twice


chipitsine:
CLA bot told me "you have signed already..."
looks like nothing left from my side

Issue #7551, opened 2024-01-08
[🐛] tests:packager:jet-reset-cache "Cannot find module"


NorseGaud:
After running all prereq steps in https://github.com/invertase/react-native-firebase/blob/main/CONTRIBUTING.md, I get to this command and see:

```
anka@ip-192-168-64-8 react-native-firebase % yarn tests:packager:jet-reset-cache
node:internal/modules/cjs/loader:1137
  throw err;
  ^

Error: Cannot find module '/Users/anka/react-native-firebase/tests/yarn'
    at Module._resolveFilename (node:internal/modules/cjs/loader:1134:15)
    at Module._load (node:internal/modules/cjs/loader:975:27)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:128:12)
    at node:internal/main/run_main_module:28:49 {
  code: 'MODULE_NOT_FOUND',
  requireStack: []
}

Node.js v18.19.0
```

The build worked before this worked.

PR #7552, opened 2024-01-08
fix: tests:packager:jet-reset-cache 'Cannot find module' - issue/7551

NorseGaud:
Issue: https://github.com/invertase/react-native-firebase/issues/7551

`node` should not be there (AFAICT)


mikehardy (approved):
Oh I think that was my fault, sorry, and good catch

PR #7565, opened 2024-01-14
feat(auth): include phoneNumber from PhoneMultiFactorInfo

mnahkies:
### Description
Expose `phoneNumber` on the MFA hints to JS.

`MultiFactorInfo` has two sub-classes in the underlying SDK's - one for TOTP and one for Phone, where the Phone one additionally includes the `phoneNumber` registered to the factor.

Before login the `phoneNumber` is redacted, and after login it is unredacted.

This is useful for showing the user a hint as to which phone number they should check, or distinguishing between multiple phone factors.

Additionally rename `enrollmentDate` -> `enrollmentTime` on iOS to match Android/Web (ref: https://firebase.google.com/docs/reference/js/auth.multifactorinfo)
- If we'd rather avoid the breaking change here, perhaps we could keep `enrollmentDate` as deprecated on iOS?

ref:
- https://firebase.google.com/docs/reference/ios/firebaseauth/api/reference/Classes/FIRPhoneMultiFactorInfo
- https://firebase.google.com/docs/reference/kotlin/com/google/firebase/auth/PhoneMultiFactorInfo
- https://firebase.google.com/docs/reference/js/auth.phonemultifactorinfo

### Release Summary
- Expose `phoneNumber` on `PhoneMultiFactorInfo`
- Rename `enrollmentDate` to `enrollmentTime` on iOS to match Android/Web

### Checklist

  - [x] Yes
  - [x] `Android`
  - [ ] `iOS`
- My change includes tests;
- This is a breaking change;
  - [x] Yes
  - [] No

### Test Plan
Tested on Android emulator, observed:
- Redacted `phoneNumber` included in hint on the login challenge
- Unredacted `phoneNumber` included in hint after login (eg: showing a manage mfa UI)

**NOT** tested on iOS - ideally could do with some help here.

---

:fire:

### DIFF
--- a/packages/auth/ios/RNFBAuth/RNFBAuthModule.m
+++ b/packages/auth/ios/RNFBAuth/RNFBAuthModule.m
@@ -892,8 +892,15 @@ RCT_EXPORT_METHOD(verifyPhoneNumberForMultiFactor
   DLog(@"verifyPhoneNumberForMultifactor phoneNumber: %@", phoneNumber);
   DLog(@"verifyPhoneNumberForMultifactor sessionId: %@", sessionId);
   FIRMultiFactorSession *session = cachedSessions[sessionId];
-  DLog(@"using instance VerifyPhoneNumberForMultifactor: %@",
-       firebaseApp.name)[[FIRPhoneAuthProvider providerWithAuth:[FIRAuth authWithApp:firebaseApp]]
+  if (session == nil) {
+    [RNFBSharedUtils rejectPromiseWithUserInfo:reject
+                                      userInfo:(NSMutableDictionary *)@{
+                                        @"code" : @"invalid-multi-factor-session",
+                                        @"message" : @"can't find session for provided key"
+                                      }];
+    return;
+  }
+  [[FIRPhoneAuthProvider providerWithAuth:[FIRAuth authWithApp:firebaseApp]]
        verifyPhoneNumber:phoneNumber
               UIDelegate:nil
       multiFactorSession:session
--- a/packages/auth/ios/RNFBAuth/RNFBAuthModule.m
+++ b/packages/auth/ios/RNFBAuth/RNFBAuthModule.m
@@ -1695,6 +1695,8 @@ RCT_EXPORT_METHOD(useEmulator
       @"enrollmentTime" : enrollmentTime,
       // @deprecated enrollmentDate kept for backwards compatibility, please use enrollmentTime
       @"enrollmentDate" : enrollmentTime,
+      // phoneNumber only present on FIRPhoneMultiFactorInfo
+      @"phoneNumber" : hint.phoneNumber == nil ? [NSNull null] : hint.phoneNumber,
     }];
   }
   return enrolledFactors;
--- a/packages/auth/ios/RNFBAuth/RNFBAuthModule.m
+++ b/packages/auth/ios/RNFBAuth/RNFBAuthModule.m
@@ -888,6 +888,9 @@ RCT_EXPORT_METHOD(verifyPhoneNumberForMultiFactor
                   : (NSString *)sessionId
                   : (RCTPromiseResolveBlock)resolve
                   : (RCTPromiseRejectBlock)reject) {
+  DLog(@"verifyPhoneNumberForMultifactor using app: %@", firebaseApp.name);
+  DLog(@"verifyPhoneNumberForMultifactor phoneNumber: %@", phoneNumber);
+  DLog(@"verifyPhoneNumberForMultifactor sessionId: %@", sessionId);
   FIRMultiFactorSession *session = cachedSessions[sessionId];
   DLog(@"using instance VerifyPhoneNumberForMultifactor: %@",
        firebaseApp.name)[[FIRPhoneAuthProvider providerWithAuth:[FIRAuth authWithApp:firebaseApp]]
--- a/packages/auth/ios/RNFBAuth/RNFBAuthModule.m
+++ b/packages/auth/ios/RNFBAuth/RNFBAuthModule.m
@@ -578,7 +578,9 @@ RCT_EXPORT_METHOD(signInWithCredential
                                                      @"has expired or is not currently supported.",
                                       }];
   }
-  DLog(@"using app SignInWithCredential: %@", firebaseApp.name)[[FIRAuth authWithApp:firebaseApp]
+  DLog(@"using app SignInWithCredential: %@", firebaseApp.name);
+
+  [[FIRAuth authWithApp:firebaseApp]
       signInWithCredential:credential
                 completion:^(FIRAuthDataResult *authResult, NSError *error) {
                   if (error) {
@@ -820,8 +822,9 @@ RCT_EXPORT_METHOD(signInWithPhoneNumber
                   : (NSString *)phoneNumber
                   : (RCTPromiseResolveBlock)resolve
                   : (RCTPromiseRejectBlock)reject) {
-  DLog(@"SignInWthPhoneNumber instance: %@",
-       firebaseApp.name)[[FIRPhoneAuthProvider providerWithAuth:[FIRAuth authWithApp:firebaseApp]]
+  DLog(@"SignInWthPhoneNumber instance: %@", firebaseApp.name);
+
+  [[FIRPhoneAuthProvider providerWithAuth:[FIRAuth authWithApp:firebaseApp]]
       verifyPhoneNumber:phoneNumber
              UIDelegate:nil
              completion:^(NSString *_Nullable verificationID, NSError *_Nullable error) {
@@ -862,8 +865,9 @@ RCT_EXPORT_METHOD(verifyPhoneNumberWithMultiFactorInfo
                                       }];
     return;
   }
-  DLog(@"using instance verifyPhoneNumberWithMultiFactorInfo: %@",
-       firebaseApp.name)[[FIRPhoneAuthProvider providerWithAuth:[FIRAuth authWithApp:firebaseApp]]
+  DLog(@"using instance verifyPhoneNumberWithMultiFactorInfo: %@", firebaseApp.name);
+
+  [[FIRPhoneAuthProvider providerWithAuth:[FIRAuth authWithApp:firebaseApp]]
       verifyPhoneNumberWithMultiFactorInfo:hint
                                 UIDelegate:nil
                         multiFactorSession:session
@@ -907,12 +911,15 @@ RCT_EXPORT_METHOD(resolveMultiFactorSignIn
                   : (NSString *)verificationCode
                   : (RCTPromiseResolveBlock)resolve
                   : (RCTPromiseRejectBlock)reject) {
-  DLog(@"using instance resolve MultiFactorSignIn: %@", firebaseApp.name)
-      FIRPhoneAuthCredential *credential =
-          [[FIRPhoneAuthProvider providerWithAuth:[FIRAuth authWithApp:firebaseApp]]
-              credentialWithVerificationID:verificationId
-                          verificationCode:verificationCode];
-  DLog(@"credential: %@", credential) FIRMultiFactorAssertion *assertion =
+  DLog(@"using instance resolve MultiFactorSignIn: %@", firebaseApp.name);
+
+  FIRPhoneAuthCredential *credential =
+      [[FIRPhoneAuthProvider providerWithAuth:[FIRAuth authWithApp:firebaseApp]]
+          credentialWithVerificationID:verificationId
+                      verificationCode:verificationCode];
+  DLog(@"credential: %@", credential);
+
+  FIRMultiFactorAssertion *assertion =
       [FIRPhoneMultiFactorGenerator assertionWithCredential:credential];
 
   [cachedResolver[sessionKey] resolveSignInWithAssertion:assertion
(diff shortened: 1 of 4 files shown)


packages/auth/ios/RNFBAuth/RNFBAuthModule.m
```diff
@@ -1667,7 +1667,8 @@ - (NSDictionary *)firebaseUserToDict:(FIRUser *)user {
       @"uid" : hint.UID,
       @"factorId" : hint.factorID == nil ? [NSNull null] : [self getJSFactorId:(hint.factorID)],
       @"displayName" : hint.displayName == nil ? [NSNull null] : hint.displayName,
-      @"enrollmentDate" : enrollmentDate,
+      @"enrollmentTime" : enrollmentDate,
+      @"phoneNumber" : hint.phoneNumber
```
mnahkies, line 1671: I guess this works without a cast given that we're looping over `FIRPhoneMultiFactorInfo *hint` - however I'd question if this is wise?

What happens if I have a user that has registered a TOTP factor - does it crash?

(not an iOS developer so hoping for a steer/hand here)

mnahkies:
>  Vercel – react-native-firebase-next — Deployment has failed 

I don't seem to be able to access the logs for this :disappointed:

packages/auth/ios/RNFBAuth/RNFBAuthModule.m
```diff
@@ -1667,7 +1667,8 @@ - (NSDictionary *)firebaseUserToDict:(FIRUser *)user {
       @"uid" : hint.UID,
       @"factorId" : hint.factorID == nil ? [NSNull null] : [self getJSFactorId:(hint.factorID)],
       @"displayName" : hint.displayName == nil ? [NSNull null] : hint.displayName,
-      @"enrollmentDate" : enrollmentDate,
+      @"enrollmentTime" : enrollmentDate,
```
mnahkies, line 1670: This is a breaking change. We might want to keep `enrollmentDate` around for a bit for backwards-compatibility in case people have worked around this discrepancy between platforms based on the OS

packages/auth/lib/index.d.ts
```diff
@@ -472,7 +472,21 @@ export namespace FirebaseAuthTypes {
...
+
+  export interface PhoneMultiFactorInfo extends MultiFactorInfoCommon {
+    factorId: 'phone';
+    /**
+     * The phone number used for this factor.
+     */
+    phoneNumber: string;
+  }
+
+  export interface TotpMultiFactorInfo extends MultiFactorInfoCommon {
```
mnahkies, line 485: There's no additional properties on this yet, but felt it is still useful to represent it for future / consistency with the underlying SDK's

packages/auth/ios/RNFBAuth/RNFBAuthModule.m
mikehardy, line 1670: Avoiding a breaking change here would be nice as changing the shape of the return isn't really an enabler of anything - so keeping backwards compat is worth it in my opinion

```suggestion
      @"enrollmentDate" : enrollmentDate, // from original implementation, kept to avoid breaking change
      @"enrollmentTime" : enrollmentDate,
```

packages/auth/ios/RNFBAuth/RNFBAuthModule.m
mikehardy, line 1671: the if-null / do-this / else do that above this line are from previous crashes, seems better to be safe than sorry by using the same idiom as above

```suggestion
      @"phoneNumber" : hint.phoneNumber == nil ? [NSNull null] : hint.phoneNumber
```

packages/auth/lib/index.d.ts
```diff
@@ -472,7 +472,21 @@ export namespace FirebaseAuthTypes {
...
    */
-  export interface MultiFactorInfo {
+  export type MultiFactorInfo = PhoneMultiFactorInfo | TotpMultiFactorInfo;
+
+  export interface PhoneMultiFactorInfo extends MultiFactorInfoCommon {
+    factorId: 'phone';
+    /**
+     * The phone number used for this factor.
+     */
+    phoneNumber: string;
```
mikehardy, line 482: You never know, and per the native idiom in iOS anyway, it *might* return null

```suggestion
    phoneNumber: string | null;
```

mikehardy (changes requested):
This looks pretty good - I left a few comments in form of code suggestions which could be adopted and this could merge, or if you've got thoughts + counter-proposals I'm happy to hear them

in general though I like the idea of backwards-compatibility to avoid the break, and following the same null-check idiom in the iOS code

packages/auth/ios/RNFBAuth/RNFBAuthModule.m
mnahkies: The documentation seems to suggest otherwise to me, https://firebase.google.com/docs/reference/ios/firebaseauth/api/reference/Classes/FIRPhoneMultiFactorInfo - `phoneNumber` is marked as non-null

I've read through the other PR https://github.com/invertase/react-native-firebase/pull/7541 regarding `factorId` being and I'm surprised this was needed. Feels like a bug in the underlying firebase SDK to me - with a null `factorId` you won't be able to tell whether to follow the phone or totp flows. Though given the auth part isn't opensource I guess there's little hope of debugging/fixing it upstream 😢

Happy to add the extra defensiveness here, but I think ideally it wouldn't be required.

mnahkies: Though as per https://github.com/invertase/react-native-firebase/pull/7565/files#r1451696979 I'm not so hot on objective c - if it will allow a `FIRMultiFactorInfo` where we've said we're looping over `FIRPhoneMultiFactorInfo` then this definitely needs a nil check, as a TOTP factor will be a `FIRMultiFactorInfo`

mikehardy:
Still very interested in getting this one in - just had a couple more things that were high priority slice into my queue in front of it - apologies

packages/auth/ios/RNFBAuth/RNFBAuthModule.m
mnahkies: I've split this part of the change to https://github.com/invertase/react-native-firebase/pull/7653 in case you'd consider including it as a breaking change for v19 - also happy to stick with keeping runtime backwards compatibility with a deprecated property though.

The rest of this PR is additive, and the iOS part around adding `phoneNumber` definitely needs more testing / adjusting, so could come as a minor release later

mikehardy (changes requested):
I think if you add the nil check on phone number, allow it to return null in the JS-level typing, and remove the enrollmentTime/Date change (which I just merged separately) this can go in? I think that would satisfy all the comments here

Again, thanks for the patience, we're moving a bit now...

packages/auth/ios/RNFBAuth/RNFBAuthModule.m
mnahkies: `nil` check added to account for `FIRMultiFactorInfo`, and also added an assertion in the e2e tests (though unfortunately it looks like these are skipped for [ios](https://github.com/invertase/react-native-firebase/blob/8b4e39baddec2f9ac0569489d47b8a3915056ef7/packages/auth/e2e/multiFactor.e2e.js#L546-L548))

packages/auth/lib/index.d.ts
mnahkies: I'm somewhat reluctant to mark this as nullable based on https://github.com/invertase/react-native-firebase/pull/7565#discussion_r1465428041 - whilst anything is possible in javascript, I believe the intention of the SDK is that this is non-nullable.

If you feel it's essential to mark as nullable then I can still

packages/auth/ios/RNFBAuth/RNFBAuthModule.m
```diff
@@ -1685,7 +1685,7 @@ - (NSDictionary *)firebaseUserToDict:(FIRUser *)user {
 - (NSArray<NSMutableDictionary *> *)convertMultiFactorData:(NSArray<FIRMultiFactorInfo *> *)hints {
   NSMutableArray *enrolledFactors = [NSMutableArray array];
 
-  for (FIRPhoneMultiFactorInfo *hint in hints) {
+  for (FIRMultiFactorInfo *hint in hints) {
```
mnahkies, line 1688: TBH I'm not 100% sure on this

mnahkies: Caused a build failure. I think it would be more accurate to loop over `FIRMultiFactorInfo` and then branch based on the specific type, but this will work since TOTP doesn't have any additional properties :shrug:

mikehardy:
Hi there! I understand the interest, however there is no reasonable way to give ETAs in open source.

I can say the fastest way to get it out, if it's very important for your work, is to determine exactly why the e2e test rig fails when the patch is applied, and propose a fix

There's a readme in the tests directory that explains how to run the tests.

mikehardy:
Actually @yuvalhermelin-fijoya please accept my apologies, I thought your comment was on the PR to get new architecture compatibility support in 🤦 - I didn't realize it was attached to this one. In fact this one is very close and I believe it's just pending me doing final review. I'm attempting to do that now, hoping to release it today! @mnahkies has been very patient as I've been focused on other things, the work in general is excellent and I'm sorry it has taken as long as it has

mnahkies:
@mikehardy TBF the e2e tests did fail on my last push and I haven't had the bandwidth to investigate properly yet 😅

mikehardy:
The CI tests are flaky for two specific reasons right now (ios messaging unenroll / enroll hanging on APNS token fetch, known issue, android one API working / one API not working, known issue) so that's not normally a problem

But I've got the PR pulled locally for final review and the android assertion was against the unmasked phone number when actually it has the masked one, so that was failing for real.

I have that fixed locally but you mention that it returns the unmasked phone number after signing in, is that true? Because the e2e test you added *is* after sign-in but it's definitely a masked number in the hints 🤔 

I'm trying to see if I can finally fix the iOS e2e test issue in order to turn those on as well - seems like it should be possible

I'm about to get pulled away in to a bunch of other things though so this may sit for one more day. It's about to go in though - I want to get this in and release it before the breaking changes in #7774 (even though they are small breaks) so that people can use this without having to think about the breaking change yet

mnahkies:
> android assertion was against the unmasked phone number when actually it has the masked one, so that was failing for real.

This was an oversight sorry - the `multiFactorResolver` from the challenge thrown during login is expected to have a masked number - since the assertion is on the challenge it should be masked.

> I have that fixed locally but you mention that it returns the unmasked phone number after signing in, is that true? Because the e2e test you added is after sign-in but it's definitely a masked number in the hints 🤔

Once you login successfully, if you read the multi-factor info from the `currentUser` you'll get an unredacted number, eg: I have:

```
            <Text variant="bodyMedium">
              You currently have {currentFactor.phoneNumber} registered for SMS
              MFA.
            </Text>
```
Where `currentFactor` is:
```
user.multiFactor?.enrolledFactors.find(
    (it) => it.factorId === "phone",
  )
```
And `user` is `auth.currentUser`

Which gives me an unredacted phone number

Screenshot_20240508-214712

(note: this particular app is using expo to target both web/native with the platform specific SDK's used, so only totp display/unenrol is supported on native until #7718 or equivalent lands :crossed_fingers:, though the ui implies otherwise right now)

mnahkies:
(if you already have a fix locally for the incorrect e2e assertion please feel free to push that up and merge)

mikehardy:
I do have a fix for that locally, thanks for the "go" on pushing once I've got things working locally.
I believe I've determined why iOS phone MFA isn't working, there is a bit of a difference in the native implementations and it appears (* based on current-but-may-change investigation results) that the enrollment isn't being done correctly at the native level and it may be that no iOS enrollment has ever been working 😬 . So that test was actually telling me something and I shouldn't merged it with iOS disabled + I think I can get it all working. That's the current aim anyway

Then this goes in - it appears to be working well on android already at least

> Once you login successfully, if you read the multi-factor info from the currentUser you'll get an unredacted number, eg: I have:

...thanks for the detailed pointer, I'll see if I can add an assertion on that, following the general idea "assert all the things you think you know..."

mikehardy:
Hey @yuvalhermelin-fijoya - it should be possible to use this pre-release using the patch set generated here + [patch-package](https://github.com/ds300/patch-package) --> https://github.com/invertase/react-native-firebase/actions/runs/9078582950?pr=7565 - the auth one is the only one you should need?

Sorry for the delay, I'm working through some testing issues getting our test suite to run against the cloud again since the auth emulator seems to be missing some functionality in the MFA area

mikehardy:
Oh of course the patch set generation fails this time because of some new CI features 🤦 - need to fix that, then rebase and re-run it, will be a bit, sorry

mikehardy:
@yuvalhermelin-fijoya okay, fixed the patchset generation it will show up as an attachment here: https://github.com/invertase/react-native-firebase/actions/runs/9086347168?pr=7565

The auth one is the only one you are interested in, but the patchset contains all merged changes since last public release + this PR, so all of them are valid.

I don't anticipate that any of the code here will change, the holdup now is test process

mikehardy:
I just fixed a major chunk of our test process issues with #7797
One more major chunk to truly verify this change - but I will likely defer the most difficult part of that change in it to get this through

packages/auth/lib/index.d.ts
mikehardy: good defense - resolved via a quick look at firebase-js-sdk types to make 100% certain, https://github.com/search?q=repo%3Afirebase%2Ffirebase-js-sdk%20phonemultifactorinfo&type=code 

it is not nullable ✅

mikehardy (approved):
This is looking good now - on course to land tonight and release, last bit of testing (enabling MFA tests for iOS) I'm going to defer as it implies a big shift to how we work since it's an auth-emulator vs cloud-auth thing

mikehardy:
still need to merge the test tweak from https://github.com/invertase/react-native-firebase/pull/7565#issuecomment-2101757645 then this can go

Issue #7483, opened 2023-11-29
[🐛] Missing firebase-js-sdk API


Slapbox:
## Issue

**Unimplemented JS SDK features include:**

- storage.getBlob
- storage.getBytesSync
- database.forceLongPolling
- database.forceWebSockets

**Also possibly:**
- auth.unenroll
- auth.addTokenListener

I located these by searching for the string `not implemented` in the portions of the project we use. I've also made an effort to document the others but this may not be comprehensive, or may contain false positives.

- **Platform that you're experiencing the issue on**:
  - [ ] iOS
  - [ ] Android
  - [ ] **iOS** but have not tested behavior on Android
  - [ ] **Android** but have not tested behavior on iOS
  - [x] Both
- **`react-native-firebase` version you're using that has this issue:**
  - `e.g. 5.4.3`
- **`Firebase` module(s) you're using that has the issue:**
  - `e.g. Instance ID`
- **Are you using `TypeScript`?**
  - `Y/N` & `VERSION`

</p>
</details>

---

- 👉 Check out [`React Native Firebase`](https://twitter.com/rnfirebase) and [`Invertase`](https://twitter.com/invertaseio) on Twitter for updates on the library.

Slapbox:
Originally mentioned in #7433.

Slapbox:
I also noticed `uploadBytes` is not implemented also. Is there anything else we can use in the meantime, or any plan to implement this in the immediate future? It would be a shame to have to revert adding `react-native-firebase` to our app. 

Unfortunately I lack the knowledge and time to help implement it.

mikehardy:
@Slapbox we handle the non-modular `put` API which is very nearly `uploadBytes` by signature, it handles the same type of input which tells me that under the covers we are already doing all the necessary things from javascript to native and back

So my first take on it is that `uploadBytes` is probably just exporting the correct symbol and API shape, and should be nearly trivial. Those are famous last words but it really does look that way to me

Need to get react-native 0.73 unexpected incompatibility #7500  fixed before I can take a look at uploadBytes but I'll check it out with priority tomorrow

Slapbox:
> Those are famous last words but it really does look that way to me

Too true!

I really appreciate your attention on this a lot. Thank you for the great project and the great assistance!

----

**Workaround:**

[...]

PR #7566, opened 2024-01-14
feat: support mfa un-enrollment

mnahkies:
### Description
implements the unenroll method, receiving either a factor uid or factor info, and then calling the underlying unenroll by factor uid overload to avoid marshaling the factor info object.

ref:
- https://firebase.google.com/docs/reference/js/auth.multifactoruser#multifactoruserunenroll
- https://firebase.google.com/docs/reference/ios/firebaseauth/api/reference/Classes/FIRMultiFactor#-unenrollwithinfo:completion:
- https://firebase.google.com/docs/reference/android/com/google/firebase/auth/MultiFactor#unenroll(java.lang.String)

### Related issues
#7483

### Release Summary
Support un-enrolling MFA factors

### Checklist

  - [x] Yes
  - [x] `Android`
  - [ ] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan
Tested on Android emulator, observed:
- Failure due to needing to re-authenticate raises the correct exception
- Success path correctly removes the factor

I've not tested that a auth state change event is emitted correctly yet.

---
:fire:

### DIFF
--- a/packages/auth/ios/RNFBAuth/RNFBAuthModule.m
+++ b/packages/auth/ios/RNFBAuth/RNFBAuthModule.m
@@ -965,6 +965,25 @@ RCT_EXPORT_METHOD(finalizeMultiFactorEnrollment
                              }];
 }
 
+RCT_EXPORT_METHOD(unenrollMultiFactor
+                  : (FIRApp *)firebaseApp
+                  : (NSString *)factorUID
+                  : (RCTPromiseResolveBlock)resolve
+                  : (RCTPromiseRejectBlock)reject) {
+
+  FIRUser *user = [FIRAuth authWithApp:firebaseApp].currentUser;
+  [user.multiFactor unenrollWithFactorUID:factorUID
+                             completion:^(NSError *_Nullable error) {
+                               if (error != nil) {
+                                 [self promiseRejectAuthException:reject error:error];
+                                 return;
+                               }
+
+                               resolve(nil);
+                               return;
+                             }];
+}
+
 RCT_EXPORT_METHOD(verifyPhoneNumber
                   : (FIRApp *)firebaseApp
                   : (NSString *)phoneNumber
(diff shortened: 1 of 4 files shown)


packages/auth/ios/RNFBAuth/RNFBAuthModule.m
```diff
@@ -965,6 +965,25 @@ - (void)invalidate {
                              }];
 }
 
+RCT_EXPORT_METHOD(unenrollMultiFactor
```
mnahkies, line 968: iOS implementation is untested, but should be _**roughly**_ correct. If someone could give it a whirl that would be helpful

mikehardy:
Just a note that I really appreciate this work @mnahkies - and I will collaborate with you to get your PRs incorporated.

I'm looking at it with an eye towards supporting TOTP as well, and I noticed that our e2e testing had all the iOS tests disabled which was very worrisome as I'd like to add e2e tests for these methods as well.

So far I haven't been able to get the local firebase auth emulator to work correctly (investigation on this branch: https://github.com/invertase/react-native-firebase/tree/%40mikehardy/auth-mfa-ios-e2e-enable) and my next step is to restore the previous non-emulator approach for auth testing where needed - this is a slightly larger job but the auth emulator apparently doesn't support TOTP with no ETA on support so I think this will be needed anyway.

That work is blocking my review + approval of these items because I want to make sure the MFA auth APIs have good testing.

But they will go in, and if you need them to work right now you may confidently use patch-package and either local patches or the sets that our CI generates, knowing we'll get these through eventually

Thanks again

mnahkies:
@mikehardy no problem - I've been a bit swamped the past few weeks and haven't had a chance to come back to these yet either, but do plan to polish them up still. I've also got appetite to implement TOTP support, but just limited on time right now

PR #7688, opened 2024-03-19
fix(app): react-native 0.74 bridgeless mode compatibility

gabrieldonadel:
### Description

When trying to use `@react-native-firebase/app` with the new architecture and bridgeless mode on (react-native version 0.74.0-rc.3), the user is faced with an unhandled promise rejection 

image

Very similar case to https://github.com/react-native-netinfo/react-native-netinfo/pull/717

----
This error happens because of the following code inside nativeModule.js
 
https://github.com/invertase/react-native-firebase/blob/b40d44c57e81a4110cbb4e6204c400cb8f78b00a/packages/app/lib/internal/registry/nativeModule.js#L62-L79

Currently we cannot copy functions  directly from the React Native's NativeModules object, given that the module object is a host object.

After this change everything works as expected and the compatibility layer allows `@react-native-firebase/app` to be used with bridgeless mode on

### Release Summary

Add support for react-native 0.74 bridgeless mode

### Checklist

  - [ ] Yes
  - [x] `Android`
  - [x] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

Tested running on Android and iOS with: 
- new arch, bridgeless off
- new arch, bridgeless on
- old arch

:fire:

### DIFF
--- a/packages/app/lib/internal/registry/nativeModule.js
+++ b/packages/app/lib/internal/registry/nativeModule.js
@@ -65,7 +65,8 @@ function nativeModuleWrapped(namespace, NativeModule, argToPrepend) {
     return NativeModule;
   }
 
-  const properties = Object.keys(NativeModule);
+  let properties = Object.keys(Object.getPrototypeOf(NativeModule));
+  if (!properties.length) properties = Object.keys(NativeModule);
 
   for (let i = 0, len = properties.length; i < len; i++) {
     const property = properties[i];
--- a/packages/app/lib/internal/registry/nativeModule.js
+++ b/packages/app/lib/internal/registry/nativeModule.js
@@ -60,7 +60,7 @@ function nativeModuleMethodWrapped(namespace, method, argToPrepend) {
  * @param argToPrepend
  */
 function nativeModuleWrapped(namespace, NativeModule, argToPrepend) {
-  const native = NativeModule;
+  const native = {};
   if (!NativeModule) {
     return NativeModule;
   }
@@ -71,6 +71,8 @@ function nativeModuleWrapped(namespace, NativeModule, argToPrepend) {
     const property = properties[i];
     if (typeof NativeModule[property] === 'function') {
       native[property] = nativeModuleMethodWrapped(namespace, NativeModule[property], argToPrepend);
+    } else {
+      native[property] = NativeModule[property];
     }
   }
 
--- a/packages/app/lib/internal/registry/nativeModule.js
+++ b/packages/app/lib/internal/registry/nativeModule.js
@@ -60,7 +60,7 @@ function nativeModuleMethodWrapped(namespace, method, argToPrepend) {
  * @param argToPrepend
  */
 function nativeModuleWrapped(namespace, NativeModule, argToPrepend) {
-  const native = {};
+  const native = NativeModule;
   if (!NativeModule) {
     return NativeModule;
   }
@@ -71,8 +71,6 @@ function nativeModuleWrapped(namespace, NativeModule, argToPrepend) {
     const property = properties[i];
     if (typeof NativeModule[property] === 'function') {
       native[property] = nativeModuleMethodWrapped(namespace, NativeModule[property], argToPrepend);
-    } else {
-      native[property] = NativeModule[property];
     }
   }
 


gabrieldonadel:
hey @mikehardy, can you take a look at this PR when you have some time?

mikehardy (dismissed):
🤔 interesting - looks good, sorry for the delay, my whole house caught a durable and pestilent plague a couple weeks back 😷  😆  and I've been behind ever since...all good now but oh, the backlog

Going to pull this locally and give it a run as well as seeing what CI does of course

mikehardy:
Something not quite right with the wrapped methods after this change when running locally, investigating...


> RNFBAppCheckModule.isTokenAutoRefreshEnabled was called with 2 arguments but expects 1 arguments. If you haven't changed this method yourself, this usually means that your versions of the native code and JavaScript code are out of sync. Updating both should make this error go away.

From experience, I know that the method wrapping is mostly to automatically inject the idea of *which* "firebase app" you are using on the native side (denoted by a string which we use to do lookups in a "firebase app name" --> "firebase app native object" map on native side

The extra argument is that native firebase app name. This is a problem in the opposite direction though, where it got 2 arguments (possibly a wrapped method...) when it expected one. Anyway, I've worked through similar issues I can figure this one out I'm certain, but there might be a little delay while I do so

fobos531:
Not stale

johnf:
I tried this patch as well. My app loads but I get the following error
```
 (NOBRIDGE) ERROR  Error getting token [TypeError: this.native.getToken is not a function (it is undefined)]
```

mikehardy:
Yeah - this is just not ready yet and I have not been able to identify exactly why yet, which is extraordinarily frustrating as the change looks perfect.

birdofpreyru:
I had a brief look at this, and I can't make any sense of this patch.

The actual problem seems to be (apart of the crazy design of this library, which is apparently hard to maintain up-to-date with RN developments), that when you come to the following line:

https://github.com/invertase/react-native-firebase/blob/9fe3e4b698ff867fcdf9689ea23e1396a94ddaba/packages/app/lib/internal/registry/app.js#L56

There is no `NATIVE_FIREBASE_APPS` array in `nativeModule`... the reason being, I believe, that it depends on an old way constants used to work in RN, where RN eagerly called `getConstants()` in native layer, and attached its result to the module. I can't find details now, apart of this note [here](https://reactnative.dev/docs/native-modules-android#exporting-constants):
image

But when I was migrating my own RN libraries to the new architecture I had to switch to [such code](https://github.com/birdofpreyru/react-native-static-server/blob/master/src/constants.ts), for example:

image

_i.e._ explicitly calling `.getConstants()` method from JS layer, and taking constant values from its result, rather than as fields on module object itself. I think that is the correct fix of this issue, but it is 01:00 AM here, so can't try it right away :sweat_smile:

harisbaig100:
I applied it as a patch but still still (No Firebase App '[DEFAULT]') error persists

birdofpreyru:
Here is a one-liner patch that works for me on Android / New Arch / RN@0.74.1.

```diff
diff --git a/node_modules/@react-native-firebase/app/lib/internal/registry/nativeModule.js b/node_modules/@react-native-firebase/app/lib/internal/registry/nativeModule.js
index 03f001c..23d467d 100644
--- a/node_modules/@react-native-firebase/app/lib/internal/registry/nativeModule.js
+++ b/node_modules/@react-native-firebase/app/lib/internal/registry/nativeModule.js
@@ -65,7 +65,7 @@ function nativeModuleWrapped(namespace, NativeModule, argToPrepend) {
     return NativeModule;
   }
 
-  const properties = Object.keys(NativeModule);
+  const properties = Object.keys(Object.getPrototypeOf(NativeModule));
 
   for (let i = 0, len = properties.length; i < len; i++) {
     const property = properties[i];
```

mikehardy:
Hey @birdofpreyru that looks fantastic locally, it works on rn73 which indicates it is backwards-compatible where the previous patch wasn't.

Separately I've done a lot of work cleaning up the tests so that CI *can* go green (it's been red on iOS in CI for a while though it was passing locally)

With those changes rebased in here, the original patch reverted, and your patch added on, I think it might work

Can you confirm that the final diff state of this branch (just a -1 / +2) is what you have working in your testing of new arch on rn74?

mikehardy:
Hmm 🤔  this works correctly for me when I apply it on top of #7774 and run it in old architecture mode

but on new architecture mode my local iOS simulator says it's loading the bundle from metro then it just hangs - I think it is a local problem though - does it work for everyone here on iOS + new arch bridgeless?

android new arch bridgeless works

mikehardy:
going to merge since it appears to be backwards compatible at least, and seems to work - if there is still a problem we can fix-forward

PR #7755, opened 2024-04-18
feat(firebase): export and fix modular types

dhenneke:
### Description

We use the JS SDK for firebase in react-native but want to switch to `@react-native-firebase/firestore` to enable offline support. To align the codebase in our react and react-native apps we prefer to use the modular API in both projects. However the types were not yet exported properly for TypeScript usage. We also discovered that certain APIs were missing in the modular implementation.

This PR includes the following changes:
1. Export the types of the modular API & extend the `packages/firestore/type-test.ts` with the modular API.
3. Add an implementation for `getDoc`, `getDocFromCache`, and `getDocFromServer`, and use it in all affected e2e tests (previously `getDocs` was called for `DocumentReference`s which worked, but is technically incorrect).
4. Fix the base type of the queries which prohibited the use of `query(collection(...), where(...))`.

> This only exports the types of the `firebase` package because that's the only one we currently use and can test.

### Checklist

  - [x] Yes
  - [x] `Android`
  - [x] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

---

:fire:

### DIFF
--- a/packages/firestore/e2e/CollectionReference/add.e2e.js
+++ b/packages/firestore/e2e/CollectionReference/add.e2e.js
@@ -58,12 +58,12 @@ describe('firestore.collection().add()', function () {
     });
 
     it('adds a new document', async function () {
-      const { getFirestore, collection, addDoc, getDocs, deleteDoc } = firestoreModular;
+      const { getFirestore, collection, addDoc, getDoc, deleteDoc } = firestoreModular;
 
       const data = { foo: 'bar' };
       const docRef = await addDoc(collection(getFirestore(), COLLECTION), data);
       should.equal(docRef.constructor.name, 'FirestoreDocumentReference');
-      const docSnap = await getDocs(docRef);
+      const docSnap = await getDoc(docRef);
       docSnap.data().should.eql(jet.contextify(data));
       docSnap.exists.should.eql(true);
       await deleteDoc(docRef);
--- a/packages/firestore/e2e/DocumentReference/delete.e2e.js
+++ b/packages/firestore/e2e/DocumentReference/delete.e2e.js
@@ -38,15 +38,15 @@ describe('firestore.doc().delete()', function () {
 
   describe('modular', function () {
     it('deletes a document', async function () {
-      const { getFirestore, doc, setDoc, getDocs, deleteDoc } = firestoreModular;
+      const { getFirestore, doc, setDoc, getDoc, deleteDoc } = firestoreModular;
 
       const ref = doc(getFirestore(), `${COLLECTION}/deleteme`);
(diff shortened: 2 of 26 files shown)


mikehardy (changes requested):
Wow! This is pretty great. Is this working for you in production now?

Can you alter the incorrect `getDocs` (plural) APIs to delegate to the correctly named ones and add a `@deprecated API named incorrectly in original implementation use getNNN instead` tag to them?

@Salakar - you mentioned a possible need to wait on typings changes, would this be included in that? Should we wait to merge here or could this merge with no hold?

mikehardy:
Received internal feedback: this is okay to merge / no types hold, so just need to deprecate old typo-APIs and delegate them to the new one

dhenneke:
@mikehardy thanks for the review. We shipped our last production build with the modular API by temporarily copying the typings to a `*.d.ts` file into our codebase. So far we didn't see any issues.

Actually, both `getDoc` and `getDocs` are needed and are present in the js-sdk. The difference is that [`getDoc*`](https://github.com/firebase/firebase-js-sdk/blob/f25b9e53e9d77a3f92e36ff62218bbcf24aae857/packages/firestore/src/api/reference_impl.ts#L110-L112) is called for `DocumentReference`s (i.e. `getDoc(doc(firestore, 'my-collection', 'my-id'))` and returns a `DocumentSnapshot` for a single document, while [`getDocs*`](https://github.com/firebase/firebase-js-sdk/blob/f25b9e53e9d77a3f92e36ff62218bbcf24aae857/packages/firestore/src/api/reference_impl.ts#L210-L212) is called for `Query`s (i.e. `getDocs(collection(firestore, 'my-collection'))`) and returns a `QuerySnapshot` with multiple documents.

Since the implementation in the context of this library is identical (they both call `.get()` on `DocumentReference` and `Query`) it worked when you are not using TypeScript. But with TypeScript, you get type errors and it is also a different API compared to `@firebase/firestore`. So I don't think there is something we could deprecate.

mikehardy:
Sorry for the delay, but thanks so much for the clear explanation @dhenneke - that makes a lot of sense.
I have a breaking change release coming up but I will try to get this out *first* so it's a very easy upgrade before contemplating the breaking change

(the breaking change isn't much on the code level, just requires Xcode 12.5+ and android minSdk 21+ in general or 23+ for auth - but still easier to just get types in first)

Thanks!

mikehardy (approved):
Looks great, very cleanly separated commit series as well, a rare treat!

dhenneke:
Thanks 🙂

PR #7826, opened 2024-06-06
build(deps): update all javascript dependencies

mikehardy:
Nothing really of interest here, just adopting new dependencies

one small note is a breaking change in the conventional-changelog-cli package so next release should have a look at the CHANGELOG generation to make sure it was complete + correct

I continue to have local issues with my main development machine and the android e2e tests but ios were perfect, hopefully CI is green


mikehardy:
maintainer note: passes e2e android for me locally but only if I remove the v8-compatibility suites from auth e2e. Why? There is some underlying process that is using up all the Binder buffer space with too many messages causing the system to kill the e2e app in my local setup which is an Android 14 google APIs emulator. This will likely vary from environment to environment but it works fine in CI so going to let it through

PR #7837, opened 2024-06-11
fix(crashlytics, ios): init w/componentsToRegister vs configureWithApp

### LINKED ISSUE #4241: Remove use of configureWithApp from iOS init sequence

mikehardy:
### Description

configureWithApp is scheduled for removal upstream, the initialization logic we need to execute appears to operate fine in componentsToRegister

@ncooke3 - it can't really be this easy can it? I'm immediately suspicious, but...the logic seems to execute just fine - the logic we execute is idempotent as well, so it's okay if it executes more than once so long as it executes at least once...

### Related issues

Fixes:
- Fixes #4241

### Release Summary

conventional commit

### Checklist

  - [x] Yes
  - [ ] `Android`
  - [x] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

If you see "RNFBCrashlyticsInit initialization successful" in the iOS log file while running `yarn tests:ios:test` with this change, the initialization logic executed

You may see that by following our standard crashlytics configuration / testing advice blog which tells you to run this:

> xcrun simctl spawn booted log stream --level debug --style compact |grep RNFBCrashlyticsInit

---

### DIFF
--- a/packages/crashlytics/ios/RNFBCrashlytics/RNFBCrashlyticsInitProvider.m
+++ b/packages/crashlytics/ios/RNFBCrashlytics/RNFBCrashlyticsInitProvider.m
@@ -18,6 +18,11 @@
 #import "RNFBCrashlyticsInitProvider.h"
 #import <Firebase/Firebase.h>
 #import <FirebaseCoreExtension/FIRAppInternal.h>
+#import <FirebaseCoreExtension/FIRComponent.h>
+#import <FirebaseCoreExtension/FIRComponentContainer.h>
+#import <FirebaseCoreExtension/FIRComponentType.h>
+#import <FirebaseCoreExtension/FIRLibrary.h>
+#import <FirebaseCrashlytics/FIRCrashlytics.h>
 #import "RNFBJSON.h"
 #import "RNFBMeta.h"
 #import "RNFBPreferences.h"
@@ -31,6 +36,19 @@ NSString *const KEY_CRASHLYTICS_IS_ERROR_GENERATION_ON_JS_CRASH_ENABLED =
     @"crashlytics_is_error_generation_on_js_crash_enabled";
 NSString *const KEY_CRASHLYTICS_JAVASCRIPT_EXCEPTION_HANDLER_CHAINING_ENABLED =
     @"crashlytics_javascript_exception_handler_chaining_enabled";
+
+/// Empty protocol to register this provider as a component with Core.
+@protocol RNFBCrashlyticsInitProviderProtocol
+@end
+
+/// The name of the Crashlytics component in FirebaseCore's component system. Reference:
+// https://github.com/firebase/firebase-ios-sdk/blob/main/Crashlytics/Crashlytics/FIRCrashlytics.m#L87-L89
+@protocol FIRCrashlyticsInstanceProvider <NSObject>
+@end
+
+/// Privately conform to the protocol for component registration.
+@interface RNFBCrashlyticsInitProvider () <RNFBCrashlyticsInitProviderProtocol, FIRLibrary>
+@end
 
 @implementation RNFBCrashlyticsInitProvider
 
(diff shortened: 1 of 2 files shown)


packages/crashlytics/ios/RNFBCrashlytics/RNFBCrashlyticsInitProvider.m
```diff
@@ -79,26 +79,13 @@ + (BOOL)isCrashlyticsJavascriptExceptionHandlerChainingEnabled {
...
- * configureWithApp is automatically invoked by Firebase as this app is a registered FIRLibrary with
- * the SDK.
- *
- * At this point "configure" has already been called on the FIRApp instance. This behavior is meant
- * to mirror what is done in ReactNativeFirebaseCrashlyticsInitProvider.java
- *
- * This pattern may not be supported long term
- * https://github.com/firebase/firebase-ios-sdk/issues/2933
- */
-+ (void)configureWithApp:(FIRApp *)app {
```
ncooke3, line 95: Thanks, @mikehardy! I need to investigate a little further but one potential issue with this is that, as I understand it, `componentToRegister` runs before `configureWithApp:`, so it may not work to move the Crashlytics bit into `componentToRegister` as the Crashlytics instance may not be ready. Were you able to run this and see if `[FIRCrashlytics crashlytics]` returns a non-nil instance?

mikehardy: Well, it ran and did not have a nil dereference as the code accessed the instance during any test runs .. perhaps it instantiates crashlytics itself? Certainly don't want to rely on winning a race condition though, happy for any guidance

mikehardy: @ncooke3 okay okay :-), in lieue of the "easy cheesy" solution (which frankly gives me anxiety since you've mentioned it could be racy and have a nil crashlytics instance) I have attempted to use the full component infrastructure to create a dependency on crashlytics, then ask the component system to instantiate our provider here eagerly for the default app, using the dependency

It appears to work in testing the same as the previous "just run it" style did, but there should be no way this can get a nil crashlytics if I understand things correctly?

Would love a set of eyes on it though if you have a chance

ncooke3: Hi @mikehardy, the one issue now is that `FIRDependency` is also staged to go away. The In App Messaging iOS SDK uses `configureWithApp:` so I've been thinking through if/how we could generalize a replacement solution for both. 

Where does the `self.isCrashlyticsCollectionEnabled` value come from?

packages/crashlytics/ios/RNFBCrashlytics/RNFBCrashlyticsInitProvider.m
```diff
@@ -79,26 +98,51 @@ + (BOOL)isCrashlyticsJavascriptExceptionHandlerChainingEnabled {
...
+      return nil;
+    }
+
+    // Ensure it's cached so it returns the same instance every time messaging is called.
+    *isCacheable = YES;
+
+    // This returns a FIRCrashlytics *, directly cast it later for practical use, reference:
+    // https://github.com/firebase/firebase-ios-sdk/blob/main/Crashlytics/Crashlytics/FIRCrashlytics.m#L282-L284
+    id<FIRCrashlyticsInstanceProvider> crashlytics =
+        FIR_COMPONENT(FIRCrashlyticsInstanceProvider, container);
```
ncooke3, line 125: This is good.

packages/crashlytics/ios/RNFBCrashlytics/RNFBCrashlyticsInitProvider.m
```diff
@@ -79,26 +98,51 @@ + (BOOL)isCrashlyticsJavascriptExceptionHandlerChainingEnabled {
...
+    // xcrun simctl spawn booted log stream --level debug --style compact |grep RNFBCrashlyticsInit
+    DLog(@"RNFBCrashlyticsInit initialization successful");
+    return nil;
+  };
+
+  // 3) ...finally: register startup block to run w/Crashlytics dependency when ready
+  FIRComponent *crashlyticsInitProvider =
+      [FIRComponent componentWithProtocol:@protocol(RNFBCrashlyticsInitProviderProtocol)
+                      instantiationTiming:FIRInstantiationTimingEagerInDefaultApp
+                             dependencies:@[ crashlyticsDep ]
```
ncooke3, line 142:
```suggestion
                             dependencies:@[]
```

But, this API is changing in Firebase v11 to remove the array parameter, so this line will need to be removed in the near future. The updated API is staged in the release-11.0 branch: https://github.com/firebase/firebase-ios-sdk/blob/69b4dba1ada7ac5ccc73ca8351158224fc56d1ab/FirebaseCore/Extension/FIRComponent.h#L72 (fyi the podspec is staged on SpecsStaging)

packages/crashlytics/ios/RNFBCrashlytics/RNFBCrashlyticsInitProvider.m
ncooke3: Okay, your latest implementation ended up needed only a few tweaks and it should be gtg. Leaving some review comments...

packages/crashlytics/ios/RNFBCrashlytics/RNFBCrashlyticsInitProvider.m
mikehardy: okay - thanks for the heads-up. I suppose I could put some dynamic responds-to-selector type logic here to check which implementation is in underlying to be forwards- + backwards-compatible but a compile error will make this fast enough to detect when v11 comes out, and our breaking changes policy states that we (react-native-firebase) expects control of firebase-ios-sdk version and we are able to do underlying breaks without surfacing them in our versions, so won't even trigger a breaking change at our layer.

TL;DR: going to defer this until compile error surfaces it then just change to match and release

packages/crashlytics/ios/RNFBCrashlytics/RNFBCrashlyticsInitProvider.m
mikehardy: Invaluable review pushing me to a more reliable implementation and then polishing that, really appreciated @ncooke3  🤜 🤛

PR #7908, opened 2024-07-15
feat(analytics): allow custom event parameters for Item in events

Pietahr:
### Description

We wanted to add custom parameters to the items for ecommerce events. Firebase supports it (https://developers.google.com/analytics/devguides/collection/ga4/item-scoped-ecommerce), but it was not yet supported by react-native-firebase. 
This MR is inspired by https://github.com/invertase/react-native-firebase/pull/5811

### Checklist

  - [x] Yes
  - [ ] `Android`
  - [ ] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

I added a test case.

---

### DIFF
--- a/packages/analytics/lib/index.d.ts
+++ b/packages/analytics/lib/index.d.ts
@@ -141,6 +141,11 @@ export namespace FirebaseAnalyticsTypes {
      * The promotion name associated with the item.
      */
     promotion_name?: string;
+    /**
+     * Custom event parameters. The parameter names can be up to 40 characters long and must start with an alphabetic character and contain only alphanumeric characters and underscores. String parameter values can be up to 100 characters long.
+     * The "firebase_", "google_" and "ga_" prefixes are reserved and should not be used for parameter names.
+     */
+    [key: string]: string | number;
   }
 
   export interface AddPaymentInfoEventParameters {
--- a/packages/analytics/lib/structs.js
+++ b/packages/analytics/lib/structs.js
@@ -15,7 +15,7 @@
  */
 import struct from '@react-native-firebase/app/lib/common/struct';
 
-const Item = struct({
+const Item = struct.interface({
   item_brand: 'string?',
   item_id: 'string?',
   item_name: 'string?',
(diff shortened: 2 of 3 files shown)


packages/analytics/lib/index.d.ts
```diff
@@ -109,6 +109,10 @@ export namespace FirebaseAnalyticsTypes {
      * See https://github.com/invertase/react-native-firebase/issues/4578#issuecomment-771703420 for more information
      */
     price?: number;
+    /**
+     * Custom event parameters.
+     */
+    [key: string]: any;
```
russellwheatley, line 115: Only string and number are allowed here, please update. Worth updating information on custom parameters:

```
Parameter names can be up to 40 characters long and must start with an alphabetic character and contain only alphanumeric characters and underscores. String parameter values can be up to 100 characters long. The
"firebase_", "google_" and "ga_" prefixes are reserved and should not be used for parameter names.
```

packages/analytics/lib/structs.js
```diff
@@ -15,7 +15,7 @@
  */
 import struct from '@react-native-firebase/app/lib/common/struct';
 
-const Item = struct({
+const Item = struct.interface({
```
russellwheatley, line 18: Not sure on just making `struct.interface`, presumably this would make all other properties permissible?

russellwheatley (changes requested):

packages/analytics/lib/structs.js
Pietahr: No, it accepts arbitrary custom event parameters while rejecting defined parameters with wrong types

dannyBies:
This is something I'm looking to use as well. Is there anything I can do to help this across the line?

mikehardy:
Hey @dannyBies 👋  the commit trail in here was a mess and it had one unaddressed commit
both were trivial to fix up though, faster than talking about it really so I just squashed the commits down to the change only, with an updated type per request from @russellwheatley 

So if Russell re-reviews and agrees this is good to go

Either way, my re-push had the side effect of re-running CI which means that the patch-package workflow re-runs and - if you follow the details link and check the summary of the patch packages worfklow you should find a patch set you can apply to your project that will work. Providing feedback on how it works is very useful

dannyBies:
Amazing, thanks for the extremely quick response, it's much appreciated!

I'll have a look at using patch-package tomorrow.

russellwheatley (approved):

Issue #7816, opened 2024-05-31
[🐛]  Android build failed 


timea-sillai:
"@react-native-firebase/auth": "^20.0.0",
"react-native": "0.72.6",

> Task :react-native-firebase_auth:compileDebugJavaWithJavac FAILED

Deprecated Gradle features were used in this build, making it incompatible with Gradle 9.0.

after: ./gradlew build --warning-mode all

> Task :react-native-firebase_auth:compileDebugJavaWithJavac FAILED
/xxx/xxx/xxx/node_modules/@react-native-firebase/auth/android/src/main/java/io/invertase/firebase/auth/ReactNativeFirebaseAuthModule.java:1620: error: pattern matching in instanceof is not supported in -source 11
                    if (exception instanceof FirebaseAuthUserCollisionException collEx) {
                                                                                ^
  (use -source 16 or higher to enable pattern matching in instanceof)
1 error

I saw that the sourceCompatibility is JavaVersion.VERSION_11 but line 1620 from /xx/xx/xx/@react-native-firebase/auth/android/src/main/java/io/invertase/firebase/auth/ReactNativeFirebaseAuthModule.java 
is in java 17 like this: if (exception instanceof FirebaseAuthUserCollisionException collEx) 
and it should be 
             if (exception instanceof FirebaseAuthUserCollisionException) {
                 FirebaseAuthUserCollisionException collEx = (FirebaseAuthUserCollisionException) exception; 
               }

ChrisYohann:
Same here

devdongwoo:
I solved it thanks to your help :)

devdongwoo:
> > I solved it thanks to your help :)
> 
> How did you fix it?

your "ReactNativeFirebaseAuthModule.java" file change from ``` if (exception instanceof FirebaseAuthUserCollisionException collEx)```to  ```if (exception instanceof FirebaseAuthUserCollisionException) {
FirebaseAuthUserCollisionException collEx = (FirebaseAuthUserCollisionException) 
```
Do you understand?

r-aurejac:
> > > I solved it thanks to your help :)
> > 
> > 
> > How did you fix it?
> 
[...]

PR #8023, opened 2024-09-18
[Fix] Android Build Failed (Task :react-native-firebase_auth:compileDebugJavaWithJavac FAILED)

sanjeevzsuper:
### Description

In Android build getting below error before this fix: -
 

> Task :react-native-firebase_auth:compileDebugJavaWithJavac FAILED
>   /xxx/xxx/xxx/node_modules/@react-native-firebase/auth/android/src/main/java/io/invertase/firebase/auth/ReactNativeFirebaseAuthModule.java:1620: error: pattern matching in instanceof is not supported in -source 11

  
Tried all the solution but not working So I have done these PR changes then I am able to run build android build

### Related issues

https://github.com/invertase/react-native-firebase/issues/7816

### DIFF
--- a/packages/auth/android/src/main/java/io/invertase/firebase/auth/ReactNativeFirebaseAuthModule.java
+++ b/packages/auth/android/src/main/java/io/invertase/firebase/auth/ReactNativeFirebaseAuthModule.java
@@ -88,14 +88,13 @@ import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import javax.annotation.Nullable;
 
-@SuppressWarnings({ "ThrowableResultOfMethodCallIgnored", "JavaDoc" })
+@SuppressWarnings({"ThrowableResultOfMethodCallIgnored", "JavaDoc"})
 class ReactNativeFirebaseAuthModule extends ReactNativeFirebaseModule {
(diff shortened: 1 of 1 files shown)


mikehardy (changes requested):
It is impossible to tell what this PR is doing since the file was reformatted
Can you push a PR that only has the functional change you desire, plus a description of what the PR is doing?

Please note that attempts to make react-native-firebase compatible with JDK17 are likely not functional + cannot be supported, as the various toolchains have minimum requirements of JDK17 now

mikehardy (commented):
I see - for the reasons stated in the related issue, this *will* fix this specific compile issue with JDK17 being required, *however* that will not result in a supported configuration. JDK17 is the minimum supported version and needs to be used throughout your android app build toolchains. From that perspective, removing this JDK17-specific feature is actually a developer regression, as this JDK17 feature usage results in a very fast + very obvious failure for developers indicating to them immediately that something is wrong (where "something" is: you must use JDK17)

mikehardy:
JDK17 is a minimum requirement, this will not be merged, rather we have put a compile time failure in place with a descriptive error message to clearly state the minimum

PR #8054, opened 2024-10-09
fix(auth, ios): demonstrate test integration of upstream fix 13772

mikehardy:
This is not for merge

This is a demonstration of how to patch a source file integrated via cocoapods in a react-native context

It should handle these cases:

- no tools installed (that is, there are things like `[cocoapods-patch](https://github.com/DoubleSymmetry/cocoapods-patch)` but you have to install things, that is unwanted
- patch is already applied (using `--forward` and the `(command || true)` construction in patch call means it will silently proceed without a non-zero exit code, even if patch is already applied)

Doing something like this in your project may help you test-integrate the upstream fix to see if it stops the crashing for you

See related discussion at https://github.com/firebase/firebase-ios-sdk/issues/13761


mikehardy:
This was never intended for merge - it was for demonstration only - and I will close it now

PR #8061, opened 2024-10-15
feat: support custom firebase.json path

CHOIMINSEOK:
### Description
Currently, the firebase.json file is configured only when it is located in the parent directory of the Android project. However, React Native can be used in different ways, especially when integrating it into an existing native project.

In my case, I’m trying to integrate a React Native app into a brownfield native app as a submodule, so retrieving firebase.json from the upper directory always fails. I think providing an option to set a custom "firebaseJsonDir" would be very useful for people in my situation.

I add "customFirebaseJsonDir" option in same way to override firebase sdk versions and reactNativeDir. 

I would appreciate any help to add this feature on iOS too. 

### Release Summary
- Support custom firebase.json path

### Checklist

  - [x] Yes
  - [x] `Android`
  - [ ] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### DIFF
--- a/packages/app/android/firebase-json.gradle
+++ b/packages/app/android/firebase-json.gradle
@@ -4,16 +4,22 @@ import groovy.json.JsonSlurper
 
 String fileName = 'firebase.json'
 String jsonRoot = 'react-native'
+String customFirebaseJsonDir = rootProject.ext.get("react-native")["options"]["firebaseJsonDir"]
 
 File jsonFile = null
-File parentDir = rootProject.projectDir
 
-for (int i = 0; i <= 3; i++) {
-  if (parentDir == null) { break }
-  parentDir = parentDir.parentFile
-  if (parentDir != null) {
-    jsonFile = new File(parentDir, fileName)
-    if (jsonFile.exists()) { break }
+if (customFirebaseJsonDir != null) {
+  jsonFile = new File(customFirebaseJsonDir, fileName)
+} else {
+  File parentDir = rootProject.projectDir
+
+  for (int i = 0; i <= 3; i++) {
+    if (parentDir == null) { break }
+    parentDir = parentDir.parentFile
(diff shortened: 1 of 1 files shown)


mikehardy:
Hmm - the iOS side does a search for firebase.json in two places:

1- one in shell script installed as a build script inside the Xcode project, during build in order to pull values that it needs to put in the Info.plist: 

https://github.com/invertase/react-native-firebase/blob/d55e891064d7343788526f688d303dd01c2484c4/packages/app/ios_config.sh#L77-L87

2- one in ruby which is pulled in to the RNFBApp podspec to get the SDK versions

https://github.com/invertase/react-native-firebase/blob/d55e891064d7343788526f688d303dd01c2484c4/packages/app/firebase_json.rb#L35-L59

I am not sure how to transmit to the Xcode build script that you have a specific firebase.json path, the only thing I can think of is an environment variable. For firebase_json.rb it could be a Podfile variable perhaps

I wonder if, in all these cases, it might be easier as part of your build / package process to determine which directories are actually being searched...and during build make sure the firebase.json is just copied/placed into a directory that is already searched?

That might be easier than complicating the build / config process here (which is already pretty touchy...)

CHOIMINSEOK:
Hi! @mikehardy  Thanks for reply. 

> I wonder if, in all these cases, it might be easier as part of your build / package process to determine which directories are actually being searched...and during build make sure the firebase.json is just copied/placed into a directory that is already searched?

Yeah that might work, but all of those who have different rn app structure have to feature out where the build script of react-native-firebase resolves firebase.json. I just want to provide a convenient option for them. Besides, in some case like me the location might be outside of project which is so weird. 

```
directory <= this is where react-native-firebase starts to resolve firebase.json
 ㄴ android project
  ㄴ react native project
```

Issue #8027, opened 2024-09-19
🐛 @react-native-firebase/firestore Need Sum in aggregation method


Ganeshguru5:
## Sum in aggregation of @react-native-firebase/firestore

In the firebase documentation. They detailed that they have the aggregation feature to sum a particular attribute of a document. But i cant found this method in @react-native-firebase/firestore. Instead everyone reccommends to fetch all data and **sum** the value using array-reduce method. I dont want to get 10 or 100 documents to obtain a single value. Those reads are unwanted.

mikehardy:
This was one of the major feature enhancements of firestore in the last long while, I think it's still a valid request as an API gap

PR #8115, opened 2024-11-05
feat(firestore): support for aggregate queries including `sum()` & `average()`

russellwheatley:
### Description

- Support for [getAggregateFromServer()](https://firebase.google.com/docs/reference/js/firestore_.md?_gl=1*1d4sq9b*_up*MQ..*_ga*MTE3MTA3NTY0NC4xNzMwOTEyNDk0*_ga_CW55HF8NVT*MTczMDkxMjQ5NC4xLjAuMTczMDkxMjQ5NC4wLjAuMA..#getaggregatefromserver_2073a74). Modular support only.

### Related issues

closes https://github.com/invertase/react-native-firebase/issues/8027

### Checklist

  - [ ] Yes
  - [ ] `Android`
  - [ ] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [ ] No

### Test Plan

---

### DIFF
--- a/packages/firestore/android/src/reactnative/java/io/invertase/firebase/firestore/ReactNativeFirebaseFirestoreCollectionModule.java
+++ b/packages/firestore/android/src/reactnative/java/io/invertase/firebase/firestore/ReactNativeFirebaseFirestoreCollectionModule.java
@@ -17,6 +17,8 @@ package io.invertase.firebase.firestore;
  *
  */
 
+import static com.google.firebase.firestore.AggregateField.average;
+import static com.google.firebase.firestore.AggregateField.sum;
 import static io.invertase.firebase.firestore.ReactNativeFirebaseFirestoreCommon.rejectPromiseFirestoreException;
 import static io.invertase.firebase.firestore.ReactNativeFirebaseFirestoreSerialize.snapshotToWritableMap;
 import static io.invertase.firebase.firestore.UniversalFirebaseFirestoreCommon.getFirestoreForApp;
@@ -205,7 +207,8 @@ public class ReactNativeFirebaseFirestoreCollectionModule extends ReactNativeFir
     ReadableArray filters,
     ReadableArray orders,
     ReadableMap options,
-    ReadableArray aggregateQueries
+    ReadableArray aggregateQueries,
+    Promise promise
   ){
     FirebaseFirestore firebaseFirestore = getFirestoreForApp(appName, databaseId);
     ReactNativeFirebaseFirestoreQuery firestoreQuery =
@@ -216,57 +219,76 @@ public class ReactNativeFirebaseFirestoreCollectionModule extends ReactNativeFir
         filters,
         orders,
         options);
+
     ArrayList<AggregateField> aggregateFields = new ArrayList<>();
 
     for (int i = 0; i < aggregateQueries.size(); i++) {
       ReadableMap aggregateQuery = aggregateQueries.getMap(i);
+
       String aggregateType = aggregateQuery.getString("aggregateType");
-      String fieldPath = aggregateQuery.getString("fieldPath");
-      if (aggregateType && fieldPath) {
+      String fieldPath = aggregateQuery.getString("field");
+
+      assert aggregateType != null;
         switch (aggregateType) {
           case "count":
-            aggregateFields.add(AggregateField.count(fieldPath));
+            aggregateFields.add(AggregateField.count());
             break;
           case "sum":
+            assert fieldPath != null;
             aggregateFields.add(AggregateField.sum(fieldPath));
             break;
           case "average":
-            aggregateFields.add(AggregateField.avg(fieldPath));
+            assert fieldPath != null;
+            aggregateFields.add(AggregateField.average(fieldPath));
             break;
           default:
-            break;
+            throw new Error("Invalid AggregateType: " + aggregateType);
         }
-      }
+    }
+      AggregateQuery firestoreAggregateQuery = firestoreQuery.query.aggregate(aggregateFields.get(0),
+      aggregateFields.subList(1, aggregateFields.size()).toArray(new AggregateField[0]));
 
-      AggregateQuery aggregateQuery = firestoreQuery.query.aggregate(aggregateFields);
-      aggregateQuery
+      firestoreAggregateQuery
         .get(AggregateSource.SERVER)
         .addOnCompleteListener(
           task -> {
             if (task.isSuccessful()) {
               WritableMap result = Arguments.createMap();
               AggregateQuerySnapshot snapshot = task.getResult();
-              aggregateFields.forEach(aggregateField -> {
-                switch (aggregateField.getOperator()) {
+
+              for (int k = 0; k < aggregateQueries.size(); k++) {
+                ReadableMap aggQuery = aggregateQueries.getMap(k);
+                String aggType = aggQuery.getString("aggregateType");
+                String field = aggQuery.getString("field");
+                String key = aggQuery.getString("key");
+                assert key != null;
+                assert aggType != null;
+                switch (aggType) {
                   case "count":
-                    result.putDouble(aggregateField.getFieldPath(), Long.valueOf(snapshot.getCount()).doubleValue());
+                    result.putDouble(key, Long.valueOf(snapshot.getCount()).doubleValue());
                     break;
                   case "sum":
-                    result.putDouble(aggregateField.getFieldPath(), snapshot.getSum(aggregateField.getFieldPath()));
+                    assert field != null;
+                    Number sum = (Number) snapshot.get(sum(field));
+                    assert sum != null;
+                    result.putDouble(key, sum.doubleValue());
                     break;
                   case "average":
-                    result.putDouble(aggregateField.getFieldPath(), snapshot.getAverage(aggregateField.getFieldPath()));
+                    assert field != null;
+                    Number average = snapshot.get(average(field));
+                    assert average != null;
+                    result.putDouble(key, average.doubleValue());
                     break;
                   default:
-                    break;
+                    throw new Error("Invalid AggregateType: " + aggType);
                 }
               }
+
               promise.resolve(result);
             } else {
               rejectPromiseFirestoreException(promise, task.getException());
             }
           });
-    }
   }
 
   @ReactMethod
--- a/packages/firestore/android/src/reactnative/java/io/invertase/firebase/firestore/ReactNativeFirebaseFirestoreCollectionModule.java
+++ b/packages/firestore/android/src/reactnative/java/io/invertase/firebase/firestore/ReactNativeFirebaseFirestoreCollectionModule.java
@@ -271,9 +271,12 @@ public class ReactNativeFirebaseFirestoreCollectionModule extends ReactNativeFir
                       break;
                     case "average":
                       Number average = snapshot.get(average(Objects.requireNonNull(field)));
-                      result.putDouble(
-                          Objects.requireNonNull(key),
-                          Objects.requireNonNull(average).doubleValue());
+                      String averageKey = Objects.requireNonNull(key);
+                      if (average == null) {
+                        result.putNull(averageKey);
+                      } else {
+                        result.putDouble(averageKey, Objects.requireNonNull(average).doubleValue());
+                      }
                       break;
                     default:
                       throw new Error("Invalid AggregateType: " + aggType);
--- a/packages/firestore/android/src/reactnative/java/io/invertase/firebase/firestore/ReactNativeFirebaseFirestoreCollectionModule.java
+++ b/packages/firestore/android/src/reactnative/java/io/invertase/firebase/firestore/ReactNativeFirebaseFirestoreCollectionModule.java
@@ -31,6 +31,7 @@ import com.google.firebase.firestore.*;
 import io.invertase.firebase.common.ReactNativeFirebaseEventEmitter;
 import io.invertase.firebase.common.ReactNativeFirebaseModule;
 import java.util.ArrayList;
+import java.util.Objects;
 
 public class ReactNativeFirebaseFirestoreCollectionModule extends ReactNativeFirebaseModule {
   private static final String SERVICE_NAME = "FirestoreCollection";
@@ -221,22 +222,18 @@ public class ReactNativeFirebaseFirestoreCollectionModule extends ReactNativeFir
 
     for (int i = 0; i < aggregateQueries.size(); i++) {
       ReadableMap aggregateQuery = aggregateQueries.getMap(i);
-
       String aggregateType = aggregateQuery.getString("aggregateType");
       String fieldPath = aggregateQuery.getString("field");
 
-      assert aggregateType != null;
-      switch (aggregateType) {
+      switch (Objects.requireNonNull(aggregateType)) {
         case "count":
           aggregateFields.add(AggregateField.count());
           break;
         case "sum":
-          assert fieldPath != null;
-          aggregateFields.add(AggregateField.sum(fieldPath));
+          aggregateFields.add(AggregateField.sum(Objects.requireNonNull(fieldPath)));
           break;
         case "average":
-          assert fieldPath != null;
-          aggregateFields.add(AggregateField.average(fieldPath));
+          aggregateFields.add(AggregateField.average(Objects.requireNonNull(fieldPath)));
           break;
         default:
           throw new Error("Invalid AggregateType: " + aggregateType);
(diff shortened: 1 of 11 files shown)


packages/firestore/android/src/reactnative/java/io/invertase/firebase/firestore/ReactNativeFirebaseFirestoreCollectionModule.java
```diff
@@ -214,6 +312,8 @@ public void collectionGet(
             orders,
             options);
     handleQueryGet(firestoreQuery, getSource(getOptions), promise);
+
+
```
mikehardy, line 316: low-value comment: that should fail `yarn lint:android` shouldn't it? (but just by running lint:android it'll auto-format it and you can just commit result)

packages/firestore/android/src/reactnative/java/io/invertase/firebase/firestore/ReactNativeFirebaseFirestoreCollectionModule.java
```diff
@@ -193,6 +198,99 @@ public void collectionCount(
...
+              WritableMap result = Arguments.createMap();
+              AggregateQuerySnapshot snapshot = task.getResult();
+
+              for (int k = 0; k < aggregateQueries.size(); k++) {
+                ReadableMap aggQuery = aggregateQueries.getMap(k);
+                String aggType = aggQuery.getString("aggregateType");
+                String field = aggQuery.getString("field");
+                String key = aggQuery.getString("key");
+                assert key != null;
+                assert aggType != null;
```
mikehardy, line 265: Are any of these asserts useful? It was my understanding that assert verifications are disabled by default so this would never really trigger anything, and if it did trigger the whole app dies (via an AssertionError) without notice of what did it since there is no exception handler set to catch Error (vs a try/catch with typical catch of Exception wrapping a block where you use Object.requireNonNull

mikehardy (commented):
exciting! couldn't but look through it even though draft+android-only+no-e2e at the moment because I contemplated implementing this myself months ago :-)

Had a couple quick thoughts but mostly just wanted to say this is exciting. Cheers

packages/firestore/ios/RNFBFirestore/RNFBFirestoreCollectionModule.m
```diff
@@ -216,6 +216,83 @@ - (void)invalidate {
...
+    } else {
+      NSMutableDictionary *snapshotMap = [NSMutableDictionary dictionary];
+      
+      for (NSDictionary *aggregateQuery in aggregateQueries) {
+        NSString *aggregateType = aggregateQuery[@"aggregateType"];
+        NSString *fieldPath = aggregateQuery[@"field"];
+        NSString *key = aggregateQuery[@"key"];
+        assert(key);
+        
+        if([aggregateType isEqualToString:@"count"]){
```
mikehardy, line 276: also really low-value comment (because, lint...) but `yarn lint:ios:fix` will clean that up for your

packages/firestore/android/src/reactnative/java/io/invertase/firebase/firestore/ReactNativeFirebaseFirestoreCollectionModule.java
russellwheatley: Removed in place of `Object.requireNonNull`. Largely to remove warnings in android studio 👍

packages/firestore/lib/modular/index.d.ts
```diff
@@ -10,6 +10,7 @@ import Query = FirebaseFirestoreTypes.Query;
 import FieldValue = FirebaseFirestoreTypes.FieldValue;
 import FieldPath = FirebaseFirestoreTypes.FieldPath;
 import PersistentCacheIndexManager = FirebaseFirestoreTypes.PersistentCacheIndexManager;
+import AggregateQuerySnapshot = FirebaseFirestoreTypes.AggregateQuerySnapshot;
```
mikehardy, line 13: I was curious where this was defined as [the types looked pretty intense](https://github.com/firebase/firebase-js-sdk/blob/e3e20784503347ba5cab1ead3a7b73df87125cb4/packages/firestore/src/lite-api/aggregate_types.ts#L68-L119) - looks like we get this for free from upstream? (if so: fantastic)

packages/firestore/e2e/Aggregate/AggregateQuery.e2e.js
```diff
@@ -0,0 +1,272 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+const COLLECTION = 'firestore';
+const { wipe } = require('../helpers');
+describe('getAggregateFromServer()', function () {
+  before(function () {
+    return wipe();
+  });
```
mikehardy, line 22: wipe is async - this one (and the `count.e2e.js` e2e file next to this new `AggregateQuery.e2e.js` one should use it async like in other areas https://github.com/invertase/react-native-firebase/blob/bb76c4d802890855cb7b1f151f422bea28101715/packages/firestore/e2e/Bundle/loadBundle.e2e.js#L25-L27

```suggestion
  before(async function () {
    return async wipe();
  });
```

packages/firestore/e2e/Aggregate/AggregateQuery.e2e.js
```diff
@@ -0,0 +1,272 @@
...
+      } = firestoreModular;
+      const firestore = getFirestore();
+
+      const colRef = collection(firestore, `${COLLECTION}/aggregate-count-field-path/collection`);
+
+      await Promise.all([
+        setDoc(doc(colRef, 'one'), { bar: 5, baz: 4 }),
+        setDoc(doc(colRef, 'two'), { bar: 5, baz: 4 }),
+        setDoc(doc(colRef, 'three'), { bar: 5, baz: 4 }),
+      ]);
```
mikehardy, line 178: I saw this and it reminded me of this incredibly traumatic issue https://github.com/invertase/react-native-firebase/issues/3004 😆 

It makes me want to poke at things a bit more to make sure we handle things correctly as javascript number handling is really picky. A classic is that 0.1 is not represented well (because it's IEEE floating point internally, not an integer or decimal)

Suggest making the average actually have to work by using 4 5 and 6
Suggest attempting an sum of 0.1, 0.1 and 0.1 to see what happens (is it 0.3, or is it 0.30000000000004? This is a javascript classic)

There are also firestore limitations where values that not numbers are ignored during the aggregation but since that is all on the backend I don't think we need to probe those sorts of limitations - but verifying our number handling (is it a "number", or an "integer" or a "double" ?) may be insightful

packages/firestore/ios/RNFBFirestore/RNFBFirestoreCollectionModule.m
```diff
@@ -216,6 +216,77 @@ - (void)invalidate {
...
+      [aggregateFields addObject:[FIRAggregateField aggregateFieldForCount]];
+    } else if ([aggregateType isEqualToString:@"sum"]) {
+      [aggregateFields addObject:[FIRAggregateField aggregateFieldForSumOfField:fieldPath]];
+    } else if ([aggregateType isEqualToString:@"average"]) {
+      [aggregateFields addObject:[FIRAggregateField aggregateFieldForAverageOfField:fieldPath]];
+    } else {
+      NSString *reason = [@"Invalid Aggregate Type: " stringByAppendingString:aggregateType];
+      @throw [NSException exceptionWithName:@"RNFB Firestore: Invalid Aggregate Type"
+                                     reason:reason
+                                   userInfo:nil];
```
mikehardy, line 251: throwing in this way creates a different error handling pathway in the javascript as it bubbles up from native through to javascript to app if I recall correctly?

I think it would be best to reject in this case so that all flow is handled in the typical resolve/reject style and it's a Promise rejection vs a separate thing you have to try/catch

You can I think take the exception you create here and just inline it as you send it in to the common rejection method e.g. `[RNFBFirestoreCommon promiseRejectFirestoreException:reject error:error];`

```suggestion
      [RNFBFirestoreCommon promiseRejectFirestoreException:reject error:[NSException exceptionWithName:@"RNFB Firestore: Invalid Aggregate Type"
                                     reason:reason
                                   userInfo:nil]];
```

(hopefully that compiles or is close to compiling, and then yarn lint:ios:fix because it looks awful in this textbox)

packages/firestore/android/src/reactnative/java/io/invertase/firebase/firestore/ReactNativeFirebaseFirestoreCollectionModule.java
```diff
@@ -193,6 +197,99 @@ public void collectionCount(
...
+            options);
+
+    ArrayList<AggregateField> aggregateFields = new ArrayList<>();
+
+    for (int i = 0; i < aggregateQueries.size(); i++) {
+      ReadableMap aggregateQuery = aggregateQueries.getMap(i);
+      String aggregateType = aggregateQuery.getString("aggregateType");
+      String fieldPath = aggregateQuery.getString("field");
+
+      switch (Objects.requireNonNull(aggregateType)) {
```
mikehardy, line 228: similar to the iOS error handling comment - this will throw NullPointerException if for some reason it evaluates to null, and that is a separate error-handling pathway for calling code vs Promise.reject so it will be unexpected.

It seems like this is just to handle "may be null?" Android Studio warnings?
Let me have a look at reshaping this so the warnings go away, but the non-null type assertions / conversions for safety are done in a try/catch where the final error handling pathway is a Promise reject if any of them are null

I'll push a commit that has that change in a moment

packages/firestore/ios/RNFBFirestore/RNFBFirestoreCollectionModule.m
```diff
@@ -216,6 +216,80 @@ - (void)invalidate {
...
+      [aggregateFields addObject:[FIRAggregateField aggregateFieldForAverageOfField:fieldPath]];
+    } else {
+      NSString *reason = [@"Invalid Aggregate Type: " stringByAppendingString:aggregateType];
+      [RNFBFirestoreCommon
+          promiseRejectFirestoreException:reject
+                                    error:[NSException exceptionWithName:
+                                                           @"RNFB Firestore: Invalid Aggregate Type"
+                                                                  reason:reason
+                                                                userInfo:nil]];
+    }
```
mikehardy, line 255: Another classic - the promise rejection doesn't actually stop native execution, so if you don't return after it, then execution continues, something else blows up and another rejection is attempted (triggering an error where promise was attempted to reject twice....) or somehow it succeeds then the resolve block is also called (another error)

I only know this because of course I have made this mistake before then painfully diagnosed + fixed it post-release some long time ago...


```suggestion
    } else {
      NSString *reason = [@"Invalid Aggregate Type: " stringByAppendingString:aggregateType];
      [RNFBFirestoreCommon
          promiseRejectFirestoreException:reject
                                    error:[NSException exceptionWithName:
                                                           @"RNFB Firestore: Invalid Aggregate Type"
                                                                  reason:reason
                                                                userInfo:nil]];
      return;
    }
```

packages/firestore/lib/modular/index.d.ts
mikehardy: Ah - I see they were split to the other (non-modular) index.d.ts since count was implemented before modular was a thing here - types look better now in that file +1

packages/firestore/e2e/Aggregate/AggregateQuery.e2e.js
mikehardy: Ah ha!! Not sure exactly what to do about this but the PR comments from the one that resolved #3004 may be instructive? This is really subtle stuff. The hunch was right though - check the result on iOS:

>      AssertionError: expected 0.30000000000000004 to equal 0.3

Devious.

mikehardy (changes requested):
- ~~has one code error from just reading in iOS, noted with suggestion to add return after executing reject block~~
- the number handling probe hit paydirt! which is kind of awful but also represents progress (known error better than unknown error...)
- ~~some new android issue ?~~

It really does look like progress tho, just this one is a real feature size-wise so has real subtleties

packages/firestore/e2e/Aggregate/AggregateQuery.e2e.js
mikehardy: Looks like all platforms reproduce the number-handling issue now, so it is consistent at least
Sorry my android push was poorly tested and had an immediate problem - had an issue running e2e android locally so just pushed without testing, and of course that's how an untested push goes...

packages/firestore/lib/FirestoreAggregate.js
```diff
@@ -36,17 +38,55 @@ export class FirestoreAggregateQuery {
...
+      return { ...this._data };
+    }
+  }
+}
+
+export const AggregateType = {
+  SUM: 'sum',
+  AVG: 'average',
+  COUNT: 'count',
+};
```
Lyokone, line 65: I was wondering if it should be an enum but maybe it's not important in JS

Lyokone (approved):

packages/firestore/lib/FirestoreAggregate.js
russellwheatley: Unfortunately, no enums in JS, although we could use enums when we move to TS...

mikehardy: Typescript has enums but javascript doesn't unless I'm way off - web search just now appeared to reconfirm that tho

mikehardy (approved):
still LGTM - just looked at the last couple test additions (nice)

joaqo:
Same here @mikehardy, can't for the life of me make `sum()` work. I read the PR's code and tried things like this:

```typescript

import firestore, { count, sum, getAggregateFromServer} from "@react-native-firebase/firestore"

// Gets permission error
const result = await getAggregateFromServer(Reference.ordersForSeller(seller.id).where("status", "==", "arrived"), { sum: sum("totalCents") })

// Works fine
const loler = await Reference.ordersForSeller(seller.id).where("status", "==", "arrived").count().get()
```

But I get weird permission errors, and as @alanschwarz said, there is no `sum` method on query objects, so I really don't know what to do.

Could you provide a working example code snippet on how to make `sum` work? Thanks for the amazing work!

mikehardy:
@joaqo there is a pile of tests right here that executes in CI all the time 

https://github.com/invertase/react-native-firebase/blob/9c6f6075900f0927cf1bd1400a49243126007366/packages/firestore/e2e/Aggregate/AggregateQuery.e2e.js#L103

Now, sometimes our typings aren't quite right - I've seen that - but these tests definitely work and I remember working through them with Russell to probe the behavior of the numbers vs firebase-js-sdk given there are lots of little gotchas in how javascript handles numbers due to IEEE doubles as the numeric type

joaqo:
Hi @mikehardy, I actually based my tests on that snippet, but I really can't get it to work. I get a `permission-denied` firestore error. This is the test I ran:

```ts
async function getOrderCount(seller: any) {
  const aggregateSpec = {
    countCollection: count(),
  }
  const ref = Reference.ordersForSeller(seller.id).where("status", "==", "arrived")
  const result = await getAggregateFromServer(ref, aggregateSpec)
  const data = result.data()
  console.log("Count", data.countCollection)  // Gets permission errors!
}

async function getOrderCountSlow(seller: any) {
  const ref = Reference.ordersForSeller(seller.id).where("status", "==", "arrived")
  const result = await ref.get()
  console.log("Count Slow", result.docs.length)  // Works just fine!
}
```

So I do have read permissions for this collection as evidenced on the second function, but for some reason the first function returns a `permission-denied` firestore error. Is it possible that these aggregators need a different permission than just a READ permission? According to the docs [it shouldn't](https://firebase.google.com/docs/firestore/query-data/aggregation-queries) though.

Thanks!

mikehardy:
I don't know, sorry, haven't played with the intersection of cloud security rules and the aggregates. This isn't react-native specific, I'd suggest using stackoverlow and tagging your issue well, the firebase support team is frequently there in official capacity

joaqo:
Here is a more concise example that shows the issue:

```ts
async function getOrderCount(seller: any) {
  const aggregateSpec = {
    countCollection: count(),
  }
  const ref = Reference.ordersForSeller(seller.id).where("status", "==", "arrived")

  const correct = await ref.count().get()
  console.log("Count correct", correct.data().count)

  const incorrect = await getAggregateFromServer(ref, aggregateSpec)
  const data = incorrect.data()
  console.log("Count incorrect", data.countCollection)
}

async function getOrderCountSlow(seller: any) {
  const ref = Reference.ordersForSeller(seller.id).where("status", "==", "arrived")
  const result = await ref.get()
  console.log("Count Slow", result.docs.length)  // Works just fine!
}
```

mikehardy:
You may be on to something here @joaqo ! Very interesting. And it looks like it may be react-native-firebase specific, contrary to my pointer towards stackoverflow for more general support

May I ask for you to open up a new issue with details on versions, platforms you reproduce with etc? (it might vary between ios and android even). This looks like something we need to investigate as an issue vs comments on a closed PR (however helpful the comments may be!)

Slight follow-on - not sure how practical this is for you, but are you able to test with the firebase-js-sdk perhaps? Just to confirm that this behavior isn't generic to the aggregate queries for whatever reason but really is react-native-firebase local

joaqo:
No problem, I can certainly create a separate issue with more details! I don't think I can provide a firebase-js example though as I'm pretty tight on time, and our app is native only. Thanks again!

joaqo:
Done @mikehardy: https://github.com/invertase/react-native-firebase/issues/8253

PR #8125, opened 2024-11-10
fix: correct type for setBackgroundMessageHandler

ethanneff:
The `async` breaks the https://typescript-eslint.io/rules/require-await/ rule. Even in the examples, you do not need to `await  remoteMessage`

### DIFF
--- a/docs/messaging/usage/index.md
+++ b/docs/messaging/usage/index.md
@@ -200,7 +200,7 @@ import messaging from '@react-native-firebase/messaging';
 import App from './App';
 
 // Register background handler
-messaging().setBackgroundMessageHandler(async remoteMessage => {
+messaging().setBackgroundMessageHandler(remoteMessage => {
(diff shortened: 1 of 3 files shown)


mikehardy:
Hmmm I don't know about this one
It has been a long time and I can't dig it up quickly - sorry - but I seem to recall many troubleshooting issues where it needed to be async and did not work unless it was. It is flagged in my mind as a pretty critical part to get the whole thing to work.

Side question, related: how can you call async APIs in the background handler and await them if the handler is not async ?

Is there a way to drive the definition and types to a place where they are more correctly async to satisfy the rule, vs taking out the async? Because in my experience at least I can't +1 that

mikehardy:
I looked at this again, and I am *certain* the handler should be async
I'm not sure about that lint rule or whether it is possible to alter the types to conform to that lint rule in some way, but I know that this change cannot be merged because the handler needs to be async to function correctly

PR #8135, opened 2024-11-14
refactor: de-duplicate deprecation warning calls and tests

mikehardy:
A quick take on reducing boilerplate - I wasn't sure it would even work so it was easier to just try it vs try to explain it, since if it didn't work it wasn't worth explaining but the code itself was just the work of a few minutes

### DIFF
--- a/packages/app/__tests__/app.test.ts
+++ b/packages/app/__tests__/app.test.ts
@@ -10,6 +10,16 @@ import firebase, {
   setLogLevel,
 } from '../lib';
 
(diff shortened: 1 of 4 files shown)


mikehardy:
I did a positive / negative visual test on this to make sure I wasn't lying to myself and if I specifically broke the test by calling deprecated in non-deprecated slot, or calling non-deprecated in deprecated slot it did trigger failures as expected

russellwheatley (approved):

PR #8128, opened 2024-11-22
feat(messaging, ios): Background Completion Handler from JS

Beat-YT:
## What feature would you like to see?

When receiving a background message, the package will calls the completion handler after fixed 25 seconds, but this has some repercussion, considering that the system limit before the app gets killed is 30 seconds.

If an app stays in the background longer than necessary, iOS may reduce how often it triggers the app for future background tasks to save on battery, affecting performance and notification reliability over time.

An option to call the completion handler directly from JavaScript would be ideal. This would allow apps to signal when they are done with background processing, instead of relying on the default timeout, enhancing responsiveness and efficiency for iOS background tasks.

By taking a look at the source code, we can see that it was planned to add this feature, but it has yet to be supported.
https://github.com/invertase/react-native-firebase/blob/227e1ed54cf668cbc701c0ed2542b8ec66957e1e/packages/messaging/ios/RNFBMessaging/RNFBMessaging%2BAppDelegate.m#L168-L173


Beat-YT:
### Description

Added  code to call the completion handler directly from JavaScript. This allow apps to signal when they are done with background processing, instead of relying on the default timeout, enhancing responsiveness and efficiency for iOS background tasks.

### Checklist

  - [x] Yes
  - [ ] `Android`
  - [x] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No



### Test Plan

I have tested this on my app, and noticed the app spends less time in background, improving performance.

🔥

packages/messaging/ios/RNFBMessaging/RNFBMessaging+AppDelegate.m
```diff
@@ -156,30 +156,34 @@ - (void)application:(UIApplication *)application
...
-          [application beginBackgroundTaskWithExpirationHandler:^{
-            if (backgroundTaskId != UIBackgroundTaskInvalid) {
-              [application endBackgroundTask:backgroundTaskId];
-              backgroundTaskId = UIBackgroundTaskInvalid;
-            }
-          }];
-      // TODO add support in a later version for calling completion handler directly from JS when
-      // user JS code complete
+      sharedInstance.backgroundTaskId = [application beginBackgroundTaskWithExpirationHandler:^{
+        if (backgroundTaskId != UIBackgroundTaskInvalid) {
```
mikehardy, line 166: do these need to be switched to `sharedInstance.backgroundTaskId` as well?

Beat-YT:
my apologies, I missed that when porting in my implementation.

I am not familiar with objective-c and not sure if this is flawless. 
In my app's implementation I made the choice to call the completion handler from the main thread like so:
```objective-c
RCT_EXPORT_METHOD(completeNotificationProcessing) {
  dispatch_async(dispatch_get_main_queue(), ^{
    RNFBMessagingAppDelegate *appDelegate = [RNFBMessagingAppDelegate sharedInstance];
    if (appDelegate.completionHandler) {
      appDelegate.completionHandler(UIBackgroundFetchResultNewData);
      appDelegate.completionHandler = nil;
    }
    if (appDelegate.backgroundTaskId != UIBackgroundTaskInvalid) {
      [[UIApplication sharedApplication] endBackgroundTask:appDelegate.backgroundTaskId];
      appDelegate.backgroundTaskId = UIBackgroundTaskInvalid;
    }
  });
}
```
I am not sure if this is needed really, but I am better off safe than sorry.
Perhaps @russellwheatley knows better than me and can give his thoughts on this.

packages/messaging/ios/RNFBMessaging/RNFBMessagingModule.m
```diff
@@ -219,6 +219,19 @@ - (NSDictionary *)constantsToExport {
   return resolve(@([RCTConvert BOOL:@(notifCenter.isHeadless)]));
 }
 
+RCT_EXPORT_METHOD(completeNotificationProcessing) {
+  RNFBMessagingAppDelegate *appDelegate = [RNFBMessagingAppDelegate sharedInstance];
+  if (appDelegate.completionHandler) {
```
mikehardy, line 224: I think this should be on the same queue as the initial set of completionHandler / backgroundTaskId otherwise there's a tiny chance via race condition that multiple threads are working on their state at the same time

It's this above, would that work here?

>                     dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_HIGH, 0), ^{`

mikehardy (changes requested):
I think this looks really good - it appears this will work without regression or behavior change as it is, and should allow fast completion otherwise.

One comment about the exact thing you mentioned - the possibility of a race condition on the sharedInstance property state was the only thing that jumped out at me

If access to those properties was thread safe, I'd say +1 for sure

mikehardy:
wait stale bot! I thought this was worthwhile, was just waiting on a question

Beat-YT:
Sorry Mike for the delay, I released my app a month ago (with the patch) and didn't pay much attention to the PR.

I think we can fire the task using the  same queue as the initial set of completionHandler / backgroundTaskId, as you said.

Beat-YT:
> I think this looks really good - it appears this will work without regression or behavior change as it is, and should allow fast completion otherwise.
> 
> One comment about the exact thing you mentioned - the possibility of a race condition on the sharedInstance property state was the only thing that jumped out at me
> 
> If access to those properties was thread safe, I'd say +1 for sure

In a previous commit, I implemented using  the same queue as the initial set of completionHandler / backgroundTaskId like you suggested.

mikehardy (approved):
this looks really good - thank you!
I haven't had time to test this but I assume you've tested it and it works well?

PR #8184, opened 2024-12-11
docs: remove version_number example in ios-notification-images docs

petetnt:
### Description

This PR fixes an issue in the `ios-notification-images.md` doc, the doc tells to update VERSION_NUMBER but there's no version set in the example. In the gif below the image you can see the VERSION_NUMBER being set if you are fast enough.

The text:

> Make sure to change the version number VERSION_NUMBER with the currently installed version (check your Podfile.lock)

Screenshot from the gif:

image

As per discussions, we removed the `VERSION_NUMBER` example from the docs

### Release Summary

docs: removed VERSION_NUMBER example in the `ios-notification-image.md` docs

### Checklist

  - [x] Yes
  - [ ] `Android`
  - [x] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

Follow the guide, add the version number and run `pod install`

### DIFF
--- a/docs/messaging/ios-notification-images.md
+++ b/docs/messaging/ios-notification-images.md
@@ -38,7 +38,6 @@ target 'ImageNotification' do
 end
 ```
 
-- Make sure to change the version number `VERSION_NUMBER` with the currently installed version (check your Podfile.lock)
 - Install or update your pods using `pod install` from the `ios` folder
 
 ![step-2](/assets/docs/messaging/ios-notification-images-step-2.gif)


mikehardy (changes requested):
Hey there!

I think this should go the other way - if the Podfile has versions for that Pod set in some other way (and it does, for the main parts of react-native-firebase) - then you should not need a version number here as it should already be resolved to the one that the rest of the app is using, right?

So what should change I think is that the docs shouldn't tell you to update any VERSION_NUMBER anywhere for the iOS notification images.

This line would go away https://github.com/invertase/react-native-firebase/pull/8184/files#diff-05f38d5729372a6ce74811f5d33eb72c9a6b290203486ad013c80d1c3bc88f26R41

What needs to happen is to confirm that your build works with no version number (it should?) and then we can just get rid of that step entirely which should simplify things for people and give them one less thing to do in the Podfile on an ongoing basis - trying to keep up with those version number

petetnt:
Hi @mikehardy!

Makes sense, I removed the `VERSION_NUMBER` and the related line about it and rebased against master in https://github.com/invertase/react-native-firebase/pull/8184/commits/a0616783405606e81bdc15934f0f2c2a8aa4944f

mikehardy (approved):
This looks right, but just to make extra sure @petetnt 

> What needs to happen is to confirm that your build works with no version number (it should?)

Can you confirm that with no version number at all, this still works for you and builds the notification extension correctly? It should but confirmation is important :-)

Thanks!

mikehardy:
Fantastic, really appreciate it

PR #8206, opened 2024-12-31
Change nativeErrorCode, nativeErrorMessage to be nullable

jvliwanag:
### Description

Update `nativeErrorCode` and `nativeErrorMessage` to reflect the possibility of having null values based [here](https://github.com/invertase/react-native-firebase/blob/9d2817be983f5e70515a2f12cb44135eac745e82/packages/app/lib/internal/NativeFirebaseError.js#L54) and [here](https://github.com/invertase/react-native-firebase/blob/9d2817be983f5e70515a2f12cb44135eac745e82/packages/app/lib/internal/NativeFirebaseError.js#L59).

### Checklist

  - [x] Yes
  - [ ] `Android`
  - [ ] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

---

### DIFF
--- a/packages/app/lib/index.d.ts
+++ b/packages/app/lib/index.d.ts
@@ -50,12 +50,12 @@ export namespace ReactNativeFirebase {
     /**
      * The native sdks returned error code, different per platform
      */
-    readonly nativeErrorCode: string | number;
+    readonly nativeErrorCode: string | number | null;
 
     /**
      * The native sdks returned error message, different per platform
      */
-    readonly nativeErrorMessage: string;
+    readonly nativeErrorMessage: string | null;
   }
 
   export type LogLevelString = 'debug' | 'verbose' | 'info' | 'warn' | 'error' | 'silent';


mikehardy:
Hmmm 🤔  no denying that they currently can be null

Upstream in firebase-js-sdk the equivalent class has code and message as non-null though https://github.com/firebase/firebase-js-sdk/blob/main/packages/util/src/errors.ts

They have a lot of machinery for making sure the error is associated with a service, and then the code can be specific (and type safe!) or just their Err.GENERIC type, with the message defaulting to a templated version of a specific error or just the string 'Error'

Moving at least closer to that by doing something similar to their GENERIC error code with perhaps something truly generic as a service/detail code combo like 'unknown/internal-error' as the default code vs null, and 'Error' as the default message vs null might start the process of us converging our errors so they may be handled more consistently for people using firebase-js-sdk on web and react-native-firebase on native in same codebase

What do you think?

jvliwanag:
This doesn't change `code` and `message` which have counterparts on firebase-js-sdk really should be non nullable. Only `nativeErrorCode` and `nativeErrorMessage`.

mikehardy:
Sorry for the delay here - we're slowly coming to consensus on how to converge our types better with firebase-js-sdk

I think the answer here is:
- alter our type `NativeFirebaseError` to extend `FirebaseError`
- by default set the nativeErrorCode and nativeErrorMessage to 'unknown' and 'an unknown error occurred'
- by default set the same into FirebaseError code and message

The alternative, to make these nullable is actually a breaking change for consumers as their build will fail if they're not handling the types - regardless if it is fixing a real bug that nulls could come through or not

So instead of doing a breaking change the wrong way - making things farther from upstream - better to make them closer to upstream so our errors fit into existing firebase-js-sdk error-handling patterns as a drop-in

mikehardy:
This has a complete analysis and needs implementation

PR #8318, opened 2025-02-11
fix(messaging): 3 statics not correctly exported

russellwheatley:
### Description

Addresses type issue discussed here: https://github.com/invertase/react-native-firebase/discussions/8311#discussioncomment-12114864
### Checklist

  - [ ] Yes
  - [ ] `Android`
  - [ ] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [ ] No

### Test Plan

---

### DIFF
--- a/packages/messaging/lib/modular/index.d.ts
+++ b/packages/messaging/lib/modular/index.d.ts
@@ -4,6 +4,10 @@ import { ReactNativeFirebase } from '@react-native-firebase/app';
 import Messaging = FirebaseStorageTypes.Module;
 import FirebaseApp = ReactNativeFirebase.FirebaseApp;
 
+export import AuthorizationStatus = FirebaseMessagingTypes.Statics.AuthorizationStatus;
+export import NotificationAndroidPriority = FirebaseMessagingTypes.Statics.NotificationAndroidPriority;
+export import NotificationAndroidVisibility = FirebaseMessagingTypes.Statics.NotificationAndroidVisibility;
+
 /**
  * Returns a Messaging instance for the given app.
  * @param app - The Firebase app instance. Optional.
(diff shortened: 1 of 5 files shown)


packages/messaging/lib/modular/index.d.ts
```diff
@@ -4,6 +4,10 @@ import { ReactNativeFirebase } from '@react-native-firebase/app';
 import Messaging = FirebaseStorageTypes.Module;
 import FirebaseApp = ReactNativeFirebase.FirebaseApp;
 
+export import AuthorizationStatus = FirebaseMessagingTypes.Statics.AuthorizationStatus;
```
mikehardy, line 7: This one seems to be following the style of just exporting these statics in modular, and *not* exporting them in namespaced. Which means storage should probably have it's namespace export of it's statics removed

Then if that style works, all three of these "statics" type PRs can go in

russellwheatley: Hmmm, disagree on this one. We need to preserve the `FirebaseMessagingTypes.Statics.AuthorizationStatus` for users still using namespaced so they have correct types from static, and reexport as a top level import (i.e. modular). I don't think we should re-export as top level from namespaced type declaration file as it serves no purpose there.

When we come to delete the namespaced type declarations, we would lose it as a type for modular exports (although unit tests would fail at this point), it makes sense to me to put it where it belongs.

I have committed again because I found a better way to reexport and tested in unit tests the individual properties.

mikehardy (approved):
Okay - I see it now and it looks good -

- shared statics definition file for use everywhere
- send those into namespaced internal registry functions to attach it to the object as the module is constructed but don't export them
- export them directly in modular type definitions + implementation

So in namespaced they're chained off the module's namespace
And in modular they're available for direct import

I believe that's all correct

PR #8319, opened 2025-02-11
fix(dynamic-links): pass dynamic links into getInitialLink()

russellwheatley:
### Description

Addresses issue raised by user here: https://github.com/invertase/react-native-firebase/discussions/8311#discussioncomment-12114750
### Checklist

  - [ ] Yes
  - [ ] `Android`
  - [ ] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [ ] No

### Test Plan

---

### DIFF
--- a/packages/dynamic-links/lib/modular/index.d.ts
+++ b/packages/dynamic-links/lib/modular/index.d.ts
@@ -80,7 +80,9 @@ export declare function buildShortLink(
  * }
  * ```
  */
-export declare function getInitialLink(): Promise<FirebaseDynamicLinksTypes.DynamicLink | null>;
+export declare function getInitialLink(
+  dynamicLinks: FirebaseDynamicLinks,
+): Promise<FirebaseDynamicLinksTypes.DynamicLink | null>;
 
 /**
  * Subscribe to Dynamic Link open events while the app is still running.
--- a/packages/dynamic-links/type-test.ts
+++ b/packages/dynamic-links/type-test.ts
@@ -1,4 +1,4 @@
-import firebase from '.';
+import firebase, { getInitialLink } from '.';
 
 console.log(firebase.default().getInitialLink);
 
@@ -38,3 +38,6 @@ console.log(
 );
 
 console.log(firebase.dynamicLinks.ShortLinkType.SHORT);
+
+
+getInitialLink(firebase.dynamicLinks());


packages/dynamic-links/lib/modular/index.d.ts
```diff
@@ -80,7 +80,9 @@ export declare function buildShortLink(
  * }
  * ```
  */
-export declare function getInitialLink(): Promise<FirebaseDynamicLinksTypes.DynamicLink | null>;
+export declare function getInitialLink(
+  dynamicLinks: FirebaseDynamicLinks,
+): Promise<FirebaseDynamicLinksTypes.DynamicLink | null>;
```
mikehardy, line 85: I'm not sure this actually addresses the problem 🤔 

There is no firebase-js-sdk dynamic links module, so we don't have a reference API unless I'm missing something? We're in control of the API

In the implementation we don't pass the initial dynamic link anywhere: we do not pass it as we chain to the namespaced call internally, and namespaced call doesn't take anything as a param, and does not pass anything to native.

So why take the param? Seems like we should continue taking no params and remove the jsdoc

However there was some mention of a possible unhandled promise rejection: that bears investigation. No idea what could be happening there since all we do is immediately call native ?

mikehardy (changes requested):

packages/dynamic-links/lib/modular/index.d.ts
russellwheatley: In the implementation of every single dynamic link function we pass in an instance of `dynamicLinks` including the API raised by the user in the discussion: https://github.com/invertase/react-native-firebase/blob/main/packages/dynamic-links/lib/modular/index.js#L63

This follows the pattern set out in every firebase-js-sdk package so it would be odd if we deviated from that. (i.e. pass an instance of package to modular function as first argument).

This fixes the user's issue exactly.

The reason why he has a promise rejection is because he isn't passing an instance of `dynamicLinks` into the function as the first argument and therefore gets the classic `Cannot read property 'getInitialLink' of undefined`.

mikehardy: Oh 🤦  of course, not sure how I forgot the standard "pass instance of module into every function as first param" given it's literally the way every function works.

mikehardy (approved):
LGTM

PR #8321, opened 2025-02-11
fix(remote-config): ensure static objects are exposed for modular API

russellwheatley:
### Checklist

  - [ ] Yes
  - [ ] `Android`
  - [ ] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [ ] No

### Test Plan

---

### DIFF
--- a/packages/remote-config/lib/modular/index.js
+++ b/packages/remote-config/lib/modular/index.js
@@ -261,3 +261,5 @@ export async function setCustomSignals(remoteConfig, customSignals) {
   }
   return remoteConfig._promiseWithConstants(remoteConfig.native.setCustomSignals(customSignals));
 }
+
+export { LastFetchStatus, ValueSource } from '../statics';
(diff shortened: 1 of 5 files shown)


packages/remote-config/lib/modular/index.js
```diff
@@ -261,3 +261,5 @@ export async function setCustomSignals(remoteConfig, customSignals) {
   }
   return remoteConfig._promiseWithConstants(remoteConfig.native.setCustomSignals(customSignals));
 }
+
+export { LastFetchStatus, ValueSource } from '../statics';
```
mikehardy, line 265: there is a slight difference in the statics export here (modular only, but used in namespaced) vs the storage one (exported in both modular and namespaced, but looks like storage statics export. Unsure which should be The Way

perhaps in storage we don't need the export in namespaced index.js ?

russellwheatley: We should only export these from modular in my view. namespaced statics should be from namespace only (e,g, RemoteConfig.Statics.ValueSource).

mikehardy: agreed - I like that style

mikehardy (approved):
LGTM - agreed with the style, seems correct, as mentioned on related messaging change

PR #8322, opened 2025-02-11
fix(storage): expose static types for modular API

russellwheatley:
### Checklist

  - [ ] Yes
  - [ ] `Android`
  - [ ] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [ ] No

### Test Plan

---

### DIFF
--- a/packages/storage/lib/index.js
+++ b/packages/storage/lib/index.js
@@ -29,7 +29,7 @@ import version from './version';
 import fallBackModule from './web/RNFBStorageModule';
 
 // import { STATICS } from '@react-native-firebase/storage';
-export const statics = {
+const statics = {
   StringFormat,
   TaskEvent,
   TaskState,
--- a/packages/storage/lib/index.js
+++ b/packages/storage/lib/index.js
@@ -28,7 +28,8 @@ import { getGsUrlParts, getHttpUrlParts, handleStorageEvent } from './utils';
 import version from './version';
 import fallBackModule from './web/RNFBStorageModule';
 
-const statics = {
+// import { STATICS } from '@react-native-firebase/storage';
+export const statics = {
   StringFormat,
   TaskEvent,
   TaskState,
@@ -191,9 +192,6 @@ class FirebaseStorageModule extends FirebaseModule {
 
 // import { SDK_VERSION } from '@react-native-firebase/storage';
 export const SDK_VERSION = version;
-
-// import { STATICS } from '@react-native-firebase/storage';
-export const STATICS = StorageStatics;
 
 // import storage from '@react-native-firebase/storage';
 // storage().X(...);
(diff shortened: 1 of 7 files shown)


packages/storage/lib/index.js
```diff
@@ -23,11 +23,18 @@ import {
...
 } from '@react-native-firebase/app/lib/internal';
 import StorageReference from './StorageReference';
-import StorageStatics from './StorageStatics';
+import { StringFormat, TaskEvent, TaskState } from './StorageStatics';
 import { getGsUrlParts, getHttpUrlParts, handleStorageEvent } from './utils';
 import version from './version';
 import fallBackModule from './web/RNFBStorageModule';
 
+// import { STATICS } from '@react-native-firebase/storage';
+export const statics = {
```
mikehardy, line 32: I did a little fixup here to make this work (by adding the export here and removing the export below)
but the other two statics PRs would not have the `export` here, they would leave it in modular only

russellwheatley: good catch 👍 . I've also updated tests for static properties which I've done for others as well (using intellisense to pick up type properties as POC 😄 )

russellwheatley: actually, I think we should remove the static export from here, it should be from modular only in my view

mikehardy:
Huh - that appears to have caused a real issue:

```
         firebase v8 compatibility
           "before all" hook in "firebase v8 compatibility":
     TypeError: Cannot read property 'StringFormat' of undefined
      at putString (/home/runner/work/react-native-firebase/react-native-firebase/packages/storage/lib/StorageReference.js:201:43)
      at seed (/home/runner/work/react-native-firebase/react-native-firebase/packages/storage/e2e/helpers.js:14:59)
      at asyncGeneratorStep (/home/runner/work/react-native-firebase/react-native-
```

enhancement thought - but not on this PR - need to get those e2e tests in to typescript and have the type check as part of our lint checks which are fast enough to run locally and faster in CI

packages/storage/lib/index.js
mikehardy: So my fixup here quieted one error but I wasn't sure on it - and I like your way of thinking better

Action - the export here should be removed, and statics should just go into the registry function for namespace attachment

mikehardy:
@russellwheatley I pushed a maybe-final commit here that adjusts this types PR to follow your thinking on the other two

 - no export of statics directly from namespaced
 - yes to your export of specific items from statics-file in modular
 
...and I adjusted the use of statics internally so they handle the new export style from the statics-file so it should pass e2e now

what do you think?

russellwheatley:
@mikehardy - yes, we're in agreement 🙏. Should hopefully go green now, there was one remaining pesky default export in that StorageReference file.

mikehardy (approved):
A thing of beauty

PR #8341, opened 2025-02-17
docs(auth, google-signin): add expo-specific instructions

gregfenton:
- explicitly point to the new Expo-specific instructions on the new  react-native-google-signin docs site

- sorted social auth providers alphabetically (i.e. moved Twitter to the end)


mikehardy:
not possible for docs change to affect native platforms so those checks don't need to complete
passes all lint checks now, merged
thanks!

PR #8342, opened 2025-02-17
docs(homepage): make Expo initial instructions

gregfenton:
Expo is now the recommended framework for use in React-Native projects.  Even the RN team [recommends that people use Expo](https://reactnative.dev/docs/environment-setup).

So Expo should be the initial/default docs people use.  If people want to continue using RN CLI, the difference is more obvious than before.

This should reduce a number of issues seen in both Expo and Firebase communities regarding setup and use of RNF.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -20,12 +20,80 @@ If you do not meet these prerequisites, follow the links below:
 
 Additionally, current versions of firebase-ios-sdk have a minimum Xcode requirement of 15.2, which implies a minimum macOS version of 13.5 (macOS Ventura).
 
-## Installation for React Native CLI projects
+## Installation for Expo projects
+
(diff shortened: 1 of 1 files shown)


mikehardy (approved):
Indeed, we've been a bit slow to adopt the new "You should use a framework " framework-first philosophy. Must admit I am personally a bit hesitant since of course there's only one framework and it's full of commercial offerings. But fighting upstream is a losing battle with regard to documentation, I'm aware of that.

Thanks

ide:
For clarity, the Expo framework is 100% free and open source under the MIT License, and EAS (Expo Application Services) is a fully optional suite of hosted services. The framework does not contain any commercial services. The related text in the React Native Framework [RFC](https://github.com/react-native-community/discussions-and-proposals/blob/main/proposals/0759-react-native-frameworks.md#dos-of-frameworks) is:  
> Recommended frameworks must be free, open source, projects. We allow for companies to provide paid services on top of the Open Source framework, but these services should not be a requirement to use the framework and should not have vendor lock-in.

PR #8363, opened 2025-02-24
fix(analytics): Type defs

MichaelVerdon:
### Checklist

  - [ ] Yes
  - [ ] `Android`
  - [ ] `iOS`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [ ] No

### Test Plan

---

### DIFF
--- a/packages/analytics/lib/modular/index.js
+++ b/packages/analytics/lib/modular/index.js
@@ -72,7 +72,7 @@ export function initializeAnalytics(app, options) {
  */
 export function getGoogleAnalyticsClientId(analytics) {
   throw new Error('getGoogleAnalyticsClientId is web-only and not yet supported.');
-  return analytics
+  return analytics;
 }
 /**
  * Log a custom event with optional params.
--- a/packages/analytics/lib/modular/index.js
+++ b/packages/analytics/lib/modular/index.js
@@ -67,11 +67,12 @@ export function initializeAnalytics(app, options) {
 /**
  * Retrieves a unique Google Analytics identifier for the web client.
  *
- * @param {AnalyticsInstance} analyticsInstance - Instance of analytics (web - only)
+ * @param {FirebaseAnalytics} analytics - Instance of analytics (web - only)
  * @returns {string}
  */
-export function getGoogleAnalyticsClientId(analyticsInstance) {
-  throw new Error('getGoogleAnalyticsClientId is only supported on Web');
+export function getGoogleAnalyticsClientId(analytics) {
+  throw new Error('getGoogleAnalyticsClientId is web-only and not yet supported.');
+  return analytics
 }
 /**
  * Log a custom event with optional params.
--- a/packages/analytics/lib/modular/index.js
+++ b/packages/analytics/lib/modular/index.js
@@ -65,6 +65,15 @@ export function initializeAnalytics(app, options) {
 }
 
 /**
+ * Retrieves a unique Google Analytics identifier for the web client.
+ *
+ * @param {AnalyticsInstance} analyticsInstance - Instance of analytics (web - only)
+ * @returns {string}
+ */
+export function getGoogleAnalyticsClientId(analyticsInstance) {
+  throw new Error('getGoogleAnalyticsClientId is only supported on Web');
+}
+/**
  * Log a custom event with optional params.
  * @param {FirebaseAnalytics} analytics - Analytics instance.
  * @param {string} name - Event name must not conflict with any Reserved Events.
--- a/packages/analytics/__tests__/analytics.test.ts
+++ b/packages/analytics/__tests__/analytics.test.ts
@@ -4,6 +4,7 @@ import {
   firebase,
   getAnalytics,
   initializeAnalytics,
+  getGoogleAnalyticsClientId,
   logEvent,
   setAnalyticsCollectionEnabled,
   setSessionTimeoutDuration,
@@ -709,6 +710,10 @@ describe('Analytics', function () {
 
     it('`initializeAnalytics` function is properly exposed to end user', function () {
       expect(initializeAnalytics).toBeDefined();
+    });
+
+    it('`getGoogleAnalyticsClientId` function is properly exposed to end user', function () {
+      expect(getGoogleAnalyticsClientId).toBeDefined();
     });
 
     it('`logEvent` function is properly exposed to end user', function () {
--- a/packages/analytics/lib/modular/index.d.ts
+++ b/packages/analytics/lib/modular/index.d.ts
@@ -24,6 +24,14 @@ export declare function initializeAnalytics(
 ): Analytics;
 
 /**
+ * Retrieves a unique Google Analytics identifier for the web client.
+ *
+ * @param analyticsInstance - Instance of analytics (web - only)
+ *
+ */
+export declare function getGoogleAnalyticsClientId(analyticsInstance: Analytics): Promise<string>;
+
+/**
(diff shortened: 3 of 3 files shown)


packages/analytics/lib/modular/index.js
```diff
@@ -64,6 +64,16 @@ export function initializeAnalytics(app, options) {
...
 }
 
+/**
+ * Retrieves a unique Google Analytics identifier for the web client.
+ *
+ * @param {FirebaseAnalytics} analytics - Instance of analytics (web - only)
+ * @returns {string}
+ */
+export function getGoogleAnalyticsClientId(analytics) {
+  throw new Error('getGoogleAnalyticsClientId is web-only and not yet supported.');
```
mikehardy, line 74: for web we have an implementation already (I'm aware because I was in there semi-recently - it works...)
https://github.com/invertase/react-native-firebase/blob/8f59579e587f8dda4fadef6360f27076007e636b/packages/analytics/lib/web/api.js#L165

if you haven't done it, running `yarn tests:macos:pod:install && yarn tests:macos:build && yarn tests:macos:test-cover` runs the "web" version, and if you put an `it.only` or `describe.only` on any part of the e2e suite it'll run just those so it's fast to test+iterate

shame not to expose it since that might be half the work that went into the "other" platform analytics!

packages/analytics/lib/modular/index.js
```diff
@@ -64,6 +64,16 @@ export function initializeAnalytics(app, options) {
...
+  return analytics;
```
mikehardy, line 75: ...but, `@returns {string}` 🤔  ?

mikehardy (changes requested):
- potential for actual implementation on web just by hooking up web plumbing
- return value question

packages/analytics/lib/modular/index.js
mikehardy: What I meant here was for `Platform.other`, it should actually return the result of the existing `_getCid()` method, because it does exist and does work. The error message `is web-only and not yet supported` would then be for the other branch of the conditional (where `Platform` was *not* `Platform.other`) and should be more like `is only supported on "other" platforms in react-native-firebase", because we do support it for "other" (macos, or anything else that uses our "other" platform support where we wrap the web SDK)

packages/analytics/e2e/analytics.e2e.js
```diff
@@ -1133,5 +1134,28 @@ describe('analytics() modular', function () {
...
+        if (!Platform.other) {
+          try {
+            const { getAnalytics } = analyticsModular;
+            await firebase.analytics.getGoogleAnalyticsClientId(getAnalytics());
+            fail('Should have thrown an error');
+          } catch (e) {
+            e.message.should.equal('getGoogleAnalyticsClientId is web-only.');
+          }
+        }
+        this.skip();
```
mikehardy, line 1149: does this work? Or should it be:

```suggestion
        } else {
          this.skip();
        }
```

🤔  (same below)

packages/analytics/e2e/analytics.e2e.js
```diff
@@ -1133,5 +1134,28 @@ describe('analytics() modular', function () {
...
+      });
+
+      it('A response is received from getGoogleAnalyticsClientId() on web platforms', async function () {
+        const { getAnalytics } = analyticsModular;
+        if (Platform.OS === 'web') {
+          await firebase.analytics.getGoogleAnalyticsClientId(getAnalytics());
+        }
+        this.skip();
```
mikehardy, line 1157: as mentioned above, I *think* should be:

```suggestion
        } else {
          this.skip();
        }
```

...but I could definitely be missing something

packages/analytics/e2e/analytics.e2e.js
```diff
@@ -1133,5 +1134,28 @@ describe('analytics() modular', function () {
...
+            e.message.should.equal('getGoogleAnalyticsClientId is web-only.');
+          }
+        }
+        this.skip();
+      });
+
+      it('A response is received from getGoogleAnalyticsClientId() on web platforms', async function () {
+        const { getAnalytics } = analyticsModular;
+        if (Platform.OS === 'web') {
+          await firebase.analytics.getGoogleAnalyticsClientId(getAnalytics());
```
mikehardy, line 1155: these follow a pattern that we can + should assert against. Not down to a really really specific level like lengths but here's the format:

https://github.com/invertase/react-native-firebase/blob/a909d1beb597dcb92035a2ee6e66995c3aee8a5a/packages/analytics/lib/web/api.js#L36

Should be a string in general, more specifically is string.string, and then even more specifically is random-number-as-string, period, timestamp as string

I think just asserting it is a non-zero length string is sufficient, but is much stronger as a test then just checking it didn't error

mikehardy (commented):
really close now - just some almost-trivial test structure questions / thoughts

packages/analytics/e2e/analytics.e2e.js
MichaelVerdon: The change shouldn't affect anything so happy to commit it.

packages/analytics/lib/modular/index.js
```diff
@@ -64,6 +65,20 @@ export function initializeAnalytics(app, options) {
...
 }
 
+/**
+ * Retrieves a unique Google Analytics identifier for the web client.
+ *
+ * @param {FirebaseAnalytics} analytics - Instance of analytics (web - only)
+ * @returns {Promise<string>}
+ */
+export async function getGoogleAnalyticsClientId(analytics) {
+  if (Platform.android || Platform.ios) {
```
mikehardy, line 75: Platform.other is defined in globals in our e2e test suite but in general usage it isn't defined well

https://github.com/invertase/react-native-firebase/blob/c26253c82f0a13f62608941d266a63071667cc16/tests/globals.js#L95-L102

When running on macOS Platform.other doesn't show up the way you expect
So I define it here basically the way we define in testing globals

tests/globals.js
```diff
@@ -357,7 +360,12 @@ Object.defineProperty(global, 'modular', {
 });
 
 if (global.Platform.other) {
-  const { initializeApp } = modular;
+  const { initializeApp, setReactNativeAsyncStorage } = modular;
+
+  // Before initializing Firebase set the Async Storage implementation
+  // that will be used to persist user sessions.
+  setReactNativeAsyncStorage(AsyncStorage);
```
mikehardy, line 367: This wasn't being done for our tests but should have been
Also, there was no modular implementation of it, so I implemented a modular API for it

mikehardy:
I suspect something about installing the async-storage has impacted auth testing - there were mentions of analytics *and* auth for installing that, and perhaps now user state is preserved where it was stateless before for other tests, causing a failure. Will examine.

Analytics is passing now though, which was the original point...

mikehardy (approved):
I think this is one is ready to go now, but since the last commits are mine needs a second set of eyes from someone else...

mikehardy:
@MichaelVerdon no rush at all intended or implied - it's a holiday! - but when you get back, if this works for you, this works for me and Other platform now passes testing

MichaelVerdon:
> @MichaelVerdon no rush at all intended or implied - it's a holiday! - but when you get back, if this works for you, this works for me and Other platform now passes testing

Woroks for me 😄

mikehardy:
I'm going to go ahead and merge this one then - always happy to do follow-on PRs if something blows up but I want to keep moving on modular

PR #8366, opened 2025-02-24
fix(app): annotate initializeApp return as a promise

henribru:
### Description
`initializeApp` in the modular API is annotated is returning the app directly, but like in the namespaced API it actually returns a promise

### Checklist

  - [x] Yes
  - [x] `Android`
  - [x] `iOS`
  - [x] `Other` (macOS, web)
- My change includes tests; N/A
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

N/A

### DIFF
--- a/packages/app/lib/modular/index.d.ts
+++ b/packages/app/lib/modular/index.d.ts
@@ -43,9 +43,9 @@ export function getApps(): FirebaseApp[];
  * Initializes a Firebase app with the provided options and name.
  * @param options - Options to configure the services used in the app.
  * @param name - The optional name of the app to initialize ('[DEFAULT]' if omitted).
- * @returns FirebaseApp - The initialized Firebase app.
+ * @returns Promise<FirebaseApp> - The initialized Firebase app.
  */
-export function initializeApp(options: FirebaseAppOptions, name?: string): FirebaseApp;
+export function initializeApp(options: FirebaseAppOptions, name?: string): Promise<FirebaseApp>;
 
 /**
  * Retrieves an instance of a Firebase app.


mikehardy (changes requested):
Unfortunately we can't accept this change 

1- it must return AppCheck, not FirebaseApp (note! currently this is an implementation change, but it's an important one and we need to do it ASAP

2- it should not return a Promise, that's a difference from firebase-js-sdk and there's no technical reason to do it

reference ref: https://firebase.google.com/docs/reference/js/app-check.md#initializeappcheck_5548dfc


Separately: currently it is not returning a Promise anyway as far as I can tell? 

1- getApp() (underlying source of the FirebaseApp that it is returning, though it *should* return AppCheck, is not async: https://github.com/invertase/react-native-firebase/blob/main/packages/app/lib/modular/index.d.ts#L55

2- implementation delegates but is not async https://github.com/invertase/react-native-firebase/blob/main/packages/app/lib/modular/index.js#L71

3- fundamental implementation is not async for getApp https://github.com/invertase/react-native-firebase/blob/8f59579e587f8dda4fadef6360f27076007e636b/packages/app/lib/internal/registry/app.js#L87

So, I don't think it *is* a Promise currently, curious how you're seeing that

Regardless, cannot be Promise and cannot return FirebaseApp to correctly implement the firebase-js-sdk and meet our goal of being a drop-in replacement for firebase-js-sdk

mikehardy (approved):
I believe this is correct - thanks @henribru

mikehardy:
@henribru I just noticed - you haven't signed the CLA for react-native-firebase yet, could you follow the details link and do so? Then I can get this merged. I believe you just need to do so once and it will cover the other PR(s) you have open at same time. Thanks!

MichaelVerdon (approved):
LGTM

henribru:
> @henribru I just noticed - you haven't signed the CLA for react-native-firebase yet, could you follow the details link and do so? Then I can get this merged. I believe you just need to do so once and it will cover the other PR(s) you have open at same time. Thanks!

Done!

ycmjason:
Hi @mikehardy,

Thanks for the work on this.

I understand that the current implementation returns a Promise. However, in the official Firebase SDK, initializeApp is actually synchronous:

https://firebase.google.com/docs/reference/js/app.md#initializeapp

I was wondering — for parity with the official SDK — would it make sense for initializeApp here to be synchronous as well?

Thanks again!

mikehardy:
React-native has a limitation where the bridge is asynchronous so we are unable to have synchronous methods that use native code. Awaiting a synchronous return is okay but not awaiting a promise is not okay, so our only broadly compatible option is to return promise for many methods (not just this one)

In the future when we port to the new architecture in our native code we'll be able to have perfect fidelity here and synchronous return as new architecture does allow it

PR #8376, opened 2025-02-26
chore: Allow type-safety for subcollections

fivecar:
### Description
Root collections currently allow strong typing:
```typescript
const allYourBase = firestore()
  .collection<SetUsUp>("are-belong")
  .get();
```
But subcollections cannot be typed:
```typescript
const allYourBase = firestore()
  .collection<SetUsUp>("are-belong")
  .doc("8675309")
  .collection<TheBomb>("to-us")  // This is not accepted by index.d.ts
  .get();
```

This PR attempts to make that possible.

### Release Summary

Support strong typing for subcollections

### Checklist

  - [X] Yes
  - [X] `Android`
  - [X] `iOS`
  - [X] `Other` (macOS, web)
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [X] No

### Test Plan

Installed local version in my project, and verified that Typescript accepts a subcollection type.

### DIFF
--- a/packages/firestore/lib/index.d.ts
+++ b/packages/firestore/lib/index.d.ts
@@ -253,7 +253,10 @@ export namespace FirebaseFirestoreTypes {
    * to the location. The document at the referenced location may or may not exist. A `DocumentReference` can also be used
    * to create a `CollectionReference` to a subcollection.
    */
-  export interface DocumentReference<T extends DocumentData = DocumentData> {
+  export interface DocumentReference<
+    T extends DocumentData = DocumentData,
+    C extends DocumentData = DocumentData,
+  > {
     /**
      * The Firestore instance the document is in. This is useful for performing transactions, for example.
      */
@@ -285,7 +288,7 @@ export namespace FirebaseFirestoreTypes {
      *
      * @param collectionPath A slash-separated path to a collection.
      */
-    collection(collectionPath: string): CollectionReference;
+    collection(collectionPath: string): CollectionReference<C>;
 
     /**
      * Deletes the document referred to by this DocumentReference.


fivecar:
@mikehardy @MichaelVerdon Would you mind taking a look, or passing to the right folks? (Apologies, wasn't sure who's best to look, but both of you seem quite active in this project)

mikehardy:
Hey @fivecar - sorry for the delay - thanks for posting this. A couple questions

1- what's the backwards-compatibility status for this, how it will it affect use of DocumentReference in existing code? Will everyone get type errors now if they are not specifying the second type? If so, is it possible to make that second type optional?
2- I looked and I think the answer is yes but - how does this affect use of DocumentReference in the modular API? If you are using this patch locally (at least while testing it...) how does it affect your code when you migrate it to the modular API ?

Thanks!

fivecar:
@mikehardy Thanks for the thoughtful response!

### What about existing code?
I believe existing `DocumentReference` code won't be affected because nobody, prior to this change, could pass a type to `DocumentReference.collection()` (i.e. no existing code can `firestore().collection("boris").doc("yeltsin").collection<Stuff>("stuff")`, prior to this PR). So with this PR, all that existing code continues to be implicitly typed `DocumentData` when chaining `collection()` calls.

### What about the modular APIs
I've tested with and without the modular API, and the results are the same with this PR. Using the following code as an example:
```typescript
  return getFirestore()
    .collection(Collections.SECURE_PROFILE)
    .doc(uid)
    .collection<PersonYouMayKnow>(Collections.PYMK);
```
the returned type is `FirebaseFirestoreTypes.CollectionReference<PersonYouMayKnow>`. But if you change the code to:
```typescript
  return getFirestore()
    .collection(Collections.SECURE_PROFILE)
    .doc(uid)
    .collection(Collections.PYMK); // OG style, the only style allowable before this PR
```
the returned type is `FirebaseFirestoreTypes.CollectionReference<FirebaseFirestoreTypes.DocumentData>` (i.e. exactly what it was prior to this PR).

LMK if that all makes sense.

mikehardy:
That does make sense - seems like this should work, thanks for your replies

mikehardy:
@fivecar with sincere apologies for how long this sat - we are nearing the end of a significant convergence effort in type systems between firebase-js-sdk and react-native-firebase - primarily via a rewrite here in react-native-firebase into typescript locally, but also with a type comparison utility to surface and force justification of any differences we have from upstream.

In react-native-firebase v24.0.0 firestore was converted and I believe this was handled - I just had a look at the collection ref class that is returned from the `.collection` function and it appears to be generic now https://github.com/invertase/react-native-firebase/blob/main/packages/firestore/lib/FirestoreCollectionReference.ts

fivecar:
@mikehardy Awesome to hear it's all fixed. Thanks!

Issue #8375, opened 2025-02-26
[🐛] Property 'firebase' doesn't exist 😱


birdofpreyru:
## Issue

The latest v21.11.0 version of the lib, after upgrading to its new, modular API, inside the `database` lib module:
![Image](https://github.com/user-attachments/assets/e88ad501-4a39-43f0-a39a-fb723acea79c)

I also see, there is the same error just above, at the line 112 of that file.

---

### Javascript

<details><summary>Click To Expand</summary>
<p>

#### `package.json`:

```json
# N/A
```

#### `firebase.json` for react-native-firebase v6:

```json
# N/A
```

</details>

### iOS

<details><summary>Click To Expand</summary>
<p>

#### `ios/Podfile`:

- [ ] I'm not using Pods
- [x] I'm using Pods and my Podfile looks like:

```ruby
# N/A
```

#### `AppDelegate.m`:

```objc
// N/A
```

</p>
</details>

---

### Android

<details><summary>Click To Expand</summary>
<p>

#### Have you converted to AndroidX?

- [ ] my application is an AndroidX application?
- [ ] I am using `android/gradle.settings` `jetifier=true` for Android compatibility?
- [ ] I am using the NPM package `jetifier` for react-native compatibility?

#### `android/build.gradle`:

```groovy
// N/A
```

#### `android/app/build.gradle`:

```groovy
// N/A
```

#### `android/settings.gradle`:

```groovy
// N/A
```

#### `MainApplication.java`:

```java
// N/A
```

#### `AndroidManifest.xml`:

```xml
<!-- N/A -->
```

</p>
</details>

---

## Environment

<details><summary>Click To Expand</summary>
<p>

**`react-native info` output:**

```
 OUTPUT GOES HERE
```

- **Platform that you're experiencing the issue on**:
  - [ ] iOS
  - [ ] Android
  - [ ] **iOS** but have not tested behavior on Android
  - [ ] **Android** but have not tested behavior on iOS
  - [ ] Both
- **`react-native-firebase` version you're using that has this issue:**
  - `e.g. 5.4.3`
- **`Firebase` module(s) you're using that has the issue:**
  - `e.g. Instance ID`
- **Are you using `TypeScript`?**
  - `Y/N` & `VERSION`

</p>
</details>

---

- 👉 Check out [`React Native Firebase`](https://twitter.com/rnfirebase) and [`Invertase`](https://twitter.com/invertaseio) on Twitter for updates on the library.

[...]

Issue #8329, opened 2025-02-13
[🐛] When migrating to the new API, what replaces database.ServerValue and storage.TaskState


nes123:
I get these warnings:

This v8 method is deprecated and will be removed in the next major release as part of move to match Firebase Web modular v9 SDK API. Please use `getApp()` instead.

I migrated most of the methods but didnt find a replacement to database.ServerValue and storage.TaskState

russellwheatley:
Database is a bit different, we didn't port over the statics because there is another way of getting the timestamp. Call this function: https://github.com/invertase/react-native-firebase/blob/vertexai/packages/database/lib/modular/index.js#L111

which is the firebase-js-sdk way of retrieving it.

e.g.
```js
// namespaced - deprecated way
const timestamp = database.ServerValue.TIMESTAMP;

// modular - latest way
const timestamp = serverTimestamp();
```

nes123:
> Database is a bit different, we didn't port over the statics because there is another way of getting the timestamp. Call this function: https://github.com/invertase/react-native-firebase/blob/main/packages/database/lib/modular/index.js#L111
> 
> which is the firebase-js-sdk way of retrieving it.
> 
> e.g.
> 
> import database, { serverTimestamp } from '@react-native-firebase/database';
> // namespaced - deprecated way
> const timestamp = database.ServerValue.TIMESTAMP;
> 
> // modular - latest way
> const timestamp = serverTimestamp();

Thank you. I dont need to TimeStamp..I need ServerValue.increment()

russellwheatley:
You need to use this function then: https://github.com/invertase/react-native-firebase/blob/main/packages/database/lib/modular/index.js#L120

```js
import database, { increment } from '@react-native-firebase/database';
// namespaced - deprecated way
const incrementUpdate = database.ServerValue.increment(1);

// modular - latest way
const incementUpdate = increment(1);
```

nes123:
In the new version, there seems to be a new problem.

export function serverTimestamp() {
  return firebase.database.ServerValue.TIMESTAMP;
}

/**
 * @param {number} delta
 * @returns {object}
 */
export function increment(delta) {
  return firebase.database.ServerValue.increment(delta);
}

"firebase" is not defined in

packages/database/lib/modular/index.js

mikehardy:
@nes123 what happens if you patch those (that is: editing file directly in node_modules as a test) to read:

```javascript
export function serverTimestamp() {
[...]

PR #8377, opened 2025-02-27
fix(database): fix increment not defined

MichaelVerdon:
### Description

closes https://github.com/invertase/react-native-firebase/issues/8375

### Checklist

  - [ ] Yes
  - [ ] `Android`
  - [ ] `iOS`
  - [ ] `Other` (macOS, web)
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [ ] No

### Test Plan

---

### DIFF
--- a/packages/database/lib/modular/index.js
+++ b/packages/database/lib/modular/index.js
@@ -1,5 +1,7 @@
 import { getApp } from '@react-native-firebase/app';
-import { firebase } from '..'; // Needed for increment() method
+import DatabaseStatics from '../DatabaseStatics';
+
+const { ServerValue } = DatabaseStatics;
 
 /**
  * @typedef {import("..").FirebaseApp} FirebaseApp
@@ -110,7 +112,7 @@ export function getServerTime(db) {
  * @returns {object}
  */
 export function serverTimestamp() {
-  return firebase.database.ServerValue.TIMESTAMP;
+  return ServerValue.TIMESTAMP;
 }
 
 /**
@@ -118,7 +120,7 @@ export function serverTimestamp() {
  * @returns {object}
  */
 export function increment(delta) {
-  return firebase.database.ServerValue.increment(delta);
+  return ServerValue.increment(delta);
 }
 
 export * from './query';
--- a/packages/database/lib/modular/index.js
+++ b/packages/database/lib/modular/index.js
@@ -1,4 +1,5 @@
 import { getApp } from '@react-native-firebase/app';
+import { firebase } from "..";
 
 /**
  * @typedef {import("..").FirebaseApp} FirebaseApp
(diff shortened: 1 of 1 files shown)


packages/database/lib/modular/index.js
```diff
@@ -1,4 +1,5 @@
 import { getApp } from '@react-native-firebase/app';
+import { firebase } from '..'; // Needed for increment() method
```
mikehardy, line 2: this brings more of the namespaced stuff into the new modular implementation...is it possible to import the ServerValue directly and use it directly (below)? There should be a way to import that ServerValue or we have bigger problems (which may also be the case: #8329 is currently a problem it appears - resolving both at once would be a big win

https://github.com/invertase/react-native-firebase/blob/f06528a64bbcf3489e6a1f3df997f3a6f0b5fd41/packages/database/lib/DatabaseStatics.js#L23

SelaseKay: @mikehardy
Adding

```js
import DatabaseStatics from '../DatabaseStatics';

const { ServerValue } = DatabaseStatics;
```
seems to work. Is this what you had in mind?

mikehardy: @SelaseKay  Yep, that's what I had in mind! Small difference in practice but it is thinking forward to the time when the non-modular index.js / index.d.ts files go away. This has it already more set up for the future

SelaseKay: @mikehardy Nice. I've added the necessary changes.

packages/database/lib/modular/index.js
```diff
@@ -1,4 +1,7 @@
 import { getApp } from '@react-native-firebase/app';
+import DatabaseStatics from '../DatabaseStatics';
+
+const { ServerValue } = DatabaseStatics;
```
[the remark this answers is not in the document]
SelaseKay: No it doesn't. It throws an error.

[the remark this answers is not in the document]
mikehardy: Huh - strange but true. I couldn't believe that didn't work since it seemed like it should but I see the same when I try it locally. I must have something wrong in my mental model of imports etc. You're totally correct. This looks as good as it's going to get then, thanks!

mikehardy (approved):
LGTM thanks!

SelaseKay:
@mikehardy I suppose it's safe to close https://github.com/invertase/react-native-firebase/issues/8329 as well.

Nantris:
`serverTimestamp` isn't working for us even after this PR. We continue to face the error reported in #8375 (albeit with `serverTimestamp`)

packages/database/lib/modular/index.js
Nantris: I believe this tweak could the reason this PR doesn't seem to have fixed the issue for us - at least, it's the best theory I've got.

More info: https://github.com/invertase/react-native-firebase/issues/8375#issuecomment-2738875112

PR #8426, opened 2025-03-25
docs(app-check): update `initializeApp` example to include proper params

Gregoirevda:
1st arg is app


mikehardy (changes requested):
oh I think you are correct - but then it shouldn't be undefined I think it should be `getApp()`, left related comments as code suggestions

what do you think?

Gregoirevda:
@mikehardy good point, added the changes

Gregoirevda:
Out of scope, but there's no AdMob package in V6?

mikehardy (approved):
Nice, I think this is a good catch and looks like a good fix now, thanks for opening the PR on it. I'll mash it through CI will little fixup commits as needed if it fails for some lint reason no worries

Cheers

PR #8450, opened 2025-04-07
feat(auth)!: use native for isSignInWithEmailLink / BREAKING CHANGE: return type now `Promise<boolean>` not `boolean`

Willham12:
Use native version of isSignInWithEmailLink instead.

### DIFF
--- a/packages/auth/android/src/main/java/io/invertase/firebase/auth/ReactNativeFirebaseAuthModule.java
+++ b/packages/auth/android/src/main/java/io/invertase/firebase/auth/ReactNativeFirebaseAuthModule.java
@@ -403,6 +403,20 @@ class ReactNativeFirebaseAuthModule extends ReactNativeFirebaseModule {
   }
 
   /**
+   * isSignInWithEmailLink
+   *
+   * @param email
+   * @param promise
+   */
+  @ReactMethod
+  public void isSignInWithEmailLink(String appName, String emailLink, final Promise promise) {
+    Log.d(TAG, "isSignInWithEmailLink");
+    FirebaseApp firebaseApp = FirebaseApp.getInstance(appName);
+    FirebaseAuth firebaseAuth = FirebaseAuth.getInstance(firebaseApp);
+    promise.resolve(firebaseAuth.isSignInWithEmailLink(emailLink));
+  }
+
+  /**
    * signInWithEmailAndPassword
    *
    * @param email
--- a/packages/auth/e2e/emailLink.e2e.js
+++ b/packages/auth/e2e/emailLink.e2e.js
@@ -81,11 +81,18 @@ describe('auth() -> emailLink Provider', function () {
       const emailLink3 = 'https://www.example.com/action?mode=signIn';
       const emailLink4 =
         'https://x59dg.app.goo.gl/?link=https://rnfirebase-b9ad4.firebaseapp.com/__/auth/action?apiKey%3Dfoo%26mode%3DsignIn%26oobCode%3Dbar';
+      const emailLink5 = 'https://www.example.com/action?mode=signIn&oobCode=oobCode&apiKey=foo';
 
-      should.equal(true, isSignInWithEmailLink(auth, emailLink1));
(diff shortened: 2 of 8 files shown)


MichaelVerdon (changes requested):
Hey there, thanks for your PR. It looks valid however I want to ask you to please make corrections to formatting and tests working (I have run the workflows so you can see where it went wrong). Also please make sure you sign the CLA. 😄

mikehardy:
Looks like the e2e test needs an async/await setup?

```

  1) React Native Firebase
       auth() -> emailLink Provider
         isSignInWithEmailLink
           should return true/false:
     AssertionError: expected true == [Promise]
      at it$argument_1 (/Users/runner/work/react-native-firebase/react-native-firebase/packages/auth/e2e/emailLink.e2e.js:85:18)
```

mikehardy:
>  android formatting error - please re-run

doing a yarn lint:android and then committing the resulting auto-format of the java code should fix that

https://github.com/invertase/react-native-firebase/blob/66ef77dda6a3f23653d2292a67226745509e5bca/package.json#L13

mikehardy (commented):
this will be a breaking change - a tiny one but we need to issue it as a breaking change since the return type will be Promise now not directly a boolean

Suggest batching it with the needed type change of modular DocumentSnapshot.exists from property to method as noted here #8460 since that is also small but is a breaking change

Additionally - need to take some care here that this continues to work on the `other` platform - seems as though it should, the API exists https://firebase.google.com/docs/reference/js/auth.md#issigninwithemaillink_db04f1d

mikehardy:
I pulled this locally, fixed up the lint error, altered the e2e test to handle the return type, and set the commit message up to properly trigger the breaking change and CHANGELOG file note I want developers to see

If this passes CI it is ready to go pending decision on what to bundle with it in order to take advantage of the breaking change / semver major

mikehardy (approved):
@Willham12 looks like the CLA still needs signing from you?

mikehardy:
Should have looked more closely at other/web support - but I didn't, so that test failed - this will still need just a little bit of repair before it's ready to go:

```

  1) React Native Firebase
       auth() -> emailLink Provider
         isSignInWithEmailLink
           should return true/false:
     TypeError: this.native.isSignInWithEmailLink is not a function. (In 'this.native.isSignInWithEmailLink(emailLink)', 'this.native.isSignInWithEmailLink' is undefined)
      at FirebaseAuthModule#isSignInWithEmailLink (/Users/runner/work/react-native-firebase/react-native-firebase/packages/auth/lib/index.js:364:44)
      at <anonymous> (/Users/runner/work/react-native-firebase/react-native-firebase/packages/app/lib/common/index.js:350:37)
      at isSignInWithEmailLink (/Users/runner/work/react-native-firebase/react-native-firebase/packages/auth/lib/modular/index.js:171:35)
      at it$argument_1 (/Users/runner/work/react-native-firebase/react-native-firebase/packages/auth/e2e/emailLink.e2e.js:90:54)
```

mikehardy (approved):
turns out the web/other implementation *also* requires apiKey, really it is only the iOS one that doesn't. Strange but true

Repaired the PR by adding in the necessary items for web/other, this should pass CI now except missing CLA

MichaelVerdon (approved):

mikehardy:
Only thing this is blocking on now: it is a breaking change, and ideally we bundle other breaking changes with it at the same time to reduce the number/frequency of them over time as a small optimization.

I haven't scanned the PR queue or our plans internally yet to see what else we could bundle with it. Once that's decided, merge all the breaking changes as a group and release then carefully check release notes / changelog to make sure it's got the required messaging on exactly what devs need to do and edit if needed to make sure things are clear for everyone

mikehardy:
Got a headsup that firebase-ios-sdk 10.12 will require Xcode 16.2+, which is also a breaking change even though it is also tiny.

Bundle those two together along with anything else that comes along in the next couple days

https://github.com/firebase/firebase-ios-sdk/pull/14606#discussion_r2052454765

PR #8501, opened 2025-04-29
chore: improve vertexai example with different media inputs

russellwheatley:
### Description

- Updated example app which shows how to use image, pdf, video and audio as inputs using base64 encoding.
- Also updated the vertexai package.json as I believe it was incorrect (it was also failing to run prepare script as a result on my machine locally). main property ought to point to commonjs and module ought to point to ECMA imports.

### Checklist

  - [ ] Yes
  - [ ] `Android`
  - [ ] `iOS`
  - [ ] `Other` (macOS, web)
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [ ] No

### Test Plan

---


packages/vertexai/package.json
```diff
@@ -3,7 +3,8 @@
   "version": "22.0.0",
   "author": "Invertase <oss@invertase.io> (http://invertase.io)",
   "description": "React Native Firebase - Vertex AI is a fully-managed, unified AI development platform for building and using generative AI",
-  "main": "./dist/module/index.js",
+  "main": "./dist/commonjs/index.js",
+  "module": "./dist/module/index.js",
```
mikehardy, line 7: this must be why I was struggling to update the firebase-js-sdk version past [11.3](https://github.com/invertase/react-native-firebase/blob/3ae9fad2284e35b4e2dde22091f639aca91de5cd/package.json#L86)

They altered how their package.json works in 11.4.0 (https://firebase.google.com/support/release-notes/js#version_1140_-_february_27_2025) and it caused all sorts of problems here

Can you update the dependency to current (try `yarn upgrade-interactive` then arrow down to firebase, and arrow over to select the new version, then return...then `yarn && yarn lerna:prepare`) with this change?

mikehardy (changes requested):
looks very promising but want to see if this unblocks firebase-js-sdk dep upgrade

packages/vertexai/package.json
russellwheatley: I didn't add to this PR as I thought that ought to be separate PR (so captured in CHANGELOGS)

mikehardy: our need for Node v22 is new as of last round of JS dep updates and I couldn't find a way to enforce it in package.json::engines as it turns out that enforces for *dependencies* not for the actual package - but node v22+ requirement is internal only so shouldn't affect consumers. I kept remembering and forgetting (then ultimately: forgetting) to socialize that to the RNFB crew though, sorry.

Going to add it to CONTRIBUTING docs and mention it on RNFB/FF slack

Fantastic this allows us to get/stay current on firebase-js-sdk though - I try to keep deps up to date in general but the firebase SDKs of course are really important to be up to date on, really bugs me when those are blocked

mikehardy: - add docs #8504 

Bumping firebase-js-sdk still doesn't work for me - the thing that's failing is not lerna:prepare, it's `yarn tests:jest` - for some reason it is selecting an incompatible bundle for the node+jest environment since they made their package.json entry point change upstream

I have been stumped on this one, the easy fix should be to force jest to select whatever bundle works the old way but I haven't figured out how to do that. The forward-looking fix would be to get Jest working in ESM / "current+future module syntax for node" way since that's the bundle it is selecting now but I also haven't figured that out.

PR #8603, opened 2025-07-08
feat(functions): turbo modules implementation

russellwheatley:
### Checklist

  - [ ] Yes
  - [ ] `Android`
  - [ ] `iOS`
  - [ ] `Other` (macOS, web)
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [ ] No

### Test Plan

---


russellwheatley:
@mikehardy - Yep, I think it would be preferable to ignore generated files actually. I think there's a number of files that don't apply to RNFB packages anyway - related to UI generated code. I'm still in the phase of getting this to work, I have iOS building successfully (although not tested yet if everything breaks once API used), android is not building successfully.

russellwheatley:
@mikehardy - Android e2e tests passing for functions. iOS has hit a snag for this reason: https://github.com/facebook/react-native/pull/49250

Note - from my testing, it is maps with properties that have a `null` value that are stripped away.

It seems like it is part of [v0.81.0-rc.2 release](https://github.com/facebook/react-native/commit/d4236791e238a614d2fadf5c5659874d983ab029) therefore it isn't in stable yet. I did try and bump test app to `0.80.0` but came across build issues for both iOS and android.

I've had a further think on not checking in generated code, and I'm not sure I agree with the idea it should be gitignored. Any codegen changes might be worth tracking for any bugs further down the line in my opinion. I haven't fixed the code quality check CI which is failing on generated android code, I will fix it if you agree generated code is worth checking in.

package.json
```diff
@@ -69,6 +70,8 @@
     "@firebase/rules-unit-testing": "^4.0.1",
     "@inquirer/prompts": "^7.4.1",
     "@octokit/core": "^6.1.5",
+    "@react-native-community/cli": "latest",
+    "@react-native/codegen": "^0.80.1",
```
russellwheatley: double check these are needed now.

packages/functions/RNFBFunctions.podspec
```diff
@@ -27,11 +27,19 @@ Pod::Spec.new do |s|
...
+  s.compiler_flags      = '-DFOLLY_NO_CONFIG -DFOLLY_MOBILE=1 -DFOLLY_USE_LIBCPP=1 -DFOLLY_CFG_NO_COROUTINES=1'
 
   # React Native dependencies
   s.dependency          'React-Core'
   s.dependency          'RNFBApp'
+  s.dependency          'ReactCodegen'
+  s.dependency          'ReactAppDependencyProvider'
+  # TODO - not sure if we need these two as I believe they're UI related
+  s.dependency          'React-Fabric'
+  s.dependency          'RCT-Folly'
```
russellwheatley, line 42: double check what can be removed from this.

packages/functions/android/src/main/java/io/invertase/firebase/functions/NativeFunctionsModule.java
```diff
@@ -0,0 +1,138 @@
...
+        code = FirebaseFunctionsException.Code.UNAVAILABLE.name();
+        message = FirebaseFunctionsException.Code.UNAVAILABLE.name();
+      }
+    }
+
+    RCTConvertFirebase.mapPutValue(CODE_KEY, code, userInfo);
+    RCTConvertFirebase.mapPutValue(MSG_KEY, message, userInfo);
+    RCTConvertFirebase.mapPutValue(DETAILS_KEY, details, userInfo);
+    promise.reject(code, message, exception, userInfo);
+  }
```
russellwheatley, line 138: double check this is faithfully ported over.

packages/functions/ios/RNFBFunctions/RNFBFunctionsModule.mm
```diff
@@ -19,54 +19,63 @@
...
+    callableData = [NSNull null];
   }
 
-  FIRHTTPSCallable *callable = [functions HTTPSCallableWithName:name];
+  std::optional<double> timeout = options.timeout();
 
-  if (options[@"timeout"]) {
-    callable.timeoutInterval = [options[@"timeout"] doubleValue];
+  if (emulatorHost != nil) {
+    [functions useEmulatorWithHost:emulatorHost port:(int)emulatorPort];
```
russellwheatley: double check this is faithfully ported over.

packages/functions/lib/HttpsError.ts
```diff
@@ -17,8 +17,21 @@
 
 import { NativeFirebaseError } from '@react-native-firebase/app/lib/internal';
 
-export default class HttpsError extends Error {
-  constructor(code, message, details, nativeErrorInstance) {
+export interface NativeError {
+  userInfo?: {
+    code?: string;
+    message?: string;
+    details?: any;
```
russellwheatley: can this be improved.

packages/functions/lib/namespaced.ts
```diff
@@ -21,13 +21,14 @@ import {
...
-import version from './version';
+import { HttpsError, type NativeError } from './HttpsError';
+import { version } from './version';
 import { setReactNativeModule } from '@react-native-firebase/app/lib/internal/nativeModule';
 import fallBackModule from './web/RNFBFunctionsModule';
-
+import type { HttpsCallableOptions } from '.';
 const namespace = 'functions';
-const nativeModuleName = 'RNFBFunctionsModule';
+// TODO: change to RNFBFunctionsModule
```
russellwheatley: cannot rename this to RNFBFunctionsModule, name has to begin with `Native`

packages/functions/lib/namespaced.ts
```diff
@@ -67,21 +68,103 @@ export const HttpsErrorCode = {
...
+
 class FirebaseFunctionsModule extends FirebaseModule {
-  constructor(...args) {
+  private _customUrlOrRegion: string;
+  private _useFunctionsEmulatorHost: string | null;
+  private _useFunctionsEmulatorPort: number;
+
+  constructor(...args: any[]) {
     super(...args);
+    // @ts-ignore
```
russellwheatley: can this be removed

packages/functions/lib/namespaced.ts
```diff
@@ -90,7 +173,7 @@ class FirebaseFunctionsModule extends FirebaseModule {
       }
     }
 
-    return data => {
+    return (data?: any) => {
```
russellwheatley, line 175: can this be improved.

packages/functions/lib/namespaced.ts
```diff
@@ -123,7 +206,7 @@ class FirebaseFunctionsModule extends FirebaseModule {
...
```
russellwheatley, line 208: same

packages/functions/lib/namespaced.ts
```diff
@@ -202,12 +287,12 @@ export default createModuleNamespace({
...
-
 // import functions, { firebase } from '@react-native-firebase/functions';
 // functions().logEvent(...);
 // firebase.functions().logEvent(...);
 export const firebase = getFirebaseRoot();
 
 // Register the interop module for non-native platforms.
 setReactNativeModule(nativeModuleName, fallBackModule);
+
+export * from '.';
```
russellwheatley: think this can be removed

packages/functions/lib/web/RNFBFunctionsModule.ts
```diff
@@ -6,6 +6,24 @@ import {
   connectFunctionsEmulator,
 } from '@react-native-firebase/app/lib/internal/web/firebaseFunctions';
 
+interface WrapperData {
+  data?: any;
+}
+
+interface CallableOptions {
+  [key: string]: any;
+}
```
russellwheatley: can these be improved.

packages/functions/lib/web/RNFBFunctionsModule.ts
```diff
@@ -58,9 +84,9 @@ export default {
       const data = wrapper['data'] ?? null;
       const result = await callable(data);
       return result;
-    } catch (error) {
+    } catch (error: any) {
```
russellwheatley, line 75: improve

packages/functions/lib/web/RNFBFunctionsModule.ts
```diff
@@ -109,9 +143,9 @@ export default {
       const callable = httpsCallableFromURL(functionsInstance, url, options);
       const result = await callable(wrapper['data']);
       return result;
-    } catch (error) {
+    } catch (error: any) {
```
russellwheatley, line 133: improve

russellwheatley (commented):
updates to improve PR

packages/functions/lib/namespaced.ts
russellwheatley: I don't think this can be really. It's up to the users discretion from what is returned from the function

packages/functions/lib/web/RNFBFunctionsModule.ts
russellwheatley: will leave as any for now

mikehardy:
Hey @russellwheatley - strange one!

> @mikehardy - Android e2e tests passing for functions. iOS has hit a snag for this reason: [facebook/react-native#49250](https://github.com/facebook/react-native/pull/49250)
> 
> Note - from my testing, it is maps with properties that have a `null` value that are stripped away.
> 
> It seems like it is part of [v0.81.0-rc.2 release](https://github.com/facebook/react-native/commit/d4236791e238a614d2fadf5c5659874d983ab029) therefore it isn't in stable yet. I did try and bump test app to `0.80.0` but came across build issues for both iOS and android.

This should not be happening. That issue does look like the symptoms you describe, however, the issue you link resulted in a PR that was first released from main in 0.79.0-rc series (if you expand the tag list you'll see it), but more importantly, it was picked into the 0.77 and 0.78 branches, so the fix was actually present in 0.78.0 (since the issue was fixed after branch cut) and then a follow-on fix was required, and it was picked into 0.78.1 - you can see those here

- original 0.78 pick https://github.com/reactwg/react-native-releases/issues/785
- follow-on fix 0.78.1 pick https://github.com/reactwg/react-native-releases/issues/850

Our e2e app is on 0.78.2, so we have those fixes. I was able to discard those patches.

So...why is this happening now, again, with TurboModules? Something really subtle must be going on. If we think something is wrong, we'll need to make a reproducer for upstream - they have a template here https://github.com/react-native-community/reproducer-react-native

it is possible to take the reproducer and add native modules to it (of any kind) and run the necessary commands to generate native code or whatever - we'll need this though as it is not possible to efficiently share firebase projects related stuff with react-native upstream with all the config and key handling and such

> I've had a further think on not checking in generated code, and I'm not sure I agree with the idea it should be gitignored. Any codegen changes might be worth tracking for any bugs further down the line in my opinion. I haven't fixed the code quality check CI which is failing on generated android code, I will fix it if you agree generated code is worth checking in.

I'll trust you on this one, you've got your sleeves rolled up and will have more taste on it right now definitely. I *usually* don't get value from seeing diffs in generated code, but you may be right in this case, and either way, it's a reversible decision. There's zero harm in trying so my default stance of "implementor's choice" (read as: 'let the cook cook') wins for sure. Let's see if it has value over time and anyone can re-evaluate if not ?

russellwheatley:
@mikehardy - I've created an issue with reproducer on react native repo: https://github.com/facebook/react-native/issues/52802

mikehardy:
Looks like they've got a patch queued for release in 0.81 already.  When I'm done traveling (still a week or so) or maybe if I catch some time while traveling I'll try to queue pick requests for older branches otherwise this will tie us to react native 0.81 as a minimum version. We probably need to separate our e2e app so macos (which lags main react native) versions aren't holding us back from testing

russellwheatley:
@mikehardy - I think we still have a problem with null values, I checked against the latest RN release candidate and it didn't make a difference.

mikehardy:
this one's up next - current thinking is to have a serializer/deserializer process for iOS that maintains an array of paths for JSON nulls for serialized objects and send then on the side of the main object, to rehydrate them correctly on the native side.

it's the only way to handle this that doesn't rely on a maybe-they-fix-it/maybe-they-don't process upstream, nor their release cadence even if they do fix it, while avoiding any collision problems with in-place canaries

packages/remote-config/e2e/config.e2e.js
```diff
@@ -697,8 +697,8 @@ describe('remoteConfig()', function () {
         (await isSupported()).should.equal(true);
       });
     });
-
-    describe('onConfigUpdate', function () {
+    // SKIPPED: Skipping suite of onConfigUpdate tests until they're fixed
+    describe.skip('onConfigUpdate', function () {
```
mikehardy, line 701: I cherry-picked this commit to a clean branch for separate PR to get it in now for all PRs until fixed, once it merges to main for everyone's benefit you'll want to rebase it out over on this PR. Cool feature, but such a hassle.

packages/functions/package.json
```diff
@@ -27,10 +46,57 @@
     "@react-native-firebase/app": "23.5.0"
   },
   "devDependencies": {
-    "@react-native-firebase/private-tests-firebase-functions": "^0.0.1"
+    "@react-native-firebase/private-tests-firebase-functions": "^0.0.1",
+    "react-native": "^0.80.1",
+    "react-native-builder-bob": "^0.40.12",
+    "typescript": "^5.8.3"
```
russellwheatley, line 52: remove this and put path to TypeScript executable. See: https://github.com/invertase/react-native-firebase/pull/8651/files#diff-d1b85f8d1f6297c72a7aa703ba620c7eb120ea77be0538faed320a306eeab865R80

packages/app/ios/RNFBApp/RNFBSharedUtils.m
```diff
@@ -164,4 +164,131 @@ + (BOOL)getConfigBooleanValue:(NSString *)tag key:(NSString *)key defaultValue:(
...
+    [stack removeLastObject];
+
+    id original = frame[@"original"];
+    id mutable = frame[@"mutable"];
+
+    if ([original isKindOfClass:[NSDictionary class]]) {
+      NSDictionary *origDict = (NSDictionary *)original;
+      NSMutableDictionary *mutDict = (NSMutableDictionary *)mutable;
+
+      for (id key in origDict) {
```
mikehardy, line 269: the only difference here between the two large logic branches which contain the child processing seem to be the concrete typing of the id to Dictionary or Array, and the last bit about null processing

Everything else seems to be the same which seems like it could be abstracted to a processChildElement method where you pass in the frame and a pointer to the in-memory stack array to de-duplicate, with a tiny type-check conditional for the "preserve primitive values" bit at the end?

This is the tough bit of the logic so might have value

packages/app/lib/internal/nullSerialization.js
```diff
@@ -0,0 +1,145 @@
...
+      keys: Object.keys(data),
+      index: 0,
+    });
+  }
+
+  while (stack.length > 0) {
+    const frame = stack[stack.length - 1];
+
+    if (frame.type === 'array') {
+      const { original, encoded } = frame;
```
mikehardy, line 99: likewise here - everything seems correct but it's the toughest part to read and the branch internals could maybe use the same extract-method type change to tighten it up

PR #8607, opened 2025-07-09
fix: add missing 'FirebaseAnalytics/Core' to podspec

radko93:
### Description

fixes #8605

This is according to https://firebase.google.com/docs/analytics/configure-data-collection?platform=ios#disable-IDFA-collection

### Checklist

  - [x] Yes
  - [ ] `Android`
  - [x] `iOS`
  - [ ] `Other` (macOS, web)
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

---

### DIFF
--- a/packages/analytics/RNFBAnalytics.podspec
+++ b/packages/analytics/RNFBAnalytics.podspec
@@ -41,6 +41,7 @@ Pod::Spec.new do |s|
   # Firebase dependencies
   if defined?($RNFirebaseAnalyticsWithoutAdIdSupport) && ($RNFirebaseAnalyticsWithoutAdIdSupport == true)
     Pod::UI.puts "#{s.name}: Not installing FirebaseAnalytics/IdentitySupport Pod, no IDFA will be collected."
+    s.dependency           'FirebaseAnalytics/Core', firebase_sdk_version
   else
     if !defined?($RNFirebaseAnalyticsWithoutAdIdSupport)
       Pod::UI.puts "#{s.name}: Using FirebaseAnalytics/IdentitySupport with Ad Ids. May require App Tracking Transparency. Not allowed for Kids apps."


packages/analytics/RNFBAnalytics.podspec
```diff
@@ -41,6 +41,7 @@ Pod::Spec.new do |s|
   # Firebase dependencies
   if defined?($RNFirebaseAnalyticsWithoutAdIdSupport) && ($RNFirebaseAnalyticsWithoutAdIdSupport == true)
     Pod::UI.puts "#{s.name}: Not installing FirebaseAnalytics/IdentitySupport Pod, no IDFA will be collected."
+    s.dependency           'FirebaseAnalytics/Core', firebase_sdk_version
```
mikehardy, line 44: In the new pod arrangement for FirebaseAnalytics pods + sub-pods, combined with our desire to allow folks to have fine-grained control with Podfile variable toggles so that all parts are optional, we need to start with `FirebaseAnalytics/Core` for *all* arrangements and then go from there

I think it may work to have the `Core` sub-spec in the conditional because of transitive dependencies bringing it in some of the conditional branches where it is needed, but...

I think it's most correct to always start with it as a dependency, outside of all conditionals, and then go

What do you think @radko93 ? I think this is most correct and if I don't hear back I'm going to commit and release it so we don't have the broken 22.3.0 version out too long...


```suggestion
  s.dependency           'FirebaseAnalytics/Core', firebase_sdk_version
  if defined?($RNFirebaseAnalyticsWithoutAdIdSupport) && ($RNFirebaseAnalyticsWithoutAdIdSupport == true)
    Pod::UI.puts "#{s.name}: Not installing FirebaseAnalytics/IdentitySupport Pod, no IDFA will be collected."
```

mikehardy (changes requested):
I think we always need `FirebaseAnalytics/Core` as mentioned in suggestion comment

packages/analytics/RNFBAnalytics.podspec
mikehardy: going with the other PR, but I appreciate this and sorry for the trouble!

PR #8629, opened 2025-07-29
feat!: bump firebase-ios-sdk to `12.0.0`

russellwheatley:
### Description

`trackingId` has been removed from firebase-ios-sdk: https://firebase.google.com/support/release-notes/ios#version_1200_-_july_15_2025

I've kept the `gaTrackingId` API in the JS code because it appears android still uses it: https://firebase.google.com/docs/reference/android/com/google/firebase/FirebaseOptions?_gl=1*dtmd6b*_up*MQ..*_ga*MzIzOTI0NDg4LjE3NTM5NjIwMjM.*_ga_CW55HF8NVT*czE3NTM5NjQxOTMkbzIkZzAkdDE3NTM5NjQxOTMkajYwJGwwJGgw#gaTrackingId()

Although we could possibly update it to match firebase-js-sdk which would be `measurementId`?

### Checklist

  - [ ] Yes
  - [ ] `Android`
  - [ ] `iOS`
  - [ ] `Other` (macOS, web)
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [ ] No

### Test Plan

---

### DIFF
--- a/packages/app/package.json
+++ b/packages/app/package.json
@@ -74,10 +74,10 @@
   },
   "sdkVersions": {
     "ios": {
-      "firebase": "11.15.0",
-      "iosTarget": "13.0",
+      "firebase": "12.0.0",
+      "iosTarget": "15.0",
       "macosTarget": "10.15",
-      "tvosTarget": "13.0"
+      "tvosTarget": "15.0"
     },
     "android": {
       "minSdk": 23,
--- a/docs/index.md
+++ b/docs/index.md
@@ -18,7 +18,7 @@ If you do not meet these prerequisites, follow the links below:
 - [React Native - Setting up the development environment](https://reactnative.dev/docs/environment-setup)
 - [Create a new Firebase project](https://console.firebase.google.com/)
 
-Additionally, current versions of firebase-ios-sdk have a minimum Xcode requirement of 15.2, which implies a minimum macOS version of 13.5 (macOS Ventura).
+Additionally, current versions of firebase-ios-sdk have a minimum Xcode requirement of 16.2, which implies a minimum macOS version of 14.5 (macOS Sequoia).
 
 ## Installation for Expo projects
 
@@ -335,7 +335,7 @@ Open your projects `/ios/Podfile` and add any of the globals shown below to the 
 
 ```ruby
 # Override Firebase SDK Version
-$FirebaseSDKVersion = '11.15.0'
+$FirebaseSDKVersion = '12.0.0'
 ```
 
 Once changed, reinstall your projects pods via pod install and rebuild your project with `npx react-native run-ios`.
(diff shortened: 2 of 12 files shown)


packages/app/package.json
```diff
@@ -74,10 +74,10 @@
   },
   "sdkVersions": {
     "ios": {
-      "firebase": "11.15.0",
-      "iosTarget": "13.0",
+      "firebase": "12.0.0",
+      "iosTarget": "15.0",
```
mikehardy, line 78: should be merged with something like 

```
BREAKING CHANGE: iOS mininum target now 15.0, was 13.0
```

Should not be too big a deal as current react-native already bumped theirs (we were running 15.1 IPHONEOS_DEPLOYMENT_TARGET in our e2e test anyway?), we supported a wider range than they do I think and are now just trimming old versions to catch up

Also, developer environment minimums were bumped:

https://github.com/invertase/react-native-firebase/blob/fe4550f362548dcb66359044715e22222a413f97/docs/index.md?plain=1#L21

It is now Xcode 16.2 as a minimum per https://firebase.google.com/support/release-notes/ios#version_1200_-_july_15_2025 which implies a minimum macOS of 14.5 (macOS Sequoia) per https://en.wikipedia.org/wiki/Xcode#Xcode_15.0_-_16.x_(since_visionOS_support)

packages/app/package.json
```diff
@@ -74,10 +74,10 @@
   },
   "sdkVersions": {
     "ios": {
-      "firebase": "11.15.0",
-      "iosTarget": "13.0",
+      "firebase": "12.0.0",
```
mikehardy, line 77: docs need an update as well 

https://github.com/invertase/react-native-firebase/blob/fe4550f362548dcb66359044715e22222a413f97/docs/index.md?plain=1#L338

mikehardy (changes requested):
as long as we're breaking I think truing up the API to measurement vs tracking ID makes sense - we don't get frequent chances to clear out cruft, let's clear it out

That *may* imply that people need to download new a new set of plist/json firebase config files if theirs are really old I think...unsure

russellwheatley:
I think all that is left to do is to add:

```
BREAKING CHANGE: iOS mininum target now 15.0, was 13.0
```

to commit body when merging.

russellwheatley:
> as long as we're breaking I think truing up the API to measurement vs tracking ID makes sense - we don't get frequent chances to clear out cruft, let's clear it out
> 
> That _may_ imply that people need to download new a new set of plist/json firebase config files if theirs are really old I think...unsure

I don't think that will be necessary, ios release notes suggest it wasn't used/obsolete so I think having that property in plist will just be unused. I had a look at service info plist from 8 years ago and don't see analytics ID: https://github.com/invertase/react-native-firebase/blob/9308994db23b3d1acf903f94f12e0b2eabe84807/tests/ios/GoogleService-Info.plist so don't think it's ever been part of the downloaded service file.

PR #8621, opened 2025-07-23
feat(auth): support TOTP enroll and unenroll

jamespb97:
### Description
The aim of these changes is to enable TOTP MFA and the ability to un-enroll from either SMS or TOTP MFA.

The motivation behind doing this was to migrate away from Firebase JS SDK implementation on our existing Expo application which extensively uses multi-factor authentication.

This is my first time ever changing native code like this. My workflow at the moment for making these changes have simply been editing the native modules in the respective code editors (Xcode and Android Studio), then using patch-package and building new development builds each time and testing live against our development environment, so nothing has been emulated.

I tried to link my forked repo locally using `npm/yarn link` and build off that instead but I couldn't get it to work so I had to go the patch-package route and copy across all of my changes to my fork manually. If you're aware of a better way to do this please let me know!

I am aware that there is a blocking issue (https://github.com/firebase/firebase-tools/issues/6224) which prevents TOTP being tested within emulation, so I guess proper tests cannot be written just yet for this feature (unsure about unenroll feature).

### Related issues
https://github.com/invertase/react-native-firebase/issues/7483
Abandoned related PR: https://github.com/invertase/react-native-firebase/pull/7718

### Release Summary
Add support for enrolling TOTP multi-factor authentication, authentication using TOTP and un-enrolling multi-factor authentication.

### Checklist

- I read the [Contributor Guide](https://github.com/invertase/react-native-firebase/blob/main/CONTRIBUTING.md) and followed the process outlined there for submitting PRs.
  - [x] Yes
- My change supports the following platforms;
  - [x] `Android`
  - [x] `iOS`
  - [x] `Other` (macOS, web) (NOTE: When using shims replacing imports with firebase js sdk, this should work as the APIs are exactly the same as the web sdk - https://github.com/invertase/react-native-firebase/issues/7921#issuecomment-3102680871)
- My change includes tests;
  - [ ] `e2e` tests added or updated in `packages/\*\*/e2e`
  - [ ] `jest` tests added or updated in `packages/\*\*/__tests__`
- [x] I have updated TypeScript types that are affected by my change.
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan
I currently only have this working in our product application which won't be possible to post videos or screenshots of on a public PR, but I can create a minimum repro app soon enough (probably in my free time over the weekend) and post videos and screenshots from that instead.

Update:
As promised, here is a demo of the changes in action along with a demo repo: https://github.com/jamespb97/firebase-totp-demo
[...]

Issue #8601, opened 2025-07-07
[🐛] RNFBAuthModule.getConstants raised an exception: -[FirebaseAuth.TOTPMultiFactorInfo phoneNumber]: unrecognized selector sent to instance 0x105492880


Daniel528:
## Issue

Recently we added TOTP MFA to our web-app using the Firebase JS SDK. The web-app shares a common firebase project with the react-native firebase project. The react-native project has been building fine but after making the web-app changes now the react-native app crashes on launch with an error alluding to TOTP. We're aware the this library does not currently support TOTP MFA, we have not attempted to implement it in the react-native portion of the project. 

The xcode error is
```
libc++abi: terminating due to uncaught exception of type facebook::jsi::JSError: RNFBAuthModule.getConstants raised an exception: -[FirebaseAuth.TOTPMultiFactorInfo phoneNumber]: unrecognized selector sent to instance 0x105492880

Error: RNFBAuthModule.getConstants raised an exception: -[FirebaseAuth.TOTPMultiFactorInfo phoneNumber]: unrecognized selector sent to instance 0x105492880
    at getConstants (native)
    at keys (native)
    at nativeModuleWrapped (http://172.16.0.9:8081/index.bundle//&platform=ios&dev=true&lazy=true&minify=false&inlineSourceMap=false&modulesOnly=false&runModule=true&excludeSource=true&sourcePaths=url-server&{appbundleid}:192817:33)
    at initialiseNativeModule (http://172.16.0.9:8081/index.bundle//&platform=ios&dev=true&lazy=true&minify=false&inlineSourceMap=false&modulesOnly=false&runModule=true&excludeSource=true&sourcePaths=url-server&{appbundleid}:192866:57)
    at getNativeModule (http://172.16.0.9:8081/index.bundle//&platform=ios&dev=true&lazy=true&minify=false&inlineSourceMap=false&modulesOnly=false&runModule=true&excludeSource=true&sourcePaths=url-server&{appbundleid}:192931:34)
    at get (http://172.16.0.9:8081/index.bundle//&platform=ios&dev=true&lazy=true&minify=false&inlineSourceMap=false&modulesOnly=false&runModule=true&excludeSource=true&sourcePaths=url-server&{appbundleid}:193288:64)
[...]

PR #8645, opened 2025-08-08
refactor: filter out any multifactor besides phone

Daniel528:
### Description

Our project has a web dashboard (implementing Firebase JS SDK) and react native app (using react-native-firebase) which share a Firebase project. We are planning on supporting MFA through TOTP & SMS. This is possible with the JS SDK, TOTP is currently unsupported in react-native-firebase. To get around this, we are planning on having the react native app open an external browser window to the web dashboard and have users authenticate through this. Once authenticated in the external browser, a custom token is passed back to the app and the function `signInWithCustomToken` in `react-native-firebase` is used to authenticate the user.

We ran into [this error](https://github.com/invertase/react-native-firebase/issues/8601#issue-3207186607) when a user account who adds a TOTP MFA method on the web dashboard then attempts to log into the react-native app will encounter that crash on iOS instead of receiving a firebase auth error with code `auth/multi-factor-auth-required`. This crash also occurs when the user attempts to authenticate through `signInWithCustomToken`. 

I have not encountered this crashing in Android when testing. The comparable java function `multiFactorInfoToMap` in `ReactNativeFirebaseAuthModule.java` does not crash.

To resolve this issue I've rewritten the `convertMultiFactorData` function in `RNFBAuthModule.m` so it no longer references the `.phoneNumber` property of a hint which does not have that field. Because only SMS is supported, I've opted to completely filter out any `FIRMultiFactorInfo` hints which do not match the `FIRPhoneMultiFactorInfo` class. The implementation is more similar to the java implementation of `multiFactorInfoToMap`.

I've also removed the `getJSFactorId` function which seemed ineffective and was no longer referenced after this refactor. 

I'm aware of [this PR](https://github.com/invertase/react-native-firebase/pull/8621) which also resolves the issue I'm trying to solve. From what I can see though that PR may take a bit longer to resolve and dependant on Android Emulator testing. If that is the case then this MR can basically be a small bridge to resolve the crash immediately at hand before working on the larger TOTP implementation. 

### Related issues

https://github.com/invertase/react-native-firebase/issues/8601#issue-3207186607

### Checklist

  - [x] Yes
  - [x] `Android`
  - [x] `iOS`
  - [ ] `Other` (macOS, web)
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

---


mikehardy (approved):

Daniel528:
Hey @mikehardy, (or anyone else with the domain knowledge) 

Do you have any ideas on what I can do to resolve [the failing Android e2e tests](https://github.com/invertase/react-native-firebase/actions/runs/16868063340/job/47846934297?pr=8645)? Considering the iOS exclusive auth module changes in this PR, I'm unsure why Android tests in the `database` module would be failing while (if I read the logs correctly) all tests in`multiFactor.e2e.js` are running successfully. 

I'm unable to replicate the failing test case on my local environment. Initially when running `yarn tests:android:test` locally they'd exit prematurely with what I assume to be an internal jet error based on this output:

```
[🟨] Jet client disconnected - for no particular reason (code = 1006).
[🟥] Exiting after an abnormal disconnect.
```

Setting `exitOnError: false` in `.jetrc.js` though allows the whole test suite to run and all tests pass. This is my first time using jet. I'm hoping the suite passing entirely with `exitOnError: false` is satisfactory or if it crashing with `exitOnError: true` is related the GitHub actions failing tests and hopefully something I can resolve. 

At this stage I'm uncertain and having my local and the logs from GitHub actions aligned would be handy for moving forward.  Any help would be very nice, thank you :)

mikehardy:
I've strugged with that test before, the test is flaky, it is not the fault of this PR

Disabling the jet exit on error means that the tests will run, however any failures encountered in the test suite will be ignored so all CI runs will be potentially false positive. The solution is for the test that's failing to be fixed so it is not flaky, or for it to be disabled in CI so it stops triggering false negatives

I'll get this merged - obviously if iOS passes on an iOS-only changeset, an android CI failure is not important

I've re-run CI since you pushed a merge commit - if iOS goes green this is merge-able, no worries

Daniel528:
Thanks for re-running the tests. The debug iOS tests have passed but release iOS tests have failed. 

Both release & debug pass completely when I run them locally. 

The failing test is in the `app-check` package rather than the `auth`. Is this likely to be related to the contents of the PR or more likely another flakey test?

mikehardy:
> The failing test is in the app-check package rather than the auth. Is this likely to be related to the contents of the PR or more likely another flakey test?

Flaky - app-check in particular puts rate limits on token exchange events in the cloud so one of our CI failure modes is for one iOS run to work but one to fail on app-check if they run concurrently, and then for the failing one to succeed if re-run separately. It wasn't flaky prior to all the modular API implementation but now all APIs are effectively run twice, on 3 platforms (android, ios release, ios debug), which may be happening for multiple PRs or main merges at once, which at the end of it all has resulted in this popping up more often. CI stability is always within reach but never quite there it seems 😅 

No worries though, this will be merged and released shortly - it's next on the list. And no flakes are the fault of the PR obviously.

mikehardy:
releasing now should be up on npmjs.com shortly barring publish problems, cheers!

PR #8656, opened 2025-08-15
build(deps): updates with required forward ports

mikehardy:
Still fuzzbusting CI but the next area of flakes to tamp down are all iOS / macOS handoff issues / races between:

- detox starting the app
- jet starting server
- app starting detox and running jest which starts jet client and running tests

I don't want to chase ghosts so I want to make sure I'm on up to date everything before I try any ideas there

So this PR is updates to all of our dependencies

1) in particular I finally solved the firebase-js-sdk + jest interaction issue by forcing the jest environment to be 'node' instead of a 'jsdom' variant, so it doesn't try to load the ESM bundle and jest is happy with the newest firebase-js-sdk bundle entry points in firebase-js-sdk v12+

2) windows compat: While I was in there I discovered how to automate the "run on unix-like or run on windows" run script choice, and how to automate the post-install prepare.

 - so now you don't have to specify the `:windows` run script variants, they will be used as needed, automatically
 - you do not have to run `yarn lerna:prepare`, it will run for you after `yarn`

3) Storage + Other platform Blob crashes

I went deep here and figured out it was a race condition, and how to mostly handle it plus handle the error without crashing if not 🚀 

4) I added screen capture and system logging to the Other platform CI (so I could diagnose 3 above...)

The rest is pretty boring forward ports

The only things not updated are react-native and it's direct transitives / related packages. react-native-macos is blocking us there unless we split our e2e app completely, and that's a bigger task I don't want to approach if I can avoid it

### DIFF
--- a/packages/analytics/e2e/analytics.e2e.js
+++ b/packages/analytics/e2e/analytics.e2e.js
@@ -74,12 +74,18 @@ describe('analytics()', function () {
         let sessionId = await firebase.analytics().getSessionId();
         // On iOS it can take ~ 3 minutes for the session ID to be generated
         // Otherwise, `Analytics uninitialized` error will be thrown
-        const retries = 240;
+        // On CI we're only going to give it a few tries
+        const retries = global.isCI ? 3 : 240;
         while (!sessionId && retries > 0) {
           await Utils.sleep(1000);
           sessionId = await firebase.analytics().getSessionId();
         }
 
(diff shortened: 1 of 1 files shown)


mikehardy:
going to merge this one since it fixes so many things (especially other platform support...) before diving more deeply into why the iOS start app / start test handoff hangs so frequently

PR #8673, opened 2025-08-23
fix: remote config e2e stabilization and error handling

mikehardy:
### Description

The flake hammer (big parallel runs of e2e workflows in CI) turned up flakiness in remote-config onConfigUpdated testing

The root cause is that the listener on one test run can (and should!) get events for updates from other test runs, but it was assuming that it would only ever see the events for it's own local test run.

The cloud function was also susceptible to failing under concurrency.

I re-worked the test suite so it uses individualized remote-config params for each local run, and only verifies if those params were updated when it triggers updates - ignoring all the other possible updates.

This turned up a few little errors in our helper suite:
-  some old non-modular syntax
- node was out of date for the functions
- there was no retention policy for the Artifact Registry containers used during the functions cloud build
- some v1 functions that hadn't been migrated to v2 and were about to run out of support
- an incorrect delete implementation for remote-config params
- the test function could orphan old remote-config test params

It also exposed that a couple Other platform APIs *still* weren't returning that they weren't supported as correct errors

Everything should be repaired now

### Checklist

  - [x] Yes
  - [x] `Android`
  - [x] `iOS`
  - [x] `Other` (macOS, web)
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

The whole point was to re-enable a big chunk of the remote-config test suite so it's almost all tests.

But the side goal was to fix any errors that it exposed in all the packages and repair them - backed by testing

If e2e passes a flake hammer run, it's good

---

### DIFF
--- a/packages/remote-config/e2e/helpers.js
+++ b/packages/remote-config/e2e/helpers.js
@@ -1,18 +1,36 @@
 const { getE2eTestProject } = require('../../app/e2e/helpers');
 
 exports.updateTemplate = async function updateTemplate(operations) {
-  // console.error('remoteConfig::helpers::updateTemplate');
-  const response = await fetch(
-    'https://us-central1-' +
-      getE2eTestProject() +
-      '.cloudfunctions.net/testFunctionRemoteConfigUpdateV2',
-    {
-      method: 'post',
-      body: JSON.stringify({ data: operations }),
-      headers: { 'Content-Type': 'application/json' },
-    },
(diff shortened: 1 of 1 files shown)


mikehardy:
The other platform bug fixes on error handling are related to work a customer is interested in so I want to merge this and keep moving but as ever, happy to entertain any comments and post a follow-on

PR #8744, opened 2025-11-07
feat(firestore): implement withConverter

### LINKED ISSUE #8611: [🐛] 🔥Firestore Typescript types are not working anymore with 22.3.0
## Issue

We noticed that our TypeScript setup doesn't work anymore with the latest `22.3.0`. This is most likely caused by https://github.com/invertase/react-native-firebase/pull/8378.

One cause could be that the following type points to `Query<AppModelType, DbModelType>`:

https://github.com/invertase/react-native-firebase/blob/7eaaf185aab15c7e4fc3c218af9be609fe7e0030/packages/firestore/lib/modular/query.d.ts#L103-L107

But that type (and similar ones in this file) was not updated and doesn't support the second parameter:

https://github.com/invertase/react-native-firebase/blob/7eaaf185aab15c7e4fc3c218af9be609fe7e0030/packages/firestore/lib/index.d.ts#L980

---

For reference this is how this is typed in `firebase-js-sdk`:

https://github.com/firebase/firebase-js-sdk/blob/d91169f061bf1dcbfe78a8c8a7f739677608fcb7/packages/firestore/src/lite-api/query.ts#L136-L140

Using this type:

https://github.com/firebase/firebase-js-sdk/blob/d91169f061bf1dcbfe78a8c8a7f739677608fcb7/packages/firestore/src/lite-api/reference.ts#L118-L121

---

<details><summary>Reproducible example with type errors</summary>
<p>

#### `rnfirebase-example.ts`:

- ✅: works as expected
- 🤔: The custom type disappeared for some reason.
- 🚨: why is this suddenly `any`?

```ts
import {
    collection,
    doc,
    FirebaseFirestoreTypes,
    getDoc,
    getDocs,
    getFirestore,
    limit,
    query,
} from '@react-native-firebase/firestore'

type UserType = { id: string }

const firestore = getFirestore()

/// OLD API ///

{
    // GET SINGLE DOCUMENT //

    // 22.2.1    `FirebaseFirestoreTypes.DocumentReference<UserType>` ✅
    // 22.3.0    `FirebaseFirestoreTypes.DocumentReference<UserType>` ✅
    // Expected: `FirebaseFirestoreTypes.DocumentReference<UserType>`
    const userReferenceWithT = doc(
        firestore,
        'users/123',
        // this cast worked in 22.2.1, should actually be `doc<UserType>`
    ) as FirebaseFirestoreTypes.DocumentReference<UserType>
    // 22.2.1    `FirebaseFirestoreTypes.DocumentSnapshot<UserType>` ✅
    // 22.3.0    `FirebaseFirestoreTypes.DocumentSnapshot<UserType>` ✅
    // Expected: `FirebaseFirestoreTypes.DocumentSnapshot<UserType>`
    const userSnapshotWithT = await getDoc(userReferenceWithT)
    // 22.2.1    `UserType | undefined` ✅
    // 22.3.0    `UserType | undefined` ✅
    // Expected: `UserType | undefined`
    const _userWithT = userSnapshotWithT.data()

    // GET MULTIPLE DOCUMENTS FROM COLLECTION //

    // 22.2.1:   `FirebaseFirestoreTypes.Query<UserType>` ✅
    // 22.3.0:   `FirebaseFirestoreTypes.Query<UserType>` ✅
    // Expected: `FirebaseFirestoreTypes.Query<UserType>`
    const usersQueryReferenceWithT = query(
        // this cast worked in 22.2.1, should actually be `query<UserType>`
        collection(firestore, 'users') as FirebaseFirestoreTypes.CollectionReference<UserType>,
        limit(10),
    )
    // 22.2.1    `FirebaseFirestoreTypes.QuerySnapshot<UserType>` ✅
    // 22.3.0    `any` 🚨
    // Expected: `FirebaseFirestoreTypes.QuerySnapshot<UserType, UserType>`
    const usersSnapshotWithT = await getDocs(usersQueryReferenceWithT)
    // 22.2.1: `UserType[]` ✅
    // 22.3.0 `any` 🚨
    // Expected: `UserType[]`
    const _usersWithT = usersSnapshotWithT.docs.map((d) => d.data())
}

/// UPDATED API ///

{
    // GET SINGLE DOCUMENT //

    // 22.2.1:   ... didn't exist ...
    // 22.3.0    `any` 🚨
    // Expected: `FirebaseFirestoreTypes.DocumentReference<UserType, UserType>`
    const userReferenceWithT = doc<UserType, UserType>(firestore, 'users/123')
    // 22.2.1:   ... didn't exist ...
    // 22.3.0    `FirebaseFirestoreTypes.DocumentSnapshot<unknown>` 🤔
    // Expected: `FirebaseFirestoreTypes.DocumentSnapshot<UserType, UserType>`
    const userSnapshotWithT = await getDoc(userReferenceWithT)
    // 22.2.1:   ... didn't exist ...
    // 22.3.0    `unknown` 🤔
    // Expected: `UserType | undefined`
    const _userWithT = userSnapshotWithT.data()

    // GET MULTIPLE DOCUMENTS FROM COLLECTION //

    // 22.2.1:   ... didn't exist ...
    // 22.3.0:   `any` 🚨
    // Expected: `FirebaseFirestoreTypes.Query<UserType, UserType>`
    const usersQueryReferenceWithT = query<UserType, UserType>(
        collection(firestore, 'users'),
        limit(10), // This isn't valid for some reason 🚨
    )
    // 22.2.1:   ... didn't exist ...
    // 22.3.0    `any` 🚨
    // Expected: `FirebaseFirestoreTypes.QuerySnapshot<UserType, UserType>`
    const usersSnapshotWithT = await getDocs(usersQueryReferenceWithT)
    // 22.2.1:   ... didn't exist ...
    // 22.3.0 `any` 🚨
    // Expected: `UserType[]`
    const _usersWithT = usersSnapshotWithT.docs.map((d) => d.data())
}
```

</details>

---

### LINKED ISSUE #8834: [🐛] `collection()` and `query()` return `any` due to wrong generic arity
## Issue

The modular API functions `collection()` and `query()` return `any` instead of properly typed `CollectionReference` and `Query` because the type definitions use 2 generic type arguments (`AppModelType`, `DbModelType`) while the actual types only accept 0-1.

In the following example, `col`and `q` have the `any`type:
```
import { collection, getFirestore, query } from "@react-native-firebase/firestore";

const col = collection(getFirestore(), 'cities');
//    ^? const col: any

const q = query(col);
//    ^? const q: any
```
---

### Javascript

<p>

#### `package.json`:

```json
    "@react-native-firebase/firestore": "^23.8.2",
```

```json
# N/A
```

</details>

## Environment

<p>

  - [ ] iOS
  - [ ] Android
  - [x] Both
  - `23.8.2`
  - `firestore`
  - `Yes` & `5.8.3`

</p>
</details>

---

I solved it by applying this patch-package:
```
diff --git a/node_modules/@react-native-firebase/firestore/lib/index.d.ts b/node_modules/@react-native-firebase/firestore/lib/index.d.ts
index 0857035..8d3c901 100644
--- a/node_modules/@react-native-firebase/firestore/lib/index.d.ts
+++ b/node_modules/@react-native-firebase/firestore/lib/index.d.ts
@@ -145,7 +145,7 @@ export namespace FirebaseFirestoreTypes {
    * A `CollectionReference` object can be used for adding documents, getting document references, and querying for
    * documents (using the methods inherited from `Query`).
    */
-  export interface CollectionReference<T extends DocumentData = DocumentData> extends Query<T> {
+  export interface CollectionReference<T extends DocumentData = DocumentData, U extends DocumentData = DocumentData> extends Query<T, U> {
     /**
      * The collection's identifier.
      */
@@ -253,7 +253,7 @@ export namespace FirebaseFirestoreTypes {
    * to the location. The document at the referenced location may or may not exist. A `DocumentReference` can also be used
    * to create a `CollectionReference` to a subcollection.
    */
-  export interface DocumentReference<T extends DocumentData = DocumentData> {
+  export interface DocumentReference<T extends DocumentData = DocumentData, U extends DocumentData = DocumentData> {
     /**
      * The Firestore instance the document is in. This is useful for performing transactions, for example.
      */
@@ -525,7 +525,7 @@ export namespace FirebaseFirestoreTypes {
    * For a DocumentSnapshot that points to a non-existing document, any data access will return 'undefined'.
    * You can use the `exists()` method to explicitly verify a document's existence.
    */
-  export interface DocumentSnapshot<T extends DocumentData = DocumentData> {
+  export interface DocumentSnapshot<T extends DocumentData = DocumentData, U extends DocumentData = DocumentData> {
     /**
      * Method of the `DocumentSnapshot` that signals whether or not the data exists. True if the document exists.
      */
@@ -600,8 +600,8 @@ export namespace FirebaseFirestoreTypes {
    * Since query results contain only existing documents, the exists() method will always be true and data() will never return 'undefined'.
    */
   export interface QueryDocumentSnapshot<
-    T extends DocumentData = DocumentData,
-  > extends DocumentSnapshot<T> {
+    T extends DocumentData = DocumentData, U extends DocumentData = DocumentData
+  > extends DocumentSnapshot<T, U> {
     /**
      * A QueryDocumentSnapshot is always guaranteed to exist.
      */
@@ -978,7 +978,7 @@ export namespace FirebaseFirestoreTypes {
    * A Query refers to a `Query` which you can read or listen to. You can also construct refined `Query` objects by
    * adding filters and ordering.
    */
-  export interface Query<T extends DocumentData = DocumentData> {
+  export interface Query<T extends DocumentData = DocumentData, U extends DocumentData = DocumentData> {
     /**
      * Calculates the number of documents in the result set of the given query, without actually downloading
      * the documents.
@@ -1447,7 +1447,7 @@ export namespace FirebaseFirestoreTypes {
    * can be accessed as an array via the `docs` property or enumerated using the `forEach` method. The number of documents
    * can be determined via the `empty` and `size` properties.
    */
-  export interface QuerySnapshot<T extends DocumentData = DocumentData> {
+  export interface QuerySnapshot<T extends DocumentData = DocumentData, U extends DocumentData = DocumentData> {
     /**
      * An array of all the documents in the `QuerySnapshot`.
      */

```

JiriHoffmann:
### Description

Hello! I'm sorry for introducing such a large PR, but I hope it's gonna be worth it. 

This PR adds the withCoverter functionality to both the namespace and modular APIs and updates types accordingly.

Both type and e2e tests are a 1:1 copy of the Firebase JS SDK to ensure feature parity.

### Related issues

- Closes #8611 
- Closes #8698 
- Closes #8672
- Closes #8834

### Release Summary

Implements `withConverter` from Firebase JS SDK

### Checklist

  - [x] Yes
  - [x] `Android`
  - [x] `iOS`
  - [ ] `Other` (macOS, web)
- My change includes tests;
- This is a breaking change;
  - [x] Yes
  - [ ] No

:fire:

### DIFF
--- a/packages/firestore/__tests__/firestore.test.ts
+++ b/packages/firestore/__tests__/firestore.test.ts
@@ -962,6 +962,7 @@ describe('Firestore', function () {
 
         collectionRefV9Deprecation(
           () => getCountFromServer(query),
+          // @ts-expect-error Combines modular and namespace API
           () => query.count(),
           'count',
         );
@@ -974,6 +975,7 @@ describe('Firestore', function () {
 
         collectionRefV9Deprecation(
           () => getCountFromServer(query),
+          // @ts-expect-error Combines modular and namespace API
           () => query.countFromServer(),
           'countFromServer',
         );
@@ -986,6 +988,7 @@ describe('Firestore', function () {
 
         collectionRefV9Deprecation(
           () => endAt('foo'),
+          // @ts-expect-error Combines modular and namespace API
           () => query.endAt('foo'),
           'endAt',
         );
@@ -998,6 +1001,7 @@ describe('Firestore', function () {
 
         collectionRefV9Deprecation(
           () => endBefore('foo'),
+          // @ts-expect-error Combines modular and namespace API
           () => query.endBefore('foo'),
           'endBefore',
         );
@@ -1010,6 +1014,7 @@ describe('Firestore', function () {
 
         collectionRefV9Deprecation(
           () => getDocs(query),
+          // @ts-expect-error Combines modular and namespace API
           () => query.get(),
           'get',
         );
@@ -1023,6 +1028,7 @@ describe('Firestore', function () {
         collectionRefV9Deprecation(
           // no equivalent method
           () => {},
+          // @ts-expect-error Combines modular and namespace API
           () => query.isEqual(query),
           'isEqual',
         );
@@ -1035,6 +1041,7 @@ describe('Firestore', function () {
 
         collectionRefV9Deprecation(
           () => limit(9),
+          // @ts-expect-error Combines modular and namespace API
           () => query.limit(9),
           'limit',
         );
@@ -1047,6 +1054,7 @@ describe('Firestore', function () {
 
         collectionRefV9Deprecation(
           () => limitToLast(9),
+          // @ts-expect-error Combines modular and namespace API
           () => query.limitToLast(9),
           'limitToLast',
         );
@@ -1059,6 +1067,7 @@ describe('Firestore', function () {
 
         collectionRefV9Deprecation(
           () => onSnapshot(query, () => {}),
+          // @ts-expect-error Combines modular and namespace API
           () => query.onSnapshot(() => {}),
           'onSnapshot',
         );
@@ -1071,6 +1080,7 @@ describe('Firestore', function () {
 
         collectionRefV9Deprecation(
           () => orderBy('foo', 'asc'),
+          // @ts-expect-error Combines modular and namespace API
           () => query.orderBy('foo', 'asc'),
           'orderBy',
         );
@@ -1083,6 +1093,7 @@ describe('Firestore', function () {
 
         collectionRefV9Deprecation(
           () => startAfter('foo'),
+          // @ts-expect-error Combines modular and namespace API
           () => query.startAfter('foo'),
           'startAfter',
         );
@@ -1095,6 +1106,7 @@ describe('Firestore', function () {
 
         collectionRefV9Deprecation(
           () => startAt('foo'),
+          // @ts-expect-error Combines modular and namespace API
           () => query.startAt('foo'),
           'startAt',
         );
@@ -1107,6 +1119,7 @@ describe('Firestore', function () {
 
         collectionRefV9Deprecation(
           () => where('foo', '==', 'bar'),
+          // @ts-expect-error Combines modular and namespace API
           () => query.where('foo', '==', 'bar'),
           'where',
         );
@@ -1119,6 +1132,7 @@ describe('Firestore', function () {
 
         collectionRefV9Deprecation(
           () => addDoc(query, { foo: 'bar' }),
+          // @ts-expect-error Combines modular and namespace API
           () => query.add({ foo: 'bar' }),
           'add',
         );
@@ -1131,6 +1145,7 @@ describe('Firestore', function () {
 
         collectionRefV9Deprecation(
           () => doc(query, 'bar'),
+          // @ts-expect-error Combines modular and namespace API
           () => query.doc('foo'),
           'doc',
         );
--- a/packages/firestore/e2e/withConverter.e2e.js
+++ b/packages/firestore/e2e/withConverter.e2e.js
@@ -1,0 +1,615 @@
+/*
+ * Copyright (c) 2021-present Invertase Limited & Contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this library except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+const COLLECTION = 'firestore';
+const { wipe } = require('./helpers');
+
+const {
+  getFirestore,
+  doc,
+  collection,
+  refEqual,
+  addDoc,
+  setDoc,
+  getDoc,
+  query,
+  where,
+  getDocs,
+  writeBatch,
+  increment,
+  initializeFirestore,
+} = firestoreModular;
+
+// Used for testing the FirestoreDataConverter.
+class Post {
+  constructor(title, author, id = 1) {
+    this.title = title;
+    this.author = author;
+    this.id = id;
+  }
+  byline() {
+    return this.title + ', by ' + this.author;
+  }
+}
+
+const postConverter = {
+  toFirestore(post) {
+    return { title: post.title, author: post.author };
+  },
+  fromFirestore(snapshot) {
+    const data = snapshot.data();
+    return new Post(data.title, data.author);
+  },
+};
+
+const postConverterMerge = {
+  toFirestore(post, options) {
+    if (
+      options &&
+      ((options && options.merge === true) ||
+        (options && Array.isArray(options.mergeFields) && options.mergeFields.length > 0))
+    ) {
+      post.should.not.be.an.instanceof(Post);
+    } else {
+      post.should.be.an.instanceof(Post);
+    }
+    const result = {};
+    if (post.title) {
+      result.title = post.title;
+    }
+    if (post.author) {
+      result.author = post.author;
+    }
+    return result;
+  },
+  fromFirestore(snapshot) {
+    const data = snapshot.data();
+    return new Post(data.title, data.author);
+  },
+};
+
+// v8 compatibility helper functions
+function modifyIgnoreUndefinedProperties(db, value) {
+  // JS SDK settings can only be called once
+  if (Platform.other) {
+    db._settings.ignoreUndefinedProperties = value;
+  } else {
+    db.settings({ ignoreUndefinedProperties: value });
+  }
+}
+
+// modular helper functions
+function withTestDb(fn) {
+  return fn(getFirestore());
+}
+
+async function withModifiedUndefinedPropertiesTestDb(fn) {
+  const db = getFirestore();
+  const previousValue = db._settings.ignoreUndefinedProperties;
+  initializeFirestore(db.app, { ignoreUndefinedProperties: false });
+  await fn(db);
+  initializeFirestore(db.app, { ignoreUndefinedProperties: previousValue });
+}
+
+function withTestCollection(fn) {
+  return withTestDb(db => fn(collection(db, COLLECTION)));
+}
+function withTestDoc(fn) {
+  return withTestDb(db => fn(doc(db, `${COLLECTION}/doc`)));
+}
+
+function withTestCollectionAndInitialData(data, fn) {
+  return withTestDb(async db => {
+    const coll = collection(db, COLLECTION);
+    for (const element of data) {
+      const ref = doc(coll);
+      await setDoc(ref, element);
+    }
+    return fn(coll);
+  });
+}
+
+describe('firestore.withConverter', function () {
+  describe('v8 compatibility', function () {
+    beforeEach(async function beforeEachTest() {
+      // @ts-ignore
+      globalThis.RNFB_SILENCE_MODULAR_DEPRECATION_WARNINGS = true;
+    });
+
+    afterEach(async function afterEachTest() {
+      // @ts-ignore
+      globalThis.RNFB_SILENCE_MODULAR_DEPRECATION_WARNINGS = false;
+    });
+
+    before(function () {
+      return wipe();
+    });
+
+    it('for collection references', function () {
+      const firestore = firebase.firestore();
+      const coll1a = firestore.collection('a');
+      const coll1b = firestore.doc('a/b').parent;
+      const coll2 = firestore.collection('c');
+
+      coll1a.isEqual(coll1b).should.be.true();
+      coll1a.isEqual(coll2).should.be.false();
+
+      const coll1c = firestore.collection('a').withConverter({
+        toFirestore: data => data,
+        fromFirestore: snap => snap.data(),
+      });
+      coll1a.isEqual(coll1c).should.be.false();
+
+      try {
+        coll1a.isEqual(firestore.doc('a/b'));
+        return Promise.reject(new Error('Did not throw an Error.'));
+      } catch (error) {
+        error.message.should.containEql('expected a Query instance.');
+        return Promise.resolve();
+      }
+    });
+
+    it('for document references', function () {
+      const firestore = firebase.firestore();
+      const doc1a = firestore.doc('a/b');
+      const doc1b = firestore.collection('a').doc('b');
+      const doc2 = firestore.doc('a/c');
+
+      doc1a.isEqual(doc1b).should.be.true();
+      doc1a.isEqual(doc2).should.be.false();
+
+      try {
+        const doc1c = firestore.collection('a').withConverter({
+          toFirestore: data => data,
+          fromFirestore: snap => snap.data(),
+        });
+        doc1a.isEqual(doc1c);
+        return Promise.reject(new Error('Did not throw an Error.'));
+      } catch (error) {
+        error.message.should.containEql('expected a DocumentReference instance.');
+      }
+
+      try {
+        doc1a.isEqual(firestore.collection('a'));
+        return Promise.reject(new Error('Did not throw an Error.'));
+      } catch (error) {
+        error.message.should.containEql('expected a DocumentReference instance.');
+      }
+      return Promise.resolve();
+    });
+
+    it('for DocumentReference.withConverter()', async function () {
+      const firestore = firebase.firestore();
+      let docRef = firestore.doc(`${COLLECTION}/doc`);
+      docRef = docRef.withConverter(postConverter);
+      await docRef.set(new Post('post', 'author'));
+      const postData = await docRef.get();
+      const post = postData.data();
+      post.should.not.be.undefined();
+      post.byline().should.equal('post, by author');
+    });
+
+    it('for DocumentReference.withConverter(null) applies default converter', function () {
+      const firestore = firebase.firestore();
+      const coll = firestore
+        .collection(COLLECTION)
+        .withConverter(postConverter)
+        .withConverter(null);
+      try {
+        return coll
+          .doc('post1')
+          .set(10)
+          .then(() => Promise.reject(new Error('Did not throw an Error.')));
+      } catch (error) {
+        error.message.should.containEql(
+          `firebase.firestore().doc().set(*) 'data' must be an object.`,
+        );
+        return Promise.resolve();
+      }
+    });
+
+    it('for CollectionReference.withConverter()', async function () {
+      const firestore = firebase.firestore();
+      let coll = firestore.collection(COLLECTION);
+      coll = coll.withConverter(postConverter);
+      const docRef = await coll.add(new Post('post', 'author'));
+      const postData = await docRef.get();
+      const post = postData.data();
+      post.should.not.be.undefined();
+      post.byline().should.equal('post, by author');
+    });
+
+    it('for CollectionReference.withConverter(null) applies default converter', function () {
+      const firestore = firebase.firestore();
+      let docRef = firestore.doc(`${COLLECTION}/doc`);
+      try {
+        docRef = docRef.withConverter(postConverter).withConverter(null);
+        return docRef.set(10).then(() => Promise.reject(new Error('Did not throw an Error.')));
+      } catch (error) {
+        error.message.should.containEql(
+          `firebase.firestore().doc().set(*) 'data' must be an object.`,
+        );
+        return Promise.resolve();
+      }
+    });
+
+    it('for Query.withConverter()', async function () {
+      const firestore = firebase.firestore();
+      const collRef = firestore.collection(COLLECTION);
+      await collRef.add({ title: 'post', author: 'author' });
+      let query1 = collRef.where('title', '==', 'post');
+      query1 = query1.withConverter(postConverter);
+      const result = await query1.get();
+      result.docs[0].data().should.be.an.instanceOf(Post);
+      result.docs[0].data().byline().should.equal('post, by author');
+    });
+
+    it('for Query.withConverter(null) applies default converter', async function () {
+      const firestore = firebase.firestore();
+      const collRef = firestore.collection(COLLECTION);
+      await collRef.add({ title: 'post', author: 'author' });
+      let query1 = collRef.where('title', '==', 'post');
+      query1 = query1.withConverter(postConverter).withConverter(null);
+      const result = await query1.get();
+      result.docs[0].should.not.be.an.instanceOf(Post);
+    });
+
+    it('keeps the converter when calling parent() with a DocumentReference', function () {
+      const db = firebase.firestore();
+      const coll = db.doc('root/doc').withConverter(postConverter);
+      const typedColl = coll.parent;
+      typedColl.isEqual(db.collection('root').withConverter(postConverter)).should.be.true();
+    });
+
+    it('drops the converter when calling parent() with a CollectionReference', function () {
+      const db = firebase.firestore();
+      const coll = db.collection('root/doc/parent').withConverter(postConverter);
+      const untypedDoc = coll.parent;
+      untypedDoc.isEqual(db.doc('root/doc')).should.be.true();
+    });
+
+    it('checks converter when comparing with isEqual()', function () {
+      const db = firebase.firestore();
+      const postConverter2 = { ...postConverter };
+
+      const postsCollection = db.collection('users/user1/posts').withConverter(postConverter);
+      const postsCollection2 = db.collection('users/user1/posts').withConverter(postConverter2);
+      postsCollection.isEqual(postsCollection2).should.be.false();
+
+      const docRef = db.doc('some/doc').withConverter(postConverter);
+      const docRef2 = db.doc('some/doc').withConverter(postConverter2);
+      docRef.isEqual(docRef2).should.be.false();
+    });
+
+    it('requires the correct converter for Partial usage', async function () {
+      const db = firebase.firestore();
+      const previousValue = db._settings.ignoreUndefinedProperties;
+      modifyIgnoreUndefinedProperties(db, false);
+
+      const coll = db.collection('posts');
+      const ref = coll.doc('post').withConverter(postConverter);
+      const batch = db.batch();
+
+      try {
+        batch.set(ref, { title: 'olive' }, { merge: true });
+        return Promise.reject(new Error('Did not throw an Error.'));
+      } catch (error) {
+        error.message.should.containEql('Unsupported field value: undefined');
+      }
+      modifyIgnoreUndefinedProperties(db, previousValue);
+      return Promise.resolve();
+    });
+
+    it('supports primitive types with valid converter', async function () {
+      const firestore = firebase.firestore();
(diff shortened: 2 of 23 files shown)


mikehardy:
> Hello! I'm sorry for introducing such a large PR, but I hope it's gonna be worth it.

That is not something I'm used to people apologizing for, this *certainly* looks worth it. Thanks! Going in the review queue - additionally - if my patch-package generation workflow is working correctly you should be able to start depending on the work relatively easily already via usage of the patch, which should apply cleanly to the current released version of `firestore` package on npmjs.com

mikehardy:
Looks like there are a couple things going on here in the CI failures -

1- there's an error in types-test with an unused 'doc' element
  - /home/runner/work/react-native-firebase/react-native-firebase/website/scripts/generate-typedoc.js:37:18

2- there's some use of non-modular code in the new modular e2e tests 
  - /Users/runner/work/react-native-firebase/react-native-firebase/packages/firestore/e2e/withConverter.e2e.js:274:17

3- Then, seemingly unrelated to this PR, there's a problem with the ios build where some sort of entropy ("things just fail over time for unrelated reasons") has taken over and resulted in 

> xcodebuild: error: Found no destinations for the scheme 'testing' and action build.

1 and 2 are for this PR, I'll tackle 3 separately and either cherry-pick or rebase it into this PR, coordinating with you as/when I do so

packages/firestore/e2e/withConverter.e2e.js
```diff
@@ -0,0 +1,595 @@
...
+        const docRef = doc(db, 'some/doc').withConverter(postConverter);
+        const docRef2 = doc(db, 'some/doc').withConverter(postConverter2);
+        refEqual(docRef, docRef2).should.be.false();
+      });
+    });
+
+    it('requires the correct converter for Partial usage', async function () {
+      return withTestDb(async db => {
+        db._settings.ignoreUndefinedProperties = false;
+        const coll = collection(db, 'posts');
```
JiriHoffmann: Tests should be fixed now, it was due to the usage of `firestore().settings()` here. This is the only test that needs the `ignoreUndefinedProperties` property to be adjusted. Please let me know if there is a better way to implement this.

mikehardy: the method to call is actually `initializeFirestore` - native lets us call that any time in order to change settings, contrary to the JS SDK documentation, but the API does match the JS API so we meet our "API compatible" goal, and are just a bit more permissive on when it may be invoked.

You should be able to use that and see the ignoreUndefinedProperties setting take hold

mikehardy:
This PR will merge as soon as it goes green, and I'll rebase this PR on top of it, at which point this PR should go green...

- #8748

packages/firestore/e2e/withConverter.e2e.js
mikehardy: I think your approach is okay for now - I'll note it to see if I can do a follow-on that uses the API I mentioned but I don't think there's much value in holding this PR up for it

mikehardy:
I've fixed the CI issues that were happening with iOS and rebased this PR on to them then force-pushed this PR back out with that new commit history built on the iOS e2e fix

I resolved the conversation on the ignoreUndefinedProperties settings and peeled it off to a separate issue so we can move on with out it

This is now just on me to do a real code review on it, then if it's good - merge and release. Will be a huge feature, there will be much rejoicing

JiriHoffmann:
Updated this PR to include Firestore settings changes for tests from https://github.com/invertase/react-native-firebase/issues/8749

mikehardy:
The other platform (which is a wrapper for firebase-js-sdk) test failure is interesting - a perfect example of "no good deed unpunished" as you tried to use the suggested initializeFirestore method, but firebase-js-sdk *does* enforce that this is called only once.

For the e2e case where `isOther` is true (I think that's how to phrase it...) I guess you just have to directly toggle the setting as it was before. Which is frustrating since you just stopped doing that!

```
           requires the correct converter for Partial usage:
     FirebaseError: Firestore can only be initialized once per app.
      at construct ([native code]:null:null)
      at _construct (/Users/runner/work/react-native-firebase/react-native-firebase/packages/app/node_modules/@babel/runtime/helpers/construct.js:4:64)
      at Wrapper (/Users/runner/work/react-native-firebase/react-native-firebase/packages/app/node_modules/@babel/runtime/helpers/wrapNativeSuper.js:15:13)
      at construct ([native code]:null:null)
      at _callSuper (http://localhost:8081/index.bundle?platform=macos&dev=true&lazy=true&minify=false&inlineSourceMap=false&modulesOnly=false&runModule=true&app=org.reactjs.native.io-invertase-testing:125161:169)
      at FirebaseError#constructor (/Users/runner/work/react-native-firebase/react-native-firebase/packages/app/node_modules/@firebase/util/dist/index.esm.js:1181:8)
      at construct ([native code]:null:null)
      at _callSuper (http://localhost:8081/index.bundle?platform=macos&dev=true&lazy=true&minify=false&inlineSourceMap=false&modulesOnly=false&runModule=true&app=org.reactjs.native.io-invertase-testing:368717:169)
      at FirestoreError#constructor (/Users/runner/work/react-native-firebase/react-native-firebase/packages/app/node_modules/@firebase/firestore/dist/lite/index.rn.esm.js:239:8)
      at initializeFirestore (/Users/runner/work/react-native-firebase/react-native-firebase/packages/app/node_modules/@firebase/firestore/dist/lite/index.rn.esm.js:4509:52)
      at guard$argument_0 (/Users/runner/work/react-native-firebase/react-native-firebase/packages/firestore/lib/web/RNFBFirestoreModule.js:169:42)
      at guard (/Users/runner/work/react-native-firebase/react-native-firebase/packages/app/lib/internal/web/utils.js:5:11)
      at default.settings (/Users/runner/work/react-native-firebase/react-native-firebase/packages/firestore/lib/web/RNFBFirestoreModule.js:168:16)
      at <anonymous> (/Users/runner/work/react-native-firebase/react-native-firebase/packages/app/lib/internal/registry/nativeModule.js:43:34)
      at FirebaseFirestoreModule#settings (/Users/runner/work/react-native-firebase/react-native-firebase/packages/firestore/lib/index.js:391:31)
      at <anonymous> (/Users/runner/work/react-native-firebase/react-native-firebase/packages/app/lib/common/index.js:729:37)
      at initializeFirestore (/Users/runner/work/react-native-firebase/react-native-firebase/packages/firestore/lib/modular/index.js:217:31)
      at asyncGeneratorStep (/Users/runner/work/react-native-firebase/react-native-firebase/tests/node_modules/@babel/runtime/helpers/asyncToGenerator.js:3:16)
      at _next (/Users/runner/work/react-native-firebase/react-native-firebase/tests/node_modules/@babel/runtime/helpers/asyncToGenerator.js:17:26)
      at Promise$argument_0 (/Users/runner/work/react-native-firebase/react-native-firebase/tests/node_modules/@babel/runtime/helpers/asyncToGenerator.js:22:11)
      at tryCallTwo (/Users/runner/work/react-native-firebase/react-native-firebase/tests/node_modules/promise/setimmediate/core.js:45:6)
      at doResolve (/Users/runner/work/react-native-firebase/react-native-firebase/tests/node_modules/promise/setimmediate/core.js:200:22)
      at Promise (/Users/runner/work/react-native-firebase/react-native-firebase/tests/node_modules/promise/setimmediate/core.js:66:11)
      at <anonymous> (/Users/runner/work/react-native-firebase/react-native-firebase/tests/node_modules/@babel/runtime/helpers/asyncToGenerator.js:14:22)
      at initializeFirestore (/Users/runner/work/react-native-firebase/react-native-firebase/packages/firestore/lib/modular/index.js:214:41)
      at withModifiedUndefinedPropertiesTestDb (/Users/runner/work/react-native-firebase/react-native-firebase/packages/firestore/e2e/withConverter.e2e.js:91:38)
```

mikehardy:
I just squashed the commits here down to a single commit since the rest were just follow-ons after the first, and rebased this to current main which should have CI deflaked plus have updated SDKs and such

This is in final review now and I'm hoping to merge it in a sec - thanks for your patience

packages/firestore/type-test.ts
```diff
@@ -147,7 +165,7 @@ getDocsFromServer(
 ).then();
 onSnapshot(doc(firebase.firestore(), 'foo', 'foo'), () => {});
 onSnapshot(doc(firebase.firestore(), 'foo', 'foo'), {
-  next: (snapshot: FirebaseFirestoreTypes.DocumentSnapshot) => {
```
mikehardy, line 152: I'm a little concerned about the typing items being breaking changes - what are your thoughts on that? Will this be drop in for people when this feature ships, or will some typescript code need updating.

We can issue breaking changes, that isn't a problem. The problem is accidentally issuing a non-major semantic release version with breaking changes - we cannot do that.

packages/firestore/e2e/withConverter.e2e.js
```diff
@@ -0,0 +1,615 @@
...
+};
+
+// v8 compatibility helper functions
+function modifyIgnoreUndefinedProperties(db, value) {
+  // JS SDK settings can only be called once
+  if (Platform.other) {
+    db._settings.ignoreUndefinedProperties = value;
+  } else {
+    db.settings({ ignoreUndefinedProperties: value });
+  }
```
mikehardy, line 91: Still can't believe that was necessary - such a trivial part of this whole work and ended up being a pain. Sorry

mikehardy (commented):
The quality here is amazing in my opinion. It is huge but I looked through it all...it seems not just like it matches firebase-js-sdk types exactly but of course implements the advertised feature and is backed by thorough type unit and e2e testing. Easy +1

My *only* question is: is this a breaking change or not? I don't think it is, but I'm curious for your thoughts. Related: if in your estimation it is a breaking change currently, is there any way we could put deprecated shims in the types so it wasn't breaking now, and then we could remove them in a future release? Cheers

packages/firestore/type-test.ts
JiriHoffmann: That is a good catch, this would be a breaking change. The JS SDK modular and namespace converter types differ slightly, so my thought process was to follow the same type structure. Anything exported through the `FirebaseFirestoreTypes` namespace belongs to v8, and v9 modular API is exported at the top level just like the JS SDK. In my opinion it would better to keep it separate and as a breaking change instead of trying to mix them together.

The release notes should probably include a comment about this change and I can add documentation as well. With that said, the types now become slightly awkward when using the modular API since `DocumentSnapshot`, `Query`, etc.. have top level exports, while, for example, `WhereFilterOp` and `Module` still have to be imported through the `FirebaseFirestoreTypes` namespace.

JiriHoffmann:
Also noticed that `packages/firestore/lib/modular/snapshot.d.ts` is missing a `DocumentData` import. Will fix that together with any necessary changes based on the types conversation.

packages/firestore/type-test.ts
mikehardy: If the types are awkward but nevertheless follow firebase-js-sdk then no problem - awkward still but we have  "taste" reference that is "just follow firebase-js-sdk". If they are awkward but divergent - and especially if we are doing a breaking change for types, then this would be a good time to true them up.

I'll reiterate that I'm not afraid of a breaking change (we're on semver major number 23 after all) I just try to be as empathetic as possible and include clear and succinct "this change, impact is that, likely adaptation is ..." so API users can take advantage of what we already know instead of spending limited daily thinking time having to derive it themselves

So the goal then - if this is a breaking change - is just to do it really well, make good docs, then send it...

JiriHoffmann: Alright, I've added a new docs section for the breaking change. Let me know if it's good.

packages/firestore/lib/modular/index.d.ts
```diff
@@ -1,20 +1,29 @@
...
 export type Primitive = string | number | boolean | undefined | null;
 
+/**
+ * A `DocumentData` object represents the data in a document.
+ * - Same as {@link FirebaseFirestoreTypes.DocumentData}
+ */
+export interface DocumentData {
+  [key: string]: any;
+}
+
```
JiriHoffmann, line 26: I've also changed some of these to be exported individually, in addition to `FirebaseFirestoreTypes` - in line with JS SDK.

`DocumentData` redefinition here is to make modular types more readable:

Image

vs 

Image

packages/firestore/lib/modular/snapshot.d.ts
```diff
@@ -1,14 +1,195 @@
...
+   * const user = await firebase.firestore().doc('users/alovelace').get();
+   *
+   * console.log('Address ZIP Code', user.get('address.zip'));
+   * ```
+   *
+   * @param fieldPath The path (e.g. 'foo' or 'foo.bar') to a specific field.
+   */
+  get<fieldType extends DbModelType[keyof DbModelType]>(
+    fieldPath: keyof DbModelType | string | FieldPath,
+  ): fieldType;
```
macksal, line 65: This doesn't work the way you expect. The function should be generic on the field path, and the return type should be the type of the document's field at that path.

[Playground link](https://www.typescriptlang.org/play/?#code/MYewdgzgLgBAYgSQKIBkAiAFAggFQBIwC8MAygJ4C2ARiADYAUARIqprnowJQDcAUFGQAOAU3gBLYbQAmGAIZQAFkRgCRIAGbxk6bPj68xYKMIBO62cFEAlYetPCwlgDxoqAWRBTJOIaOEAPYzApCBgAbwBfAD5w3l4YBJgAc2EoJ3UJaR8RGACgkJhXDy9abOEAbQBrYTINQvdPb18AXSj6eMSMyRl5BQAuGGrazSLG0t8YAB8YaBNDJKnxbrlFABoOzgGurN99CLjDYzMLUTcyNBBgAFcKB1gwjoTZEig5sCSB2fm+RJhZADkblRTAMwEDTHx9l5gLRZCZRKBILB4eoBjY7PDHMInGcLtdbkYovoAPTEmAAASgEAAtAERMAoLSTCYQCZkqkoPMVApRAB3FnvFQTCAKEBXaRgADksF5rMqvER0Bg-lB4LZxBRADoUlAmM9XvMuCSyZSaXThAymSy2TrOYLFHyBQtVKIRWKJdKYLKTPLFbAyJ8DYKNbZtak9YDqKYjUA)

packages/firestore/lib/modular/snapshot.d.ts
```diff
@@ -1,14 +1,195 @@
...
+/**
+ * A QueryDocumentSnapshot contains data read from a document in your Firestore database as part of a query.
+ * The document is guaranteed to exist and its data can be extracted with .data() or .get(:field) to get a specific field.
+ *
+ * A QueryDocumentSnapshot offers the same API surface as a DocumentSnapshot.
+ * Since query results contain only existing documents, the exists() method will always be true and data() will never return 'undefined'.
+ */
+export class QueryDocumentSnapshot<
+  AppModelType = DocumentData,
+  DbModelType extends DocumentData = DocumentData,
```
macksal, line 94: Most users won't be using withConverter (since it doesn't currently exist).

Would it make more sense to type this and all related types as:

`class Query<DbModelType extends DocumentData = DocumentData, AppModelType = DbModelType>`

In this way, if I don't use withConverter, and rely on the types in the DB being correct, I can just provide a single type parameter e.g. `Query<{field: string; otherField: number}>` and the AppModelType will be the same as whatever's in the DB. (This is how the library worked before this feature, although the types have been broken for several years somehow).

mikehardy: the "somehow" on types being broken is the inevitable result of the library existing since dinosaurs roamed the earth and *Flow* typing was more popular than typescript. When you code in javascript and bolt on types later, it's never pretty and the internal type tests are only an attempt to catch the most basic sort of problems like "the type simply isn't exported at all".

To the point on types and breaking change release or not I'm inclined to just go breaking, make sure the types are perfect fidelity with firebase-js-sdk (within our capacity as fallable humans...) and go with it. I see your note above that you've add docs about the breaking change - that satisfies the only pre-requisite I've got for that, so we are good to merge there.

I see a couple other comments indicating there may be a few little options to resolve, I'll try to work through those as best I can and get this thing merged today

packages/firestore/lib/modular/snapshot.d.ts
mikehardy: @macksal do you have a concrete code suggestion for it to resolve it - apologies if this is well known for you, but if you hit the little +- doc button in the all files diff view, you can suggest exactly what you're thinking so we can see it directly and then directly apply it as a commit in the series of commits on the PR

packages/firestore/lib/modular/snapshot.d.ts
mikehardy: @macksal one vital note:

For any question of this format:

> Would it make more sense to type this and all related types as:

The answer is always 100% "our goal for typing is 100% fidelity with firebase-js-sdk and any typing question should be resolved by consulting the firebase-js-sdk types and copying them here directly"

The only exceptions to that rule are where:

-  we as a wrapper for native SDKs have different APIs than firebase-js-sdk (e.g., things JS SDK doesn't even support, like crashlytics...) - at that point we will formulate local types inspired by firebase-js-sdk style
- or we as a react-native library have be asynchronous where firebase-js-sdk is synchronous, because we have constraints related to the react-native bridge and cannot talk to native methods synchronously until everything is migrated to TurboModules. The alteration here is that we'll have the same types as firebase-js-sdk but our return values may be wrapped in Promise vs being sync

packages/firestore/lib/modular/snapshot.d.ts
JiriHoffmann: @macksal is correct. `fieldType` should extend `DocumentFieldType` instead, as per the current namespace implementation:

https://github.com/invertase/react-native-firebase/blob/4d5463609edd2831edc34cfe8a39b0fd1f6fe8fa/packages/firestore/lib/index.d.ts#L575

Although the js-sdk doesn't use a generic

https://github.com/firebase/firebase-js-sdk/blob/5511b4fa7de0e98f3a3b933645584ba758de4405/packages/firestore/src/api/snapshot.ts#L486-L504

@mikehardy, up to you which one you prefer.

macksal, line 65: Thx for the nudge @mikehardy, I would suggest this:

```suggestion
  get<K extends keyof DbModelType | FieldPath>(
    fieldPath: K,
  ): K extends keyof DbModelType ? DbModelType[K] : any;
```

If the caller passes a string literal, they get back the type of the field with that key. It seems unnecessary to include `string`: DocumentData is the default for DbModelType, and DocumentData is Record<string, any>. So that will allow K to be any string and return any. Meanwhile if a specific type is used for DbModelType, that will restrict K to known keys of that type or FieldPath objects.

Passing a FieldPath can access nested paths, and there's no way for the types to resolve what type is something like `get(new FieldPath(["my", "nested", "property"]))`. FieldPath could also be FieldPath.documentId() in which case the return type is a string. So `any` is the simplest thing to return there. One could make FieldPath generic and retain type info about the full range of keys, and then you could resolve the nested properties with a recursive type helper, but I imagine the complexity isn't worth it.

packages/firestore/lib/modular/snapshot.d.ts
macksal: @mikehardy Thank you for clarifying re firebase-js-sdk, matching those types is a logical goal. In the past I have found myself hoping for common types across admin/rnfb/web sdks. In that case the current types are totally fine.

packages/firestore/lib/modular/snapshot.d.ts
mikehardy: Close inspection of the types prior to merge was on my list today as part of final review, and at least initially I noticed that they are not following firebase-js-sdk exactly but I hadn't commented here yet as I wanted to be certain first. What I saw is that they were not using the generic type either though

In all things we should follow firebsae-js-sdk so where ever there was a difference I was going to alter them here to line them up then put this PR in

I'm fine taking that work on - I need to coordinate this PR's landing across 2 PRs and a typescript conversion so I grabbing control of it and getting it done is the most expedient way

JiriHoffmann: Sounds good, go ahead and take over, if you need anything just let me know!

mikehardy:
quick status report for 2026 - types have been shored up quite a bit separately, and I got the VectorValue PR polished+posted for final review, which means this is next. Then after this, the whole firestore module will get a conversion to native typescript. So this may seem like it's just sitting but in general firestore and react-native-firebase is extremely active at the moment - and this PR will get it's turn

Cheers

mikehardy:
So the 23.8.x release was rougher than I though - but now it's time at long last!
Got this rebased on to main and fixed up, I think - going to be doing final review and get this one so @russellwheatley can get busy with the typescript conversion

packages/firestore/__tests__/firestore.test.ts
```diff
@@ -962,6 +962,7 @@ describe('Firestore', function () {
 
         collectionRefV9Deprecation(
           () => getCountFromServer(query),
+          // @ts-expect-error Combines modular and namespace API
```
russellwheatley, line 965: We probably should create a namespaced query so these annotations aren't needed I think. e.g.

```js
const namespaceQuery = firestore.collection('test');
collectionRefV9Deprecation(
          () => endBefore('foo'),
          () => namespaceQuery.endBefore('foo'),
          'endBefore',
        );
```

mikehardy: In fact that was the most troublesome part, for me - even in firebase-js-sdk they appear to have multiple Query (and Document/Collection/etc) types, one in compat that of the original form before this PR, that is something like `Thing<U>: ReturnedThing<U>` and one that is what we're doing after this PR that is more like `Thing<AppModelType, DbModelType>: ReturnedThing<AppModelType, DbModelType>`

in the interest of moving towards the future and accepting that it's breaking, I just switched everything to the future-state firebase-js-sdk style since our types are sort of welded together between namespaced and modular

I perceived that as acceptable because it is efficient (one break as a DX, one type system - and it's the modern one - as maintainers) vs trying to separate them. But I'm open to ideas

If those ideas are implementing a compat set of types as well, fine by me but I saw it as low-value vs current backlog 🤷 . I accept there will be affected devs that will disagree and I do empathize and would happily merge related PRs.

mikehardy:
Alright! We're merging this thing 🤞 the types are in fidelity with firebase-js-sdk and where they aren't we'll be happily closing gaps with any PRs people submit that get us to firebase-js-sdk fidelity.

@JiriHoffmann thank you very much for doing this. Any other interested parties, please please 🙏 collaborate to smooth out rough edges if there are any left. We'll be releasing as a breaking change once we have a few other type-related things sorted out here in other packages.

ycmjason:
Thank you so much for pushing this! This is such a milestone! 🔥🔥

mikehardy:
My pleasure in so far as I could get it merged - it's been something I wish we had for a long long time, and was amazing that @JiriHoffmann posted it up. For your part @ycmjason I understand from your previous PRs that you're keen on having the types match - I am as well, I will do my best to shepherd any fixup PRs through if you notice anything, or collaborate on anything in that area. Please post up anything you see - it would be remarkable if the types etc were all correct first time so I expect something was missed

PR #8771, opened 2025-11-24
fix(messaging, android): prevent OOM by limiting stored notifications

### LINKED ISSUE #8770: [🐛] [Android] OOM Crash due to excessive notification storage in SharedPreferences
## Issue

🔥 **[Android] OOM Crash due to excessive notification storage in SharedPreferences (ReactNativeFirebaseMessagingStoreImpl)**

We are encountering `java.lang.OutOfMemoryError` on Android devices.
The root cause is that `ReactNativeFirebaseMessagingStoreImpl` stores up to 100 notifications in `SharedPreferences` by default.
When the app receives many notifications with large payloads, the `SharedPreferences` file becomes too large. Since `SharedPreferences` loads the entire file into memory, this leads to OOM crashes, especially on devices with limited heap memory.

**Stack Trace:**
```java
java.lang.OutOfMemoryError: Failed to allocate a 24 byte allocation with 425280 free bytes and 415KB until OOM...
    at java.util.HashMap.newNode(HashMap.java:1907)
    at java.util.HashMap.putVal(HashMap.java:636)
    at java.util.HashMap.putMapEntries(HashMap.java:521)
    at java.util.HashMap.<init>(HashMap.java:491)
    at android.app.SharedPreferencesImpl$EditorImpl.commitToMemory(SharedPreferencesImpl.java:539)
    at android.app.SharedPreferencesImpl$EditorImpl.apply(SharedPreferencesImpl.java:486)
    at io.invertase.firebase.common.UniversalFirebasePreferences.setStringValue(UniversalFirebasePreferences.java:66)
    at io.invertase.firebase.messaging.ReactNativeFirebaseMessagingStoreImpl.storeFirebaseMessage(ReactNativeFirebaseMessagingStoreImpl.java:43)
```

**Analysis:**
The `ReactNativeFirebaseMessagingStoreImpl.java` has a constant `MAX_SIZE_NOTIFICATIONS = 100`.
Every time a notification is received (whether in foreground or background), it is stored in SharedPreferences.
If the notification data is large, storing 100 entries can easily exceed the available heap size during the `SharedPreferences` load/commit process.

**Proposed Solution & Fix:**
We have successfully patched the library locally with the following changes:
1.  **Reduce the default limit**: 100 seems excessive. We reduced it to 20.
2.  **Make it configurable**: Added support for `rn_firebase_messaging_max_stored_notifications` in `AndroidManifest.xml`.
3.  **Safety Cap**: Even if configured higher, we capped the limit at 100 to prevent OOM.
4.  **Improve cleanup logic**: Changed `if` to `while` loop to aggressively clean up excess notifications.

**Example Patch Code:**

```java
// In ReactNativeFirebaseMessagingStoreImpl.java

private static int MAX_SIZE_NOTIFICATIONS = 20; // Default reduced to 20
private static boolean isInitialized = false;

private int getMaxNotificationSize() {
    if (isInitialized) return MAX_SIZE_NOTIFICATIONS;
    try {
        Context context = ReactNativeFirebaseApp.getApplicationContext();
        ApplicationInfo appInfo = context.getPackageManager().getApplicationInfo(context.getPackageName(), PackageManager.GET_META_DATA);
        if (appInfo.metaData != null) {
            int configValue = appInfo.metaData.getInt("rn_firebase_messaging_max_stored_notifications", 20);
            // Safety cap: never exceed 100 to prevent OOM
            MAX_SIZE_NOTIFICATIONS = Math.min(configValue, 100);
        }
    } catch (Exception e) {
        // Ignore
    }
    isInitialized = true;
    return MAX_SIZE_NOTIFICATIONS;
}

@Override
public void storeFirebaseMessage(RemoteMessage remoteMessage) {
    // ...
    // check and remove old notifications message
    List<String> allNotificationList = convertToArray(notifications);

    // Changed 'if' to 'while' for aggressive cleanup
    while (allNotificationList.size() > getMaxNotificationSize()) {
        String firstRemoteMessageId = allNotificationList.get(0);
        preferences.remove(firstRemoteMessageId);
        notifications = removeRemoteMessage(firstRemoteMessageId, notifications);
        allNotificationList.remove(0);
    }
    // ...
}
```

---

### Javascript

<p>

#### `package.json`:

```json
{
  "dependencies": {
    "react-native": "0.81.5",
    "@react-native-firebase/app": "21.7.1",
    "@react-native-firebase/messaging": "21.7.1"
  }
}
```

```json
# N/A
```

</details>

### iOS

<p>

#### `ios/Podfile`:

- [ ] I'm not using Pods

```ruby
# N/A
```

#### `AppDelegate.m`:

```objc
// N/A
```

</p>
</details>

---

### Android

<p>

- [x] I am using `android/gradle.settings` `jetifier=true` for Android compatibility?

#### `android/build.gradle`:

```groovy
// N/A
```

```groovy
// N/A
```

```groovy
// N/A
```

#### `MainApplication.java`:

```java
// N/A
```

#### `AndroidManifest.xml`:

```xml
<!-- We added this configuration to solve the issue -->
<meta-data
    android:name="rn_firebase_messaging_max_stored_notifications"
    android:value="20" />
```

</p>
</details>

---

## Environment

<p>

```
System:
    OS: macOS
Binaries:
    Node: 20.x
    Yarn: N/A
    npm: 10.x
    Watchman: Yes
SDKs:
    iOS SDK:
      Platforms: iOS
    Android SDK:
      API Levels: 34
IDEs:
    Android Studio: Yes
    Xcode: Yes
Languages:
    Java: 17
    Python: N/A
npmPackages:
    @react-native-community/cli: Not Found
    react: 18.3.1
    react-native: 0.81.5
```

  - [ ] iOS
  - [x] Android
  - [ ] Both
  - `21.7.1`
  - `Messaging`
  - `Y`

</p>
</details>

sean5940:
## Description

This PR addresses an `OutOfMemoryError` (OOM) crash on Android caused by `ReactNativeFirebaseMessagingStoreImpl` storing up to 100 notifications in `SharedPreferences`.

When a device receives many notifications with large payloads, the `SharedPreferences` file grows significantly. Since `SharedPreferences` loads the entire file into memory, this can exceed the heap size on some devices, leading to a crash.

**Related Issue:**
Closes #8770

## Changes

1.  **Reduced Default Limit**: The default `MAX_SIZE_NOTIFICATIONS` has been reduced from **100** to **20**.
    - *Reasoning*: `getInitialNotification` typically only needs the single notification that launched the app. A buffer of 20 is sufficient for most use cases while significantly reducing memory pressure.

2.  **Configurable Limit**: Added support for configuring the limit via `AndroidManifest.xml`.
    - Developers can now set `rn_firebase_messaging_max_stored_notifications` in their `AndroidManifest.xml` to adjust this limit without forking the library.
    ```xml
    <meta-data
      android:name="rn_firebase_messaging_max_stored_notifications"
      android:value="50" />
    ```

3.  **Safety Cap**: Added a hard cap of **100** to the configurable limit.
    - Even if a developer sets the value higher in the manifest, it will be capped at 100 to prevent re-introducing the OOM risk.

4.  **Aggressive Cleanup**: Changed the cleanup logic from an `if` statement to a `while` loop.
    - *Reasoning*: If a user already has >20 notifications stored (e.g., 100 from the previous version), the previous logic only removed *one* old notification per new message. The `while` loop ensures the storage is immediately drained down to the limit upon receiving the next message.

## Verification

- Verified locally that `getInitialNotification` still functions correctly with the reduced limit.
- Verified that the limit is correctly read from `AndroidManifest.xml`.
- Verified that the safety cap prevents values > 100.
- Verified that the `while` loop correctly cleans up excess notifications.

## Stack Trace (Reference)

```java
java.lang.OutOfMemoryError: Failed to allocate a 24 byte allocation with 425280 free bytes and 415KB until OOM...
    at android.app.SharedPreferencesImpl$EditorImpl.commitToMemory(SharedPreferencesImpl.java:539)
    at io.invertase.firebase.messaging.ReactNativeFirebaseMessagingStoreImpl.storeFirebaseMessage(ReactNativeFirebaseMessagingStoreImpl.java:43)
```

### DIFF
--- a/packages/app/firebase-schema.json
+++ b/packages/app/firebase-schema.json
@@ -119,11 +119,9 @@
           "description": "On iOS, indicating how to present a notification in a foreground app.",
           "type": "array"
         },
-        "rn_firebase_messaging_max_stored_notifications": {
-          "description": "Controls how many notifications are retained on-device for background storage. Priority order is SharedPreferences -> firebase.json -> AndroidManifest. Values outside the 20-100 range are clamped to prevent excessive deletions or OOM.",
-          "type": "number",
-          "minimum": 20,
-          "maximum": 100
+        "messaging_max_stored_notifications": {
+          "description": "Controls how many notifications are retained on-device for background storage. Configuration priority order is SharedPreferences -> firebase.json -> AndroidManifest. Default 100. Very small values may cause unwanted excessive deletions, large values or large message contents run the risk of OutOfMemoryErrors.",
+          "type": "number"
         },
         "perf_auto_collection_enabled": {
           "description": "Disable or enable auto collection of performance monitoring data collection.\n This is useful for opt-in-first data flows, for example when dealing with GDPR compliance.\nThis can be overridden in JavaScript.",
--- a/packages/app/firebase-schema.json
+++ b/packages/app/firebase-schema.json
@@ -119,6 +119,12 @@
           "description": "On iOS, indicating how to present a notification in a foreground app.",
           "type": "array"
         },
+        "rn_firebase_messaging_max_stored_notifications": {
+          "description": "Controls how many notifications are retained on-device for background storage. Priority order is SharedPreferences -> firebase.json -> AndroidManifest. Values outside the 20-100 range are clamped to prevent excessive deletions or OOM.",
+          "type": "number",
+          "minimum": 20,
+          "maximum": 100
+        },
         "perf_auto_collection_enabled": {
           "description": "Disable or enable auto collection of performance monitoring data collection.\n This is useful for opt-in-first data flows, for example when dealing with GDPR compliance.\nThis can be overridden in JavaScript.",
           "type": "boolean"
--- a/packages/messaging/android/src/main/java/io/invertase/firebase/messaging/ReactNativeFirebaseMessagingStoreImpl.java
+++ b/packages/messaging/android/src/main/java/io/invertase/firebase/messaging/ReactNativeFirebaseMessagingStoreImpl.java
@@ -8,31 +8,65 @@ import static io.invertase.firebase.messaging.ReactNativeFirebaseMessagingSerial
 import com.facebook.react.bridge.ReadableMap;
 import com.facebook.react.bridge.WritableMap;
 import com.google.firebase.messaging.RemoteMessage;
-import io.invertase.firebase.common.UniversalFirebasePreferences;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.List;
+
 import org.json.JSONException;
 import org.json.JSONObject;
 
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+import io.invertase.firebase.common.ReactNativeFirebaseJSON;
+import io.invertase.firebase.common.ReactNativeFirebaseMeta;
+import io.invertase.firebase.common.ReactNativeFirebasePreferences;
+import io.invertase.firebase.common.UniversalFirebasePreferences;
+
 public class ReactNativeFirebaseMessagingStoreImpl implements ReactNativeFirebaseMessagingStore {
 
+  private static final String KEY_MAX_STORED_NOTIFICATIONS = "rn_firebase_messaging_max_stored_notifications";
   private static final String S_KEY_ALL_NOTIFICATION_IDS = "all_notification_ids";
-  private static final int MAX_SIZE_NOTIFICATIONS = 100;
   private final String DELIMITER = ",";
+  private static final int DEFAULT_MAX_SIZE_NOTIFICATIONS = 100;
+  private static final int maxNotificationSize = resolveMaxNotificationSize();
+
+  private static int resolveMaxNotificationSize() {
+    int maxSize = DEFAULT_MAX_SIZE_NOTIFICATIONS;
+    ReactNativeFirebaseJSON json = ReactNativeFirebaseJSON.getSharedInstance();
+    ReactNativeFirebaseMeta meta = ReactNativeFirebaseMeta.getSharedInstance();
+    ReactNativeFirebasePreferences prefs = ReactNativeFirebasePreferences.getSharedInstance();
+
+    try {
+      // Priority: SharedPreferences -> firebase.json -> AndroidManifest
+      // Note: ReactNativeFirebaseMeta doesn't have getIntValue, so we check meta.contains
+      // and read from AndroidManifest directly if needed
+      if (prefs.contains(KEY_MAX_STORED_NOTIFICATIONS)) {
+        maxSize = prefs.getIntValue(KEY_MAX_STORED_NOTIFICATIONS, DEFAULT_MAX_SIZE_NOTIFICATIONS);
+      } else if (json.contains(KEY_MAX_STORED_NOTIFICATIONS)) {
+        maxSize = json.getIntValue(KEY_MAX_STORED_NOTIFICATIONS, DEFAULT_MAX_SIZE_NOTIFICATIONS);
+      } else if (meta.contains(KEY_MAX_STORED_NOTIFICATIONS)) {
+        maxSize = meta.getIntValue(KEY_MAX_STORED_NOTIFICATIONS, DEFAULT_MAX_SIZE_NOTIFICATIONS);
+      }
+
+      // Safety cap: prevent values > 100 to avoid re-introducing OOM risk
+      return Math.min(maxSize, DEFAULT_MAX_SIZE_NOTIFICATIONS);
+    } catch (Exception e) {
+      // Ignore and use default
+      return DEFAULT_MAX_SIZE_NOTIFICATIONS;
+    }
+  }
 
   @Override
   public void storeFirebaseMessage(RemoteMessage remoteMessage) {
     try {
       String remoteMessageString =
-          reactToJSON(remoteMessageToWritableMap(remoteMessage)).toString();
+        reactToJSON(remoteMessageToWritableMap(remoteMessage)).toString();
       //      Log.d("storeFirebaseMessage", remoteMessageString);
       UniversalFirebasePreferences preferences = UniversalFirebasePreferences.getSharedInstance();
 
       // remove old notifications message before store to free space as needed
       String notificationIds = preferences.getStringValue(S_KEY_ALL_NOTIFICATION_IDS, "");
       List<String> allNotificationList = convertToArray(notificationIds);
-      while (allNotificationList.size() > MAX_SIZE_NOTIFICATIONS - 1) {
+      while (allNotificationList.size() > maxNotificationSize - 1) {
         clearFirebaseMessage(allNotificationList.get(0));
         allNotificationList.remove(0);
       }
@@ -61,7 +95,7 @@ public class ReactNativeFirebaseMessagingStoreImpl implements ReactNativeFirebas
   @Override
   public WritableMap getFirebaseMessageMap(String remoteMessageId) {
     String remoteMessageString =
-        UniversalFirebasePreferences.getSharedInstance().getStringValue(remoteMessageId, null);
+      UniversalFirebasePreferences.getSharedInstance().getStringValue(remoteMessageId, null);
     if (remoteMessageString != null) {
       //      Log.d("getFirebaseMessage", remoteMessageString);
       try {
--- a/packages/messaging/android/src/main/java/io/invertase/firebase/messaging/ReactNativeFirebaseMessagingStoreImpl.java
+++ b/packages/messaging/android/src/main/java/io/invertase/firebase/messaging/ReactNativeFirebaseMessagingStoreImpl.java
@@ -27,6 +27,7 @@ public class ReactNativeFirebaseMessagingStoreImpl implements ReactNativeFirebas
   private static final String S_KEY_ALL_NOTIFICATION_IDS = "all_notification_ids";
   private final String DELIMITER = ",";
   private static final int DEFAULT_MAX_SIZE_NOTIFICATIONS = 100;
+  private static final int MIN_SIZE_NOTIFICATIONS = 20;
   private static final int maxNotificationSize = resolveMaxNotificationSize();
 
   private static int resolveMaxNotificationSize() {
@@ -47,8 +48,8 @@ public class ReactNativeFirebaseMessagingStoreImpl implements ReactNativeFirebas
         maxSize = meta.getIntValue(KEY_MAX_STORED_NOTIFICATIONS, DEFAULT_MAX_SIZE_NOTIFICATIONS);
       }
 
-      // Safety cap: prevent values > 100 to avoid re-introducing OOM risk
-      return Math.min(maxSize, DEFAULT_MAX_SIZE_NOTIFICATIONS);
+      // Clamp to allowed range to avoid OOM (upper) or mass deletions (lower)
+      return Math.max(MIN_SIZE_NOTIFICATIONS, Math.min(maxSize, DEFAULT_MAX_SIZE_NOTIFICATIONS));
     } catch (Exception e) {
       // Ignore and use default
       return DEFAULT_MAX_SIZE_NOTIFICATIONS;
--- a/packages/messaging/android/src/main/java/io/invertase/firebase/messaging/ReactNativeFirebaseMessagingStoreImpl.java
+++ b/packages/messaging/android/src/main/java/io/invertase/firebase/messaging/ReactNativeFirebaseMessagingStoreImpl.java
@@ -12,6 +12,8 @@ import com.google.firebase.messaging.RemoteMessage;
 import org.json.JSONException;
 import org.json.JSONObject;
 
+import android.util.Log;
+
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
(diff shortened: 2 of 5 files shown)


packages/messaging/android/src/main/java/io/invertase/firebase/messaging/ReactNativeFirebaseMessagingStoreImpl.java
```diff
@@ -18,8 +18,31 @@
 public class ReactNativeFirebaseMessagingStoreImpl implements ReactNativeFirebaseMessagingStore {
 
   private static final String S_KEY_ALL_NOTIFICATION_IDS = "all_notification_ids";
-  private static final int MAX_SIZE_NOTIFICATIONS = 100;
   private final String DELIMITER = ",";
+  private static int MAX_SIZE_NOTIFICATIONS = 20;
```
mikehardy: I don't believe changing this is a good idea - not for 100% technical reasons - 100 may not be practical for every case - you obviously have an OOM so this is critical to address somehow - but this feature has existed for 5 years (ref: https://github.com/sean5940/react-native-firebase/commit/a7cafc92f4250d2ccb16cb181a4029f14b76b8ca) and you are the first report of a problem.

So while 100 may be impractical sometimes, it's once in 5 years out of maybe tens of thousands of apps and millions and millions of users.

Weigh that versus even the slim possibility of some app relying on this behavior (or some unknown amount less than 100 and greater than 20) and that implies this would be a breaking change.

I don't think a fix here is worth a breaking change given it has existed for 5 years without any other issues.

So I am a "no" on changing this constant

packages/messaging/android/src/main/java/io/invertase/firebase/messaging/ReactNativeFirebaseMessagingStoreImpl.java
```diff
@@ -18,8 +18,31 @@
...
+  private static boolean isInitialized = false;
+
+  private int getMaxNotificationSize() {
+    if (isInitialized) return MAX_SIZE_NOTIFICATIONS;
+    try {
+      android.content.Context context =
```
mikehardy: I don't like the fully qualified package names here vs imports alphabetic at the top and just the names here. Fully qualified package names when they are not critical to disambiguate similarly named objects do not match style

packages/messaging/android/src/main/java/io/invertase/firebase/messaging/ReactNativeFirebaseMessagingStoreImpl.java
```diff
@@ -18,8 +18,31 @@
...
+          io.invertase.firebase.app.ReactNativeFirebaseApp.getApplicationContext();
+      android.content.pm.ApplicationInfo appInfo =
+          context
+              .getPackageManager()
+              .getApplicationInfo(
+                  context.getPackageName(), android.content.pm.PackageManager.GET_META_DATA);
+      if (appInfo.metaData != null) {
+        int configValue =
+            appInfo.metaData.getInt("rn_firebase_messaging_max_stored_notifications", 20);
+        MAX_SIZE_NOTIFICATIONS = Math.min(configValue, 100);
```
mikehardy: If you want to get a configurable item from firebase.json (transmitted through AndroidManifest.xml) you should use the existing + tested utilities we provide for all modules.

Here is an example that controls a configuration item from one key via priority order of shared prefs, then firebase.json then AndroidManifest

https://github.com/invertase/react-native-firebase/blob/b38802203086d30f5bb98e76892e5023e02b010e/packages/app-check/android/src/main/java/io/invertase/firebase/appcheck/ReactNativeFirebaseAppCheckModule.java#L40-L79

There's a potential problem here that loading the SharedPrefs at all - even just to fetch this configuration value - may cause an OOM. 

Worse - and potentially causing an unrecoverable boot crash loop however - our crashlytics implements an Android Provider (which starts before all other parts of app) - and it uses SharedPreferences for configuration so if people have that installed and have this OOM problem there is no way around it besides having user 'Clear Storage' in app settings on device. Even an attempt to purge old messages that cause an OOM 

I regret now that these messages were not put into their own shared preferences file

So: the solution is to modify the usage of SharedPreferences  in our *app* module to do these things:

- catch OutOfMemoryError: not something an app should normally do but this is recoverable, if we catch it
- in the catch block there, call system.gc and try one more time, catching again
- in the final catch block, assume we are in an unrecoverable boot crash loop and our only alternative is to either have the user manually clear storage for the app, or for us to programmatically wipe the shared preferences. Neither is great, but one is automated. So the second OutOfMemoryError catch block where SharedPreferences is accessed should purge the shared preferences file. the clear/apply style should work https://stackoverflow.com/a/3687333/9910298

At that point you will be using shared code and actually improving it so that all paths through it are safe from OutOfMemoryError and users may recover

packages/messaging/android/src/main/java/io/invertase/firebase/messaging/ReactNativeFirebaseMessagingStoreImpl.java
```diff
@@ -35,10 +58,11 @@ public void storeFirebaseMessage(RemoteMessage remoteMessage) {
 
       // check and remove old notifications message
       List<String> allNotificationList = convertToArray(notifications);
-      if (allNotificationList.size() > MAX_SIZE_NOTIFICATIONS) {
+      while (allNotificationList.size() > getMaxNotificationSize()) {
```
mikehardy: Previously this was not a value that could change. However, if the value is made configurable, then people could change it and this would be the only way to trim messages down at once.

So I like this change

mikehardy (changes requested):
I mentioned in more detail in specific comments but a summary:

1- I really like the idea of making the serialized notification count configurable

But, I believe strongly this should use our existing code though, and if you want to make existing boot-crash-loop devices recoverable you'll need to add error handling to that common code

2- I do not believe we should change the default, we should avoid a potentially breaking change

3- I like the trim-until-we-are-the-value vs one-at-a-time change

I think there's the core of something that can go in here if you work through the info I put out for item 1 here in this list

sean5940:
@mikehardy 

```
@Override
  public void storeFirebaseMessage(RemoteMessage remoteMessage) {
    try {
      String remoteMessageString =
          reactToJSON(remoteMessageToWritableMap(remoteMessage)).toString();
      //      Log.d("storeFirebaseMessage", remoteMessageString);
      UniversalFirebasePreferences preferences = UniversalFirebasePreferences.getSharedInstance();
      preferences.setStringValue(remoteMessage.getMessageId(), remoteMessageString);
      // save new notification id
      String notifications = preferences.getStringValue(S_KEY_ALL_NOTIFICATION_IDS, "");
      notifications += remoteMessage.getMessageId() + DELIMITER; // append to last

      // check and remove old notifications message
      List<String> allNotificationList = convertToArray(notifications);
      if (allNotificationList.size() > MAX_SIZE_NOTIFICATIONS) {
        String firstRemoteMessageId = allNotificationList.get(0);
        preferences.remove(firstRemoteMessageId); // ------ here
        notifications = removeRemoteMessage(firstRemoteMessageId, notifications);
      }
      preferences.setStringValue(S_KEY_ALL_NOTIFICATION_IDS, notifications);
    } catch (JSONException e) {
      e.printStackTrace();
    }
  }

```

While preferences.remove(firstRemoteMessageId) is invoked when the notification list exceeds MAX_SIZE_NOTIFICATIONS, the editor never calls apply()/commit(). 
As a result, the entry stored under remoteMessage.getMessageId() is never actually removed from io.invertase.firebase.xml, so the file keeps growing. 
Once the XML becomes large enough, calling preferences.setStringValue(S_KEY_ALL_NOTIFICATION_IDS, notifications) can trigger OOM because the entire SharedPreferences file must be loaded into memory before the update. 
Is this understanding correct?


<video src="https://github.com/user-attachments/assets/2ae98c7a-ea93-4e37-90b8-071460821710" width=720 />

mikehardy:
@sean5940 your analysis appears correct, regardless of the other feedback I gave, your conversion of the "if" to a "while", and adding an `apply()` there is an immediate bug fix. I will post a PR and release it independently of this here so that it not blocked and the fix can be out with no delay

mikehardy:
@sean5940 I can merge and release the PR quickly, but review always helps. If you could take a look?

- #8776

I would still like to pursue the idea of making the storage limit configurable here, and working through the idea of OOM recovery by catching an OOM error. Although I think the 5th commit in my PR - "purge old messages before storing new" may actually recover people sufficiently,.

mikehardy:
This was mostly superceded by #8776 which I wanted to get out as soon as possible because it was in "crash fix" territory

The remaining part of this is the configurable maximum message parameter - what do you think @sean5940 is it still worth it to implement that, or do you want to see if #8776 clears out the problem?

sean5940:
> This was mostly superceded by #8776 which I wanted to get out as soon as possible because it was in "crash fix" territory
> 
> The remaining part of this is the configurable maximum message parameter - what do you think @sean5940 is it still worth it to implement that, or do you want to see if #8776 clears out the problem?

@mikehardy 

One concern I have is that the shared preferences file is already extremely large, which significantly increases the risk of OOM errors. I’m exploring ways to deal with this — do you have any suggestions on how we might mitigate the issue?

For example, if we assume a situation where the XML file manages only 100 IDs while several thousand FCM messages are stored, the cleanup process can operate only on those 100 known IDs. As a result, any data associated with IDs outside that range cannot be deleted.

I’m not seeing a clear way to address this at the library level. In this situation, would it make sense for the app to check the file size in Application.onCreate() and delete it when it becomes abnormally large?

packages/messaging/android/src/main/java/io/invertase/firebase/messaging/ReactNativeFirebaseMessagingStoreImpl.java
sean5940: Updated to maintain a maximum value of 100, based on comments
[41ca8ef](https://github.com/invertase/react-native-firebase/pull/8771/commits/41ca8efb3e20177a93dae834fa808209d7a3603b)

packages/messaging/android/src/main/java/io/invertase/firebase/messaging/ReactNativeFirebaseMessagingStoreImpl.java
sean5940: Thanks for the detailed feedback. I’ve updated the implementation to read the configuration value using the approach you mentioned at the beginning (SharedPreferences → firebase.json → AndroidManifest).

[41ca8ef](https://github.com/invertase/react-native-firebase/pull/8771/commits/41ca8efb3e20177a93dae834fa808209d7a3603b)

mikehardy:
🤔 

> do you have any suggestions on how we might mitigate the issue

Don't send huge data chunks in your messages, have the data in some other location and send unambiguous references to it, for fetch after getting the notification? But that's a design issue and not really your question

> For example, if we assume a situation where the XML file manages only 100 IDs while several thousand FCM messages are stored, the cleanup process can operate only on those 100 known IDs. As a result, any data associated with IDs outside that range cannot be deleted.

Ah, but if you mean that somehow the "accounting" (that is, the assumption in the code that each and every actual message stored has an id stored as well, with no ids for messages that don't exist and no messages without an id entry) is broken then I can see what you mean and we would need some sort of cleanup process.

Indeed, this is the easiest to understand representation of the logic that is currently deployed it is after my refactor for clarity but before any logic changes from the PR I just merged

https://github.com/invertase/react-native-firebase/blob/2ca86dfd589216760f3fff052fd793dc9a4553aa/packages/messaging/android/src/main/java/io/invertase/firebase/messaging/ReactNativeFirebaseMessagingStoreImpl.java#L31-L43


So,

1- store the message as messageId + contents key/value pair
2- then store the messageId list
3- then purge one old message if there are more than the limit

We can assume that 1 may work, and 2 may fail which would lead to an "accounting" failure and an orphaned message yes.

I believe a combination of [`getAll()`](https://developer.android.com/reference/android/content/SharedPreferences#getAll()) on the SharedPreferences, followed by a search for keys that are a) in the known pattern of firebase message ids* and b) are *not* in the list of messageIds would allow for cleaning old messages. Similarly if there is an id but no related message, that id could be removed from the list, though that is less likely since the messages are stored first.

However, the `*` is important - perhaps you are sending messages from some other source?

Judging from your recording above, a regex like (untested...) `^\d\:\d{16}%\d{17}$` *should be* a good match for message ids so you assert that a particular SharedPreferences key is a message (assuming you test that and make sure I got it correct...)

Then you can do a look up in the notificationIds list, and if it is not there, `remove().apply()` on it.
There is a note in the API docs for `getAll()` to be careful with how you treat the Map that is returned, so you will likely want to just use it to get all the keys for storage into an fresh intermediate variable as one step, for direct key-by-key processing as a second pass.

This would be a positive change in my opinion, if you implemented it I would be happy to merge it. If you didn't I would probably implement it myself but with a considerable delay - it could be weeks

packages/app/firebase-schema.json
```diff
@@ -119,6 +119,12 @@
           "description": "On iOS, indicating how to present a notification in a foreground app.",
           "type": "array"
         },
+        "rn_firebase_messaging_max_stored_notifications": {
```
mikehardy: we prefix with the specific name traditionally - it is reasonable to argue that since SharedPreferences is a global namespace we should be more specific but we're using `messaging_` so far for messaging items, and consistency does have value

```suggestion
        "messaging_max_stored_notifications": {
```

packages/app/firebase-schema.json
```diff
@@ -119,6 +119,12 @@
...
+          "description": "Controls how many notifications are retained on-device for background storage. Priority order is SharedPreferences -> firebase.json -> AndroidManifest. Values outside the 20-100 range are clamped to prevent excessive deletions or OOM.",
```
mikehardy: I do not believe in restricting developers capabilities without reason. Someone may have reason to set it to zero for zero data retention for some reason. Someone may have reason to set it to 1000 (small messages!) for their particular use case.

Our job is to pick a sensible default then allow people to do what they want, and just expose the risks.

100 was already selected as a reasonable default for most cases (and as mentioned has caused no problems for 5 years...) so we have done that. The rest is up to developers.

```suggestion
          "description": "Controls how many notifications are retained on-device for background storage. Configuration priority order is SharedPreferences -> firebase.json -> AndroidManifest. Default 100. Very small values may cause unwanted excessive deletions, large values or large message contents run the risk of OutOfMemoryErrors.",
```

packages/app/firebase-schema.json
```diff
@@ -119,6 +119,12 @@
...
+          "type": "number",
+          "minimum": 20,
+          "maximum": 100
```
mikehardy:
```suggestion
          "type": "number"
```

packages/messaging/android/src/main/java/io/invertase/firebase/messaging/ReactNativeFirebaseMessagingStoreImpl.java
```diff
@@ -8,31 +8,66 @@
...
+import io.invertase.firebase.common.UniversalFirebasePreferences;
 
 public class ReactNativeFirebaseMessagingStoreImpl implements ReactNativeFirebaseMessagingStore {
 
+  private static final String KEY_MAX_STORED_NOTIFICATIONS = "rn_firebase_messaging_max_stored_notifications";
   private static final String S_KEY_ALL_NOTIFICATION_IDS = "all_notification_ids";
-  private static final int MAX_SIZE_NOTIFICATIONS = 100;
   private final String DELIMITER = ",";
+  private static final int DEFAULT_MAX_SIZE_NOTIFICATIONS = 100;
+  private static final int MIN_SIZE_NOTIFICATIONS = 20;
```
mikehardy: As mentioned in the edit for the key description - please do not restrict developers via "good intentions" as those are frequently just our idea of what a developer needs but sometimes restrict valid use cases.

Just maintain the 100 original default, take the value in from configuration (if set) and log the final setting and source it came from for diagnostics just in case. No clamping

mikehardy (changes requested):
Thanks for the config change - nice to give people control over this one
As mentioned I don't think a clamp is a positive move but that's an easy change

And if you wanted to implement an all-messageIds-iterator in a separate method like `preenMessageStorage` or similar then call it before storing a new message, I'd be open to that

sean5940:
@mikehardy 
I've addressed all the review feedback in commit [d91c462](https://github.com/invertase/react-native-firebase/pull/8771/commits/d91c462799751134e4efbe7c1dcb42e724164a40).

Key changes:
- Renamed the config key to messaging_max_stored_notifications for consistency with other messaging configs
- Removed all clamping logic (no min/max restrictions) so developers have full control
- Added logging to show the final setting value and where it came from (SharedPreferences/firebase.json/AndroidManifest/default)
- Updated the schema description to warn about risks without restricting choices
- Kept the default value at 100

mikehardy:
Thanks again for the collaboration here - I'll put this in my personal review queue

Mostly updating to say that the first batch of sort-of-emergency OOM fixes have been released, so there should be some relief anyway, while we work through the extra stuff here

mikehardy (commented):
most recent two commits look good on a scan - thanks for the bugfix on Meta integer key handling - approving workflows now as the code looks good and just needs a review of all commits together as a batcnh

just1and0:
Hey @sean5940 👋 — great work on this PR, the configurable limit approach is solid and mikehardy's last review was positive. I'm picking this up since it's been stale for a while and we'd like to get it shipped as part of the v26 push. I'll rebase on current main, fix the lint failure, and address any remaining review feedback in a fresh PR — will credit your work. Thanks for the contribution!

just1and0:
Now resolved in #9111

PR #8774, opened 2025-11-25
fix(ios): update core logic to use sentinel value for preserving `null` 

russellwheatley:
### Description

- **iOS TurboModules Only**: Solution targets iOS TurboModules exclusively where null values in object properties get stripped during JSI serialization (arrays preserve nulls, legacy Native Modules unaffected)
- **Sentinel Value Encoding (JS)**: JavaScript side detects TurboModules via !NativeModules[moduleName] and encodes null object properties as { __rnfbNull: true } sentinel objects before crossing the bridge
- **Automatic Runtime Initialization**: Objective-C +load method called automatically by the OS runtime at app launch (before main()) to set up interception with zero manual initialization
- **Method Swizzling (iOS)**: Dynamically discovers and swizzles all RCTCxxConvert methods matching pattern JS_*Module_Spec*Data: to intercept TurboModule data conversions, decoding sentinels back to NSNull before they reach module implementations
- **Zero Implementation Changes**: Module methods remain completely unchanged - they receive properly decoded data, making this a drop-in workaround that's easy to remove when (**if**) React Native fixes the underlying issue
- **Centralized in App Package**: All encoding/decoding logic lives in @react-native-firebase/app, automatically benefiting all packages (Functions, Firestore, Database, etc.) without per-package modifications

### Checklist

  - [ ] Yes
  - [ ] `Android`
  - [ ] `iOS`
  - [ ] `Other` (macOS, web)
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [ ] No

### Test Plan

---

### DIFF
--- a/packages/app/ios/RNFBApp/RNFBSharedUtils.m
+++ b/packages/app/ios/RNFBApp/RNFBSharedUtils.m
@@ -165,14 +165,6 @@ static NSString *const RNFBErrorDomain = @"RNFBErrorDomain";
 }
 
 /**
- * This works around iOS TurboModule's limitation where null values in object properties get
- * stripped. The JavaScript side converts nulls to { __rnfbNull: true } sentinel objects, this
- * method converts them back to NSNull before passing to react-native-firebase binding code
- *
- * @param value - The value to decode (can be any type)
- * @returns The decoded value with sentinels replaced by NSNull
- */
-/**
  * Decodes null sentinel objects back to NSNull values.
  * Uses iterative stack-based traversal to avoid stack overflow on deeply nested structures.
  *


packages/app/ios/RNFBApp/RNFBNullSentinelInterceptor.h
```diff
@@ -0,0 +1,47 @@
...
+
+#ifndef RNFBNullSentinelInterceptor_h
+#define RNFBNullSentinelInterceptor_h
+
+#import <Foundation/Foundation.h>
+
+/**
+ * Intercepts TurboModule conversions to automatically decode null sentinels.
+ *
+ * iOS TurboModules strip null values from object properties during serialization.
```
mikehardy, line 26: add upstream link for context + future status seeking

```suggestion
 * iOS TurboModules strip null values from object properties during serialization.
 * See: https://github.com/facebook/react-native/issues/52802
```

packages/app/ios/RNFBApp/RNFBSharedUtils.m
```diff
@@ -164,4 +164,40 @@ + (BOOL)getConfigBooleanValue:(NSString *)tag key:(NSString *)key defaultValue:(
   return enabled;
 }
 
+/**
+ * Recursively converts sentinel objects back to NSNull.
+ * This works around iOS TurboModule's limitation where null values in object properties get stripped.
+ * The JavaScript side converts nulls to { __rnfbNull: true } sentinel objects,
+ * and this method converts them back to NSNull before passing to Firebase SDKs.
```
mikehardy: The goal is that all our native binding code between JS<->SDK gets the nulls vs the sentinels
so just re-worded slightly to emphasize that this conversion is before the binding code, before the SDK code... WDYT?

```suggestion
 * The JavaScript side converts nulls to { __rnfbNull: true } sentinel objects,
 * this method converts them back to NSNull before passing to react-native-firebase binding code
```

packages/app/ios/RNFBApp/RNFBSharedUtils.m
```diff
@@ -164,4 +164,40 @@ + (BOOL)getConfigBooleanValue:(NSString *)tag key:(NSString *)key defaultValue:(
...
++ (id)decodeNullSentinels:(id)value {
+  if ([value isKindOfClass:[NSDictionary class]]) {
+    NSDictionary *dict = (NSDictionary *)value;
+    
+    // Check if this is a null sentinel object
+    if ([dict count] == 1 && dict[@"__rnfbNull"] != nil && [dict[@"__rnfbNull"] boolValue]) {
+      return [NSNull null];
+    }
+    
+    // Recursively process dictionary values
```
mikehardy: Very concerned here. Recursion is elegant but we have arbitrarily nested input here so using general recursion to process it implies that we will suffer from a stack overflow crash at some point and it will be a nightmare to debug. We cannot use it.

I looked into how Objective-C toolchain handles the special case of tail-recursion, which we could probably elegantly refactor into, but I was unable to satisfy myself it would be crashproof. In particular this article mentions that with Objective-C memory management even seemingly clean tail-recursion implementations will still consume a stack frame because of language object autorelease requirements on recursive call return, so it doesn't tail-call optimization and eliminate stack use. Still a crash. http://devetc.org/code/2014/05/24/tail-recursion-objc-and-arc.html

There is a nice screenshot in there that shows how you can verify if the toolchain is doing the correct thing now (it's 2025 after all...) with a breakpoint and thread stack inspection....but...

...the only safe thing is the nasty thing I think. This recursive idea is excellent and elegant but I think it needs to go through the "ugly recursive -> iterative" re-render transform to be safe.

I was intellectually curious since I hadn't done work in this area for a while and wanted to offer a concrete suggestion, so:

I think the search term is "preorder iterative tree traversal"

And the implementation is you have an actual stack object where you put "data" initially as setup. Then in a while loop while stack not empty, pop item, if item is object, process item by transforming non-objects and putting pointer to any objects on the stack.

That transforms recursive stack frame use into stack usage in memory

packages/app/lib/internal/registry/nativeModule.js
```diff
@@ -38,9 +39,14 @@ function nativeModuleKey(module) {
...
  * @returns {Function}
  */
-function nativeModuleMethodWrapped(namespace, method, argToPrepend) {
+function nativeModuleMethodWrapped(namespace, method, argToPrepend, nativeModuleName) {
   return (...args) => {
-    const allArgs = [...argToPrepend, ...args];
+    // if the module is not a NativeModule, it is a TurboModule
+    const isTurboModule = !!NativeModules[nativeModuleName];
+    // For iOS TurboModules, encode null values in arguments to work around
+    // the limitation where null values in object properties get stripped during serialization
```
mikehardy, line 47: Always trying to leave a breadcrumb for my forgetful future self

```suggestion
    // the limitation where null values in object properties get stripped during serialization
    // See: https://github.com/facebook/react-native/issues/52802
```

packages/app/ios/RNFBApp/RNFBSharedUtils.m
```diff
@@ -164,4 +164,40 @@ + (BOOL)getConfigBooleanValue:(NSString *)tag key:(NSString *)key defaultValue:(
   return enabled;
 }
 
+/**
+ * Recursively converts sentinel objects back to NSNull.
+ * This works around iOS TurboModule's limitation where null values in object properties get stripped.
+ * The JavaScript side converts nulls to { __rnfbNull: true } sentinel objects,
```
mikehardy: going to make a suggestion to constant-ify this, so remove the specific value here

```suggestion
 * The JavaScript side converts nulls to sentinel objects,
```

packages/app/ios/RNFBApp/RNFBNullSentinelInterceptor.h
```diff
@@ -0,0 +1,47 @@
...
+ * The JavaScript side encodes nulls as { __rnfbNull: true } sentinel objects,
```
mikehardy: removing specific sentinel value 

```suggestion
 * The JavaScript side encodes nulls as sentinel objects,
```

tests/local-tests/storage/https-callable.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+export function HttpsCallableTestComponent(): React.JSX.Element {
+  const handleCallFunction = async (): Promise<void> => {
+    try {
+      const functionRunner = httpsCallable<unknown[], unknown>(
+        getFunctions(getApp()),
+        'testFunctionDefaultRegionV2',
+      );
+      const response = await functionRunner([1, 2, 3, 4, null, { yo: 1, nooo: null }]);
+      console.log('response', response.data);
```
mikehardy: console logging is going away (hello "React Native DevTools"), suggest make test fixture data a const, and display the original on screen plus the return value for visual inspection / comparison

tests/local-tests/index.js
```diff
@@ -36,6 +37,7 @@ const testComponents = {
   'Firestore onSnapshotInSync Test': FirestoreOnSnapshotInSyncTest,
   'VertexAI Generation Example': VertexAITestComponent,
   'Auth MFA Demonstrator': AuthMFADemonstrator,
+  'HttpsCallable Test': HttpsCallableTestComponent,
```
mikehardy, line 40: Honestly, just makes me a little happy to see the local test infra used for a quick otherwise-one-off test that still may have value in future, but mostly you just need some way to run a little thing fast and here it is

packages/app/lib/internal/nullSerialization.js
```diff
@@ -0,0 +1,60 @@
...
+export function encodeNullValues(data) {
+  if (data === null) {
+    return NULL_SENTINEL;
+  }
+
+  if (Array.isArray(data)) {
+    // Arrays preserve null values, but we still need to process nested objects
+    return data.map(item => {
+      if (item !== null && typeof item === 'object') {
+        return encodeNullValues(item);
```
mikehardy: same problem with user-provided data possibly not terminating recursion prior to stack exhaustion - we can't guarantee that the data graph is balanced so it could crash us

same conversion to iterative preorder tree traversal should work ?

package.json
```diff
@@ -9,6 +9,7 @@
     "lerna:clean": "lerna clean",
     "build:all:clean": "lerna run build:clean",
     "build:all:build": "lerna run build",
+    "build:functions:build": "cd .github/workflows/scripts/functions && yarn build",
```
mikehardy: suggest grouping this test-only code in test prefix of run scripts - fits my mental model better?

```suggestion
    "test:functions:build": "cd .github/workflows/scripts/functions && yarn build",
```

package.json
```diff
@@ -46,7 +47,7 @@
     "tests:android:test:jacoco-report": "cd tests/android && ./gradlew jacocoAndroidTestReport",
     "tests:ios:build": "cd tests && yarn detox build --configuration ios.sim.debug",
     "tests:ios:build:release": "cd tests && yarn detox build --configuration ios.sim.release",
-    "tests:ios:manual": "cd tests && SIMCTL_CHILD_GULGeneratedClassDisposeDisabled=1 CC=clang CPLUSPLUS=clang++ LD=clang LDPLUSPLUS=clang++ yarn react-native run-ios",
+    "tests:ios:manual": "cd tests && SIMCTL_CHILD_GULGeneratedClassDisposeDisabled=1 CC=clang CPLUSPLUS=clang++ LD=clang LDPLUSPLUS=clang++ yarn react-native run-ios --simulator='iPhone 15 Pro'",
```
mikehardy: This is not a common device definition and in fact doesn't appear in my default list of devices, so will fail for me.

Is there any reason to specify one here when any could do?
If we must specify can we use same definition as we use for e2e runs (currently iPhone 16 which may also be non-standard but at least we have to have it to run e2e, and that's driven by runner availability of device definitions...)

https://github.com/invertase/react-native-firebase/blob/b38802203086d30f5bb98e76892e5023e02b010e/tests/.detoxrc.js#L47-L53

packages/app/ios/RNFBApp/RNFBSharedUtils.m
```diff
@@ -164,4 +164,40 @@ + (BOOL)getConfigBooleanValue:(NSString *)tag key:(NSString *)key defaultValue:(
...
+ *
+ * @param value - The value to decode (can be any type)
+ * @returns The decoded value with sentinels replaced by NSNull
+ */
++ (id)decodeNullSentinels:(id)value {
+  if ([value isKindOfClass:[NSDictionary class]]) {
+    NSDictionary *dict = (NSDictionary *)value;
+    
+    // Check if this is a null sentinel object
+    if ([dict count] == 1 && dict[@"__rnfbNull"] != nil && [dict[@"__rnfbNull"] boolValue]) {
```
mikehardy: I had some thoughts on sentinel collision-with-real data robustness, and extracting this to a string constant and exposing that to JS via react-native native constants propagation to JS layer but in the end I didn't think they really added value. I don't think any changes here really add value vs being more "academically correct" and even then only maybe.

So no changes proposed for either the value or the constant in both layers. Just leaving this note to myself to say "you already had a deep think on it because it's vital to get right, and you couldn't really improve it. Stop thinking about it."

mikehardy (changes requested):
This is a great solution I think.

A smattering of nitpicks that you can take or leave

But a real concern about stack overflow crashes. I don't mean to handwave on it but I do feel strongly about it. I don't like to feel strongly and request changes in people's otherwise-awesome stuff without offering a concrete suggestion as balm in case they don't feel so strongly - so that making the change doesn't suck as much 😆 . So I attempted to do so with basic transform-to-iterative info.

WDYT?

packages/app/ios/RNFBApp/RNFBSharedUtils.m
mikehardy: Just above where this page anchor drops you is a nice clean one if you're visual on code vs my text description https://nilmamano.com/blog/iterativetreetraversal#preorder-traversal-practice-problems

package.json
russellwheatley: It kept trying to use my actual iphone device and I found it annoying so I added a device. I'll remove

mikehardy:
This remoteConfig onConfigUpdate flakiness is terrible - I've spent enough time in that area that I'm surprised it is failing now but also I should be able to fix it. Either way, it's the cause of the android and ios release failures

the ios debug failure was a different flake - simulator booted, but not happily and app failed to start.

I don't think any of the CI failures were *actual* failures.

packages/app/ios/RNFBApp/RNFBSharedUtils.m
```diff
@@ -172,6 +172,23 @@ + (BOOL)getConfigBooleanValue:(NSString *)tag key:(NSString *)key defaultValue:(
  * @param value - The value to decode (can be any type)
  * @returns The decoded value with sentinels replaced by NSNull
  */
+/**
```
mikehardy, line 167: there is a double comment in this file now - the new comment is great but the old one up above ☝️  still exists

russellwheatley:
@mikehardy - I'm going to merge this into the functions turbomodules PR as I want to refactor naming and update the pattern matching logic for turbomodule methods. Any further updates to serialisation/deserialisation can be made there: https://github.com/invertase/react-native-firebase/pull/8603 👍

PR #8779, opened 2025-11-27
feat(auth): cache password policy per-app in validatePassword

gyula-s:
Cache the fetched password policy per Firebase app to avoid
redundant API calls to the identity toolkit endpoint.

### Description

Every time you call `validatePassword()`, the library was making an API call to Firebase's Identity Toolkit
to fetch the password policy. When a user types their password character by character (e.g.,
"P-a-s-s-w-o-r-d-1-2-3-$"), that's 12 API calls just for one password entry. If users backspace and
retype, it adds up fast.

Firebase has rate limits on this endpoint, so you hit:
`QUOTA_EXCEEDED: Exceeded quota for getting password policy`

### The Fix (Aligned with Firebase JS SDK)

This implementation mirrors the [Firebase JS SDK's approach](https://github.com/nickclaw/firebase-js-sdk/blob/master/packages/auth/src/core/auth/auth_impl.ts#L462) for password policy caching:

**Instance-level caching**
- Password policy is now cached on the Auth instance after first fetch
- Separate caches for project-level (`_projectPasswordPolicy`) and tenant-specific policies (`_tenantPasswordPolicies`)

**Lazy loading**
- Policy is fetched only on the first `validatePassword()` call
- Subsequent calls use the cached policy (zero API calls)

**Multi-tenant support**
- Each tenant gets its own cached policy
- Project-level policy used when `tenantId` is null

**Reactive cache invalidation**
- Cache is refreshed when `PASSWORD_DOES_NOT_MEET_REQUIREMENTS` error occurs
- This handles the edge case where policy changes mid-session

**Schema version validation**
- Rejects unsupported schema versions (matching JS SDK behavior)

### Related issues
https://github.com/invertase/react-native-firebase/issues/8780

### Release Summary
Fixed `validatePassword()` causing `QUOTA_EXCEEDED` errors by caching the password policy
per Auth instance. Implementation matches Firebase JS SDK behavior including tenant support
and reactive cache invalidation.

### Checklist

  - [x] Yes
  - [x] `Android`
  - [x] `iOS`
  - [x] `Other` (macOS, web)
- My change includes tests;
  - [x] `e2e` tests added or updated in `packages/\*\*/e2e` (no e2e behaviour changed)
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

Jest tests added covering:
- Caching behavior (lazy loading, subsequent calls use cache)
- Tenant isolation (separate caches per tenant)
- Project vs tenant cache separation
- Schema version validation
- Cache invalidation via `_recachePasswordPolicy()`
- Input validation (null/undefined password)

### DIFF
--- a/packages/auth/__tests__/validatePassword.test.js
+++ b/packages/auth/__tests__/validatePassword.test.js
@@ -279,8 +279,21 @@ describe('validatePassword (modular API)', () => {
     validatePassword = modular.validatePassword;
 
     mockAuth = {
+      app: { name: '[DEFAULT]', options: { apiKey: 'test-api-key' } },
       validatePassword: jest.fn(),
     };
+  });
+
+  it('should throw error for undefined auth', async () => {
+    await expect(validatePassword(undefined, 'Password123$')).rejects.toThrow(
+      "firebase.auth().validatePassword(*) 'auth' must be a valid Auth instance with an 'app' property",
+    );
+  });
+
+  it('should throw error for auth without app property', async () => {
+    await expect(validatePassword({}, 'Password123$')).rejects.toThrow(
+      "firebase.auth().validatePassword(*) 'auth' must be a valid Auth instance with an 'app' property",
+    );
   });
 
   it('should throw error for null password', async () => {
--- a/packages/auth/lib/modular/index.js
+++ b/packages/auth/lib/modular/index.js
@@ -635,6 +635,12 @@ export function getCustomAuthDomain(auth) {
  * @returns {Promise<PasswordValidationStatus>}
  */
 export async function validatePassword(auth, password) {
+  if (!auth || !auth.app) {
+    throw new Error(
+      "firebase.auth().validatePassword(*) 'auth' must be a valid Auth instance with an 'app' property. Received: undefined",
+    );
+  }
+
   if (password === null || password === undefined) {
     throw new Error(
       "firebase.auth().validatePassword(*) expected 'password' to be a non-null or a defined value.",


packages/auth/__tests__/validatePassword.test.js
```diff
@@ -0,0 +1,118 @@
...
+
+describe('validatePassword', () => {
+  let validatePassword;
+  let mockAuthDefault;
+  let mockAuthSecondary;
+
+  beforeEach(async () => {
+    jest.resetModules();
+
+    mockFetchPasswordPolicy.mockClear();
```
mikehardy: these might fit better in an afterEach?

mikehardy:
🤔  Hmmm - I really appreciate the effort here, the tests are great as well. I'm just not sure we can make caching decisions like this for library consumers. I believe this would fit well at the app level as a React useState var set via useEffect in the component where passwords were set. I am just not sure there is any caching policy we can pick that will be correct in 100% of the consuming app use cases

gyula-s:
> 🤔 Hmmm - I really appreciate the effort here, the tests are great as well. I'm just not sure we can make caching decisions like this for library consumers. I believe this would fit well at the app level as a React useState var set via useEffect in the component where passwords were set. I am just not sure there is any caching policy we can pick that will be correct in 100% of the consuming app use cases

I respectfully disagree - caching is important here, otherwise it will (and currently does) spam the server on every keystroke during password validation on mobile, which breaks the user experience.

The Firebase JS SDK actually implements this same caching pattern, but more robustly:
  - Cache stored at the Auth instance level (per-tenant support)
  - Lazy-loaded on first validatePassword() call
  - Validation runs 100% client-side using the cached policy
  - Cache is only refreshed reactively when PASSWORD_DOES_NOT_MEET_REQUIREMENTS errors occur from actual auth operations

This is a "fetch once, validate locally, refresh on failure" pattern - no TTL, no complexity, just pragmatic caching that ensures the policy is always fresh when it matters.

I'm currently working on an implementation that mimics the JS SDK much more closely, including tenant support and reactive cache invalidation.

gyula-s:
PR is now updated @mikehardy

packages/auth/lib/index.js
```diff
@@ -338,13 +344,37 @@ class FirebaseAuthModule extends FirebaseModule {
...
+          return this._recachePasswordPolicy()
+            .catch(() => {
+              // Silently ignore recache failures - the original error matters more
+            })
+            .then(() => {
+              throw error;
+            });
+        }
+        throw error;
+      });
```
mikehardy: I find the triple repetition of this recaching catch block distasteful but it is exactly how firebase-js-sdk did it and there is a value in that. I bet you had the same thought. Best to leave it, but I'm surprised they did that upstream

packages/auth/__tests__/validatePassword.test.js
```diff
@@ -0,0 +1,269 @@
...
+  },
+  allowedNonAlphanumericCharacters: ['!', '@', '#', '$', '%'],
+  enforcementState: 'ENFORCE',
+};
+
+const mockFetchPasswordPolicy = jest.fn().mockResolvedValue(mockPasswordPolicy);
+
+jest.unstable_mockModule('../lib/password-policy/passwordPolicyApi', () => ({
+  fetchPasswordPolicy: mockFetchPasswordPolicy,
+}));
```
mikehardy: These are in the beforeEach, so why do them here as well? 🤔

mikehardy:
does appear there is a legitimate issue that the tests have probed and either the test isn't quite correct or there may be a small implementation detail? If you check the CI run logs you'll see it...

packages/auth/__tests__/validatePassword.test.js
gyula-s: Removed the redundant mockAuth._tenantId = null assignments. The beforeEach already initializes mockAuth with _tenantId: null, so setting it again was unnecessary.

packages/auth/__tests__/validatePassword.test.js
gyula-s: Added afterEach(() => { jest.clearAllMocks(); }); to properly clean up mocks between tests.

packages/auth/lib/index.js
gyula-s: I agree! When upstream they will change this, maybe we can to! :)

packages/auth/__tests__/validatePassword.test.js
gyula-s: Removed the redundant mockAuth._tenantId = null assignments. The beforeEach already initializes mockAuth with _tenantId: null, so setting it again was unnecessary.

packages/auth/__tests__/validatePassword.test.js
gyula-s: Added afterEach(() => { jest.clearAllMocks(); }); to properly clean up mocks between tests.

gyula-s:
sorry @mikehardy 
these tests are a pain and I can't get them run locally, same with codecov.
have a look at the changes and approve if you're happy... 
let's see what fails this time

mikehardy:
🤔 

```

  1 failing

  1) React Native Firebase
       auth() -> validatePassword()
         validatePassword throws an error if given a bad auth instance:
     AssertionError: expected 'Cannot read property \'validatePassword\' of undefined' to contain 'app'
      at Assertion.prototype.fail (/Users/runner/work/react-native-firebase/react-native-firebase/tests/node_modules/should/cjs/should.js:275:22)
      at Object.defineProperty$argument_2.value (/Users/runner/work/react-native-firebase/react-native-firebase/tests/node_modules/should/cjs/should.js:356:22)
      at it$argument_1 (/Users/runner/work/react-native-firebase/react-native-firebase/packages/auth/e2e/validatePassword.e2e.js:75:33)
      at asyncGeneratorStep (/Users/runner/work/react-native-firebase/react-native-firebase/packages/auth/node_modules/@babel/runtime/helpers/asyncToGenerator.js:3:16)
      at _throw (/Users/runner/work/react-native-firebase/react-native-firebase/packages/auth/node_modules/@babel/runtime/helpers/asyncToGenerator.js:20:26)

```

So, on line 75 of validatePassword.e2e.js, it is failing

When confronted with that I typically zoom in to that specific test, change the `it` to `it.only`, then rerun the test (`yarn tests:ios:test` or android - whichever you prefer) and it will only run that single test which is much faster. If you change RNFBDebug to true in tests/globals.js it won't do retries either so failure is extremely quick.

That allows for rapid iteration on development and test troubleshooting, ideally it allows for a quick discovery + fix on whatever's going on with that test after the changes in this PR

gyula-s:
> 🤔
> 
> ```
> 
>   1 failing
> 
>   1) React Native Firebase
>        auth() -> validatePassword()
>          validatePassword throws an error if given a bad auth instance:
>      AssertionError: expected 'Cannot read property \'validatePassword\' of undefined' to contain 'app'
>       at Assertion.prototype.fail (/Users/runner/work/react-native-firebase/react-native-firebase/tests/node_modules/should/cjs/should.js:275:22)
>       at Object.defineProperty$argument_2.value (/Users/runner/work/react-native-firebase/react-native-firebase/tests/node_modules/should/cjs/should.js:356:22)
>       at it$argument_1 (/Users/runner/work/react-native-firebase/react-native-firebase/packages/auth/e2e/validatePassword.e2e.js:75:33)
>       at asyncGeneratorStep (/Users/runner/work/react-native-firebase/react-native-firebase/packages/auth/node_modules/@babel/runtime/helpers/asyncToGenerator.js:3:16)
>       at _throw (/Users/runner/work/react-native-firebase/react-native-firebase/packages/auth/node_modules/@babel/runtime/helpers/asyncToGenerator.js:20:26)
> ```
> 
> So, on line 75 of validatePassword.e2e.js, it is failing
> 
> When confronted with that I typically zoom in to that specific test, change the `it` to `it.only`, then rerun the test (`yarn tests:ios:test` or android - whichever you prefer) and it will only run that single test which is much faster. If you change RNFBDebug to true in tests/globals.js it won't do retries either so failure is extremely quick.
> 
> That allows for rapid iteration on development and test troubleshooting, ideally it allows for a quick discovery + fix on whatever's going on with that test after the changes in this PR

updated. 
as far as I can tell the tests are passing... but I had the same thought before this one failed. so I don't know what's different on my machine to make if fail. I'm hoping that this will be solved now. 
having said that, codecov might be still an issue... idk.

gyula-s:
> I think we've worked this PR over enough 😅 - CI looks good I just got publishing working again, this will be the first feature that goes out the door hopefully with modern and fixed publishing (as opposed to modern and not fixed, or not modern but fixed - both states the repo was in recently).
> 
> Thanks for your patience as the publishing issues pulled my attention away from anything like responsiveness to the work here. Solid work though and I think anyone with password policies will appreciate it. Cheers

Thank you!
Happy to help!
can't wait for the release so we can incorporate in our app ;)

PR #8788, opened 2025-12-04
feat(crashlytics): add firebase.json support for NDK crash reporting

### LINKED ISSUE #8784: 🐛 [@react-native-firebase/crashlytics] Add support for enabling gwpAsanMode in android config plugin
The crashlytics dashboard is alerting me to enable GWP-ASan:
> Your app has NDK crashes that can lead to security vulnerabilities. Enable GWP-ASan to see additional debugging data in Crashlytics [Learn more](https://firebase.google.com/docs/crashlytics/customize-crash-reports#gwp-asan)

The firebase BoM v31.3.0+ supports this feature, and react-antive-firebase uses a newer version:
https://github.com/invertase/react-native-firebase/blob/c4959a1dcba577a8f6f0f87c16e5d247169cd383/packages/app/package.json#L86

https://github.com/invertase/react-native-firebase/blob/c4959a1dcba577a8f6f0f87c16e5d247169cd383/packages/crashlytics/android/build.gradle#L34

I'm considering writing a config plugin to do just that, and was considering contributing it to react-native-firebase.

My instinct is to put the config in the expo config plugin's parameters, and not the firebase.json file, as the feature I'm suggesting is a config plugin only.

Thoughts?

MichaelVerdon:
### Description

closes https://github.com/invertase/react-native-firebase/issues/8784 

### Checklist

  - [ ] Yes
  - [ ] `Android`
  - [ ] `iOS`
  - [ ] `Other` (macOS, web)
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [ ] No

### Test Plan

---

### DIFF
--- a/docs/crashlytics/android-setup.md
+++ b/docs/crashlytics/android-setup.md
@@ -57,12 +57,30 @@ apply plugin: 'com.google.firebase.crashlytics'
 // ..
 ```
 
-## 4. (Optional) Enable Crashlytics NDK reporting
+## 4. (Optional) Configure Crashlytics NDK and GWP-ASan support
+
+### Enable NDK Crash Reporting
 
 Crashlytics NDK reporting allows you to capture Native Development Kit crashes, e.g. in React Native this will capture
 crashes originating from the Yoga layout engine.
 
-Add the `firebaseCrashlytics` block line to the `android/app/build.gradle` file:
+To enable NDK crash reporting, add the following to your `firebase.json` file:
+
+```json
+{
+  "react-native": {
+    "crashlytics_ndk_enabled": true
+  }
+}
+```
+
+When `crashlytics_ndk_enabled` is set to `true`, React Native Firebase will automatically:
+- Enable NDK crash reporting in the manifest
+- Configure automatic native symbol upload for all release build variants
+
+> **Note**: The automatic symbol upload configuration works with standard release builds and custom build flavors (e.g., `playRelease`, `premiumRelease`). Any build variant with "release" in its name will have symbol upload enabled.
+
+If you need to manually configure symbol upload or override the automatic configuration, you can add the `firebaseCrashlytics` block to your `android/app/build.gradle` file:
 
 ```groovy
 android {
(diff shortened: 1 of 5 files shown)


mikehardy (changes requested):
I believe this will be a little more involved than the current state of the PR. A couple things:

1- the initial issue was about adding configuration support to optionally enable GWP-ASan support, this PR does not address that at all
2- this PR does add configuration support to optionally enable NDK support, but does so incompletely

For item 1 - GWP-ASan support, the configured AndroidManifest item is this one https://developer.android.com/ndk/guides/gwp-asan#opt-in and assuming we can configure it using the manifest placeholder style I suggest having it be a string value in firebase.json with 'default' as the default value (since that's what it is if not specified - our current state) but never and always as possible other values. That would actually close the bug and warrant the `closes #nnnn` tag in the PR description

For item 2 - NDK symbol upload is a real need for developers as well but is distinct from the logged issue. Simply enabling it isn't so bad, it's a similar placeholder but is in addition to the current one - https://firebase.google.com/docs/crashlytics/android/get-started-ndk#add-extension

That is half the job with regard to config, but the second half will take most of the work - you have to configure automatic symbol uploading.

Auto upload is trickier though because it is linked to gradle's internal + dynamic model of the build variants for an app, which are app-specific and may be customized (e.g., `playRelease`, `fdroidRelease` or `premiumRelease`, `liteRelease` if you have build flavors for different app modes vs the more common `release`). So there needs to be some gradle logic that hooks the gradle build *post-config* to detect all variant names with case insensitive `release` in the name, and add the upload as a `finalizes` dependency to the build for those variants

If it is not feasible to get all the release variants of an app (it should be, but may be difficult, and in fact the variants API in gradle just changed so it is particularly annoying...) an attempt to simply hard-code the `release` variant and pull it from config directly in a try/catch with a warning if it failed could be a first attempt

And the task to hook after the build of the release type(s) is then https://firebase.google.com/docs/crashlytics/android/get-started-ndk#set-up-automatic-native-symbols-upload

Issue #8210, opened 2025-01-04
🐛 [@react-native-firebase/functions] Missing `stream` implementation for `httpsCallable` functions


harveyappleton:
Hey amazing invertase team! Happy new year!

In Dec 2024, a [PR was merged](https://github.com/firebase/firebase-js-sdk/pull/8609) into `firebase-js-sdk` that allows you to stream responses from Firebase functions which is awesome! I've tried it out on web and it works wonderfully. Would be awesome if this could be implemented in RN Firebase Functions!!! :D

mikehardy:
Neat, wasn't aware of that work.

I believe we'll need underlying SDK support prior to us being able to do anything at this layer above them

iOS: https://github.com/firebase/firebase-ios-sdk/pull/14290
Android: https://github.com/firebase/firebase-android-sdk/pull/6602

After that we'll have a fair bit of work here to wrap them and present them in a manner compatible with firebase-js-sdk (or we'd happily accept a PR of course...)

The only thing I can think of in the meantime is that you can actually use the firebase-js-sdk alongside react-native-firebase if you like. It is bundled with react-native-firebase as it is used as a fallback for "other" platform support already internally here. You'll have to initialize an app manually after react-native-app boot just like you do on web, and if you use AppCheck you'll have to set the firebase-js-sdk functions headers with it after fetching it from react-native-firebase/app-check, similar with auth - but those headers shouldn't be too hard to set ?

harveyappleton:
I looked into using `firebase-js-sdk` alongside, but I think the issue is that on React Native, the `fetch` implementation doesn't support `ReadStream`. Seems like a common problem. However I found that Expo SDK 52 provides a fetch implementation that is winter cg compliant that supports streaming, so I'm currently updating my app to use that instead and will just manually stream from the firebase function I think :) I can get the Bearer token from firebase auth using `getIdToken()` so should be all good to go 🙏 would be awesome if the RN firebase supported it but I understand its probably a big chunk of work.

mikehardy:
Is a big chunk of work to support it here but most importantly we can't do anything not provided by the underling native android and iOS SDKs, so we are necessarily blocked on them regardless.

Very nifty that it looks like it'll be possible to "weld together" the firebase-js-sdk library support for it while still using react-native-firebase for it's native-only stuff though

harveyappleton:
Still requires attention

mikehardy:
[...]

PR #8799, opened 2025-12-09
feat(functions): httpsCallable.stream support

MichaelVerdon:
### Description

@russellwheatley taken over:

- I had to create a swift implementation for callables as the Swift firebase-ios-sdk is the only way to set limited app check token.
- As mentioned internally, the `HttpsCallableStreamOptions` will only apply to web streaming. Use `HttpsCallableOptions` for iOS and android http rest calls/streaming.
- I created an asynciterator for cancelling streams on web as AbortSignal isn't available on any version of RN lower than `v.81`.
- refactored android and removed [UniversalFirebaseFunctionsModule.java‎](https://github.com/invertase/react-native-firebase/pull/8799/changes#diff-939d7d869846fb89ea02b820bd062db2981f6a355975c4b48f72d32335334441). This required making `TaskExecutorService` constructor public on Firebase App so we can still use it.
- refactored android, ios and TS to remove code duplications.

### Related issues

fixes https://github.com/invertase/react-native-firebase/issues/8210

### Checklist

  - [ ] Yes
  - [ ] `Android`
  - [ ] `iOS`
  - [ ] `Other` (macOS, web)
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [ ] No

### Test Plan

---

### DIFF
--- a/packages/functions/RNFBFunctions.podspec
+++ b/packages/functions/RNFBFunctions.podspec
@@ -27,7 +27,7 @@ Pod::Spec.new do |s|
   s.ios.deployment_target = firebase_ios_target
   s.macos.deployment_target = firebase_macos_target
   s.tvos.deployment_target = firebase_tvos_target
-  s.swift_version = '6.0'
+  s.swift_version = '5.10'
   s.source_files        = 'ios/**/*.{h,m,mm,cpp,swift}'
   s.private_header_files = "ios/**/*.h"
   s.exclude_files       = 'ios/generated/RCTThirdPartyComponentsProvider.*', 'ios/generated/RCTAppDependencyProvider.*', 'ios/generated/RCTModuleProviders.*', 'ios/generated/RCTModulesConformingToProtocolsProvider.*', 'ios/generated/RCTUnstableModulesRequiringMainQueueSetupProvider.*'


packages/functions/__tests__/functions.test.ts
```diff
@@ -24,7 +25,24 @@ type FirebaseApp = ReactNativeFirebase.FirebaseApp;
...
+// but we verify and add removeFunctionsStreaming if needed
+try {
+  const module = getReactNativeModule('NativeRNFBTurboFunctions');
+  if (module && !module.removeFunctionsStreaming) {
+    module.removeFunctionsStreaming = () => {};
+  }
+} catch (_e) {
+  // Module not registered yet - register it ourselves as fallback
+  // This shouldn't happen if namespaced.ts imported correctly
+  setReactNativeModule('NativeRNFBTurboFunctions', {
```
mikehardy: This reads like a workaround - if there is a "should" and we're in a test, then it seems better to transform the expectation into a more concrete "must" and verify it or fail, then fix any underlying issue so the "must" level of API / function registration contract is met

workarounds in tests are a red flag because if we have to workaround in our own tests, what are library consumers going to have to do ?

packages/functions/android/src/main/java/io/invertase/firebase/functions/UniversalFirebaseFunctionsModule.java
```diff
@@ -95,4 +105,352 @@ Task<Object> httpsCallableFromUrl(
...
+                    FirebaseFunctions.getInstance(firebaseApp, region);
+
+                if (host != null) {
+                  functionsInstance.useEmulator(host, port);
+                  android.util.Log.d("RNFBFunctions", "Using emulator: " + host + ":" + port);
+                }
+
+                URL parsedUrl = new URL(url);
+                HttpsCallableReference httpReference =
+                    functionsInstance.getHttpsCallableFromUrl(parsedUrl);
```
mikehardy: is there any required functional difference between this method from an Url and the above method *except* the setup before to get the `HttpsCallableReference`? It doesn't seem like there is much of a difference except augmented logging which the non-Url method could probably benefit from anyway

perhaps an extract-method refactor and then have both the Url version and non-Url version call a single implementation of all the subscription / event emit etc logic ?

packages/functions/ios/RNFBFunctions/RNFBFunctionsModule.mm
```diff
@@ -241,4 +257,282 @@ - (NSString *)getErrorCodeName:(NSError *)error {
...
+        }
+      };
+      [[RNFBRCTEventEmitter shared] sendEventWithName:@"functions_streaming_event" body:eventBody];
+      [self.streamSubscriptions removeObjectForKey:listenerIdNumber];
+    }
+  });
+}
+
+- (void)
+    httpsCallableStreamFromUrl:(NSString *)appName
```
mikehardy, line 204: similar to the java thought on largely-duplicated implementation - it looks like the largest portion of code which is actual code requiring implementation + maintenance on our part (that is, not boiler-plate, or method definitions, or generated code) is right here, and it's in 2 sizable chunks that appear nearly identical. Looks like an opportunity to reduce the future maintenance significantly by eliminating the repetition with a parameterized helper method that does the concrete implementation and is called by the two defined API methods we serve up to the upper layers

packages/functions/lib/web/types.d.ts
```diff
@@ -0,0 +1,86 @@
...
+  export * from 'firebase/app';
+  export * from 'firebase/functions';
+}
+
+/**
+ * Type declaration for RNFBAppModule to include eventsSendEvent method
+ * used for streaming events on web platform.
+ */
+declare module '@react-native-firebase/app/lib/internal/web/RNFBAppModule' {
+  interface RNFBAppModuleType {
```
mikehardy: it seems very strange to have this definition for an app thing here in this functions area 🤔

packages/functions/lib/web/RNFBFunctionsModule.ts
```diff
@@ -5,13 +5,23 @@ import {
...
+import { emitEvent } from '@react-native-firebase/app/lib/internal/web/utils';
 import type { HttpsCallableOptions } from '../index';
 import type { NativeError } from '../HttpsError';
 
 interface WrapperData {
   data?: any;
 }
 
+// Map to store active streaming subscriptions by listenerId
+// Subscription type from RxJS has an unsubscribe() method
```
mikehardy: what is RxJS ?

packages/functions/lib/web/RNFBFunctionsModule.ts
```diff
@@ -143,4 +153,391 @@ export default {
...
+    options: HttpsCallableOptions,
+    listenerId: number,
+  ): void {
+    // Wrap entire function to catch any synchronous errors
+    try {
+      const app = getApp(appName);
+      let functionsInstance;
+      if (regionOrCustomDomain) {
+        functionsInstance = getFunctions(app, regionOrCustomDomain);
+        // Hack to work around custom domain and region not being set on the instance.
```
mikehardy: I see Hack and want to either fix the underlying issue / raise a problem upstream or downgrade the word "hack" to an explanation about the external limitation that cannot be removed and how/why we handle it here. Just to say that the word "hack" without a concrete explanation of the deeper "why" creates a reflexive reaction in me to want to fix it

packages/functions/lib/web/RNFBFunctionsModule.ts
```diff
@@ -143,4 +153,391 @@ export default {
...
+        functionsInstance.region = 'us-central1';
+        functionsInstance.customDomain = null;
+      }
+      if (host) {
+        connectFunctionsEmulator(functionsInstance, host, port);
+        // Hack to work around emulator origin not being set on the instance.
+        functionsInstance.emulatorOrigin = `http://${host}:${port}`;
+      }
+
+      const callable = httpsCallableFromURL(functionsInstance, url, options);
```
mikehardy: same note on duplication through here - did this code come from somewhere and needs attribution ?

packages/functions/lib/namespaced.ts
```diff
@@ -89,6 +90,21 @@ class FirebaseFunctionsModule extends FirebaseModule {
     this._customUrlOrRegion = customUrlOrRegion || 'us-central1';
     this._useFunctionsEmulatorHost = null;
     this._useFunctionsEmulatorPort = -1;
+    this._id_functions_streaming_event = 0;
+
+    // @ts-ignore - emitter and eventNameForApp exist on FirebaseModule
```
mikehardy: `ts-ignore` --> but we just implemented types noooooo 😱  - we should fix the types so they're correct vs immediately corrupting / working around the type system

packages/functions/lib/namespaced.ts
```diff
@@ -122,6 +138,60 @@ class FirebaseFunctionsModule extends FirebaseModule {
...
+      if (streamOptions.timeout) {
+        if (isNumber(streamOptions.timeout)) {
+          streamOptions.timeout = streamOptions.timeout / 1000;
+        } else {
+          throw new Error('HttpsCallableOptions.timeout expected a Number in milliseconds');
+        }
+      }
+
+      const listenerId = this._id_functions_streaming_event++;
+      // @ts-ignore
```
mikehardy: same comment will apply to all ts-ignores - with a sparkly clean + fresh typescript implementation, and the ability to create types for things we need, ts-ignores should feel unacceptable at this point, with strong preference to fix types or create definitions as needed so we can use the type system instead of working around it

packages/functions/lib/namespaced.ts
```diff
@@ -155,6 +225,55 @@ class FirebaseFunctionsModule extends FirebaseModule {
         );
       });
     };
+
+    callableFunction.stream = (
```
mikehardy: looks duplicate when lined up with the above stream implementation ?

tests/local-tests/functions/streaming-callable.tsx
```diff
@@ -0,0 +1,335 @@
...
+
+export function StreamingCallableTestComponent(): React.JSX.Element {
+  const [logs, setLogs] = useState<string[]>([]);
+  const [isStreaming, setIsStreaming] = useState(false);
+  const [stopFunction, setStopFunction] = useState<(() => void) | null>(null);
+
+  const addLog = (message: string) => {
+    const timestamp = new Date().toLocaleTimeString();
+    setLogs(prev => [`[${timestamp}] ${message}`, ...prev].slice(0, 50));
+    console.log(`[StreamingTest] ${message}`);
```
mikehardy: console.log is something that's going to become more difficult to access in the future - it will go away from the metro console and only be visible in a separate / relatively heavy DevTools instance that you have to open independently

since the test is an active screen with a UI, could the logs been a scroll area directly visible on the screen instead? I think that'll be a much better DX going forward - we're only one dependency update (react-native-macos release...) from upgrading to a version of react-native in the test app where this will be our reality

packages/functions/RNFBFunctions.podspec
```diff
@@ -27,10 +27,29 @@ Pod::Spec.new do |s|
   s.ios.deployment_target = firebase_ios_target
   s.macos.deployment_target = firebase_macos_target
   s.tvos.deployment_target = firebase_tvos_target
+  s.swift_version = '5.0'
+  # Exclude Swift from source_files to avoid C++ dependency scanner issues
+  # Swift files must be added via Podfile post_install hook (see README or Podfile example)
+  # This is required because CocoaPods doesn't support automatic hooks from podspecs
```
mikehardy: Is there some explanation of what's going on here - that is, what 'C++ dependency scanner issues' exactly are being avoided, and is there some upstream issue logged?

This looks like an awful (tempted to say unacceptable really) developer experience to require library consumers to do a Podfile modification just to use this functions module after this. It implies more manual integration steps to document and answer issues for as well as some sort of Expo plugin that will need to be created and maintained

Especially if it's just because of using some Swift, when we are certainly not the first module to use Swift vs Objective-C 🤔 

Additionally, there is some infra in react-native itself to run things that may help, for instance there is an ability to add things to build phases so scripts can run

If there truly is some insurmountable problem with the Obj-C++/Swift interop then perhaps the whole module could be ported to Swift - effort-wise that would probably be similar to the effort to correctly document this, answer the integration failure issues over time, and make the expo plugin

Basically, we need to do anything we can to avoid this

.github/workflows/scripts/functions/src/testStreamingCallable.ts
```diff
@@ -0,0 +1,297 @@
...
+ * This function specifically accepts null and returns success: true
+ */
+export const testStreamingCallableWithNull = onCall(
+  async (req: CallableRequest<any>, response?: CallableResponse<any>) => {
+    logger.info('testStreamingCallableWithNull called', { data: req.data });
+
+    // Send a chunk indicating null data was received
+    if (response) {
+      await response.sendChunk({
+        message: 'Null data received',
```
mikehardy, line 219: It's just a test function, so not really important, but if the "dataType" below varies on whether it received real data or not, shouldn't the message?

```suggestion
        message: `${req.data === null || req.data === undefined ? 'Null' : 'Real'} data received`,
```

.github/workflows/scripts/functions/src/testStreamingCallable.ts
```diff
@@ -0,0 +1,297 @@
...
+      );
+    }
+
+    return outputData;
+  },
+);
+
+/**
+ * Streaming callable that throws HttpsError (for testing stream-from-URL).
+ * Same behaviour as testStreamWithHttpsError; separate export for httpsCallableFromUrl.stream() e2e.
```
mikehardy, line 266: does seem to be an identical implementation - could be one method wrapped by the two exports, as a DRY refactor. Same as above suggestion, fairly trivial in the grand scheme, but I see it so I'll mention at least

packages/functions/RNFBFunctions.podspec
```diff
@@ -27,7 +27,9 @@ Pod::Spec.new do |s|
   s.ios.deployment_target = firebase_ios_target
   s.macos.deployment_target = firebase_macos_target
   s.tvos.deployment_target = firebase_tvos_target
-  s.source_files        = 'ios/**/*.{h,m,mm,cpp}'
+  s.swift_version = '5.0'
```
mikehardy: verified Xcode 10.x first supported Swift v5.0, and we already require a higher Xcode  (16.2+ for rnfb 23+ to be precise)

packages/functions/RNFBFunctions.podspec
```diff
@@ -27,7 +27,9 @@ Pod::Spec.new do |s|
...
+  s.source_files        = 'ios/**/*.{h,m,mm,cpp,swift}'
+  s.private_header_files = "ios/**/*.h"
   s.exclude_files       = 'ios/generated/RCTThirdPartyComponentsProvider.*', 'ios/generated/RCTAppDependencyProvider.*', 'ios/generated/RCTModuleProviders.*', 'ios/generated/RCTModulesConformingToProtocolsProvider.*', 'ios/generated/RCTUnstableModulesRequiringMainQueueSetupProvider.*'
   # Turbo modules require these compiler flags
   s.compiler_flags      = '-DFOLLY_NO_CONFIG -DFOLLY_MOBILE=1 -DFOLLY_USE_LIBCPP=1 -DFOLLY_CFG_NO_COROUTINES=1'
```
mikehardy: I believe the compiler_flags is already handled by the `install_module_dependencies()` call (which is...frankly non-negotiable so should be taken out of the `if(defined?(install_modules_dependencies())` conditional and be the only path)

suggest experiment with removing this line and removing the conditional so that `install_modules_dependencies` is always called, and I think TurboModules will be set up correctly for all supported react-native

jest.setup.ts
```diff
@@ -144,6 +144,16 @@ jest.doMock('react-native', () => {
           checkForUpdate: jest.fn(),
           signOutTester: jest.fn(),
         },
+        RNFBFunctionsModule: {
```
mikehardy: does this exist at same time as `NativeRNFBTurboFunctions` ? or does just one of them exist now

packages/functions/ios/RNFBFunctions/RNFBFunctionsStreamHandler.swift
```diff
@@ -0,0 +1,289 @@
...
+ */
+
+import Foundation
+import FirebaseFunctions
+import FirebaseCore
+
+/// Swift wrapper for Firebase Functions streaming that's accessible from Objective-C
+/// This is necessary because Firebase's streaming API uses Swift's AsyncStream which
+/// doesn't have Objective-C bridging
+@available(iOS 15.0, macOS 12.0, *)
```
mikehardy, line 25: this `@available` annotation shouldn't be necessary unless the pbxproj IPHONEOS_DEPLOYMENT_TARGET is wrong

In fact no, it is set to 10 ! That's not correct

4 places, here is an example https://github.com/invertase/react-native-firebase/blob/13d6bb3dfa3328fe987179d6e342d1551a208a38/packages/functions/ios/RNFBFunctions.xcodeproj/project.pbxproj#L162

confirmed iOS 15 is already our minimum supported iOS version - we're issuing the merge result of this as a breaking change but it's related to functions going new architecture only, this iOS version doesn't have to be mentioned.

As a separate issue / peeled from this review (right-click on this comment and create issue - I formatted it well for that), we should raise the deploy target to 15 in the package pbxproj's to match current requirements

packages/functions/lib/namespaced.ts
```diff
@@ -122,7 +307,32 @@ class FirebaseFunctionsModule extends FirebaseModule {
...
         );
       });
+    }) as HttpsCallable<unknown, unknown, unknown>;
+
+    callableFunction.stream = async (
+      data?: unknown,
+      streamOptions?: HttpsCallableStreamOptions,
+    ) => {
+      const platformOptions =
+        isAndroid || isIOS
```
mikehardy: could also be written maybe more concisely as `!isOther` but `isOther` is defined as `!isAndroid && !isIOS` so means literally the same thing. Philosophically I think isOther is technically more accurate as perhaps we add more native platforms that need the non-web type, but isOther is specifically how we say "firebase-js-sdk implementation" so is maybe best?

```suggestion
        !isOther
```

packages/functions/lib/namespaced.ts
```diff
@@ -15,7 +15,7 @@
  *
  */
 
-import { isAndroid, isNumber } from '@react-native-firebase/app/dist/module/common';
+import { isAndroid, isIOS, isNumber } from '@react-native-firebase/app/dist/module/common';
```
mikehardy: paired with comment below - I think isOther is more precise as a firebase-js-sdk backs the implementation signifier

minor / semantically identical as implementation at the moment though

```suggestion
import { isAndroid, isNumber, isOther } from '@react-native-firebase/app/dist/module/common';
```

packages/functions/lib/namespaced.ts
```diff
@@ -155,7 +365,31 @@ class FirebaseFunctionsModule extends FirebaseModule {
...
         );
       });
+    }) as HttpsCallable<unknown, unknown, unknown>;
+
+    callableFunction.stream = async (
+      data?: unknown,
+      streamOptions?: HttpsCallableStreamOptions,
+    ) => {
+      const platformOptions =
+        isAndroid || isIOS
```
mikehardy:
```suggestion
        !isOther
```

packages/functions/ios/RNFBFunctions/RNFBFunctionsModule.mm
```diff
@@ -131,114 +148,179 @@ - (void)httpsCallableFromUrl:(NSString *)appName
...
+    double timeoutValue = timeout.has_value() ? timeout.value() : 0;
+
+    void (^eventCallback)(NSDictionary *) = ^(NSDictionary *event) {
+      NSMutableDictionary *normalisedEvent = @{
+        @"appName" : appName,
+        @"eventName" : @"functions_streaming_event",
+        @"listenerId" : listenerIdNumber,
+        @"body" : event
+      };
+      [[RNFBRCTEventEmitter shared] sendEventWithName:@"functions_streaming_event"
```
mikehardy, line 265: I have noticed event batching behavior on iOS only, and I trace it to this method call (so far).
Specifically if I add an `NSLog` call here before the `RNFBRCTEventEmitter` call, and I alter the `Basic Stream` local-test to wait 5 seconds, I get no events emitted for 20 seconds, then the first 4 come all at once. Then nothing, then the last event and the terminate event (you can see it in the timestamps):

Image

The function itself (running in the emulator) emits logs as expected, every 5 seconds when configured for a 5 second delay.

I haven't determined why yet but it's worth noting as I believe batching these is unwanted, and android doesn't exhibit this behavior.

It could be in the functions emulator batching the chunk sends to iOS for some reason (but...the native/emulator event emitting doesn't do that on android?) or it could be in firebase-ios-sdk, or it could be in our native code / firebase-ios-sdk setup / callbacks somehow.

packages/functions/ios/RNFBFunctions/RNFBFunctionsStreamHandler.swift
```diff
@@ -0,0 +1,289 @@
...
+    
+    do {
+      let encodedParams = AnyEncodable(parameters)
+
+      let stream = try callableStream.stream(encodedParams)
+
+      for try await response in stream {
+        switch response {
+        case .message(let message):
+          eventCallback([
```
mikehardy, line 116: related to investigating why events are delayed/batched on iOS, they are also batched here - this isn't called every time the emulator function actually sends a chunk

packages/functions/RNFBFunctions.podspec
russellwheatley: bumped to to Swift v6 in light of min Xcode 16.2
Screenshot 2026-02-11 at 11 39 25

jest.setup.ts
russellwheatley: Removed both, they're nested in NativeModules object which meant neither were used

packages/functions/RNFBFunctions.podspec
russellwheatley: I've made `install_module_dependencies ()` always run. 

I've tried a number of different ways not setting all headers private (nuclear option) but no matter what, the Swift module build blows up when it encounters C++ stdlib header from codegen.

packages/functions/ios/RNFBFunctions/RNFBFunctionsStreamHandler.swift
russellwheatley: Ok, if I understood you correctly - raise min deployment target in Functions package (and possibly across the board) in a separate PR(s)?

packages/functions/lib/namespaced.ts
russellwheatley: done

.github/workflows/scripts/functions/src/testStreamingCallable.ts
russellwheatley: I'm going to leave suggestions for updating cloud functions for testing

.github/workflows/scripts/functions/src/testStreamingCallable.ts
russellwheatley: same reason as above

packages/functions/ios/RNFBFunctions/RNFBFunctionsStreamHandler.swift
mikehardy: yep exactly that - just peel this to another issue directly via the comment tools and table it since it's good for hygiene but that's low priority and this PR needs to go in

packages/functions/ios/RNFBFunctions/RNFBFunctionsModule.mm
russellwheatley: Can't see anything on our iOS setup that would batch the responses to JS side. We can infer from android working fine, it isn't a problem with JS internal implementation. Sounds like a firebase-ios-sdk issue

mikehardy:
CI being pesky. 

- Simulator boot issue on one of the iOS. Unrelated flake
- compile error on debug attempt, while release still worked? That's new
image

russellwheatley:
@mikehardy - I think the build issue is related to changes made to podspec from PR review. I've reran debug build to confirm. Keeping my eye on it 👍

russellwheatley:
@mikehardy - failed. I think sendable protocol is related to Swift v6 that causes issue perhaps 🤔

mikehardy:
swift to 5 did it, ios debug was a simulator pre-boot flake, re-running

ftaibi:
is stream() available now in main?

mikehardy:
@ftaibi yes on main and I believe if you look at the PR where it went in, you should be able to grab patch-package packages that apply to the current release (and bring it up to main+that PR at that point in time) if you would like to use it pre-release.

PR #8814, opened 2025-12-23
refactor(crashlytics): migrate to TypeScript.

russellwheatley:
### Checklist

  - [ ] Yes
  - [ ] `Android`
  - [ ] `iOS`
  - [ ] `Other` (macOS, web)
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [ ] No

### Test Plan

---


packages/crashlytics/package.json
```diff
@@ -46,5 +50,50 @@
...
+    },
+    "./package.json": "./package.json"
+  },
+  "files": [
+    "lib/modular",
+    "dist",
+    "!**/__tests__",
+    "!**/__fixtures__",
+    "!**/__mocks__"
+  ],
```
mikehardy, line 77: In an open-source context, with a semi-frequent need for "patch-package" I was semi-frequently annoyed when typescript packages didn't ship with all the source needed to recreate the build after doing a yarn inside their module inside node_modules

Is there any reason not to include all the source files? It appears some are missing in the event someone wanted to patch the typescript and transpile it post-install vs patching the generated blobbage

>   "include": ["lib/**/*", "../app/lib/internal/global.d.ts"],

packages/crashlytics/tsconfig.json
```diff
@@ -0,0 +1,40 @@
+{
+  "compilerOptions": {
```
mikehardy, line 3: this appears to be 100% boilerplate, any reason not to have a common definition in app and use extends to centralize management of these configs?

ref: https://www.typescriptlang.org/tsconfig/#extends

packages/crashlytics/package.json
russellwheatley: @mikehardy - I've included source in the files included: https://github.com/invertase/react-native-firebase/pull/8814/changes/2efcaa5273400307e87a8c607c1c33ff9a29bcac

packages/crashlytics/tsconfig.json
russellwheatley: @mikehardy - I've updated to have a base config whilst some boilerplate code regarding relative paths has to remain: https://github.com/invertase/react-native-firebase/pull/8814/changes/66d2d8939224ad090a57f88532a49bd46627fbf8

russellwheatley: Once this PR is in, I'll update app check and messaging tsconfigs to extend the base config.

packages/crashlytics/package.json
russellwheatley: If this is what you had in mind, I'll update the app check and messaging PRs.

mikehardy: Yes! The basic idea is that a user with the installed module in node_modules can hack on the .ts files if they want, run `yarn` there in package and `yarn build` and it will actually result in their .ts changes being present in the .js files and their change running next time app executes

I think of this as the gating factor for contributions that I refer to as "drive-by" contributions, like - they will potentially not become a long-term contributor or something but I've seen lots of bugs fixed via little one-off contributions in other modules when you make it easy to hack on source in situ. Then they PR that little tweak and voila, everyone gets an improvement

this change seems like it implements what I was hoping there

packages/crashlytics/tsconfig.json
mikehardy: per usual - will always defer to implementor's experience, I'm not in there making it work after all - but this looks so much cleaner! Should be easier to maintain hopefully

PR #8861, opened 2026-01-27
fix(analytics): security_storage is no more a boolean

mlecoq:
### Description

In the version 23.8.3 of @react-native-firebase/analytics, `security_storage` has been modified from boolean to 'granted' or 'denied' string

so it raises the issue:
`firebase.analytics().setConsent(*) 'consentSettings' value for parameter 'security_storage' is invalid, expected a boolean.`

### Checklist

  - [x] Yes
  - [x] `Android`
  - [x] `iOS`
  - [ ] `Other` (macOS, web)
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

I have added a jest test case

---

### DIFF
--- a/packages/analytics/__tests__/analytics.test.ts
+++ b/packages/analytics/__tests__/analytics.test.ts
@@ -205,7 +205,7 @@ describe('Analytics', function () {
     it('throws if security_storage of consentSettings is not one of granted or denied', function () {
       // @ts-ignore test
       expect(() => firebase.analytics().setConsent({ security_storage: 'invalid' })).toThrow(
-        "'consentSettings' value for parameter 'security_storage' is invalid, expected one of 'granted', 'denied', or 'unknown'.",
+        "'consentSettings' value for parameter 'security_storage' is invalid, expected one of 'granted' or 'denied'.",
       );
     });
 
--- a/packages/analytics/lib/namespaced.ts
+++ b/packages/analytics/lib/namespaced.ts
@@ -277,7 +277,7 @@ class FirebaseAnalyticsModule extends FirebaseModule {
       if (key === 'security_storage') {
         if (!isOneOf(value, ['granted', 'denied'])) {
           throw new Error(
-            `firebase.analytics().setConsent(*) 'consentSettings' value for parameter '${key}' is invalid, expected one of 'granted', 'denied', or 'unknown'.`,
+            `firebase.analytics().setConsent(*) 'consentSettings' value for parameter '${key}' is invalid, expected one of 'granted' or 'denied'.`,
           );
         }
         continue;


mikehardy:
Interesting @mlecoq - thanks for the PR - do you know when this change happened? Returning different values is a breaking change and this may have been an accident / unintended break in the transition to typescript

Stated differently: the right thing to do may be to revert it to taking a boolean vs the string - I'm just unsure on when it happened / what version last took a boolean

mlecoq:
Hi @mikehardy, I noticed the change in version 23.8.3

mlecoq:
here is the commit: https://github.com/invertase/react-native-firebase/commit/804508b0b454e46f79fd53b96da53551bccd0fb3#diff-aab8943bcad984933668d5965f97d561f15f5fd49fdeb951ceb84633d0a9b333

mikehardy:
Thanks! Was pretty sure it was the typescript migration

https://github.com/invertase/react-native-firebase/commit/804508b0b454e46f79fd53b96da53551bccd0fb3#diff-8d94eedad8b38468862c5182acebd24c02736c21e230e09a47db6059454b83d8R131

So now we're taking a type as the param - `AnalyticsConsent`

that type is https://github.com/invertase/react-native-firebase/commit/804508b0b454e46f79fd53b96da53551bccd0fb3#diff-f6aacd634286d9189996438dd10e92bad8433de110aa08aeb02262f8812f76aeR26-R29

a map of strings now

We're actually out of spec with firebase-js-sdk, but only slight. It *is* supposed to be `granted` or `denied`:

https://firebase.google.com/docs/reference/js/analytics.md#consentstatusstring

but it should be typed as ConsentStatusString, and it should be a param called `ConsentSettings` https://firebase.google.com/docs/reference/js/analytics.consentsettings

and the key ad_personalization shouldn't be a special case boolean, it should also be a `ConsentStatusString`

This whole type needs repair here

@russellwheatley I didn't notice this during review - but was there a reason the types here diverged from firebase-js-sdk 🤔  ?

russellwheatley:
@mikehardy @mlecoq 

`AnalyticsConsent` is web only because we only support one argument at the moment. It is internal, custom and not user facing: https://github.com/invertase/react-native-firebase/blob/main/packages/analytics/lib/web/RNFBAnalyticsModule.ts#L94

`ConsentSettings` is the exact same before TS migration, it looked like this:

```ts
export interface ConsentSettings {
    /** Enables storage, such as cookies, related to advertising */
    ad_storage?: boolean;
    /** Sets consent for sending user data to Google for online advertising purposes */
    ad_user_data?: boolean;
    /** Sets consent for personalized advertising */
    ad_personalization?: boolean;
    /** Enables storage, such as cookies, related to analytics (for example, visit duration) */
    analytics_storage?: boolean;
    /**
     * Enables storage that supports the functionality of the website or app such as language settings
     */
    functionality_storage?: boolean;
    /** Enables storage related to personalization such as video recommendations */
    personalization_storage?: boolean;
    /**
     * Enables storage related to security such as authentication functionality, fraud prevention,
     * and other user protection.
     */
    security_storage?: ConsentStatusString;

    [key: string]: unknown;
  }
```

Except `ConsentStatusString` didn't exist therefore resolved to `any` I believe. With TS migration I gave it the actual type of:

```js
export type ConsentStatusString = 'granted' | 'denied';
```

which is sort of a mistake from the refactor but only because it didn't have a type previously.

I didn't change `ConsentSettings` parameters from booleans to strings to match firebase-js-sdk because that would constitute a breaking change as it is user facing (even if I did with `ConsentStatusString` albeit because there was a bug with our existing types).

I would also say that `security_storage` isn't actually used anywhere (i.e. web, android or iOS). As linked up top, we only pass in `ad_personalization` for web, and `security_storage` is web only.


but this PR is definitely worthwhile, thank you for picking up. `security_storage` type now matches firebase-js-sdk which is the aim, but we cannot change the others unless we make a breaking change on existing type parameters to `ConsentStatusString` on `ConsentSettings`.

mikehardy (approved):
Based on that analysis from Russell which had way more context than I was aware of (thanks!) this looks good to me, thanks @mlecoq

mikehardy:
iOS is failing for unrelated reasons, I'm working on that now 

- #8865

PR #8865, opened 2026-01-29
test(ios): workaround simulator availability issues

mikehardy:
### Description

Our Simulator runtimes were just deprecated and removed: https://github.com/actions/runner-images/issues/13570

Also, there are intermittent failures on iOS runners related to the build failing which seem to be related to simulators not being consistently available. Works most of the time, fails sometimes.

Per discussion in the upstream repo as linked in the comment in this diff, simply listing the simulators as a step before the ones that fail may be a viable workaround, although there are many indications that it does not 100% fix the issue.

Notes:
1- we may need to download the runtime but this may apparently add a lot of CI time https://github.com/actions/runner-images/issues/12758#issuecomment-3206748945
2- it may be that the runtime already exists but is simply not loaded, and reloading it may be done without adding too much CI time - https://github.com/getsentry/sentry-cocoa/pull/6837/changes#diff-0e892ebad2a2b4f630bce02ab66a0a7ead71801904e24dd1cc30b94ce907e598
3- it may be that we still need to create the simulator even then - example https://github.com/siteline/swiftui-introspect/pull/482/changes
4- ~~it may be that switching from a `-destination <simulator name>` in build to a `-target <iphoneos>` and using any simulator would work best though there are reports that there are zero simulators available at times~~ [edit: we already do this]

Of those options,  
- 1 appears to be overkill but is also easy to implement
- 2 appears to be supported by evidence and the most lightweight solution. It's in use by Sentry. But it only helps if the runtime already exists, and in our case it does not
- 3 appears unnecessary if we know a simulator is supposed to exist based on installed software and only intermittent failure.
- 4 is done but doesn't help, 

✅  *So, option 1 it is - I download the runtime now if it is not available. I skip the download if it isn't.*

This adds ~5mins to CI but lets us whatever runtime we want.

This is important because the current runtimes - 26.2 - are *awful*. So slow with performance problems that they flake all the time. 26.0.1 was the last one that performed decently. But in order to use it we have to download the runtime now.

👀 - while I was in there I tweaked coverage as well. It is not useful to have codecov project level stats give us a failing CI run status when we don't gate PRs on that status. So I set it to informational mode. I want to see the coverage, but not have it fail the status checks.
👀  - I also disabled a test that was failing in ios release mode, it will skip now if it is ios + ci. Runs fine locally

There were a couple other tiny workflow tweaks in there that each got their own commit but aren't worth mentioning - timeouts and workflow step labels etc

### Related issues

- https://github.com/actions/runner-images/issues/13459#issuecomment-3681674842

### Release Summary

All test

### Checklist

  - [x] Yes
  - [ ] `Android`
  - [x] `iOS`
  - [ ] `Other` (macOS, web)
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

This was developed while watching the iOS e2e test carefully.

I'll run a 30-iteration flake hammer on it as well to make sure it didn't regress from prior to the runtimes disappearing

---

### DIFF
--- a/tests/ios/Podfile.lock
+++ b/tests/ios/Podfile.lock
@@ -1804,7 +1804,7 @@ PODS:
     - Yoga
   - RNDeviceInfo (15.0.1):
     - React-Core
-  - RNFBAnalytics (23.8.3):
+  - RNFBAnalytics (23.8.5):
     - DoubleConversion
     - FirebaseAnalytics/Core (= 12.8.0)
     - FirebaseAnalytics/IdentitySupport (= 12.8.0)
@@ -1829,7 +1829,7 @@ PODS:
     - ReactCommon/turbomodule/core
     - RNFBApp
     - Yoga
-  - RNFBApp (23.8.3):
+  - RNFBApp (23.8.5):
     - DoubleConversion
     - Firebase/CoreOnly (= 12.8.0)
     - glog
@@ -1851,7 +1851,7 @@ PODS:
     - ReactCommon/turbomodule/bridging
     - ReactCommon/turbomodule/core
     - Yoga
-  - RNFBAppCheck (23.8.3):
+  - RNFBAppCheck (23.8.5):
     - DoubleConversion
     - Firebase/AppCheck (= 12.8.0)
     - glog
@@ -1874,7 +1874,7 @@ PODS:
     - ReactCommon/turbomodule/core
     - RNFBApp
     - Yoga
-  - RNFBAppDistribution (23.8.3):
+  - RNFBAppDistribution (23.8.5):
     - DoubleConversion
     - Firebase/AppDistribution (= 12.8.0)
     - glog
@@ -1897,7 +1897,7 @@ PODS:
     - ReactCommon/turbomodule/core
     - RNFBApp
     - Yoga
-  - RNFBAuth (23.8.3):
+  - RNFBAuth (23.8.5):
     - DoubleConversion
     - Firebase/Auth (= 12.8.0)
     - glog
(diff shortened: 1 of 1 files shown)


mikehardy:
Indeed, that did not work.

> xcodebuild: error: Found no destinations for the scheme 'testing' and action build.

Will try the runtime reloading script next

russellwheatley:
@mikehardy - this worked for my callable stream PR: https://github.com/invertase/react-native-firebase/pull/8799/changes/9fce0befb34de79de4466457581c4496af4c1822

edit- I should add that I have only ran once so maybe it's intermittent like existing.

mikehardy:
@russellwheatley that will help because as I understand it you will now be in sync with the current installed tuples of Xcode/SDK/Simulator-Runtime

But it still may be intermittent on Runtime availability, or on performance of the runtime (the current ones have problems and caused sim boot / timeout flakes - which is why I pinned Xcode to a lower version before).

If we want to stray at all from the recently-shrunk officially supported version windows of the Xcode/SDK/Runtime tuple then we'll have the problem again

I'm looking for something more like what sentry implemented where they can specify arbitrary tuples and it works with no availability flakes

...then hopefully the performance flakes in the newer runtimes are fixed (apparently Apple is working on it, per note from Github runner image maintainers)

It's all just messy.

mikehardy:
I've got the iOS tests passing again, and we need this in quickly to unblock all the other work here, so merging it now
Just re-self-reviewed everything and it looks good to me, will obviously handle any follow-ups as needed / you break it you buy it

Issue #8512, opened 2025-05-03
[🐛] Module '"@react-native-firebase/app-check"' has no exported member 'ReactNativeFirebaseAppCheckProvider'


fracico-tech:
I'm trying to configure app check as documented, but when I import it I have the typescript error:
`Module '"@react-native-firebase/app-check"' has no exported member 'ReactNativeFirebaseAppCheckProvider'. Did you mean to use 'import ReactNativeFirebaseAppCheckProvider from "@react-native-firebase/app-check"' instead?ts(2614)
`

It's only a typescript error, because the feature works as expected

- **`react-native-firebase` version you're using that has this issue:**
  - `21.14.0`
- **`Firebase` module(s) you're using that has the issue:**
  - `AppCheck`
- **Are you using `TypeScript`?**
  - `Y` & `5.0.4`

</p>
</details>

anudit:
@mikehardy @lasfito   Getting this too on expo@53.0.7, RN@0.79.2 and @react-native-firebase/app@^22.1.0, @react-native-firebase/app-check@^22.1.0

```bash

✖ Prebuild failed
Error: [ios.dangerous]: withIosDangerousBaseMod: Cannot add Firebase code to AppDelegate of language "swift"
    at modifyAppDelegateAsync (/Users/anudit/Documents/GitHub/app/node_modules/@react-native-firebase/app-check/plugin/build/ios/appDelegate.js:69:15)
    at /Users/anudit/Documents/GitHub/app/node_modules/@react-native-firebase/app-check/plugin/build/ios/appDelegate.js:77:19
    at action (/Users/anudit/Documents/GitHub/app/node_modules/@expo/config-plugins/build/plugins/withMod.js:199:29)
    at interceptingMod (/Users/anudit/Documents/GitHub/app/node_modules/@expo/config-plugins/build/plugins/withMod.js:104:27)
    at action (/Users/anudit/Documents/GitHub/app/node_modules/@expo/config-plugins/build/plugins/withMod.js:204:14)
    at async interceptingMod (/Users/anudit/Documents/GitHub/app/node_modules/@expo/config-plugins/build/plugins/withMod.js:104:21)
```

anudit:
@mikehardy looks like @jeremyplt made a pr for this https://github.com/invertase/react-native-firebase/pull/8521

fracico-tech:
@mikehardy  I really apology for not adding any example, I thought it was already exhaustive.
I have used the same code in the documentation
```
[...]

PR #8871, opened 2026-02-02
fix(app-check): remove ReactNativeFirebaseAppCheckProvider from type-only export

kimchi-developer:
## Problem

When importing `ReactNativeFirebaseAppCheckProvider` as documented:

```typescript
import { ReactNativeFirebaseAppCheckProvider } from '@react-native-firebase/app-check';

const provider = new ReactNativeFirebaseAppCheckProvider(); // ❌ TS2693
```

TypeScript throws:
> 'ReactNativeFirebaseAppCheckProvider' only refers to a type, but is being used as a value here. (TS2693)

**Note:** This is a different issue from #8512 (TS2614: "has no exported member"), which was fixed by #8529. That PR added the modular exports but kept `ReactNativeFirebaseAppCheckProvider` in the `export type {}` block, causing this new conflict.

## Root Cause

In `packages/app-check/lib/index.ts`:

```typescript
export type {
  // ...
  ReactNativeFirebaseAppCheckProvider,  // ← exported as TYPE
} from './types/appcheck';

export * from './modular';  // ← also exports as VALUE (class)
```

When the same name is exported as both a type (via `export type`) and a value (via `export *`), TypeScript prioritizes the type-only export, making it impossible to use as a constructor.

## Solution

Remove `ReactNativeFirebaseAppCheckProvider` from the `export type {}` block.

The class will still be exported via `export * from './modular'`, and type inference will work correctly.

## Affected Versions

- Tested on `@react-native-firebase/app-check@23.8.5`

### DIFF
--- a/packages/app-check/lib/index.ts
+++ b/packages/app-check/lib/index.ts
@@ -33,7 +33,6 @@ export type {
   ReactNativeFirebaseAppCheckProviderWebOptions,
   ReactNativeFirebaseAppCheckProviderAppleOptions,
   ReactNativeFirebaseAppCheckProviderAndroidOptions,
-  ReactNativeFirebaseAppCheckProvider,
   AppCheck,
   FirebaseAppCheckTypes,
 } from './types/appcheck';


mikehardy:
This will be superceded by #8889 - app-check types and docs have been problematic, no argument there, apologies for that and we should have them shaped up shortly so they match firebase-js-sdk much more closely and behave well with typescript

PR #8910, opened 2026-03-02
ci: static analyser for checking latest firebase-js-sdk types against react native firebase

russellwheatley:
### Description

Cherry picked the comparison script for static analysis so all TS PRs can leverage it.

### Checklist

  - [ ] Yes
  - [ ] `Android`
  - [ ] `iOS`
  - [ ] `Other` (macOS, web)
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [ ] No

### Test Plan

---

### DIFF
--- a/.github/scripts/compare-types/src/registry.ts
+++ b/.github/scripts/compare-types/src/registry.ts
@@ -13,8 +13,6 @@
 import path from 'path';
 import type { PackageConfig } from './types';
 
-import remoteConfigConfig from '../packages/remote-config/config';
-
 const SCRIPT_DIR = path.resolve(__dirname, '..');
 const REPO_ROOT = path.resolve(SCRIPT_DIR, '..', '..', '..');
 
@@ -52,23 +50,5 @@ function rnDist(packageName: string): string {
 }
 
 export const packages: PackageEntry[] = [
-  {
-    name: 'remote-config',
-    firebaseSdkTypesPath: path.join(
-      SCRIPT_DIR,
-      'packages',
-      'remote-config',
-      'firebase-sdk.d.ts',
-    ),
-    rnFirebaseModularFiles: [
-      path.join(rnDist('remote-config'), 'types', 'modular.d.ts'),
-      path.join(rnDist('remote-config'), 'modular.d.ts'),
-    ],
-    rnFirebaseSupportFiles: [
-      path.join(rnDist('remote-config'), 'statics.d.ts'),
-      path.join(rnDist('remote-config'), 'types', 'namespaced.d.ts'),
-      path.join(rnDist('remote-config'), 'types', 'internal.d.ts'),
-    ],
-    config: remoteConfigConfig,
-  },
+
 ];


.github/scripts/compare-types/src/registry.ts
```diff
@@ -0,0 +1,54 @@
+/**
+ * Package registry — defines which packages are compared and where to find
+ * their type files.
+ *
+ * To add a new package:
+ *  1. Create .github/scripts/compare-types/packages/<name>/firebase-sdk.d.ts
+ *     with the public types copied from the firebase-js-sdk release.
```
mikehardy, line 7: > with the public types copied from the firebase-js-sdk release

@russellwheatley am I correct in thinking that, e.g., if I were to try to add analytics as a user of this tool, I would grab this file: https://github.com/firebase/firebase-js-sdk/blob/main/packages/analytics/src/public-types.ts

and so on for any other packages ?

.github/scripts/compare-types/README.md
```diff
@@ -0,0 +1,208 @@
...
+
+Copy the public type declarations for the package from the firebase-js-sdk release into a new file:
+
+```
+packages/<package-name>/firebase-sdk.d.ts
+```
+
+The file should contain only the **modular** (tree-shakeable) public exports — interfaces, type aliases, and `declare function` declarations. Strip internal/private types.
+
+> In the future this step will be automated: a CI job will clone the firebase-js-sdk, run `yarn && yarn build`, and extract the generated `.d.ts` files automatically.
```
mikehardy, line 107: No need to build, and in fact that would be problematic as we could be comparing against unreleased changes and it would break our build before there was even a release

A search through the result of a `yarn` in root to get the types from test app install is sufficient. Question is, which file is best? Here are two styles

```
mike@isabela:~/work/invertase/react-native-firebase (@mikehardy/repro-function-streaming-ios-batch-issue *) % for PACKAGE in analytics app-check app; do find tests/node_modules/@firebase/$PACKAGE | grep \/$PACKAGE.d.ts; done
tests/node_modules/@firebase/analytics/dist/analytics.d.ts
tests/node_modules/@firebase/app-check/dist/app-check.d.ts
tests/node_modules/@firebase/app/dist/app.d.ts
mike@isabela:~/work/invertase/react-native-firebase (@mikehardy/repro-function-streaming-ios-batch-issue *) % for PACKAGE in analytics app-check app; do find tests/node_modules/@firebase/$PACKAGE | grep public-types.d.ts; done
tests/node_modules/@firebase/analytics/dist/esm/src/public-types.d.ts
tests/node_modules/@firebase/analytics/dist/src/public-types.d.ts
tests/node_modules/@firebase/app-check/dist/esm/src/public-types.d.ts
tests/node_modules/@firebase/app-check/dist/src/public-types.d.ts
tests/node_modules/@firebase/app/dist/app/src/public-types.d.ts
tests/node_modules/@firebase/app/dist/esm/app/src/public-types.d.ts
```

Then a little post-processing, or perhaps ts-morph can be configured to ignore internal/private types, or perhaps the comparator can do so, such that no file-processing is needed, it could happen at the static analysis level 🤔

.github/scripts/compare-types/README.md
```diff
@@ -0,0 +1,208 @@
...
+yarn compare:types
+```
+
+---
+
+## Updating a package's firebase-sdk snapshot
+
+When a new firebase-js-sdk version ships with type changes:
+
+1. Copy the updated public types from `node_modules/@firebase/<package>/dist/index.d.ts` (or the equivalent built output) into `packages/<package-name>/firebase-sdk.d.ts`.
```
mikehardy, line 204: I suppose if we need to build our types anyway, and that implies an install of node_modules in the test app, where the types as released from firebase-js-sdk already exist...that means we don't really need to copy them at all do we? We could just reference into the test app's node_modules ?

Issue #8911, opened 2026-03-02
🐛 [@react-native-firebase/analytics] Expose logTransaction method


asnaseer:
## What feature would you like to see?

We have an Expo 53 React Native app written in TypeScript. We use RevenueCat for purchases and are using `StoreKitV2`.

We want to log the trial start event. `StoreKitV2` breaks Firebase's ability to automatically detect and log trial / purchase events. We were told by both RevenueCat and Google that when we know that a trial has been started, we should call `Analytics.logTransaction(transaction)` and this triggers Firebase to gather and log the event. In theory it automatically sucks up information like the product ID, price, trial vs not trial, etc, so we don't need to set any attributes. We have also been told that this special event only works with `logTransaction`. `logEvent` doesn't work the same.

Your `@react-native-firebase/analytics` SDK exposes the `logEvent` method. Was there a reason for not exposing the `logTransaction` method as well? If not, we would really appreciate this being exposed.

asnaseer:
Actually I just discovered that the `Store.Transaction` object is not serialisable and hence cannot be passed between the native and JS layers. However, we could pass in the transaction ID and, I believe, you could use that to look up the `Store.Transaction` object.

russellwheatley:
Hi - Thank you for reaching out. This is definitely something we can implement, but I think it is only available on iOS.

russellwheatley:
For reference, FlutterFire has a similar PR for this API: https://github.com/firebase/flutterfire/pull/17995

asnaseer:
> Hi - Thank you for reaching out. This is definitely something we can implement, but I think it is only available on iOS.

Yes, we are aware that it is only for iOS and that is the platform we would like this for please.

MichaelVerdon:
Hi there @asnaseer just letting you know I have a PR out for it so it should be soon merged 😄 https://github.com/invertase/react-native-firebase/pull/8921

PR #8824, opened 2026-01-08
refactor(storage)!: migrate to TypeScript and match firebase-js-sdk API

russellwheatley:
### Description

- First TS migration to completely separate namespaced types from modular. namespaced types live in `types/namespaced.ts` and are almost identical to previous index.d.ts. This make it a really easy clean up when we switch off namespaced.
- Matches firebase-js-sdk as close as possible now which includes below noted breaking changes:

## Breaking Changes
- **breaking** rename StorageReference to Reference to match firebase-js-sdk. [See commit](https://github.com/invertase/react-native-firebase/pull/8824/changes/f2478aab578175b5fc2cb4d849a4b683c5863b04)
- **breaking** remove methods from Reference and put on internal Reference type for modular [See commit](https://github.com/invertase/react-native-firebase/pull/8824/changes/f826a6d9384c117afe65344f863c0835b435a989)
- **breaking** remove methods from Storage and put on an internal Storage type for modular. [See commit](https://github.com/invertase/react-native-firebase/pull/8824/changes/2b72c0edbbe637824512fd9dc229d85059d173ce)
- **breaking**  Type Reference is now StorageReference. [See commit]
- **breaking** rename StorageReference class to Reference. [See commit](https://github.com/invertase/react-native-firebase/pull/8824/changes/f2478aab578175b5fc2cb4d849a4b683c5863b04)
- **breaking** ListOptions updated to match firebase-js-sdk. [See commit](https://github.com/invertase/react-native-firebase/pull/8824/changes/05775dff22626ec4d7afdc39fa447d30d6a21ee9)
- **breaking** update ListResult to match firebase-js-sdk. [See commit](https://github.com/invertase/react-native-firebase/pull/8824/changes/733e8fbbe0f89574b38ccd4f8ae5439e4a5832b1)
- **breaking** match SettableMetadata, UploadMetadata & FullMetadata. Also using NativeFirebaseError rather than Error. [See commit](https://github.com/invertase/react-native-firebase/pull/8824/changes/a79265084048dc1f75aaed9271ffbfa18433ccac) 
- **breaking** updated TaskState and TaskEvent types to match firebase-js-sdk. [See commit](https://github.com/invertase/react-native-firebase/pull/8824/changes/f409528cb4690b80756257051b44df7ce242166d)
- **breaking** remove `storage.statics.*` types from modular. Need to import individually. [See commit](https://github.com/invertase/react-native-firebase/pull/8824/changes/4adb561320d55522e988e69eba083bbf97a0987c)
[...]

PR #8921, opened 2026-03-06
feat(analytics, iOS): add support for 'logTransaction'

MichaelVerdon:
### Description

https://firebase.google.com/docs/analytics/ios/measure-in-app-purchases#swift

Only supported for Swift IOS

closes https://github.com/invertase/react-native-firebase/issues/8911

### Checklist

  - [ ] Yes
  - [ ] `Android`
  - [ ] `iOS`
  - [ ] `Other` (macOS, web)
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [ ] No

### Test Plan

---

### DIFF
--- a/packages/analytics/lib/modular.ts
+++ b/packages/analytics/lib/modular.ts
@@ -425,9 +425,8 @@ export function logTransaction(analytics: Analytics, transaction_id: string): Pr
   if (Platform.OS !== 'ios') {
     return Promise.reject(new Error('logTransaction is only available on iOS'));
   }
-  const native = (analytics as AnalyticsInternal).native;
-  // logTransaction is optional on RNFBAnalyticsModule (Android/web don't implement it); on iOS it exists
-  return (native as { logTransaction(id: string): Promise<void> }).logTransaction(transaction_id);
+
+  return (analytics as AnalyticsInternal).native.logTransaction(transaction_id);
 }
 
 /**
--- a/packages/analytics/lib/types/internal.ts
+++ b/packages/analytics/lib/types/internal.ts
@@ -35,7 +35,7 @@ export interface RNFBAnalyticsModule {
   getAppInstanceId(): Promise<string | null>;
   getSessionId(): Promise<number | null>;
   /** iOS only (StoreKit 2). Not present on Android native. */
-  logTransaction?(transactionId: string): Promise<void>;
+  logTransaction(transactionId: string): Promise<void>;
   /** iOS only. */
   initiateOnDeviceConversionMeasurementWithEmailAddress?(emailAddress: string): Promise<void>;
   /** iOS only. */


packages/analytics/android/src/reactnative/java/io/invertase/firebase/analytics/ReactNativeFirebaseAnalyticsModule.java
```diff
@@ -192,6 +192,13 @@ public void setConsent(ReadableMap consentSettings, Promise promise) {
             });
   }
 
+  /** Rejects with unimplemented; logTransaction is iOS-only (StoreKit 2). */
+  @ReactMethod
+  public void logTransaction(String transactionId, Promise promise) {
+    rejectPromiseWithCodeAndMessage(
+        promise, "unimplemented", "logTransaction is only available on iOS");
+  }
```
russellwheatley: this shouldn't be here, it shouldn't make it to the native side i.e. throw in the JS side

packages/analytics/lib/web/RNFBAnalyticsModule.ts
```diff
@@ -132,4 +132,9 @@ export const RNFBAnalyticsModule: RNFBAnalyticsModuleType = {
       return null;
     });
   },
+
+  logTransaction(_transactionId: string): Promise<null> {
+    // Unimplemented on web; iOS-only (StoreKit 2).
+    return Promise.reject(new Error('logTransaction is only available on iOS')) as Promise<null>;
+  },
```
russellwheatley: should just throw for iOS and web from logTransaction method immediately before reaching here

packages/analytics/ios/RNFBAnalytics/RNFBAnalyticsLogTransaction.swift
```diff
@@ -0,0 +1,79 @@
...
+      }
+
+      guard let transaction = foundTransaction else {
+        await MainActor.run { reject(Self.kCode, "Transaction not found", nil) }
+        return
+      }
+
+      Analytics.logTransaction(transaction)
+      await MainActor.run { resolve(NSNull()) }
+    } catch {
```
russellwheatley: Can remove the do/catch block as it is unreachable according to the warning below

packages/analytics/lib/modular.ts
```diff
@@ -416,6 +416,16 @@ export function logEvent(
   return analytics.logEvent.call(analytics, name, params, options, MODULAR_DEPRECATION_ARG);
 }
 
+/** Logs verified in-app purchase events in Google Analytics for Firebase
+ * after a purchase is successful.
+ * Modular API only; iOS only (StoreKit 2). On Android, the native module rejects with unimplemented.
+ */
+export function logTransaction(analytics: Analytics, transaction_id: string): Promise<void> {
+  return (
+    analytics as unknown as { native: { logTransaction(id: string): Promise<void> } }
```
russellwheatley: could you do what you had to do for other TS migrations so native object is "known"? add this: https://github.com/invertase/react-native-firebase/pull/8824/changes#diff-c22886beb4b9b413fea6ca9ebfbbf0a50736bdb540e76139dea26b5e3e032a5aR51

i.e.:
```
extends FirebaseModule<typeof nativeModuleName>
```
and update the native method types for analytics in internal.ts like in storage here: https://github.com/invertase/react-native-firebase/pull/8824/changes#diff-e91b09f6262f2813089ef252c4f1f327e6e84bffb14bfd3b8b0d5598066ca487R122-R188

packages/analytics/ios/RNFBAnalytics/RNFBAnalyticsModule.m
```diff
@@ -18,6 +18,7 @@
 #import <Firebase/Firebase.h>
 #import <React/RCTUtils.h>
 
+#import <RNFBAnalytics/RNFBAnalytics-Swift.h>
```
mikehardy, line 23: Okay this is outrageously specific and subtle but we have to change this for an interop reason:

react-native 0.84+ and Expo 55+ use this new thing called "pre-built rncore" - a nice compile-time optimization but it is not built for `use_frameworks!`

So our modules are *not* frameworks in that case and angle-bracket / framework imports do not work in that case.

I just stumbled into this and I'm fixing it for the existing functions import here:

- #8925 

Here's the fix for the issue that hasn't even shown up yet but would affect almost everyone 😅 


```suggestion
#if __has_include(<RNFBAnalytics/RNFBAnalytics-Swift.h>)
// This import will work in situations where `use_frameworks!` is in use
#import <RNFBAnalytics/RNFBAnalytics-Swift.h>
#elif __has_include("RNFBFunctions-Swift.h")
// If `use_frameworks!` is not in use (for example, while using pre-built
// react-native core) then header imports based on frameworks assumptions fail.
// So, if frameworks are not available, fall back to importing the header directly, it
// should be findable from a header search path pointing to the build
// directory. See firebase-ios-sdk#12611 for more context.
#import "RNFBAnalytics-Swift.h"
#endif
```

packages/analytics/lib/modular.ts
```diff
@@ -416,6 +417,21 @@ export function logEvent(
...
+ */
+export function logTransaction(analytics: Analytics, transaction_id: string): Promise<void> {
+  if (Platform.OS !== 'ios') {
+    return Promise.reject(new Error('logTransaction is only available on iOS'));
+  }
+  const native = (analytics as AnalyticsInternal).native;
+  if (!native.logTransaction) {
+    return Promise.reject(new Error('logTransaction is only available on iOS'));
+  }
+  return native.logTransaction(transaction_id);
```
mikehardy: 🤔  I don't understand how `native.logTransaction` could be null now that we know `Platform.OS === 'ios'` - is this just to satisfy the type checker that doesn't understand the `.native` object has these methods ? Genuinely curious, I'm likely missing something

MichaelVerdon: Hi there, had a think I dont think there is a need so this redundent check is removed as the guard would already rule everything else out

mikehardy:
Just waking up and re checking- I had a copy paste issue in my suggestion RNFBFunctions string still in it in one spot not RNFBAnalytics sorry

packages/analytics/lib/modular.ts
```diff
@@ -416,6 +417,19 @@ export function logEvent(
...
+ * after a purchase is successful.
+ * Modular API only; iOS only (StoreKit 2). Throws on Android and web before reaching native.
+ */
+export function logTransaction(analytics: Analytics, transaction_id: string): Promise<void> {
+  if (Platform.OS !== 'ios') {
+    return Promise.reject(new Error('logTransaction is only available on iOS'));
+  }
+  const native = (analytics as AnalyticsInternal).native;
+  // logTransaction is optional on RNFBAnalyticsModule (Android/web don't implement it); on iOS it exists
+  return (native as { logTransaction(id: string): Promise<void> }).logTransaction(transaction_id);
```
russellwheatley: you should be able to just do something like this:

```js
(analytics as AnalyticsInternal).logTransaction()
```

You definitely should not have to do:

```js
(native as { logTransaction(id: string): Promise<void> })
```

as this defeats the purpose of: https://github.com/invertase/react-native-firebase/pull/8921/changes#diff-c85491d0a0fe609029cef293c942215971c7a32d7534aa448506df4f1701211dR38

Issue #8883, opened 2026-02-12
[🐛] compile errors with react-native CLI + pre-compiled iOS distribution default on RN 0.84 - workaround - disable pre-compiled react-native or use Podfile preinstall hook


birdofpreyru:
**Maintainer edit:**

 workaround is to either 

1) disable prebuild rncore with environment variable `RCT_USE_PREBUILT_RNCORE=0` in your Podfile (or the equivalent) or that set in your environment as you run `pod install` to disable the pre-compiled react-native build]  -- see https://github.com/mikehardy/rnfbdemo/blob/main/make-demo.sh for an example, specifically https://github.com/mikehardy/rnfbdemo/blob/270b7ea0df6a6899fcff1e0ca9b7f4389078fbe5/make-demo.sh#L310-L312 for one way to disable prebuilt rncore

or 

2) use Podfile preinstall hook to force static link our Pods --> https://github.com/invertase/react-native-firebase/issues/8883#issuecomment-3991498731

--------

## Issue

I am upgrading an existing, barebone-RN app using RN Firebase from RN@0.83 to RN@0.84. All looks fine for Android, however for iOS it fails to build with the following errors:

<img width="2648" height="1018" alt="Image" src="https://github.com/user-attachments/assets/fdd9aa1f-209b-4ab6-9156-89e3eccf2aa0" />

I'm not sure out of my head what is the purpose of `<React/RCTConvert.h>`, however in other places it similarly complains about imports of `<React/RCTBridgeModule.h>` and `<React/RCTEventEmitter.h>`, which I don't have to import inside my own libraries to support communication between JS and native layers in the latest RN versions... so guess that is just some legacy code which relies on legacy stuff removed (or hidden) now from RN?

---

### Javascript

<details><summary>Click To Expand</summary>
<p>

#### `package.json`:

```json
# N/A
```

#### `firebase.json` for react-native-firebase v6:

```json
# N/A
```

</details>

### iOS

<details><summary>Click To Expand</summary>
<p>

#### `ios/Podfile`:

- [ ] I'm not using Pods
- [x] I'm using Pods and my Podfile looks like:

```ruby
# N/A
```

#### `AppDelegate.m`:

```objc
// N/A
```

</p>
</details>

---

### Android

<details><summary>Click To Expand</summary>
<p>

#### Have you converted to AndroidX?

- [ ] my application is an AndroidX application?
- [ ] I am using `android/gradle.settings` `jetifier=true` for Android compatibility?
[...]

PR #8925, opened 2026-03-09
fix(ios): fix objective-c / swift interop in the non-`use_frameworks!` case for pre-built rncore

mikehardy:
### Description

this should resolve an issue seen with my expo demo script where static linking wasn't working correctly via per-module directives and a functions import was incorrect while using [the new prebuilt react-native core](https://reactnative.dev/blog/2026/02/11/react-native-0.84#precompiled-binaries-on-ios-by-default)

I ask Xcode to generate a module map for the App package now, and I have a fallback Swift interop header import style for the Functions Swift header, inspired by:

- https://github.com/firebase/firebase-ios-sdk/pull/12659
- specifically https://github.com/firebase/firebase-ios-sdk/blob/f7b3d6a615f3c02b907c1d3f3ca107756799338b/Crashlytics/Crashlytics/FIRCrashlytics.m#L63-L72

### Related issues

Nothing logged - this was seen with my Expo demo script which has these currently unique features, from the perspective of how we test:

- it uses the new prebuilt react-native core, which is non-static linkage (default in react-native 0.84+ / Expo 55+)
- it configures the iOS build to static link only specific modules one by one via expo-build-properties

When you do this, and you do not have this fix, you see:

> [!] The following Swift pods cannot yet be integrated as static libraries:
> 
> The Swift pod `RNFBFunctions` depends upon `RNFBApp`, which does not define modules. To opt into those targets generating module maps (which is necessary to import them from Swift when building as static libraries), you may set `use_modular_headers!` globally in your Podfile, or specify `:modular_headers => true` for particular dependencies.

With this, it builds farther but still fails later with a functions import that isn't found:

> #import <RNFBFunctions/RNFBFunctions-Swift.h>

Note that this is *not* needed on current release v23.8.6, I only see this failure on main, so it isn't breaking for people...yet

Additionally, if you opt out of the prebuilt react-native core you won't see it, but everyone wants to use that and the issues in this repo about it are *very* active.

### Checklist

  - [x] Yes
  - [ ] `Android`
  - [x] `iOS`
  - [ ] `Other` (macOS, web)
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

- generate packages from main via `.github/workflows/scripts/create_npm_packages.sh packages ~/packages`
- run https://github.com/mikehardy/rnfbdemo/blob/main/make-expo-demo.sh

You'll see the compile error I mention until you add this module map compile flag, then you'll see the swift bridging header import issue until you add the fallback direct import style

---

### DIFF
--- a/packages/functions/ios/RNFBFunctions/RNFBFunctionsModule.mm
+++ b/packages/functions/ios/RNFBFunctions/RNFBFunctionsModule.mm
@@ -18,12 +18,23 @@
 #import <Firebase/Firebase.h>
 #import <React/RCTUtils.h>
 
-#import <RNFBFunctions/RNFBFunctions-Swift.h>
 #import "NativeRNFBTurboFunctions.h"
 #import "RNFBApp/RCTConvert+FIRApp.h"
 #import "RNFBApp/RNFBRCTEventEmitter.h"
 #import "RNFBApp/RNFBSharedUtils.h"
 #import "RNFBFunctionsModule.h"
+
+#if __has_include(<RNFBFunctions/RNFBFunctions-Swift.h>)
+// This import will work in situations where `use_frameworks!` is in use
+#import <RNFBFunctions/RNFBFunctions-Swift.h>
+#elif __has_include("RNFBFunctions-Swift.h")
+// If `use_frameworks!` is not in use (for example, while using pre-built
+// react-native core) then header imports based on frameworks assumptions fail.
+// So, if frameworks are not available, fall back to importing the header directly, it
+// should be findable from a header search path pointing to the build
+// directory. See firebase-ios-sdk#12611 for more context.
+#import "RNFBFunctions-Swift.h"
+#endif
 
 static __strong NSMutableDictionary *streamListeners;
 


mikehardy:
I believe the swift bridging header file issue will be fixed with a solution similar to this:

https://github.com/firebase/firebase-ios-sdk/blob/f7b3d6a615f3c02b907c1d3f3ca107756799338b/Crashlytics/Crashlytics/FIRCrashlytics.m#L63-L72

Specifically - I think the problem is that importing with angle brackets:

> #import <RNFBFunctions/RNFBFunctions-Swift.h>

...requires frameworks, but react-native builds using the pre-built (*not `use_frameworks!`*) react-native core - which is the react-native 0.83+ / Expo 55+ default case won't have RNFBFunctions as a framework so it can't find it's own header file inside a module.

However, the header search path should...I hope...have it so doing like:

> #import "RNFBFunctions-Swift.h"

Should work?

This will likely also affect #8921 since that PR implements a Swift file for the first time, implying the need to import the bridging header file to Objective-C inside it's own package also

mikehardy:
@ncooke3 / @paulb777 credit it where it is due, I used an import trick from a firebase-ios-sdk PR to fix objective-C/swift interop in the non-use_frameworks case, thank you - I'm abruptly tagging you and mentioning it though to raise awareness: react-native ships a pre-built core binary now that is not setup for `use_frameworks!`, so all the compile tricks people have been attempting over the years are all the sudden being put to a real test. For react-native-firebase I expect the majority of firebase-ios-sdk consumers/linkages now will be non-use_frameworks and will force static linkage for the firebase pods only, via little tricks like 

https://github.com/invertase/react-native-firebase/issues/8883#issuecomment-3991498731

I don't know what this means for firebase-ios-sdk but just be aware that it is effectively impossible for react-native-firebase to continue swimming upstream in the "just `use_frameworks!` and be done..." battle. it's all workarounds for that requirement from here on out...

mikehardy:
release management note: this is not needed for release-v23 branch, the Functions stream implementation was after the breaking changes planned for v24, and it is what created the need for a Swift/Objective-C interop header

PR #8929, opened 2026-03-12
feat(firestore): add support for local source snapshot listeners

### LINKED ISSUE #8126: Implement ListenSource / SnapshotListenOptions for firestore (new SDK feature March 2024)
### Discussed in https://github.com/invertase/react-native-firebase/discussions/8124

<div type='discussions-op-text'>

<sup>Originally posted by **cSieves** November 10, 2024</sup>
Hi, 
I was wondering if there are any plans to allow a user to specify where to get the listener data from. At the moment as far as I understand when setting up a listener data data will be pulled from both the cache and server as changes occur. However, from what I can tell, (correct me if I am wrong) doing it this way creates a number of issues; 

- It drives up costs as every change to a query/doc results in you being charged for a read. This is bad for a few reasons, but if you manage and situation like a social media app where lots of users are updating a single doc e.g. liking a post, the cost grows exponentially. 
- Frequent updates can cause re-renders in react if you are storing the new state in a react hook. This can mess with the user experience 

I noticed in the other [SDK](https://firebase.google.com/docs/firestore/query-data/listen#:~:text=Client%20SDKs%20allow%20you%20to,reflects%20local%20data%20and%20mutations.)'s there is an option to specify where the data is coming from, is this feature going to become available in the react-native sdk soon too?

In the meantime is there some other way I can achieve the same thing i.e. get realtime updates for changes in my local cache? </div>

astanb:
### Related issues

- Fixes #8126

### Release Summary

Adds the ability to specify the source ('cache'/'default') for snapshot listeners.

### Checklist

  - [x] Yes
  - [x] `Android`
  - [x] `iOS`
  - [x] `Other` (macOS, web)
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

---

New tests added for functionality.
All tests for both platforms passing.

Screenshot 2026-03-12 at 17 45 47

🔥

---

> [!NOTE]
> **Medium Risk**
> Changes Firestore listener behavior on both Android and iOS by wiring a new `source` option into native `addSnapshotListener` APIs, which could affect realtime update delivery and caching semantics. Coverage is improved with new unit and e2e tests, but it still touches core data subscription paths.
> 
> **Overview**
> Adds support for specifying snapshot listener `source` (`'default'` or `'cache'`) for `DocumentReference.onSnapshot` and `Query.onSnapshot` across the JS, Android, and iOS layers.
> 
> Updates JS/TS types and argument parsing (`SnapshotListenOptions`, new `ListenSource`) to validate and default the option, and updates native modules to build `SnapshotListenOptions`/`FIRSnapshotListenOptions` with the requested source. Expands Jest/type tests and adds emulator e2e coverage (including out-of-band server writes) to verify cache-only vs default listener behavior, and tightens the type-compare tooling to fail on *stale* registry/config entries (with clearer reporting).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7541dec0fa07ce2b3f0f513a1a62e466a20d547d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/firestore/e2e/DocumentReference/onSnapshot.e2e.js
+++ b/packages/firestore/e2e/DocumentReference/onSnapshot.e2e.js
@@ -16,7 +16,7 @@
  */
 const COLLECTION = 'firestore';
 const NO_RULE_COLLECTION = 'no_rules';
-const { wipe } = require('../helpers');
+const { wipe, setDocumentOutOfBand } = require('../helpers');
 
 describe('firestore().doc().onSnapshot()', function () {
   before(function () {
@@ -302,6 +302,108 @@ describe('firestore().doc().onSnapshot()', function () {
           "'options' SnapshotOptions.includeMetadataChanges must be a boolean value",
         );
         return Promise.resolve();
+      }
+    });
+
+    it("throws if SnapshotListenerOptions.source is invalid ('server')", function () {
+      try {
+        firebase.firestore().doc(`${NO_RULE_COLLECTION}/nope`).onSnapshot({
+          source: 'server',
+        });
+        return Promise.reject(new Error('Did not throw an Error.'));
+      } catch (error) {
+        error.message.should.containEql(
+          "'options' SnapshotOptions.source must be one of 'default' or 'cache'",
+        );
+        return Promise.resolve();
+      }
+    });
+
+    it('accepts source-only SnapshotListenerOptions', async function () {
+      if (Platform.other) {
+        return;
+      }
+      const callback = sinon.spy();
+      const unsub = firebase.firestore().doc(`${COLLECTION}/source-only`).onSnapshot(
+        {
+          source: 'cache',
+        },
+        callback,
+      );
+
+      await Utils.spyToBeCalledOnceAsync(callback);
+      unsub();
+    });
+
+    it('accepts source + includeMetadataChanges SnapshotListenerOptions', async function () {
+      if (Platform.other) {
+        return;
+      }
+      const callback = sinon.spy();
+      const unsub = firebase.firestore().doc(`${COLLECTION}/source-with-metadata`).onSnapshot(
+        {
+          source: 'default',
+          includeMetadataChanges: true,
+        },
+        callback,
+      );
+
+      await Utils.spyToBeCalledOnceAsync(callback);
+      unsub();
+    });
+
+    it('cache source listeners ignore out-of-band server writes', async function () {
+      if (Platform.other) {
+        return;
+      }
+
+      const docPath = `${COLLECTION}/${Utils.randString(12, '#aA')}`;
+      const docRef = firebase.firestore().doc(docPath);
+      await docRef.set({ value: 1 });
+      await docRef.get();
+
+      const callback = sinon.spy();
+      const unsub = docRef.onSnapshot({ source: 'cache' }, callback);
+      try {
+        await Utils.spyToBeCalledOnceAsync(callback);
+
+        await setDocumentOutOfBand(docPath, { value: 2 });
+        await Utils.sleep(1500);
+        callback.should.be.callCount(1);
+
+        await docRef.set({ value: 3 });
+        await Utils.spyToBeCalledTimesAsync(callback, 2);
+        callback.args[1][0].get('value').should.equal(3);
+      } finally {
+        unsub();
+      }
+    });
+
+    it('default source listeners receive out-of-band server writes', async function () {
+      if (Platform.other) {
+        return;
+      }
+
+      const docPath = `${COLLECTION}/${Utils.randString(12, '#aA')}`;
+      const docRef = firebase.firestore().doc(docPath);
+      await docRef.set({ value: 1 });
+      await docRef.get();
+
+      const callback = sinon.spy();
+      const unsub = docRef.onSnapshot(
+        { source: 'default', includeMetadataChanges: true },
+        callback,
+      );
+      try {
+        await Utils.spyToBeCalledOnceAsync(callback);
+
+        await setDocumentOutOfBand(docPath, { value: 2 });
+        await Utils.spyToBeCalledTimesAsync(callback, 2, 8000);
+
+        const latestSnapshot = callback.args[callback.callCount - 1][0];
+        latestSnapshot.get('value').should.equal(2);
+      } finally {
+        unsub();
       }
     });
 
@@ -614,6 +716,58 @@ describe('firestore().doc().onSnapshot()', function () {
         );
         return Promise.resolve();
       }
+    });
+
+    it("throws if SnapshotListenerOptions.source is invalid ('server')", function () {
+      const { getFirestore, doc, onSnapshot } = firestoreModular;
+      try {
+        onSnapshot(doc(getFirestore(), `${NO_RULE_COLLECTION}/nope`), {
+          source: 'server',
+        });
+        return Promise.reject(new Error('Did not throw an Error.'));
+      } catch (error) {
+        error.message.should.containEql(
+          "'options' SnapshotOptions.source must be one of 'default' or 'cache'",
+        );
+        return Promise.resolve();
+      }
+    });
+
+    it('accepts source-only SnapshotListenerOptions', async function () {
+      if (Platform.other) {
+        return;
+      }
+      const { getFirestore, doc, onSnapshot } = firestoreModular;
+      const callback = sinon.spy();
+      const unsub = onSnapshot(
+        doc(getFirestore(), `${COLLECTION}/mod-source-only`),
+        {
+          source: 'cache',
+        },
+        callback,
+      );
+
+      await Utils.spyToBeCalledOnceAsync(callback);
+      unsub();
+    });
+
+    it('accepts source + includeMetadataChanges SnapshotListenerOptions', async function () {
+      if (Platform.other) {
+        return;
+      }
+      const { getFirestore, doc, onSnapshot } = firestoreModular;
+      const callback = sinon.spy();
+      const unsub = onSnapshot(
+        doc(getFirestore(), `${COLLECTION}/mod-source-with-metadata`),
+        {
+          source: 'default',
+          includeMetadataChanges: true,
+        },
+        callback,
+      );
+
+      await Utils.spyToBeCalledOnceAsync(callback);
+      unsub();
     });
 
     it('throws if next callback is invalid', function () {
--- a/packages/firestore/ios/RNFBFirestore/RNFBFirestoreDocumentModule.m
+++ b/packages/firestore/ios/RNFBFirestore/RNFBFirestoreDocumentModule.m
@@ -97,13 +97,19 @@ RCT_EXPORT_METHOD(documentOnSnapshot
   };
 
   BOOL includeMetadataChanges = NO;
-  if (listenerOptions[@"includeMetadataChanges"] != nil) {
-    includeMetadataChanges = [listenerOptions[@"includeMetadataChanges"] boolValue];
+  FIRListenSource source = FIRListenSourceDefault;
+  if (listenerOptions[KEY_INCLUDE_METADATA_CHANGES] != nil) {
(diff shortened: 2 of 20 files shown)


packages/firestore/lib/types/firestore.ts
```diff
@@ -203,6 +203,7 @@ export type QueryConstraintType =
 
 export interface SnapshotListenOptions {
   readonly includeMetadataChanges?: boolean;
+  readonly source?: 'default' | 'cache';
```
mikehardy: This is correct in practice it seems, but I think should be typed as a `ListenSource` where ListenSource then has the default or cache value

https://firebase.google.com/docs/reference/js/firestore_.snapshotlistenoptions#snapshotlistenoptionssource
https://firebase.google.com/docs/reference/js/firestore_.md#listensource

A basically verbatim copy of this might work? 

https://github.com/firebase/firebase-js-sdk/blob/c44bb9a62b3c3e324677f14b55e6ca8578d4f411/packages/firestore/src/api/reference_impl.ts#L88-L101

packages/firestore/lib/types/internal.ts
```diff
@@ -94,9 +94,10 @@ export interface FirestoreAggregateQueryResultInternal {
   [key: string]: unknown;
 }
 
-/** Options for snapshot listeners (includeMetadataChanges). */
+/** Options for snapshot listeners (includeMetadataChanges, source). */
 export interface FirestoreSnapshotListenOptionsInternal {
   includeMetadataChanges?: boolean;
+  source?: 'default' | 'cache';
```
mikehardy: same thought that for perfect firebase-js-sdk type fidelity, in the absence of a reason not to, type like they do with a ListenSource wrapping the two string values

https://github.com/firebase/firebase-js-sdk/blob/c44bb9a62b3c3e324677f14b55e6ca8578d4f411/packages/firestore/src/api/reference_impl.ts#L88-L101

packages/firestore/type-test.ts
```diff
@@ -196,13 +206,15 @@ const nsArrayRemove = firebase.firestore.FieldValue.arrayRemove(1);
...
-}).then(() => {});
+nsDocRef
+  .set({
+    name: 'x',
+    deleted: nsDelete,
+    ts: nsServerTs,
+    arr: nsArrayUnion,
+    cnt: nsIncrement,
+  })
+  .then(() => {});
```
mikehardy, line 469: there are a few changes in here - I think this is the 4th one - that are spacing-only, not semantic changes and not really part of the PR. Maybe an editor config thing not picking up our prettier config ? Unsure, but it will likely cause issues with the linter - `yarn lint:js --fix` will probably set things right

packages/firestore/e2e/DocumentReference/onSnapshot.e2e.js
```diff
@@ -305,6 +305,97 @@ describe('firestore().doc().onSnapshot()', function () {
...
+      await docRef.set({ enabled: true });
+      await docRef.get();
+
+      let unsub = () => {};
+      try {
+        await firebase.firestore().disableNetwork();
+        const callback = sinon.spy();
+        unsub = docRef.onSnapshot({ source: 'cache' }, callback);
+        await Utils.spyToBeCalledOnceAsync(callback);
+        callback.args[0][0].metadata.fromCache.should.equal(true);
```
mikehardy: I love the added test - thank you - I wonder if this truly probes it though - I think it would be false-positive in that ListenSource.default and ListenSource.cache would have the same behavior in this scenario wouldn't they (since default listens to server and cache) ? So this doesn't really differentiate if the implementation is correct

The only thing I can think of that would truly differentiate is to have a function outside of the test app update firestore (so it didn't go in the test app local cache during set) and then have two listeners with the two different sources, do the out-of-band server set and make sure server listener was called but cache was not. Then do an in-band set and make sure they were both called

Cloud function to run in emulator could be added in to this area https://github.com/invertase/react-native-firebase/blob/main/.github/workflows/scripts/functions/src/index.ts ) where you told it a document path to update and a key/value pair perhaps (or anything that worked) via a post from the test app, and post from the test app could look like the auth ones do https://github.com/invertase/react-native-firebase/blob/main/packages/auth/e2e/helpers.js

A bit heavy but it is the only idea I have to really differentiate between the two source options and prove they work as defined vs just appear to function without error

Thoughts?

packages/firestore/ios/RNFBFirestore/RNFBFirestoreDocumentModule.m
```diff
@@ -97,13 +97,20 @@ - (void)invalidate {
   };
 
   BOOL includeMetadataChanges = NO;
-  if (listenerOptions[@"includeMetadataChanges"] != nil) {
-    includeMetadataChanges = [listenerOptions[@"includeMetadataChanges"] boolValue];
```
mikehardy, line 101: actually improved the area by extracting the other string to header constant as well, thanks

mikehardy (changes requested):
The implementation here looks really solid at the native level
The typescript types have a request for more strict coherence with upstream as mentioned in comments
The test support in types and unit look great, e2e I have a false-positive worry but solving it looks annoyingly "heavy", there is prior art to copy though so is perhaps not a large task?

This looks like just a little polish away from being merge-able though and I imagine it is already working in practice, fantastic, thank you

mikehardy:
Java failed lint and objective-c may as well

```
Attempt 1
  Error: google-java-format exited with exit code 1.
      at errorFromExitCode (/home/runner/work/react-native-firebase/react-native-firebase/node_modules/google-java-format/index.js:19:10)
      at ChildProcess.<anonymous> (/home/runner/work/react-native-firebase/react-native-firebase/node_modules/google-java-format/index.js:130:14)
      at ChildProcess.emit (node:events:519:28)
      at maybeClose (node:internal/child_process:1101:16)
      at ChildProcess._handle.onexit (node:internal/child_process:304:5)
  ```

for java: `yarn lint:android` and it will error but also write the fixes in ready for commit while checking
for ios: `yarn lint:ios:check` and/or `yarn lint:ios:fix` to see and/or write lint fixes in automatically

We're strict on formatting so it doesn't turn into an opinion battle but the fixes are all automated at least so shouldn't require thought except how to run them

File: packages/firestore/e2e/DocumentReference/onSnapshot.e2e.js (hunk 4 above)
Comment (astanb) [reply packages/firestore/e2e/DocumentReference/onSnapshot.e2e.js]: Good point, the existing test didn't really prove anything! 

Instead of going the callable function route I added an out-of-band helper function that uses direct API access to firestore, adds a little boilerplate but avoids the dependency on functions and could potentially be re-used in the future. Let me know if you'd prefer me to go the callable function route and I can do that.

File: packages/firestore/lib/types/namespaced.ts [hunk 6]
```diff
@@ -1764,7 +1764,11 @@ export namespace FirebaseFirestoreTypes {
     /**
      * Include a change even if only the metadata of the query or of a document changed. Default is false.
      */
-    includeMetadataChanges: boolean;
+    includeMetadataChanges?: boolean;
+    /**
+     * Set the source the query listens to. Default is 'default'.
+     */
+    source?: 'default' | 'cache';
```
mikehardy: should be `ListenSource` with accompanying import above now?

packages/firestore/lib/utils/index.ts
```diff
@@ -151,7 +151,10 @@ function isPartialObserver(
 }
 
 export interface ParseSnapshotArgsResult {
-  snapshotListenOptions: { includeMetadataChanges?: boolean };
+  snapshotListenOptions: {
+    includeMetadataChanges?: boolean;
+    source?: 'default' | 'cache';
```
mikehardy: should be `ListenSource` with accompanying import above now?

packages/firestore/lib/utils/index.ts
```diff
@@ -163,7 +166,10 @@ export function parseSnapshotArgs(args: unknown[]): ParseSnapshotArgsResult {
   }
 
   const NOOP = (): void => {};
-  const snapshotListenOptions: { includeMetadataChanges?: boolean } = {};
+  const snapshotListenOptions: {
+    includeMetadataChanges?: boolean;
+    source?: 'default' | 'cache';
```
mikehardy: should be `ListenSource` with accompanying import above now?

packages/firestore/lib/FirestoreDocumentReference.ts
```diff
@@ -188,7 +188,7 @@ export default class DocumentReference {
   }
 
   onSnapshot(...args: unknown[]): () => void {
-    let snapshotListenOptions: { includeMetadataChanges?: boolean };
+    let snapshotListenOptions: { includeMetadataChanges?: boolean; source?: 'default' | 'cache' };
```
mikehardy: should be `ListenSource` with accompanying import above now?

packages/firestore/lib/FirestoreQuery.ts
```diff
@@ -331,7 +331,7 @@ export default class Query {
   }
 
   onSnapshot(...args: unknown[]): () => void {
-    let snapshotListenOptions: { includeMetadataChanges?: boolean };
+    let snapshotListenOptions: { includeMetadataChanges?: boolean; source?: 'default' | 'cache' };
```
mikehardy: should be `ListenSource` with accompanying import above now?

mikehardy (changes requested):
that e2e test is exactly what I was hoping for, fantastic
It is done IMHO to just the last places to thread the `ListenSource` type change through then the firestore package here has a great new feature

packages/firestore/lib/utils/index.ts
```diff
@@ -31,6 +31,7 @@ import type {
   PartialSnapshotObserverInternal,
 } from '../types/internal';
 import FieldPath, { fromDotSeparatedString } from '../FieldPath';
+import type { ListenSource } from 'lib/types/firestore';
```
mikehardy: 🤔  most of the other intra-package references are using relative path annotation, I suppose this should be 

```suggestion
import type { ListenSource } from '../types/firestore';
```

?

packages/firestore/lib/utils/index.ts
```diff
@@ -189,9 +196,10 @@ export function parseSnapshotArgs(args: unknown[]): ParseSnapshotArgsResult {
   }
 
   if (isObject(args[0]) && !isPartialObserver(args[0])) {
-    const opts = args[0] as { includeMetadataChanges?: boolean };
+    const opts = args[0] as { includeMetadataChanges?: boolean; source?: 'default' | 'cache' };
```
mikehardy: missed a possible use of `ListenSource` type vs direct strings here

```suggestion
    const opts = args[0] as { includeMetadataChanges?: boolean; source?: ListenSource };
```

mikehardy (changes requested):
this looks fantastic - I believe it is down to two remaining spots still missing the requested ListenSource type addition, and that's it ? I think ready for merge after that

astanb:
@mikehardy should be ready for merge now

mikehardy:
We have done a *lot* of work on types in firestore since this was worked on, and that's not a bad thing - I can fix this one up and get it merged in, I'd like to see it released and I should be able to rebase main into it quickly

mikehardy (approved):
- squashed to a single feature commit after running all lints
- rebased to main and fixed up conflicts
- updated new-since-this-PR compare-types script to recognize stale entries as failures, and fixed up the now-stale entries for ListenSource since they are implemented now

This passes `yarn test:full` locally which is every test we have, running in CI now

this should be good to go, thanks again for the feature, the excellent e2e testing, and the patience

mikehardy:
not sure how an iOS compile bug snuck in there when I swear I had this testing locally and working before, but pushed a fix/tweak regardless.

mikehardy:
iOS simulators are very flaky at the moment for reasons unrelated to this PR, seeing one of the two go green (and with tests passing locally) this is good to go

PR #8931, opened 2026-03-13
feat(firestore):  Support for Firestore pipelines API 

russellwheatley:
## Support for Firestore pipelines API

- Adds a new public Firestore Pipelines API at `@react-native-firebase/firestore/pipelines` so consumers can build pipelines with `db.pipeline()` and run them with `execute(...)`.
- Keeps pipeline construction in JS and defers all native work until a single `pipelineExecute(...)` bridge call at execution time.
- Treats pipelines as a cross-platform feature for Android, iOS, and web, with platform-specific parsing/building/execution behind the same serialized payload.

## Public JS & runtime shape

- `import '@react-native-firebase/firestore/pipelines'` installs `pipeline()` onto the Firestore runtime prototype as a side effect. Similar to how it works on firebase-js-sdk.
- The new entrypoint exports the pipeline builder types, execute helper, result types, stage option types, and a large set of expression/helper functions aligned with the JS SDK naming.
- Source builders now support `collection`, `collectionGroup`, `database`, `documents`, and `createFrom(query)`.
- Pipeline instances are immutable; each stage append returns a new pipeline.
- JS runtime normalizes overloads and option aliases before serialization, so native sees a consistent shape even when the public API has overloads.
- Runtime guards prevent invalid combinations early, especially:
  - mixing pipelines/queries/references from different Firestore instances
  - invalid `documents(...)` inputs
  - `union(...)` self-cycles or cross-instance pipelines

## `pipelineExecute()` JS -> native contract

- `packages/firestore/lib/types/internal.ts` is the source of truth for the data sent over the wire to native.
- JS no longer sends loose  dictionaries; it serializes explicit node types such as:
  - field expressions
  - constant expressions
  - function expressions
  - orderings
  - aliased expressions
  - aliased aggregates
- The request shape is:
  - `pipeline.source`: one of `collection | collectionGroup | database | documents | query`
  - `pipeline.stages`: ordered `{ stage, options }[]`
  - `options`: currently `indexMode?: 'recommended'` and `rawOptions?: Record<string, unknown>`
- Query-backed sources are serialized as a typed query payload (`path`, `queryType`, `filters`, `orders`, `options`) rather than as an already-built native query.
- The response shape is `executionTime` plus `results[]`, which are rehydrated back into `PipelineSnapshot` / `PipelineResult`.

## Native architecture by platform

- Android now follows a clear `parse -> build native SDK objects -> execute -> serialize snapshot` flow.
- Android initially relied on brittle raw map plumbing for most pipeline work; now the parser produces typed DTOs and the node builder lowers those to Firestore SDK pipeline objects.
- iOS now follows the same broad separation, but via `parse -> bridge factory/node builder -> PipelineBridge.execute -> snapshot serializer`.
- Web was refactored to mirror native conceptually: parse serialized payload, rebuild a JS SDK pipeline, execute with the web SDK, then serialize the snapshot back into RNFB's internal response shape.

## Parsing & query building

- Native parsers validates the serialized payload up front instead of coercing deep raw dictionaries during execution which it was doing initially which made it nigh on impossible to reason with.
- Query source support was added so `createFrom(query)` can be serialized in JS and rebuilt natively/web later.
- Android rebuilds query sources using existing query infrastructure, then feeds that query into `pipelineSource.createFrom(...)`.
- iOS also rebuilds query sources through existing query infrastructure, but does so through typed parser DTOs plus bridge-factory glue.

## Known limitations & current guards

- `execute({ indexMode, rawOptions })` is rejected in JS on all platforms, because the execute-option surface is not available yet (It just throws an error on all platforms).
- iOS also rejects `pipeline.source.rawOptions` for source builders because the linked iOS pipeline bridge does not currently expose source options.
- iOS keeps a JS-side unsupported-function allowlist/denylist guard for pipeline functions that the current iOS lowering/runtime path does not yet support.
- That guard is intentional and should not be removed until the linked iOS native runtime is verified and the iOS node builder actually supports those functions. The public API shape is ahead of full iOS/native support in a few places; some options/functions exist in types and serialization but are intentionally blocked on native platforms.

## Snapshot & result handling

- Native and web responses are normalized into the same internal snapshot/result format before becoming public `PipelineSnapshot` / `PipelineResult` objects.
- Extra work was added around timestamp/result serialization so returned values preserve Firestore-like types consistently.
- iOS snapshot serialization includes fallback metadata handling for cases where the bridge does not surface all document metadata directly.

## Tests

- The branch adds unit tests for runtime serialization/guards, web rebuild/execute behavior, and native parity checks.
- Large E2E coverage was added for source builders, `createFrom(query)`, documents source, stage execution, expression behavior, unhappy paths, and platform-specific option failures.

## Why the compare-types script changed

- The compare-types tooling was extended so pipeline types can be checked as a separate public surface, not mixed into the main Firestore modular API.
- A dedicated `firestore-pipelines` SDK snapshot/config was added so CI can detect drift between RNFB pipeline types and the Firestore JS SDK pipeline types.
- The parser/comparer was hardened to avoid noisy false positives from formatting-only differences, inline comments, and class-vs-interface declaration differences (This is the reason why about 20 different Firestore diffs were included, the parser was ignoring firebase-js-sdk class v react-native-firebase interfaces, not it detects them and checks).
- The registry now supports multiple primary/support declaration files, which was needed because the pipelines API is exported from its own path and re-exports several supporting types.

### Checklist

  - [ ] Yes
  - [ ] `Android`
  - [ ] `iOS`
  - [ ] `Other` (macOS, web)
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [ ] No

### Test Plan

---

### DIFF
--- a/packages/firestore/android/src/reactnative/java/io/invertase/firebase/firestore/ReactNativeFirebaseFirestorePipelineExecutor.java
+++ b/packages/firestore/android/src/reactnative/java/io/invertase/firebase/firestore/ReactNativeFirebaseFirestorePipelineExecutor.java
@@ -53,8 +53,8 @@ import com.google.firebase.firestore.pipeline.RawStage;
 import com.google.firebase.firestore.pipeline.SampleStage;
 import com.google.firebase.firestore.pipeline.Selectable;
 import com.google.firebase.firestore.pipeline.UnnestOptions;
-import java.util.Arrays;
 import java.util.ArrayDeque;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Locale;
@@ -191,7 +191,9 @@ class ReactNativeFirebaseFirestorePipelineExecutor {
         Pipeline sourcePipeline = buildSourcePipeline(enterFrame.request.source);
         List<PendingPipelineStage> pendingStages =
             new java.util.ArrayList<>(enterFrame.request.stages.size());
-        List<Map.Entry<ReactNativeFirebaseFirestorePipelineParser.ParsedPipelineRequest, PipelineBox>>
+        List<
+                Map.Entry<
+                    ReactNativeFirebaseFirestorePipelineParser.ParsedPipelineRequest, PipelineBox>>
             nestedRequests = new java.util.ArrayList<>();
 
         for (ReactNativeFirebaseFirestorePipelineParser.ParsedPipelineStage stage :
@@ -221,8 +223,7 @@ class ReactNativeFirebaseFirestorePipelineExecutor {
       Pipeline currentPipeline = exitFrame.sourcePipeline;
       for (PendingPipelineStage pendingStage : exitFrame.stages) {
         if (pendingStage instanceof ReadyPipelineStage) {
-          currentPipeline =
-              applyStage(currentPipeline, ((ReadyPipelineStage) pendingStage).stage);
+          currentPipeline = applyStage(currentPipeline, ((ReadyPipelineStage) pendingStage).stage);
           continue;
         }
 
@@ -631,7 +632,8 @@ class ReactNativeFirebaseFirestorePipelineExecutor {
   private Pipeline applyUnionStage(
       Pipeline pipeline, ReactNativeFirebaseFirestorePipelineParser.ParsedUnionStage stage)
       throws PipelineValidationException {
-    throw new PipelineValidationException("pipelineExecute() failed to build nested union pipeline.");
+    throw new PipelineValidationException(
+        "pipelineExecute() failed to build nested union pipeline.");
   }
 
   private Pipeline applyUnnestStage(
--- a/packages/firestore/android/src/reactnative/java/io/invertase/firebase/firestore/ReactNativeFirebaseFirestorePipelineNodeBuilder.java
+++ b/packages/firestore/android/src/reactnative/java/io/invertase/firebase/firestore/ReactNativeFirebaseFirestorePipelineNodeBuilder.java
@@ -163,7 +163,8 @@ final class ReactNativeFirebaseFirestorePipelineNodeBuilder {
     final String fieldName;
     final LoweredExpressionBox box;
 
-    EnterObjectVectorExpressionValueFrame(Object value, String fieldName, LoweredExpressionBox box) {
+    EnterObjectVectorExpressionValueFrame(
+        Object value, String fieldName, LoweredExpressionBox box) {
       this.value = value;
       this.fieldName = fieldName;
       this.box = box;
@@ -191,9 +192,7 @@ final class ReactNativeFirebaseFirestorePipelineNodeBuilder {
     final List<String> pendingUnaryFunctions;
 
     ExitApplyPendingUnaryBooleanFrame(
-        LoweredExpressionBox box,
-        LoweredBooleanBox childBox,
-        List<String> pendingUnaryFunctions) {
+        LoweredExpressionBox box, LoweredBooleanBox childBox, List<String> pendingUnaryFunctions) {
       this.box = box;
       this.childBox = childBox;
       this.pendingUnaryFunctions = pendingUnaryFunctions;
@@ -258,7 +257,8 @@ final class ReactNativeFirebaseFirestorePipelineNodeBuilder {
     final LoweredExpressionBox box;
     final List<LoweredExpressionBox> childBoxes;
 
-    ExitObjectArrayExpressionFrame(LoweredExpressionBox box, List<LoweredExpressionBox> childBoxes) {
+    ExitObjectArrayExpressionFrame(
+        LoweredExpressionBox box, List<LoweredExpressionBox> childBoxes) {
       this.box = box;
       this.childBoxes = childBoxes;
     }
@@ -1099,16 +1099,22 @@ final class ReactNativeFirebaseFirestorePipelineNodeBuilder {
               }
               stack.push(
                   new ExitBooleanLogicalFrame(
-                      enterFrame.box, "AND".equals(normalizedOperator), childBoxes, currentFieldName));
+                      enterFrame.box,
+                      "AND".equals(normalizedOperator),
+                      childBoxes,
+                      currentFieldName));
               for (int i = queries.size() - 1; i >= 0; i--) {
                 stack.push(
                     new EnterObjectBooleanFrame(
-                        queries.get(i), currentFieldName + ".queries[" + i + "]", childBoxes.get(i)));
+                        queries.get(i),
+                        currentFieldName + ".queries[" + i + "]",
+                        childBoxes.get(i)));
               }
               continue;
             }
 
-            Object fieldValue = map.get("fieldPath") != null ? map.get("fieldPath") : map.get("field");
+            Object fieldValue =
+                map.get("fieldPath") != null ? map.get("fieldPath") : map.get("field");
             if (fieldValue == null) {
               throw new ReactNativeFirebaseFirestorePipelineExecutor.PipelineValidationException(
                   "pipelineExecute() expected " + currentFieldName + ".fieldPath to be provided.");
@@ -1121,7 +1127,11 @@ final class ReactNativeFirebaseFirestorePipelineNodeBuilder {
                     ? map.get("value")
                     : map.containsKey("right") ? map.get("right") : map.get("operand"));
             scheduleBooleanFunctionLowering(
-                mapOperatorToFunctionName(normalizedOperator), args, currentFieldName, enterFrame.box, stack);
+                mapOperatorToFunctionName(normalizedOperator),
+                args,
+                currentFieldName,
+                enterFrame.box,
+                stack);
             continue;
           }
 
@@ -1152,7 +1162,8 @@ final class ReactNativeFirebaseFirestorePipelineNodeBuilder {
               continue;
             }
 
-            scheduleBooleanFunctionLowering((String) name, args, currentFieldName, enterFrame.box, stack);
+            scheduleBooleanFunctionLowering(
+                (String) name, args, currentFieldName, enterFrame.box, stack);
             continue;
           }
         }
@@ -1171,10 +1182,13 @@ final class ReactNativeFirebaseFirestorePipelineNodeBuilder {
           continue;
         }
         if (containsLowerableExpression(enterFrame.value)) {
-          stack.push(new EnterObjectExpressionFrame(enterFrame.value, enterFrame.fieldName, enterFrame.box));
+          stack.push(
+              new EnterObjectExpressionFrame(
+                  enterFrame.value, enterFrame.fieldName, enterFrame.box));
           continue;
         }
-        enterFrame.box.value = constantExpression(resolveConstantValue(enterFrame.value, enterFrame.fieldName));
+        enterFrame.box.value =
+            constantExpression(resolveConstantValue(enterFrame.value, enterFrame.fieldName));
         continue;
       }
 
@@ -1185,11 +1199,13 @@ final class ReactNativeFirebaseFirestorePipelineNodeBuilder {
           continue;
         }
         if (containsLowerableExpression(enterFrame.value)) {
-          stack.push(new EnterObjectBooleanFrame(enterFrame.value, enterFrame.fieldName, enterFrame.box));
+          stack.push(
+              new EnterObjectBooleanFrame(enterFrame.value, enterFrame.fieldName, enterFrame.box));
           continue;
         }
 
-        Expression expression = constantExpression(resolveConstantValue(enterFrame.value, enterFrame.fieldName));
+        Expression expression =
+            constantExpression(resolveConstantValue(enterFrame.value, enterFrame.fieldName));
         if (expression instanceof BooleanExpression) {
           enterFrame.box.value = (BooleanExpression) expression;
           continue;
@@ -1210,7 +1226,8 @@ final class ReactNativeFirebaseFirestorePipelineNodeBuilder {
           LoweredExpressionBox expressionBox = new LoweredExpressionBox();
           stack.push(new ExitSetObjectFromExpressionFrame(enterFrame.box, expressionBox));
           stack.push(
-              new EnterObjectExpressionValueFrame(enterFrame.value, enterFrame.fieldName, expressionBox));
+              new EnterObjectExpressionValueFrame(
+                  enterFrame.value, enterFrame.fieldName, expressionBox));
           continue;
         }
         enterFrame.box.value = resolveConstantValue(enterFrame.value, enterFrame.fieldName);
@@ -1258,7 +1275,9 @@ final class ReactNativeFirebaseFirestorePipelineNodeBuilder {
           continue;
         }
 
-        stack.push(new EnterObjectExpressionValueFrame(currentValue, enterFrame.fieldName, enterFrame.box));
+        stack.push(
+            new EnterObjectExpressionValueFrame(
+                currentValue, enterFrame.fieldName, enterFrame.box));
         continue;
       }
 
@@ -1361,7 +1380,8 @@ final class ReactNativeFirebaseFirestorePipelineNodeBuilder {
           expressions[i] = exitFrame.childBoxes.get(i).value;
         }
         exitFrame.box.value =
-            Expression.rawFunction(normalizeExpressionFunctionName(exitFrame.functionName), expressions);
+            Expression.rawFunction(
+                normalizeExpressionFunctionName(exitFrame.functionName), expressions);
         continue;
       }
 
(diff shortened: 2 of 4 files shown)


packages/firestore/lib/pipelines/pipeline-result.ts
```diff
@@ -0,0 +1,77 @@
...
+import type Timestamp from '../FirestoreTimestamp';
+import type { Field } from './expressions';
+
+/**
+ * @beta
+ * Result of running a pipeline. One per document (or per group with aggregation, or single for global aggregation).
+ */
+export interface PipelineResult<T = DocumentData> {
+  data(): T;
+  get(fieldPath: string | import('../modular/FieldPath').FieldPath | Field): unknown;
```
russellwheatley: update this

packages/firestore/lib/pipelines/expressions.ts
```diff
@@ -0,0 +1,2317 @@
...
+]);
+
+function createNodeProxy<T extends RuntimeNode>(target: T): T {
+  const targetRecord = target as unknown as Record<PropertyKey, unknown>;
+  const cached = targetRecord[RUNTIME_PROXY_SYMBOL];
+  if (cached) {
+    return cached as T;
+  }
+
+  const proxy = new Proxy(targetRecord, {
```
russellwheatley: refactor this to remove proxy

packages/firestore/lib/pipelines/pipeline.ts
```diff
@@ -0,0 +1,92 @@
...
+import type { FirestorePipelineSerializedInternal } from '../types/internal';
+
+/** @internal Serialized pipeline payload shape forwarded to native/web executors. */
+export type SerializedPipeline = FirestorePipelineSerializedInternal;
+
+/**
+ * @beta
+ * Pipeline with chained stages. Each stage returns a new Pipeline (immutable chain).
+ */
+export interface Pipeline<T = import('../types/firestore').DocumentData> {
```
russellwheatley: update this

packages/firestore/__tests__/pipelines.test.ts
```diff
@@ -0,0 +1,427 @@
...
+  trunc,
+} from '../lib/pipelines';
+import '../lib/pipelines';
+import { ConstantExpression } from '../lib/pipelines/expressions';
+import { getIOSUnsupportedPipelineFunctions } from '../lib/pipelines/pipeline_support';
+
+describe('Firestore pipelines runtime', function () {
+  beforeAll(function () {
+    // @ts-ignore test
+    globalThis.RNFB_SILENCE_MODULAR_DEPRECATION_WARNINGS = true;
```
mikehardy: 🤔  I'm confused by any interaction with the namespaced API in this instance - we're modular-only for new features aren't we?

packages/firestore/android/src/reactnative/java/io/invertase/firebase/firestore/ReactNativeFirebaseFirestorePipelineExecutor.java
```diff
@@ -0,0 +1,1070 @@
...
+        case String:
+          output.put(key, readableMap.getString(key));
+          break;
+        case Map:
+          ReadableMap nestedMap = readableMap.getMap(key);
+          output.put(key, nestedMap == null ? null : readableMapToJava(nestedMap));
+          break;
+        case Array:
+          ReadableArray nestedArray = readableMap.getArray(key);
+          output.put(key, nestedArray == null ? null : readableArrayToJava(nestedArray));
```
mikehardy: Oh no 😆  is this another case of possibly unbounded recursion? And last I read, even if it is tail-recursion in Java because of GC constraints holding on to pointers and such it can run out of stack and needs the stack-based -> heap-based conversion to eliminate risk on these two recursive calls made on user-supplied data structures

packages/firestore/android/src/reactnative/java/io/invertase/firebase/firestore/ReactNativeFirebaseFirestorePipelineExecutor.java
```diff
@@ -0,0 +1,1070 @@
...
+        case String:
+          output.add(readableArray.getString(i));
+          break;
+        case Map:
+          ReadableMap nestedMap = readableArray.getMap(i);
+          output.add(nestedMap == null ? null : readableMapToJava(nestedMap));
+          break;
+        case Array:
+          ReadableArray nestedArray = readableArray.getArray(i);
+          output.add(nestedArray == null ? null : readableArrayToJava(nestedArray));
```
mikehardy: same concerned thought about doing recursion on user-supplied data here

packages/firestore/android/src/reactnative/java/io/invertase/firebase/firestore/ReactNativeFirebaseFirestorePipelineNodeBuilder.java
```diff
@@ -0,0 +1,1026 @@
...
+    if (value instanceof Map) {
+      Map<?, ?> map = (Map<?, ?>) value;
+      Object directPath = map.get("path");
+      if (directPath instanceof String && !((String) directPath).isEmpty()) {
+        return (String) directPath;
+      }
+
+      Object fieldPath = map.get("fieldPath");
+      if (fieldPath != null && fieldPath != map) {
+        return coerceFieldPath(fieldPath, fieldName);
```
mikehardy: recursion, I believe this is on user-supplied structures? Unsure on this one

packages/firestore/android/src/reactnative/java/io/invertase/firebase/firestore/ReactNativeFirebaseFirestorePipelineNodeBuilder.java
```diff
@@ -0,0 +1,1026 @@
...
+
+    throw new ReactNativeFirebaseFirestorePipelineExecutor.PipelineValidationException(
+        "pipelineExecute() expected " + fieldName + " to resolve to a field path string.");
+  }
+
+  double[] coerceVectorValue(Object value)
+      throws ReactNativeFirebaseFirestorePipelineExecutor.PipelineValidationException {
+    if (value instanceof Map) {
+      Object nestedValues = ((Map<?, ?>) value).get("values");
+      return coerceVectorValue(nestedValues);
```
mikehardy: a broken record on the theme of recursion, I believe this is on user-supplied structures? Unsure on this one - also, these stack->heap transforms are performant but they are developer-time expensive. I'll note that in my analysis it was the only way to be sure in Java and Obj-C/Swift that you wouldn't StackOverflowError, but I'm open to being wrong if there's an analysis that shows that tail-recursion can handle arbitrary levels (and that all our recursion is tail-recursion)

packages/firestore/android/src/reactnative/java/io/invertase/firebase/firestore/ReactNativeFirebaseFirestorePipelineParser.java
```diff
@@ -0,0 +1,1243 @@
...
+        case String:
+          output.put(key, readableMap.getString(key));
+          break;
+        case Map:
+          ReadableMap nestedMap = readableMap.getMap(key);
+          output.put(key, nestedMap == null ? null : readableMapToJava(nestedMap));
+          break;
+        case Array:
+          ReadableArray nestedArray = readableMap.getArray(key);
+          output.put(key, nestedArray == null ? null : readableArrayToJava(nestedArray));
```
mikehardy, line 1326: hope I didn't miss at least marking these, assuming it's an issue - but this looks like tail-recursion which I think doesn't avoid StackOverflowException risk

packages/firestore/android/src/reactnative/java/io/invertase/firebase/firestore/ReactNativeFirebaseFirestorePipelineParser.java
```diff
@@ -0,0 +1,1243 @@
...
+        case String:
+          output.add(readableArray.getString(i));
+          break;
+        case Map:
+          ReadableMap nestedMap = readableArray.getMap(i);
+          output.add(nestedMap == null ? null : readableMapToJava(nestedMap));
+          break;
+        case Array:
+          ReadableArray nestedArray = readableArray.getArray(i);
+          output.add(nestedArray == null ? null : readableArrayToJava(nestedArray));
```
mikehardy, line 1363: StackOverflowException mark

packages/firestore/ios/RNFBFirestore/RNFBFirestoreCollectionModule.m
```diff
@@ -16,9 +16,11 @@
  */
 
 #import <RNFBApp/RNFBRCTEventEmitter.h>
+#import <RNFBFirestore/RNFBFirestore-Swift.h>
```
mikehardy, line 21: warning that this is a use_frameworks/modular style header but it will fail in newly-typical build situations where the RNFB module is not use_frameworks with modular headers and thus fails to find it's own auto-generated Swift bridging header - needs this treatment:

https://github.com/invertase/react-native-firebase/pull/8925/changes#diff-49931ad4a2ab4fdce3492a3b5b98d90e001f83196af76aeffaa587d301b9a99dR27-R38

packages/firestore/ios/RNFBFirestore/RNFBFirestorePipelineBridgeFactory.swift
```diff
@@ -0,0 +1,244 @@
...
+final class RNFBFirestorePipelineBridgeFactory {
+  private let nodeBuilder = RNFBFirestorePipelineNodeBuilder()
+
+  func buildStageBridges(
+    firestore: Firestore,
+    request: RNFBFirestoreParsedPipelineRequest
+  ) throws -> [StageBridge] {
+    var stageBridges = try buildSourceStageBridges(firestore: firestore, source: request.source)
+    for stage in request.stages {
+      stageBridges.append(try buildStageBridge(stage, firestore: firestore))
```
mikehardy: This calls buildStageBridge, and it appeared to me that buildStageBridge had at least one case that called buildStageBridges, a recursive cycle

Not sure on the provenance or possible size of the input here for risk of stack overflow during recursion

packages/firestore/ios/RNFBFirestore/RNFBFirestorePipelineBridgeFactory.swift
```diff
@@ -0,0 +1,244 @@
...
+    case let .sampleStage(parsed):
+      if let documents = parsed.documents {
+        return SampleStageBridge(count: Int64(try nodeBuilder.coerceInt(documents, fieldName: "stage.options.documents")))
+      }
+      if let percentage = parsed.percentage {
+        return SampleStageBridge(percentage: try coerceNumber(percentage, fieldName: "stage.options.percentage"))
+      }
+      throw PipelineValidationError("pipelineExecute() expected sample stage to include documents or percentage.")
+    case let .unionStage(parsed):
+      let otherStages = try buildStageBridges(firestore: firestore, request: parsed.other)
```
mikehardy: here's the other side of the recursive cycle that goes back to buildStageBridges from buildStageBridge

packages/firestore/ios/RNFBFirestore/RNFBFirestorePipelineBridgeFactory.swift
```diff
@@ -0,0 +1,244 @@
...
+        "__kind": "expression",
+        "exprType": "Constant",
+        "value": buildQuerySourceValue(value),
+      ]
+    case let .function(name, args):
+      return [
+        "__kind": "expression",
+        "exprType": "Function",
+        "name": name,
+        "args": buildQuerySourceArray(args),
```
mikehardy: I'm actually annoyed with myself for leaving all these recursion comments, I'm reading all the lines these are just the only things I see right now
And I'm annoyed because if they are an issue...zero fun

packages/firestore/ios/RNFBFirestore/RNFBFirestorePipelineNodeBuilder.swift
```diff
@@ -0,0 +1,939 @@
...
+
+    guard let map = value as? [String: Any] else {
+      throw PipelineValidationError("pipelineExecute() could not convert \(fieldName) into a pipeline expression.")
+    }
+
+    if let nested = map["expr"] {
+      return try coerceExpression(nested, fieldName: "\(fieldName).expr")
+    }
+    if let nested = map["expression"] {
+      return try coerceExpression(nested, fieldName: "\(fieldName).expression")
```
mikehardy: recursive calls

packages/firestore/ios/RNFBFirestore/RNFBFirestorePipelineNodeBuilder.swift
```diff
@@ -0,0 +1,939 @@
...
+      "arraycontains", "arraycontainsany", "arraycontainsall", "equalany", "notequalany",
+    ]
+
+    if comparisonFunctions.contains(normalized) {
+      guard args.count >= 2 else {
+        throw PipelineValidationError(
+          "pipelineExecute() expected \(fieldName).args to include left and right operands.")
+      }
+
+      let left = try coerceExpression(args[0], fieldName: "\(fieldName).args[0]")
```
mikehardy: coerceExpression may call coerceFunctionExpression which can then recur here, not sure on this specific provenance/size for data tho, maybe not an issue

packages/firestore/ios/RNFBFirestore/RNFBFirestorePipelineNodeBuilder.swift
```diff
@@ -0,0 +1,939 @@
...
+    case let .primitive(primitive):
+      return primitive
+    case let .list(values):
+      return values.map(serializeValueNode)
+    case let .map(values):
+      return values.reduce(into: [String: Any]()) { result, entry in
+        result[entry.key] = serializeValueNode(entry.value)
+      }
+    case let .expression(expression):
+      return serializeExpressionNode(expression)
```
mikehardy: serializeExpressionNode->serializeValueNode->serializeExpressionNode ♾️

packages/firestore/ios/RNFBFirestore/RNFBFirestorePipelineSnapshotSerializer.swift
```diff
@@ -0,0 +1,98 @@
...
+  func serializeSnapshot(
+    _ snapshot: __PipelineSnapshotBridge?,
+    metadata: PipelineExecutionMetadata
+  ) throws -> [AnyHashable: Any] {
+    var output: [AnyHashable: Any] = [:]
+    let results = snapshot?.results ?? []
+    let resultCount = results.count
+
+    output["results"] = results.enumerated().map { index, result in
+      serializeResult(result, fallbackPath: metadata.fallbackPath(for: index, resultCount: resultCount))
```
mikehardy, line 42: are we reading all results into heap then sending them over to JS and forwarding to client?

standard question about size of data vs expectation of memory behavior, possible need to stream through to API callers ? Not sure it is even possible to stream, but we may be at risk of heap exhaustion here (same thought would apply to Java implementation, perhaps JS as well if our layer is a process/forward intermediary)

tests/local-tests/firestore/pipelines-e2e.tsx
```diff
@@ -0,0 +1,688 @@
...
+const COLLECTION = 'pipeline-collection';
+
+function randString(length: number): string {
+  const chars = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789';
+  let result = '';
+  for (let i = 0; i < length; i++) {
+    result += chars.charAt(Math.floor(Math.random() * chars.length));
+  }
+  return result;
+}
```
mikehardy, line 26: this helper exists already I think ?

tests/local-tests/firestore/pipelines-e2e.tsx
```diff
@@ -0,0 +1,688 @@
...
+
+interface TestCase {
+  name: string;
+  run: TestFn;
+}
+
+export function PipelinesE2ETestComponent() {
+  const [result, setResult] = useState<string>('');
+  const [loading, setLoading] = useState<string | null>(null);
+  // "default" database does not seem to work
```
mikehardy, line 38: 🤔  curious what the error is, may be useful to put the error in and currently firebase-tools / SDK version combinations where it was seen, to prompt / inform possible future inspection

russellwheatley: "default" database does not work for pipelines, you need to use "enterprise" as it uses Enterprise Firestore db. See documentation: https://firebase.google.com/docs/firestore/pipelines/get-started-with-pipelines#initialization.

packages/firestore/ios/RNFBFirestore/RNFBFirestorePipelineNodeBuilder.swift
mikehardy: still a recursive cycle here, however, while data may be arbitrarily nested, perhaps there is a limit to depth of pipeline operations? Such that stack exhaustion isn't a worry ?

packages/firestore/lib/web/pipelines/pipeline_bridge_factory.ts
```diff
@@ -0,0 +1,292 @@
...
+  revivePipelineValue,
+  type WebPipelineInstance,
+  type WebPipelineSource,
+} from './pipeline_node_builder';
+import type { WebParsedPipelineRequest } from './pipeline_parser';
+import { buildQuery } from '../query';
+import type { FilterSpec, OrderSpec, QueryOptions } from '../query';
+
+const FIRESTORE_LITE_UNSUPPORTED_SUFFIX =
+  ' This operation is unavailable because the web runtime uses Firestore Lite.';
```
mikehardy: Just an idle thought - related to an issue in the tracker I'm still thinking on and maybe you have an idea - it may be possible to optionally depend on firestore, and let the user specify their own firebase-js-sdk dependency. We have currently non-optionally chosen for users that they get firestore-lite but there are people that want full firestore. If it were possible to alter the dependency at the app package.json level and for us to tolerate both cases, that would be ideal and a change I hope to / wish to make

How would it affect this area? (looks like it's just a couple places here - would need to be a conditional vs a hard-throw-error type thing I guess - not sure what we'd test to see if it was lite or not 🤔 )

packages/firestore/lib/pipelines/pipeline_support.ts
```diff
@@ -0,0 +1,134 @@
...
+  'findNearest',
+  'replaceWith',
+  'sample',
+  'union',
+  'unnest',
+  'rawStage',
+] as const;
+
+export const PIPELINE_UNSUPPORTED_BASE_MESSAGE =
+  'Firestore pipelines are not supported by this native implementation yet.';
```
mikehardy: base message seems conclusive and "all not supported"-like to me, perhaps slightly softer / less binary-sounding "not supported" base message to flow better with the suffixes that are tacked on about specific stages / sources ?

```suggestion
  'Some Firestore pipeline features are not supported by this native implementation yet.';
```

packages/firestore/ios/RNFBFirestore/RNFBFirestorePipelineNodeBuilder.swift
russellwheatley: Not quite finished on android yet, I've been doing it piecemeal and rebuilding/testing so that it's easier to catch

packages/firestore/__tests__/pipelines.test.ts
russellwheatley: sorry, this is an oversight, I'll get it removed.

packages/firestore/lib/web/pipelines/pipeline_bridge_factory.ts
russellwheatley: I've updated this error message to more accurate reason why it would fail at this point: [`4ed337e` (this PR)](https://github.com/invertase/react-native-firebase/pull/8931/commits/4ed337eca15ff05eaacb6d82a33fd96bc147cf7a)

i.e. it shouldn't fail because of lite but because pipelines hadn't been imported properly.

PR #8942, opened 2026-03-27
ci: update AI package workflow to use compare script

russellwheatley:
## Summary
- Added AI package coverage to the `compare-types` workflow by registering `ai`, adding the `ai-sdk.d.ts` snapshot, and creating an AI-specific `config.ts`.
- Improved the compare script so it handles enums and enum-like SDK constant objects correctly, and filters out private/internal API noise when comparing public types.
- Documented the known AI package differences in `config.ts` and removed stale config entries uncovered by the parser fixes.
- Removed the AI Cursor rule markdown files from this repo because that guidance has been moved to the private skill repository.

## Why
- This makes AI API drift detectable in CI without needing a local `firebase-js-sdk` checkout, so new web SDK features or type changes are easier to spot and evaluate for RNFB parity.
- The compare script changes were needed because AI exposes enum-heavy types and generated declarations include private/internal members that previously produced misleading false positives.
- Populating the AI config and removing stale entries makes the compare output actionable: only real, intentional public API differences remain documented.
- Moving the AI rules out of this repo keeps the public branch focused on the compare-types workflow while retaining the porting guidance in the private skill system.

### Checklist

  - [ ] Yes
  - [ ] `Android`
  - [ ] `iOS`
  - [ ] `Other` (macOS, web)
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [ ] No

### Test Plan

---

### DIFF
--- a/.github/scripts/compare-types/packages/ai/config.ts
+++ b/.github/scripts/compare-types/packages/ai/config.ts
@@ -101,6 +101,31 @@ const config: PackageConfig = {
       reason:
         'Chrome Prompt API prompt options type used by browser-only on-device language model integration.',
     },
+    {
+      name: 'AnyOfSchema',
+      reason:
+        'RN Firebase schema-builder does not currently expose the `anyOf` helper class, so union-schema composition is not part of the public RN AI API.',
+    },
+    {
+      name: 'LiveServerGoingAwayNotice',
+      reason:
+        'RN Firebase live sessions do not currently surface the server `goingAwayNotice` message type in the public API.',
+    },
+    {
+      name: 'ObjectSchemaRequest',
+      reason:
+        'RN Firebase exposes `ObjectSchemaInterface` for schema helper typing, but does not separately export the raw request-shape `ObjectSchemaRequest` type.',
+    },
+    {
+      name: 'SingleRequestOptions',
+      reason:
+        'RN Firebase does not currently expose per-call request overrides such as `AbortSignal`; requests are configured via model-level `RequestOptions` only.',
+    },
+    {
+      name: 'ThinkingLevel',
+      reason:
+        'RN Firebase supports thinking budgets but does not currently expose the JS SDK `ThinkingLevel` preset constants/type.',
+    },
   ],
   extraInRN: [
     {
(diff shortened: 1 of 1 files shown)


mikehardy:
kicked off android CI again, appeared to be a flake, rerun will provide another data point

macOS failure is #8947 - unrelated, non-blocking but at least a local test with this PR rebased after 8947 goes in would be good since the AI package makes heavy use of firebase-js-sdk which exposes problems between pure-web and web-APIs-in-react-native-execution context which macOS / Other platform probes well

.github/scripts/compare-types/packages/ai/config.ts
```diff
@@ -0,0 +1,271 @@
...
+    },
+    {
+      name: 'HybridParams',
+      reason:
+        'Hybrid mode combines Chrome on-device AI with cloud inference, which is a browser-only feature and intentionally absent in React Native.',
+    },
+    {
+      name: 'OnDeviceParams',
+      reason:
+        'On-device inference params are only used by the browser Chrome adapter / hybrid flow, which is not supported in React Native.',
```
mikehardy, line 52: I was curious about this assertion (I'm aware of firebase-ml, and some native / on-device SDK work in the google AI area and was sure they had something ?) and sure enough there is stuff for Android:

https://firebase.google.com/docs/ai-logic/hybrid/android/get-started?api=dev

> On-device inference only supports single-turn text generation (not chat), with streaming or non-streaming output. It supports the following text-generation capabilities:
> 
> Generating [text from text-only input](https://firebase.google.com/docs/ai-logic/hybrid/android/get-started?api=dev#text-in-text-out)
> 
> Generating [text from text-and-image input](https://firebase.google.com/docs/ai-logic/hybrid/android/get-started?api=dev#image-in-text-out), specifically a single Bitmap image as input

Something to think about as enabling-via-bindings this sort of native integration is the reason react-native-firebase exists vs pure firebase-js-sdk

.github/scripts/compare-types/src/compare.ts
```diff
@@ -11,6 +11,7 @@ import type {
   InterfaceShape,
   TypeAliasShape,
   VariableShape,
+  EnumShape,
```
mikehardy, line 14: Happy to see this - was one of the few things that seemed low signal-to-noise

PR #8964, opened 2026-04-08
fix(analytics): coerce JS string or double params to native integral values as needed

brojor:
### Description

The React Native bridge converts all JS numbers to `double`, but the native Firebase SDK expects `FirebaseAnalytics.Param.INDEX` (which maps to `item_list_index`) as an integer/long. When a `double` is forwarded, Firebase does not recognise it as the standard ecommerce parameter — `item_list_index` ends up as `(not set)` in GA4 / BigQuery, breaking attribution of clicks/views to list positions.

The library already handles this exact problem for `Param.QUANTITY` in both native modules. This PR mirrors that conversion for `Param.INDEX`:

- **Android** (`ReactNativeFirebaseAnalyticsModule.java`): `putLong(FirebaseAnalytics.Param.INDEX, (long) number)`
- **iOS** (`RNFBAnalyticsModule.m`): `item[kFIRParameterIndex] = @([item[kFIRParameterIndex] integerValue])`

No JS / TS changes — the public API already accepts `index: number`, only the native coercion was missing.

### Related issues

None filed upstream; happy to open one if preferred.

### Release Summary

Fix `item_list_index` showing as `(not set)` in GA4 / BigQuery when `index` is provided on `items[]`.

### Checklist

  - [x] Yes
  - [x] `Android`
  - [x] `iOS`
  - [ ] `Other` (macOS, web)
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

No tests added — the existing `QUANTITY` conversion (the precedent this PR follows) is also not covered by jest/e2e tests, since the coercion happens in the native modules below the JS bridge and there is no native test infra in the analytics package. Happy to add coverage if maintainers point me at the right place.

### Test Plan

Verified in production via \`patch-package\` against \`@react-native-firebase/analytics@22.3.0\` in a live e-commerce app:

- **Before:** \`item_list_index\` arrives as \`(not set)\` in GA4 / BigQuery despite \`index\` being set on every item.
- **After:** \`item_list_index\` is populated with the correct integer position for \`view_item_list\`, \`select_item\`, and related ecommerce events.

:fire:

---

> [!NOTE]
> **Medium Risk**
> Touches native Android/iOS analytics parameter marshaling and tightens JS/TS validation for `success`, which could affect apps relying on previously accepted string values. Changes are localized to analytics event parameter conversion but can impact GA4/BigQuery reporting if incorrect.
> 
> **Overview**
> Fixes GA4 parameter typing issues by coercing specific numeric event params (e.g., `items[].index` and other GA4 long-only keys) from React Native bridge `double` values into native integral types on both Android and iOS, including within `items` arrays.
> 
> Also normalizes `LevelEnd.success` to be a boolean in JS/TS validation/types (and updates tests accordingly) while converting the native `success` value into the expected integer/long representation for the Firebase SDK.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 005267186cb6499ba9fc2c7423141e8199e28dee. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/analytics/android/src/reactnative/java/io/invertase/firebase/analytics/ReactNativeFirebaseAnalyticsModule.java
+++ b/packages/analytics/android/src/reactnative/java/io/invertase/firebase/analytics/ReactNativeFirebaseAnalyticsModule.java
@@ -211,6 +211,10 @@ public class ReactNativeFirebaseAnalyticsModule extends ReactNativeFirebaseModul
               double number = itemBundle.getDouble(FirebaseAnalytics.Param.QUANTITY);
               itemBundle.putInt(FirebaseAnalytics.Param.QUANTITY, (int) number);
             }
+            if (itemBundle.containsKey(FirebaseAnalytics.Param.INDEX)) {
+              double number = itemBundle.getDouble(FirebaseAnalytics.Param.INDEX);
+              itemBundle.putLong(FirebaseAnalytics.Param.INDEX, (long) number);
+            }
             validBundles.add(itemBundle);
           }
         }
--- a/packages/analytics/android/src/reactnative/java/io/invertase/firebase/analytics/ReactNativeFirebaseAnalyticsModule.java
+++ b/packages/analytics/android/src/reactnative/java/io/invertase/firebase/analytics/ReactNativeFirebaseAnalyticsModule.java
@@ -26,6 +26,7 @@ import com.facebook.react.bridge.ReadableMap;
 import com.google.firebase.analytics.FirebaseAnalytics;
 import io.invertase.firebase.common.ReactNativeFirebaseModule;
 import java.util.ArrayList;
+import java.util.Locale;
 import javax.annotation.Nullable;
 
 public class ReactNativeFirebaseAnalyticsModule extends ReactNativeFirebaseModule {
@@ -233,6 +234,7 @@ public class ReactNativeFirebaseAnalyticsModule extends ReactNativeFirebaseModul
     }
 
     coerceLongNumericParams(bundle);
+    coerceSuccessParamToLong(bundle);
 
     if (bundle.containsKey(FirebaseAnalytics.Param.EXTEND_SESSION)) {
       double number = bundle.getDouble(FirebaseAnalytics.Param.EXTEND_SESSION);
@@ -248,5 +250,23 @@ public class ReactNativeFirebaseAnalyticsModule extends ReactNativeFirebaseModul
         bundle.putLong(key, (long) number);
       }
     }
+  }
+
+  private static void coerceSuccessParamToLong(Bundle bundle) {
+    if (!bundle.containsKey(FirebaseAnalytics.Param.SUCCESS)) {
+      return;
+    }
+    Object value = bundle.get(FirebaseAnalytics.Param.SUCCESS);
+    bundle.remove(FirebaseAnalytics.Param.SUCCESS);
+    long asLong = 0L;
+    if (value instanceof Boolean) {
+      asLong = (Boolean) value ? 1L : 0L;
+    } else if (value instanceof Number) {
+      asLong = ((Number) value).longValue() != 0L ? 1L : 0L;
+    } else if (value instanceof String) {
+      String s = ((String) value).trim().toLowerCase(Locale.ROOT);
+      asLong = ("1".equals(s) || "true".equals(s) || "yes".equals(s)) ? 1L : 0L;
+    }
+    bundle.putLong(FirebaseAnalytics.Param.SUCCESS, asLong);
   }
 }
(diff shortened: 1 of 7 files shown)


packages/analytics/android/src/reactnative/java/io/invertase/firebase/analytics/ReactNativeFirebaseAnalyticsModule.java
```diff
@@ -211,6 +211,10 @@ private Bundle toBundle(ReadableMap readableMap) {
               double number = itemBundle.getDouble(FirebaseAnalytics.Param.QUANTITY);
               itemBundle.putInt(FirebaseAnalytics.Param.QUANTITY, (int) number);
             }
+            if (itemBundle.containsKey(FirebaseAnalytics.Param.INDEX)) {
+              double number = itemBundle.getDouble(FirebaseAnalytics.Param.INDEX);
+              itemBundle.putLong(FirebaseAnalytics.Param.INDEX, (long) number);
+            }
```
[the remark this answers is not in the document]
brojor: Regarding the `QUANTITY` inconsistency — I'd prefer to keep that out of this PR to stay focused on the `INDEX` fix. If `QUANTITY` should indeed be a `long` per the Firebase SDK, that's a pre-existing issue and deserves its own PR so it can be reviewed and reverted independently.

For the variable naming, I intentionally kept `number` to stay consistent with the existing `QUANTITY` block right above, which uses the same pattern. Renaming only the new lines would introduce inconsistency within the same method.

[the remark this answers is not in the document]
brojor: @gemini-code-assist I responded inline — keeping QUANTITY out of scope and preserving the `number` variable for consistency with the surrounding block. Anything else that should block this PR from your side?

brojor:
Friendly ping @mikehardy @Ehesp — this is a small, low-risk native fix that mirrors the existing `QUANTITY` → int cast for `INDEX`. Both platforms are covered, CI is green, and it's been validated in production. Would appreciate a review when you get a chance. 🔥

mikehardy:
this is a good change and I respect that you didn't want to perturb other params so you could stay focused on the one you were affected by

However, I had a minute to do a quick audit since I saw you mentioned quantity should be long (not int) - I was curious how many other params needed special handling for the "javascript is doubles -> ios/java needs integral values" case

There were a lot! So I refactored it into an array of them so they were all handled, and also handled "success" (a level end param) which needed a boolean -> integral value coercion

I'm going to push that as a second commit, once CI goes green I can merge and release this

Cheers

packages/analytics/ios/RNFBAnalytics/RNFBAnalyticsModule.m
```diff
@@ -268,20 +286,47 @@ - (NSDictionary *)cleanJavascriptParams:(NSDictionary *)params {
...
     newParams[kFIRParameterExtendSession] = @YES;
   }
   return [newParams copy];
 }
 
+- (void)rnfb_coerceLongNumericParametersInMutableDictionary:(NSMutableDictionary *)dict {
+  for (NSString *key in RNFBAnalyticsLongNumericParameterKeys()) {
+    id value = dict[key];
+    if (value != nil && value != [NSNull null]) {
+      dict[key] = @([value intValue]);
```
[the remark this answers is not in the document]
mikehardy: I was suspicious of this myself. re-pushing in a moment with `integerValue` in place of `intValue`

mikehardy (approved):
I liked the original change, and appreciate that it mentioned there were other things wrong in the area - that led to the wider change which I've reviewed and Cursor reviewed, I think it's good to go. Pending a final set of eyes

packages/analytics/lib/types/analytics.ts
```diff
@@ -282,7 +282,7 @@ export interface LevelEndEventParameters {
   /**
    * The result of an operation.
    */
-  success?: string;
+  success?: boolean;
```
russellwheatley, line 285: Wouldn't this constitute a breaking change?

russellwheatley (changes requested):

packages/analytics/lib/types/analytics.ts
mikehardy: I considered this, however, and normally would say definitely yes that is a breaking change.
But during consideration I also realized what I put in the commit message:

> additionally the level end "success" param could never have worked as it was a string
in JS level when firebase-js-sdk has it as a boolean, and in native it is also an integral
value - so attempting to use a string natively as before would never register in analytics

...that is, there is just no way it worked before. It was broken out of the box. Is it a breaking change to break something broken? Historically we've released things like that as patch releases - it's rare but we've done it

russellwheatley: Yep, this sounds reasonable to me. No need to make breaking change.

russellwheatley (approved):

packages/analytics/lib/types/analytics.ts
mikehardy: After some thought, I split the commits for coercion such that level end "success" has a separate commit with a prescriptive changelog message and discussion, for anyone impacted

PR #8999, opened 2026-04-30
docs(repo): add AGENTS.md

dackers86:
## Summary
- add a root `AGENTS.md` using the `agents.md` example structure adapted for React Native Firebase
- document repo layout, focused validation guidance, and PR expectations for coding agents
- keep the guidance concise and specific to the monorepo and package workflow

## Test plan
- [ ] Docs-only change

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -17,7 +17,7 @@
 ## Testing instructions
 
 - Start with focused validation for the area you changed instead of running the entire suite.
-- For JavaScript and TypeScript changes, run `yarn lint:js`.
+- For code changes, run the relevant linting scripts (e.g., yarn lint:js, yarn lint:android, or yarn lint:ios:check).
 - For package behavior changes, run targeted Jest coverage with `yarn tests:jest <path-to-test-file>`.
 - For public API changes, update and run the relevant `packages/<name>/type-test.ts` checks and nearby Jest tests in `packages/<name>/__tests__/`.
 - Run platform-specific e2e commands in `tests/` only when the change affects Android, iOS, or macOS behavior.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,30 @@
+# React Native Firebase AGENTS Guide
+
+## Project overview
+
+- This repository is a Yarn workspaces monorepo for React Native Firebase packages.
+- Most library code lives in `packages/*`.
+- The end-to-end test app lives in `tests/`.
+- Contributor expectations and review norms are documented in `CONTRIBUTING.md`.
+
+## Dev environment tips
+
+- Install dependencies once at the repository root with `yarn`.
+- Prefer working from the repository root so package scripts and workspace resolution behave consistently.
+- Check the affected package's `type-test.ts`, `__tests__/`, and plugin directories before changing public APIs or platform-specific behavior.
+- Prefer consistency with existing patterns in the package you are editing instead of introducing new structure.
+
+## Testing instructions
+
+- Start with focused validation for the area you changed instead of running the entire suite.
+- For JavaScript and TypeScript changes, run `yarn lint:js`.
+- For package behavior changes, run targeted Jest coverage with `yarn tests:jest <path-to-test-file>`.
+- For public API changes, update and run the relevant `packages/<name>/type-test.ts` checks and nearby Jest tests in `packages/<name>/__tests__/`.
+- Run platform-specific e2e commands in `tests/` only when the change affects Android, iOS, or macOS behavior.
+- Before merging substantial code changes, run the relevant root scripts from `package.json` for the touched area.
+
+## PR instructions
+
+- Keep pull requests scoped to one package or one repo concern where possible.
+- Use Conventional Commit style PR titles, following the examples in `CONTRIBUTING.md`.
+- When public APIs or observable behavior change, update tests, docs, and types in the same PR.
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -0,0 +1,30 @@
+# React Native Firebase AGENTS Guide
+
+## Project overview
+
+- This repository is a Yarn workspaces monorepo for React Native Firebase packages.
+- Most library code lives in `packages/*`.
```
mikehardy:
```suggestion
- Most library code lives in `packages/**/lib`.
```

AGENTS.md
```diff
@@ -0,0 +1,30 @@
...
+
+- Install dependencies once at the repository root with `yarn`.
+- Prefer working from the repository root so package scripts and workspace resolution behave consistently.
+- Check the affected package's `type-test.ts`, `__tests__/`, and plugin directories before changing public APIs or platform-specific behavior.
+- Prefer consistency with existing patterns in the package you are editing instead of introducing new structure.
+
+## Testing instructions
+
+- Start with focused validation for the area you changed instead of running the entire suite.
+- For code changes, run the relevant linting scripts (e.g., yarn lint:js, yarn lint:android, or yarn lint:ios:check).
```
mikehardy, line 20: I haven't had great luck with verification selection so am leaning towards generic verification scripts that are umbrellas for the sub-scripts, e.g. just `yarn lint` is the guidance vs trusting that the correct type of lint was selected and verified

AGENTS.md
```diff
@@ -0,0 +1,30 @@
...
+- This repository is a Yarn workspaces monorepo for React Native Firebase packages.
+- Most library code lives in `packages/*`.
+- The end-to-end test app lives in `tests/`.
+- Contributor expectations and review norms are documented in `CONTRIBUTING.md`.
+
+## Dev environment tips
+
+- Install dependencies once at the repository root with `yarn`.
+- Prefer working from the repository root so package scripts and workspace resolution behave consistently.
+- Check the affected package's `type-test.ts`, `__tests__/`, and plugin directories before changing public APIs or platform-specific behavior.
```
mikehardy, line 14: @russellwheatley maybe a good time to mention `yarn tsc:compile:consumer` (or the general idea of "consumer type test", or perhaps an umbrella type test run script that runs all type checks - consumer and non-consumer ?)

AGENTS.md
```diff
@@ -0,0 +1,30 @@
...
+- Prefer consistency with existing patterns in the package you are editing instead of introducing new structure.
+
+## Testing instructions
+
+- Start with focused validation for the area you changed instead of running the entire suite.
+- For code changes, run the relevant linting scripts (e.g., yarn lint:js, yarn lint:android, or yarn lint:ios:check).
+- For package behavior changes, run targeted Jest coverage with `yarn tests:jest <path-to-test-file>`.
+- For public API changes, update and run the relevant `packages/<name>/type-test.ts` checks and nearby Jest tests in `packages/<name>/__tests__/`.
+- Run platform-specific e2e commands in `tests/` only when the change affects Android, iOS, or macOS behavior.
+- Before merging substantial code changes, run the relevant root scripts from `package.json` for the touched area.
```
mikehardy, line 24: Maybe I'm paranoid but I have a low-level of trust in test selection / avoidance of unexpected issues in unrelated areas from code changes, so made a `yarn test:full` target I expect LLMs to use before they attempt to tell me they're "done"

https://github.com/invertase/react-native-firebase/blob/b6af4545232d010cdd4bc7dc335c58adb8935bd5/package.json#L25

It's heavy but if the code change and correctness verification is automated, I'd prefer the verification to be thorough at the cost of compute before consuming reviewer time (which will be the future bottleneck)

mikehardy (commented):
some thoughts on specific verification targets agents should use

AGENTS.md
russellwheatley: I think "consumer type test" should be the one. Non-consumer types, I think is covered when we compile each package, right? It wouldn't compile if there's an issue

AGENTS.md
```diff
@@ -0,0 +1,30 @@
...
+- Prefer working from the repository root so package scripts and workspace resolution behave consistently.
+- Check the affected package's `type-test.ts`, `__tests__/`, and plugin directories before changing public APIs or platform-specific behavior.
+- Prefer consistency with existing patterns in the package you are editing instead of introducing new structure.
+
+## Testing instructions
+
+- Start with focused validation for the area you changed instead of running the entire suite.
+- For code changes, run the relevant linting scripts (e.g., yarn lint:js, yarn lint:android, or yarn lint:ios:check).
+- For package behavior changes, run targeted Jest coverage with `yarn tests:jest <path-to-test-file>`.
+- For public API changes, update and run the relevant `packages/<name>/type-test.ts` checks and nearby Jest tests in `packages/<name>/__tests__/`.
```
russellwheatley, line 22: So this would have to change as well to use `yarn tsc:compile:consumer` as this checks the public types

russellwheatley (changes requested):

AGENTS.md
dackers86: @russellwheatley Could you propose a suggested change, not quite sure what this could include?

russellwheatley: In fact, `yarn tsc:compile:consumer` only runs on firestore at the moment, but will be updated in the future for all packages. So you can leave this for now and we can update it in the future.

russellwheatley (approved):

PR #9018, opened 2026-05-12
fix(firestore): serverTimestamps implement correct behavior for 'estimate' | 'previous' | 'none'

### LINKED ISSUE #9016: [🐛] withConverter.fromFirestore + snapshot.data({ serverTimestamps: 'estimate' }) returns null for pending serverTimestamp() fields
# `withConverter.fromFirestore` + `snapshot.data({ serverTimestamps: 'estimate' })` returns `null` for pending `serverTimestamp()` fields

## Issue

When a `FirestoreDataConverter.fromFirestore` callback calls `snapshot.data({ serverTimestamps: 'estimate' })`, fields written with `serverTimestamp()` come back as `null` for any emission where `metadata.hasPendingWrites === true` — even though `'estimate'` is supposed to substitute pending sentinels with a client-clock-estimated Timestamp.

Repro: https://github.com/harelw/rnfb-timestamp-issue-repro

## Versions

- `@react-native-firebase/app`: `24.0.0`
- `@react-native-firebase/firestore`: `24.0.0`
- `expo`: `~55.0.23`
- `react-native`: `0.83.6`
- iOS simulator (iPhone 17 Pro, iOS 26.x); same shape observed on Android emulator.
- Reproduced against the Firestore emulator (`firebase-tools` `14.27.0`), so no production project state involved.

## Setup

The repo is wired for the Firestore emulator. After `pnpm install && pnpm prebuild`:

```bash
pnpm emulators   # terminal 1
pnpm ios         # terminal 2
```

A real `GoogleService-Info.plist` is required just to clear Firebase iOS SDK's startup validation; all Firestore traffic from the running app goes to `localhost:8080`.

## What the test does

```ts
const ref = collection(db, 'rnfb_timestamp_repro').withConverter(sampleConverter);
const docRef = doc(ref);  // pre-generate client-side id

await new Promise<void>((resolve, reject) => {
  const unsub = onSnapshot(docRef, { includeMetadataChanges: true }, (snap) => {
    // snap.data() runs through fromFirestore which calls
    // snap.data({ serverTimestamps: 'estimate' })
    if (snap.exists() && !snap.metadata.hasPendingWrites) {
      unsub();
      resolve();
    }
  }, reject);

  // Fire-and-forget — observe local-write echo, then server-acked frame
  setDoc(docRef, {}).catch(reject);
});
```

`sampleConverter`'s `toFirestore` writes both `createdAt` and `updatedAt` as `serverTimestamp()`. Its `fromFirestore` reads `snap.data({ serverTimestamps: 'estimate' })`.

## Expected

Per [`SnapshotOptions.serverTimestamps`](https://firebase.google.com/docs/reference/js/firestore_.snapshotoptions), `'estimate'` should substitute pending serverTimestamps with the client clock. So every emission — including the `hasPendingWrites=true` ones — should carry `Timestamp(<estimated>)` for both fields.

## Observed

```
Emissions: 3

[0] exists=true  hasPendingWrites=true   fromCache=true
     createdAt: null
     updatedAt: null

[1] exists=true  hasPendingWrites=true   fromCache=false
     createdAt: null
     updatedAt: null

[2] exists=true  hasPendingWrites=false  fromCache=false
     createdAt: Timestamp(2026-05-11T19:22:38.776Z)
     updatedAt: Timestamp(2026-05-11T19:22:38.776Z)
```

Both pending emissions return `null` despite `'estimate'` being passed to `data()`. The settled emission resolves correctly. Two pending emissions, not just one — the bug persists even after the snapshot has been refreshed from network (`fromCache=false`), as long as `hasPendingWrites=true`.

## Why this matters

Any consumer that validates the converter output (Zod, runtime type assertion, downstream `.toMillis()`) crashes on the first emission of a freshly-written document. In our app, this surfaced as account-setup failure on first login after sign-up: an immediate read of a doc written with `serverTimestamp()` failed schema validation because both timestamp fields were `null`.

The documented `'estimate'` workaround is supposed to prevent this exact scenario but doesn't, at least under `withConverter`.

## Current workaround

Inside `fromFirestore`, check `snapshot.metadata.hasPendingWrites` and substitute the current `Date` for known-server-set fields when true. This duplicates what `'estimate'` is supposed to do natively, but it's the only way to keep a strict schema in front of the converter today.

russellwheatley:
### Description

Fixes #9016: `snapshot.data({ serverTimestamps: 'estimate' | 'previous' | 'none' })` (and `get()`) ignored the per-call option, so pending `serverTimestamp()` fields came back as `null` even with `'estimate'`.

- Wire `SnapshotOptions.serverTimestamps` end-to-end on Android, iOS, and the web/Lite shim so `data()` / `get()` honour `'estimate'`, `'previous'`, and `'none'`
- Preserve the pending-write timestamp variants across the native bridge, and only compute the extra maps when `hasPendingWrites` is true (settled docs don't need them)
- Throw a clear unsupported error for `onSnapshot` on the Lite SDK (macOS / Other), matching the rest of the lite surface
- Cover the option through converters, unit tests, and e2e (including `'previous'` on an updated doc, not just a brand-new one)

Last commit (`test(firestore): skip onSnapshot argument-validation e2e tests on lite SDK`): once `onSnapshot` throws on Other, the old argument-validation e2es on that platform started failing with "Not supported in the lite SDK" instead of the expected validation message and blew the Other job past its 20 minute timeout. Same skip pattern every sibling listener test already uses; Android/iOS still run the full validation coverage.

### Related issues

- Fixes #9016

### Release Summary

fix(firestore): honour `SnapshotOptions.serverTimestamps` (`estimate` / `previous` / `none`) on document snapshots, including under `withConverter`.

### Checklist

  - [x] Yes
  - [x] `Android`
  - [x] `iOS`
  - [x] `Other` (macOS, web)
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

- [x] `yarn jest packages/firestore`
- [x] CI green on Android, iOS, and Other


mikehardy:
I think this is just waiting for prioritization and should not be stale-closed

packages/firestore/__tests__/firestore.test.ts
```diff
@@ -357,6 +397,92 @@ describe('Firestore', function () {
     });
   });
 
+  describe('FirestoreDocumentSnapshot (unit)', function () {
+    it('data() respects SnapshotOptions.serverTimestamps', function () {
+      const firestore = getFirestore();
+      const snapshot = new FirestoreDocumentSnapshot(
+        // @ts-expect-error calling a private constructor directly which expects FirestoreInternal type
```
mikehardy, line 409: After analysis it seemed using internal APIs was unavoidable here at the unit test level so I left a comment in order to avoid the doubt of "why?" for future intrepid explorers

packages/firestore/e2e/withConverter.e2e.js
```diff
@@ -89,9 +92,12 @@ function withTestDb(fn) {
 async function withModifiedUndefinedPropertiesTestDb(fn) {
   const db = getFirestore();
   const previousValue = db._settings.ignoreUndefinedProperties;
-  initializeFirestore(db.app, { ignoreUndefinedProperties: false });
-  await fn(db);
-  initializeFirestore(db.app, { ignoreUndefinedProperties: previousValue });
+  db._settings.ignoreUndefinedProperties = false;
```
mikehardy: This was using internal APIs but public APIs could work here (whereas they could not in unit tests as noted above) - so I switched these to public APIs

mikehardy (approved):
changed a couple of minor testing things, nothing to do with the actual PR functionality

actual PR functionality is +1 - just waiting on CI for merge

PR #8933, opened 2026-03-17
feat(ios): add SPM dependency resolution support alongside CocoaPods

jsnavarroc:
## Summary

- Add dual SPM/CocoaPods dependency resolution for Firebase iOS SDK via a centralized `firebase_dependency()` helper (`packages/app/firebase_spm.rb`)
- When React Native >= 0.75 is detected, Firebase dependencies are resolved via Swift Package Manager (`spm_dependency`). For older versions or when explicitly disabled (`$RNFirebaseDisableSPM = true`), CocoaPods is used as fallback.
- Solves Xcode 26 compilation errors caused by explicit modules not finding Firebase internal modules (`FirebaseCoreInternal`, `FirebaseSharedSwift`) when using CocoaPods

## Changes

- **`packages/app/firebase_spm.rb`** — New helper with `firebase_dependency()` function that auto-detects SPM support
- **`packages/app/package.json`** — Added `firebaseSpmUrl` field as single source of truth
- **16 podspecs** — Updated to use `firebase_dependency()` instead of direct `s.dependency`
- **43 native iOS files** — Added `#if __has_include` guards for dual SPM/CocoaPods imports
- **CI matrix** — Extended E2E workflow with `dep-resolution: ['spm', 'cocoapods']` dimension (4 job combinations)
- **Unit tests** — Added Ruby Minitest suite for `firebase_spm.rb` logic
- **Documentation** — Added `docs/ios-spm.md` with architecture, integration guides, and troubleshooting

## How it works

```ruby
# Each podspec calls:
firebase_dependency(s, version, ['FirebaseAuth'], 'Firebase/Auth')

# Internally decides:
# - SPM path:      spm_dependency(spec, url: ..., products: ['FirebaseAuth'])
# - CocoaPods path: spec.dependency 'Firebase/Auth', version
```

**Decision logic:**
1. Is `spm_dependency()` defined? (RN >= 0.75 injects it) → YES → use SPM
2. Is `$RNFirebaseDisableSPM` set in Podfile? → YES → force CocoaPods
3. Neither available → fall back to CocoaPods

## User-facing configuration

**SPM mode (default for RN >= 0.75):**
```ruby
# ios/Podfile
linkage = 'dynamic'
use_frameworks! :linkage => linkage.to_sym
```

**CocoaPods mode (legacy/opt-out):**
```ruby
# ios/Podfile
$RNFirebaseDisableSPM = true
linkage = 'static'
use_frameworks! :linkage => linkage.to_sym
```

[...]

PR #9024, opened 2026-05-16
fix(ios): wire add_rncore_dependency + allow non-modular includes for RN 0.84+ prebuilt RNCore (#8883)

wneel:
### Description

Closes the producer-side half of #8883 in two podspec changes:

**1. `add_rncore_dependency(s)` in every RNFB podspec**, guarded by `defined?` so older RN keeps working. This is React Native's own helper from `react_native_pods.rb`: when `RCT_USE_PREBUILT_RNCORE=1` (the default on 0.84+), it adds `React-Core-prebuilt` as a dependency and wires `$(PODS_ROOT)/React-Core-prebuilt/React.xcframework/Headers` into `HEADER_SEARCH_PATHS`. Without this, RNFB pods can't resolve `<React/RCTConvert.h>` and friends under prebuilt RNCore. The helper is opt-in per podspec and RNFB never opted in.

**2. `CLANG_ALLOW_NON_MODULAR_INCLUDES_IN_FRAMEWORK_MODULES = YES`** in every RNFB pod's `pod_target_xcconfig`. RNFBApp's umbrella header re-exports `<React/...>` imports through its public `.h` files. Without this flag, RNFBApp's own framework module compile trips `-Werror,-Wnon-modular-include-in-framework-module` under `use_frameworks!` (Expo's default and any project using static framework linkage).

Together these unblock `pod install` and the producer-side RNFB framework compile under stock RN 0.84 defaults (prebuilt RNCore on) for both bare-RN and `use_frameworks!: static` projects, with no Podfile workarounds: no `$RNFirebaseAsStaticFramework`, no `RCT_USE_PREBUILT_RNCORE=0`, no preinstall hook.

**Honest scope:** this PR does NOT yet close the consumer-side chain. When downstream RNFB pods (Storage, Auth, Messaging, etc.) `@import RNFBApp` as a Clang module, RNFBApp's umbrella still re-exports `<React/...>` imports, and Clang module visibility breaks `RCT_EXPORT_METHOD` expansion and `RCTBridgeModule` protocol visibility on the consuming side. Under `use_frameworks!` Expo today, even with this PR applied, users still need `RCT_USE_PREBUILT_RNCORE=0` or `$RNFirebaseAsStaticFramework=true` as a workaround until follow-up work lands. Details and an attempted source-level fix are discussed in the issue thread.

### Related issues

Fixes (partial, producer-side) #8883.

### Release Summary

iOS: RNFB podspecs now call `add_rncore_dependency` and allow non-modular includes in framework modules, so the producer-side RNFB build works on RN 0.84+ with prebuilt RNCore enabled (the default) on both bare-RN and `use_frameworks!` projects. A consumer-side architectural follow-up is needed for full `use_frameworks!` support.

### Checklist

  - [x] Yes
  - [ ] `Android`
  - [x] `iOS`
  - [ ] `Other` (macOS, web)
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

Verified against my app (Animalert Debug && Release, production RN 0.84.1 app, 8 RNFB packages: app, analytics, app-check, auth, crashlytics, firestore, messaging, storage). Setup: `use_frameworks! :linkage => :static`, Xcode 26.5, iOS 26.5 Simulator (iPhone 17 Pro), physical iPhone 13 on iOS 16.2 aswell.

**With this PR's patched podspecs overlaid into `node_modules/@react-native-firebase/*/`:**

- `pod install` completes cleanly. Generated `Pods/Target Support Files/RNFB*/RNFB*.debug.xcconfig` files for all 8 packages contain `${PODS_XCFRAMEWORKS_BUILD_DIR}/React-Core-prebuilt` and `${PODS_XCFRAMEWORKS_BUILD_DIR}/ReactNativeDependencies` in `FRAMEWORK_SEARCH_PATHS`, plus `${PODS_ROOT}/Headers/Public/React-Core-prebuilt` in `HEADER_SEARCH_PATHS`, and `CLANG_ALLOW_NON_MODULAR_INCLUDES_IN_FRAMEWORK_MODULES = YES`.
- Direct compile of `RNFBApp` via `xcodebuild -project Pods/Pods.xcodeproj -target RNFBApp -sdk iphonesimulator26.5 -arch arm64 build` succeeds in Debug and Release. `RCTConvert+FIRApp.m`, `RCTConvert+FIROptions.m`, `RNFBAppModule.m`, `RNFBRCTEventEmitter.m`, `RNFBSharedUtils.m`, `RNFBUtilsModule.m` all compile (these are the files that `#import <React/RCTConvert.h>`, `<React/RCTBridgeModule.h>`, `<React/RCTEventEmitter.h>`, the exact failure mode of #8883).
- Zero `-Wnon-modular-include-in-framework-module` errors or warnings in either Debug or Release build logs.

**Limitations of testing:**

- Producer-side only. Workspace-level `xcodebuild -scheme Animalert build` against the full app fails further along at the consumer-side chain documented in the issue thread. Workspace builds succeed when `$RNFirebaseAsStaticFramework = true` is re-enabled, confirming the workaround still works and what this PR doesn't yet remove.
- Not tested with a vanilla (no `use_frameworks!`) RN 0.84 app, but that's the simpler case for this PR's scope and the `add_rncore_dependency` helper is documented for it.

CI should validate the full matrix.

---

### DIFF
--- a/packages/analytics/RNFBAnalytics.podspec
+++ b/packages/analytics/RNFBAnalytics.podspec
@@ -39,6 +39,12 @@ Pod::Spec.new do |s|
     s.dependency "React-Core"
   end
 
+  # Wire up prebuilt React-Core (RN 0.83+, default on 0.84+) so the legacy
+  # <React/...> header imports resolve when RCT_USE_PREBUILT_RNCORE=1.
+  if defined?(add_rncore_dependency)
+    add_rncore_dependency(s)
+  end
+
   if defined?($FirebaseSDKVersion)
     Pod::UI.puts "#{s.name}: Using user specified Firebase SDK version '#{$FirebaseSDKVersion}'"
     firebase_sdk_version = $FirebaseSDKVersion
--- a/packages/analytics/RNFBAnalytics.podspec
+++ b/packages/analytics/RNFBAnalytics.podspec
@@ -45,6 +45,12 @@ Pod::Spec.new do |s|
     add_rncore_dependency(s)
   end
 
+  # RNFB public headers re-export non-modular <React/...> imports. Required so
+  # the framework module validates when consumers build with use_frameworks!.
+  s.pod_target_xcconfig = {
+    "CLANG_ALLOW_NON_MODULAR_INCLUDES_IN_FRAMEWORK_MODULES" => "YES",
+  }
+
   if defined?($FirebaseSDKVersion)
     Pod::UI.puts "#{s.name}: Using user specified Firebase SDK version '#{$FirebaseSDKVersion}'"
     firebase_sdk_version = $FirebaseSDKVersion
--- a/packages/analytics/ios/RNFBAnalytics/RNFBAnalyticsModule.m
+++ b/packages/analytics/ios/RNFBAnalytics/RNFBAnalyticsModule.m
@@ -16,6 +16,7 @@
  */
 
 #import <Firebase/Firebase.h>
+#import <React/RCTBridgeModule.h>
 #import <React/RCTUtils.h>
 
 #if __has_include(<RNFBAnalytics/RNFBAnalytics-Swift.h>)
--- a/packages/app-check/RNFBAppCheck.podspec
+++ b/packages/app-check/RNFBAppCheck.podspec
@@ -38,6 +38,12 @@ Pod::Spec.new do |s|
     s.dependency "React-Core"
   end
 
+  # Wire up prebuilt React-Core (RN 0.83+, default on 0.84+) so the legacy
+  # <React/...> header imports resolve when RCT_USE_PREBUILT_RNCORE=1.
+  if defined?(add_rncore_dependency)
+    add_rncore_dependency(s)
+  end
+
   if defined?($FirebaseSDKVersion)
     Pod::UI.puts "#{s.name}: Using user specified Firebase SDK version '#{$FirebaseSDKVersion}'"
     firebase_sdk_version = $FirebaseSDKVersion
--- a/packages/app-check/RNFBAppCheck.podspec
+++ b/packages/app-check/RNFBAppCheck.podspec
@@ -44,6 +44,12 @@ Pod::Spec.new do |s|
     add_rncore_dependency(s)
   end
 
+  # RNFB public headers re-export non-modular <React/...> imports. Required so
+  # the framework module validates when consumers build with use_frameworks!.
+  s.pod_target_xcconfig = {
+    "CLANG_ALLOW_NON_MODULAR_INCLUDES_IN_FRAMEWORK_MODULES" => "YES",
+  }
+
   if defined?($FirebaseSDKVersion)
     Pod::UI.puts "#{s.name}: Using user specified Firebase SDK version '#{$FirebaseSDKVersion}'"
     firebase_sdk_version = $FirebaseSDKVersion
--- a/packages/app-check/ios/RNFBAppCheck/RNFBAppCheckModule.m
+++ b/packages/app-check/ios/RNFBAppCheck/RNFBAppCheckModule.m
@@ -17,6 +17,7 @@
 
 #import <Firebase/Firebase.h>
 #import <FirebaseAppCheck/FIRAppCheck.h>
+#import <React/RCTBridgeModule.h>
 
 #import <React/RCTUtils.h>
 
--- a/packages/app-distribution/RNFBAppDistribution.podspec
+++ b/packages/app-distribution/RNFBAppDistribution.podspec
@@ -36,6 +36,12 @@ Pod::Spec.new do |s|
     s.dependency "React-Core"
   end
 
+  # Wire up prebuilt React-Core (RN 0.83+, default on 0.84+) so the legacy
+  # <React/...> header imports resolve when RCT_USE_PREBUILT_RNCORE=1.
+  if defined?(add_rncore_dependency)
+    add_rncore_dependency(s)
+  end
+
   if defined?($FirebaseSDKVersion)
     Pod::UI.puts "#{s.name}: Using user specified Firebase SDK version '#{$FirebaseSDKVersion}'"
     firebase_sdk_version = $FirebaseSDKVersion
--- a/packages/app-distribution/RNFBAppDistribution.podspec
+++ b/packages/app-distribution/RNFBAppDistribution.podspec
@@ -42,6 +42,12 @@ Pod::Spec.new do |s|
     add_rncore_dependency(s)
   end
 
+  # RNFB public headers re-export non-modular <React/...> imports. Required so
+  # the framework module validates when consumers build with use_frameworks!.
+  s.pod_target_xcconfig = {
+    "CLANG_ALLOW_NON_MODULAR_INCLUDES_IN_FRAMEWORK_MODULES" => "YES",
+  }
+
   if defined?($FirebaseSDKVersion)
     Pod::UI.puts "#{s.name}: Using user specified Firebase SDK version '#{$FirebaseSDKVersion}'"
     firebase_sdk_version = $FirebaseSDKVersion
--- a/packages/app-distribution/ios/RNFBAppDistribution/RNFBAppDistributionModule.m
+++ b/packages/app-distribution/ios/RNFBAppDistribution/RNFBAppDistributionModule.m
@@ -16,6 +16,7 @@
  */
 
 #import <Firebase/Firebase.h>
+#import <React/RCTBridgeModule.h>
 #import <React/RCTUtils.h>
 
 #import "RNFBApp/RNFBSharedUtils.h"
--- a/packages/app/RNFBApp.podspec
+++ b/packages/app/RNFBApp.podspec
@@ -50,6 +50,12 @@ Pod::Spec.new do |s|
     s.dependency "React-Core"
   end
 
+  # Wire up prebuilt React-Core (RN 0.83+, default on 0.84+) so the legacy
+  # <React/...> header imports resolve when RCT_USE_PREBUILT_RNCORE=1.
+  if defined?(add_rncore_dependency)
+    add_rncore_dependency(s)
+  end
+
   if (ENV.include?('FIREBASE_SDK_VERSION'))
     Pod::UI.puts "#{s.name}: Found Firebase SDK version in environment '#{ENV['FIREBASE_SDK_VERSION']}'"
     $FirebaseSDKVersion = ENV['FIREBASE_SDK_VERSION']
--- a/packages/app/RNFBApp.podspec
+++ b/packages/app/RNFBApp.podspec
@@ -38,9 +38,13 @@ Pod::Spec.new do |s|
     Pod::UI.puts '[react-native-firebase] '.yellow + "Suppress this with environment variable RNFB_SUPPRESS_NEW_ARCHITECTURE_WARNING=1"
   end
 
-  # App must define modules for static framework integration of other packages to work
+  # App must define modules for static framework integration of other packages to work.
+  # CLANG_ALLOW_NON_MODULAR_INCLUDES_IN_FRAMEWORK_MODULES is required so RNFB's umbrella,
+  # which re-exports <React/...> imports, validates when consumers build with use_frameworks!
+  # (the default on Expo, and required for the firebase-ios-sdk).
   s.pod_target_xcconfig = {
     "DEFINES_MODULE" => "YES",
+    "CLANG_ALLOW_NON_MODULAR_INCLUDES_IN_FRAMEWORK_MODULES" => "YES",
   }
 
   # React Native dependencies
--- a/packages/app/ios/RNFBApp/RNFBAppModule.m
+++ b/packages/app/ios/RNFBApp/RNFBAppModule.m
@@ -16,6 +16,7 @@
  */
 
 #import <Firebase/Firebase.h>
+#import <React/RCTBridgeModule.h>
 #import <React/RCTUtils.h>
 
 #import "RNFBAppModule.h"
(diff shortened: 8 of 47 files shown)


wneel:
Update from the deep dive on the consumer-side caveat.

The mechanic, now isolated: under `use_frameworks!` Clang auto-modularizes `<React/...>` framework imports (even with quote-style or an empty/textual-header modulemap), and Clang modules don't propagate `#define` macros across boundaries. So `RCT_EXTERN` and `RCT_EXPORT_METHOD` go undefined when consumer `.m` files reach them.

I tried everything project-side (custom modulemaps with `textual header`, stripping `React-Core-prebuilt` from FRAMEWORK_SEARCH_PATHS, removing the prebuilt `Modules/` dir, source-patching node_modules React-Core to quote-style internal imports, `-DRCT_EXTERN=extern` predefines, `-fno-implicit-module-maps`). Clang's framework-module detection re-asserts every time. The true fix has to be upstream in React Native and I'm opening a discussion at facebook/react-native#56862 to flag it.

Side-finding worth a separate PR: 22 RNFB `.m` files using `RCT_EXPORT_*` macros never directly `#import <React/RCTBridgeModule.h>` and rely on transitive includes that are getting unreliable on newer RN (#9026). Overlaps a little with the +12-line diff you mentioned but applied broadly to all packages instead of just firestore.

Scope of this PR unchanged.

wneel:
Thanks @gemini-code-assist for the review.

Applied suggestion 1 (idiomatic `defined?` check) in 5dc5fc61c. `defined?(add_rncore_dependency()) != nil` → `defined?(add_rncore_dependency)` across all 17 podspecs and the template.

Deferring suggestion 2 (centralizing duplicated config into a shared helper) to a follow-up. RNFB doesn't currently have a shared-helper file pattern for podspecs, so introducing one expands scope beyond what this PR aims to fix. Happy to send it as a separate refactor once the maintainer signals appetite for the broader cleanup.

mikehardy:
prior to personal review, what I don't understand is why this is a problem when this appears to work (in expo build properties, at least - rn cli case still needs a podfile post-install to effectively do the same)

that is: if we have something that works cleanly with Expo config (and perhaps a post_install), what *is* all of this build churn around header reordering and module maps and on and on ?

```json
[
  "expo-build-properties",
  {
    "ios": {
      "useFrameworks": "static",
      "forceStaticLinking": [
        "RNFBAnalytics",
        "RNFBApp",
        "RNFBAppCheck",
        "RNFBAuth",
        "RNFBCrashlytics",
        "RNFBFirestore",
        "RNFBMessaging",
        "RNFBRemoteConfig",
        "RNFBStorage"
      ]
    }
  }
]
```

wneel:
I'm not 100% sure your snippet alone is enough, because the workaround I actually verified on my app (no expo) needed two things together, not one:

```ruby
ENV['RCT_USE_PREBUILT_RNCORE'] = '0'
```

`forceStaticLinking` is the Expo equivalent of the first line. It flips RNFB pods from framework modules to static libs, which avoids the modulemap evaluation path where `<React/RCTBridgeModule.h>` auto-modularizes and the `RCT_EXPORT_*` macros get lost across the module boundary. But in my testing, that alone wasn't enough on RN 0.84+ with the new default `RCT_USE_PREBUILT_RNCORE=1`. I also had to opt back to the source React-Core to get clean builds.

So before I say anything stronger, a genuine question: when your Expo config builds cleanly, is `RCT_USE_PREBUILT_RNCORE` actually `1` at compile time? If Expo's prebuild flow is leaving it at `0` (or `forceStaticLinking` is somehow steering header resolution to the source symlinks), that would explain the difference: the macro bug only fires when the prebuilt `React.xcframework` modulemap is the one Clang loads. The repro at wneel/rnfbdemo hits it reliably on RN 0.85.3 + RNFB 24.0.0 with the 0.84+ default.

Independent of that: #9024 stands on its own. `add_rncore_dependency` is the API RN added specifically for podspecs consuming prebuilt RNCore (RNFB doesn't call it today on 0.84+), and the CLANG_ALLOW xcconfig clears a warning every clean `use_frameworks!` build emits. The upstream RN work in #56862 is what would let users drop the workaround entirely; this PR is useful regardless of whether that lands I guess.

wneel:
> [!NOTE]
> Quick update for the thread: the upstream RN PR (facebook/react-native#56862) was merged today at `90c303fa`. So the consumer-side macro propagation problem will be fixed in a future RN release. That makes the producer-side cleanup in this PR (`add_rncore_dependency` + `CLANG_ALLOW_NON_MODULAR_INCLUDES_IN_FRAMEWORK_MODULES`) the remaining piece for clean `use_frameworks!` + prebuilt RNCore builds. Independent of the workaround discussion, happy to address any further review feedback whenever you have cycles.

mikehardy:
> I'm not 100% sure your snippet alone is enough

I don't make the assertion lightly, it is demonstrably enough, at least in certain situations:

https://github.com/mikehardy/rnfbdemo/blob/53af017e81f78eb503a171db4238476828ed16f3/make-expo-demo.sh#L105-L114

https://github.com/mikehardy/rnfbdemo/blob/53af017e81f78eb503a171db4238476828ed16f3/make-expo-demo.sh#L152-L165

That is a from-scratch build test harness and works 100% of the time

So if that is not sufficient in all cases, the critical question is: what is missing that would demonstrate a failure where it is not sufficient?

Otherwise you can understand why I'm paying a lot of attention but I'm also skeptical, i.e., what *exactly* is going on and what *minimal* thing do we need to do to smooth this rough edge out ? It may be everything you've proposed is minimal but I'm still curious then to demonstrate why by making some addition to the build test - otherwise we may regress in the future

mikehardy:
(after just a bit of reflection: perhaps the main benefit here is that after this PR, `use_frameworks!` need to be toggled on but all the `forceStaticLinking` expo-build-properties jank may be removed? likewise with RN CLI builds no need for a post_install that has the same effect? If so, that's obviously an improvement and sufficient motivation for this PR...)

wneel:
> perhaps the main benefit here is that after this PR, `use_frameworks!` need to be toggled on but all the `forceStaticLinking` expo-build-properties jank may be removed?

Yes, that's the right read, and I think that framing is cleaner than anything I wrote earlier.

To be precise about the sequencing: this PR (producer side) + facebook/react-native#56862 (consumer side, now merged) together are what close the gap. This PR alone doesn't eliminate the workaround for users on RN versions that predate 56862 landing in a release. But once a RN release ships with that fix, RNFB podspecs with `add_rncore_dependency` should let users run `use_frameworks!` without `forceStaticLinking` or a `post_install` workaround.

Your build test harness is a good fit for validating that: once there's an RN release containing 56862, a variant of `make-expo-demo.sh` without the `forceStaticLinking` array (or the equivalent RN CLI test without the `$RNFirebaseAsStaticFramework` post_install) and just `use_frameworks!` should build clean. If it does, that's the regression guard you were asking about, and it demonstrates the before/after difference concretely.

I take your point that the current workaround demonstrably works, and I should have framed the motivation around the end state (workaround removal) earlier rather than the failure mode.

I filled the pick requests already on react-native and I'll update you when they get approved :)

mikehardy:
okay - glad I appear to understand everything then - that's job 1 of course - removing that jank is sufficiently motivating, this is great then, and I noticed those upstream pick requests for previous stable after your merge, really well done all around on this, I appreciate it. Will work to shepherd this into merge+release here

mikehardy:
merged the direct include work from #9026 over here so there was just one commit doing the same thing
rebased onto main and squashed commits here to put review comment handling directly into the original commit that spawned the comment
I've rerun validation locally (passes)
this should be good to go pending CI

- [ ] rnfbdemo make-demo.sh with rn0.85.3 and this PR
- [ ] rnfbdemo make-demo.sh with rn0.86.0-rc.3 and this PR
- [ ] rnfbdemo make-demo.sh with rn0.86.0-rc.3 and this PR and removing the build workarounds
- [ ] rnfbdemo make-expo-demo.sh with expo 56 and this PR

wneel:
Thanks @mikehardy, glad you squashed the review handling directly into the original commits. And good call absorbing #9026 here rather than landing two PRs for the same area.

Let CI run and I'll keep an eye on the checklist :) I'm here if anything comes up.

wneel:
On the header order thing, your read was actually right. In the upstream RN commit [ff2562e6](https://github.com/react/react-native/commit/ff2562e6e170bab24c7ed55869a418e576eb70de) they hit the exact same issue and ended up with the same pragma to lock RCTDefines.h in first. The clang-format off/on comments in there weren't mine by the way, Meta added those during landing.

wneel:
Good news from the trenches: I just got a real production app (RN 0.86.0, new arch, `use_frameworks! :linkage => :static`, `$RNFirebaseAsStaticFramework = true`, prebuilt RNCore at its default, Xcode 26.5) building AND running with this PR. App boots, Firestore queries stream, messaging initializes. So the combination works end to end on 0.86, which I believe maps to items 2 and 3 of your checklist. I am genuinely happy about this one :)

It needed one more producer-side commit on top of your squash though, and you will likely hit it when you run rnfbdemo on 0.86, so here is the heads-up.

**The failure.** With the PR as-is, RNFBMessaging fails to compile under `use_frameworks!`:

```
declaration of 'RCTConvert' must be imported from module 'RNFBApp.RCTConvert_FIRApp' before it is required
declaration of 'RCTPromiseRejectBlock' must be imported from module 'RNFBApp.RNFBAppModule' before it is required
RCT_EXPORT_MODULE(); ... error: type specifier missing, defaults to 'int'
```

**The mechanism.** `CLANG_ALLOW_NON_MODULAR_INCLUDES_IN_FRAMEWORK_MODULES` (which this PR adds, and which we need) has a side effect: when clang builds the RNFBApp framework module, the React headers that RNFBApp's public headers include get absorbed into per-header submodules. `RCTConvert.h` lands in `RNFBApp.RCTConvert_FIRApp`, `RCTBridgeModule.h` in `RNFBApp.RNFBAppModule`. In a consumer .m, importing any `<RNFBApp/...>` header loads only that one submodule, after which clang considers the React declarations owned by hidden sibling submodules. A later textual `#import <React/RCTConvert.h>` can no longer make them visible, and the `RCT_EXPORT_*` macros vanish with it.

The trigger is import order, and the clang-format alphabetical sort works against us here: `RNFBApp` sorts before `React`, so after the #9026 merge every affected file imports `<RNFBApp/...>` first. This is the same family of bug as the umbrella ordering we fixed upstream in facebook/react-native#56862.

**The fix.** One commit, 13 files (messaging and firestore translation units that import RNFBApp headers): import the React headers the TU needs before any RNFBApp header, wrapped in `// clang-format off/on` with a comment explaining why, mirroring what Meta did for `RCTDefines.h` in ff2562e6. The other packages already import React first or do not use React symbols, I audited all of them.

**Validation.** Two ways. On this branch's HEAD plus the fix: clean build from wiped DerivedData, zero errors across all 88 RNFB ObjC files. And cherry-picked (the three PR commits plus the fix, all apply cleanly) onto v24.1.1, which is what my app actually consumes: rebuild, install, run on the iOS 26.5 simulator, live Firestore traffic on screen.

Pushed here as dbb11cda4. It fast-forwards on top of your squash, so your history is untouched. Happy to adjust anything when you run the checklist.

mikehardy:
Interesting - unfortunately I ran out of timeblock after suffering some sort of local environmental issue with rnfbdemo and what I had as the full stack of changes, and then also when attempting to baseline my setup *without* the changes - meaning I lost confidence in any reproduction until I could pick it up again. I'll be retrying again shortly though - and as soon as we get these iOS build items sorted it'll be time to start fresh with iOS build items to convert to SPM 😆

wneel:
Hi @mikehardy good news !

I ran a Claude agent in auto mode to drive the debugging, all findings below are from that run and I reviewed them.

On your **current** main ([mikehardy/rnfbdemo@53af017](https://github.com/mikehardy/rnfbdemo/commit/53af017e81f78eb503a171db4238476828ed16f3)), with RN 0.86.0 stable (npm `latest`) + RNFB 24.1.1 and the `RCT_USE_PREBUILT_RNCORE=0` / `RCT_USE_RN_DEP=0` workarounds removed (the "removing the build workarounds" checklist row):

- **Without this PR**: FAILED. `RNFBFiamModule.h:22: declaration of 'RCTBridgeModule' must be imported from module 'RNFBApp.RNFBAppModule' before it is required`, then the `RCT_EXPORT_*` cascade.
- **With this PR's head** overlaid + clean `pod install`: **BUILD SUCCEEDED**.

Setup, exact commands and logs: [VALIDATION.md](https://github.com/wneel/rnfbdemo/blob/validate/rn086-rnfb2411-pr9024/VALIDATION.md) on the [`validate/rn086-rnfb2411-pr9024`](https://github.com/wneel/rnfbdemo/tree/validate/rn086-rnfb2411-pr9024) branch of [wneel/rnfbdemo](https://github.com/wneel/rnfbdemo).

One thing possibly related to your lost timeblock: make-demo.sh deletes the Gemfile, then runs `cd ios && bundle exec pod install && cd ..`. If bundler finds no Gemfile up-tree it fails, `set -e` does not catch mid-list failures, and the script keeps running with cwd stranded in `ios/`, so everything downstream fails with confusing "No such file" errors. The branch swaps in the `npx pod-install` variant you had commented next to it.

Happy to detail any of this further, just ask.

mikehardy:
Just an update on this - we know this is a big pain point for the repository but we're working on SPM support with all available capacity at the moment:

- #8933 

...following our TurboModules migration (which also touched all the Apple files...) we have to land these one at a time and the all conflict but LLMs make commit fixups easy at least.

We will get to this though

russellwheatley:
Thanks for this 🙏

#9200 was the reimplementation of this: Clang flag on `pod_target_xcconfig` before `install_modules_dependencies`, and no extra `add_rncore_dependency` on live pods.

Closing as superseded.

Issue #8988, opened 2026-04-21
🔥 [🐛] iOS build fails: cannot find 'RCTPromiseRejectBlock' in scope in RNFBFirestoreCommon.h (v24 + Expo SDK 55 + use_frameworks!:static)


cathandnya:
## Issue

On iOS, building an Expo SDK 55 / React Native 0.83.4 app that uses `@react-native-firebase/firestore@24.0.0` with `use_frameworks!: :static` fails with:

```
node_modules/@react-native-firebase/firestore/ios/RNFBFirestore/RNFBFirestoreCommon.h:40:42:
  error: cannot find 'RCTPromiseRejectBlock' in scope
+ (void)promiseRejectFirestoreException:(RCTPromiseRejectBlock)reject error:(NSError *)error;
                                         ^
```

Similar errors are also raised in `RNFBFirestoreCollectionModule.h` and the other Firestore headers that expose methods taking `RCTPromiseRejectBlock` / `RCTResponseSenderBlock`.

### Root cause

Two separate issues in the shipped iOS sources of `@react-native-firebase/firestore@24.0.0`:

1. **Header import order.** In every `.h` under `ios/RNFBFirestore/`, `#import <Firebase/Firebase.h>` is placed *before* `#import <React/RCTBridgeModule.h>`:

    ```objc
    // RNFBFirestoreCommon.h (current)
    #import <Firebase/Firebase.h>
    #import <React/RCTBridgeModule.h>
    ```

    With Expo's default `use_frameworks!: :static` + Xcode 26, resolving Firebase first seems to prevent `React/RCTBridgeModule.h` from contributing its block typedefs into scope, so `RCTPromiseRejectBlock` / `RCTResponseSenderBlock` are not visible to later declarations in the same file.

2. **Missing import in `.m` sources.** Several implementation files call `RCTPromiseRejectBlock` / `RCTResponseSenderBlock` but never import `<React/RCTBridgeModule.h>` directly — they were relying on the header chain picking it up. Affected files:

    - `RNFBFirestoreCollectionModule.m`
    - `RNFBFirestoreCommon.m`
    - `RNFBFirestoreDocumentModule.m`
    - `RNFBFirestoreModule.m`
    - `RNFBFirestoreTransactionModule.m`

### Why this is not #8883

[...]

PR #9026, opened 2026-05-16
fix(ios): import RCTBridgeModule.h directly in .m files using RCT_EXPORT_* macros

wneel:
### Description

22 RNFB `.m` files use `RCT_EXPORT_METHOD`, `RCT_EXPORT_MODULE`, or `RCT_REMAP_METHOD` but never directly `#import <React/RCTBridgeModule.h>`. They've been picking up the macros transitively through the package's own `.h` files. That chain is becoming unreliable under newer RN + `use_frameworks!` because Clang modules scope `#define`s and don't propagate them across boundaries.

This PR adds the explicit `#import <React/RCTBridgeModule.h>` at the top of each affected `.m`. One line per file, pure hygiene, no behavior change. Makes the macros reliably available regardless of how transitive includes resolve, now or in the future.

Related to #8988 (firestore-specific symptom) and complements #8883. Applied uniformly across all RNFB packages instead of just firestore. Happy to defer to a narrower scope if maintainers prefer.

### Related issues

Helps #8883 (consumer-side chain). Related to #8988.

### Release Summary

iOS: every RNFB `.m` file using `RCT_EXPORT_*` macros now directly imports `<React/RCTBridgeModule.h>`. Removes a brittle transitive-include dependency.

### Checklist

  - [x] Yes
  - [ ] `Android`
  - [x] `iOS`
  - [ ] `Other` (macOS, web)
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

Verified on Animalert (RN 0.84.1, 8 RNFB packages). Setup: `use_frameworks! :linkage => :static`, `$RNFirebaseAsStaticFramework = true`, `ENV['RCT_USE_PREBUILT_RNCORE'] = '0'` (the documented workaround until #8883's consumer side closes). Xcode 26.5. Tested on iPhone 17 Pro iOS 26.5 simulator and iPhone 13 physical device.

Before: Debug fails at every RNFB module's `RCT_EXPORT_METHOD` with cascading parse errors.

After: clean Debug and Release builds of the full Animalert workspace, `.app` bundles produced for both configurations.

---

### DIFF
--- a/packages/analytics/ios/RNFBAnalytics/RNFBAnalyticsModule.m
+++ b/packages/analytics/ios/RNFBAnalytics/RNFBAnalyticsModule.m
@@ -16,6 +16,7 @@
  */
 
 #import <Firebase/Firebase.h>
+#import <React/RCTBridgeModule.h>
 #import <React/RCTUtils.h>
 
 #if __has_include(<RNFBAnalytics/RNFBAnalytics-Swift.h>)
--- a/packages/app-check/ios/RNFBAppCheck/RNFBAppCheckModule.m
+++ b/packages/app-check/ios/RNFBAppCheck/RNFBAppCheckModule.m
@@ -17,6 +17,7 @@
 
 #import <Firebase/Firebase.h>
 #import <FirebaseAppCheck/FIRAppCheck.h>
+#import <React/RCTBridgeModule.h>
 
 #import <React/RCTUtils.h>
 
--- a/packages/app-distribution/ios/RNFBAppDistribution/RNFBAppDistributionModule.m
+++ b/packages/app-distribution/ios/RNFBAppDistribution/RNFBAppDistributionModule.m
@@ -16,6 +16,7 @@
  */
 
 #import <Firebase/Firebase.h>
+#import <React/RCTBridgeModule.h>
 #import <React/RCTUtils.h>
 
 #import "RNFBApp/RNFBSharedUtils.h"
--- a/packages/app/ios/RNFBApp/RNFBAppModule.m
+++ b/packages/app/ios/RNFBApp/RNFBAppModule.m
@@ -16,6 +16,7 @@
  */
 
 #import <Firebase/Firebase.h>
+#import <React/RCTBridgeModule.h>
 #import <React/RCTUtils.h>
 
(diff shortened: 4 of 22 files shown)


wneel:
Thanks @gemini-code-assist for the review.

On the suggestions:

1. **Grouping React framework headers together**: noted. Kept the addition as a single line at the top of the existing import block to keep the diff minimal and uniformly reviewable across 22 files (one-line add per file). Happy to reorganize if maintainers prefer the grouped style.

2. **Primary module headers at top (Obj-C convention)**: out of scope here. The existing files don't follow that convention either, so reordering would inflate this PR with unrelated style churn. Better suited to a dedicated formatting pass on RNFB's iOS sources if there's appetite.

mikehardy (approved):
I thought the analysis was great, provided sufficient motivation for the fix and the fix was targeted well

Unfortunately in local validation run it appeared the real CI problem was not an android lint failure (I think that was transient / unrelated? unsure) but a clang failure - that was real and just masked by the android failure since clang runs after

the problem is that clang-format with the Google format standard expects alphabetic imports, but "fixing" that reverts the explicit ordering that *is* the fix

So, we could make a custom clang-format file that allows this specific alphabetic mis-ordering or we could put a pragma in. I opted for the pragma.

Rebased to current main. Did the pragma fixup. It passes locally. Pushed, hopefully goes through CI

Truly appreciate the work here - this header expansion think is a really subtle problem

mikehardy:
I re-thought my analysis on location of the header - I was conflating the header order issue here with the upstream RN PR you made - had that stuck in my head where you needed the defines to be prior to use, but that's not in play here

The clang-format thing was bugging me so I removed my lint fix commit, ran clang-format on your original commit during a rebase edit, and pushed that. Much cleaner

mikehardy:
...final evolution of this as I'm working #9024 in parallel - that one includes a commit that does similar to what is happening here, best to combine them both. I've pulled the changes here into the commit there and will close this in preference of 9024 - that one's in final validation as well

wneel:
Makes sense, thanks for pulling it all together. I'll close this one out. All further discussion in #9024 :)

PR #9031, opened 2026-05-26
docs: update generated reference styling

Ehesp:
Add custom css overrides to the typedoc generated output to align with new docs

### DIFF
--- a/docs/reference.css
+++ b/docs/reference.css
@@ -1,0 +1,104 @@
+:root {
+  --light-color-background: lab(100% 0 0);
+  --light-color-background-secondary: lab(100% 0 0);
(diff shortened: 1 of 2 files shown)


mikehardy (approved):
LGTM functionality wise and I trust your CSS a precise 100.000% more than mine so if you're happy with the CSS that works for me.

Typedoc publishes on release so it matches shipped code, and we're due for one shortly so we'll get to see this live soon but not immediately on this merge

mikehardy:
close/open cycle to restart actions after github actions outage

mikehardy (approved):
no idea what's going on with the CI that won't fix itself but I pulled this locally and verified (`yarn reference:api`) that it looks good and the links will work. It is a real visual improvement, merging

PR #9045, opened 2026-06-10
feat(android): add support for Firebase Phone Number Verification (PNV)

### LINKED ISSUE #9013: Support for Firebase Phone Number Verification as it is generally available now
Firebase PNV is generally available now.

https://firebase.blog/posts/2026/05/firebase-pnv-generally-available/

When will it be available in the Auth Firebase RN SDK?

Thanks.

just1and0:
### Description

Adds a new `@react-native-firebase/phone-number-verification` package implementing the Firebase Phone Number Verification (PNV) SDK, which is now Generally Available.

PNV enables carrier-level phone number verification directly from the device's SIM — no SMS required. This is Android-only as the Firebase PNV SDK (`com.google.firebase:firebase-pnv`) has no iOS counterpart.

**API:**
- `enableTestSession(token)` — enable SIM-less testing with a Firebase Console token
- `getVerificationSupportInfo()` — check if the device's SIM supports PNV
- `getVerifiedPhoneNumber()` — full verification flow (consent UI + carrier check)
- `getDigitalCredentialPayload(nonce)` — for custom flows using Android Credential Manager
- `exchangeCredentialResponseForPhoneNumber(dcApiResponse)` — exchange a Credential Manager JWT for a verified phone number

Both namespaced (`firebase.phoneNumberVerification()`) and modular APIs are supported. iOS gracefully rejects all calls with an "only supported on Android" error.

### Related issues

- Fixes #9013

### Release Summary

Added `@react-native-firebase/phone-number-verification` — a new package for Firebase's Phone Number Verification (PNV) SDK. Supports carrier-level phone number verification on Android without SMS.

### Checklist

  - [x] Yes
  - [x] `Android`
  - [ ] `iOS`
  - [ ] `Other` (macOS, web)
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

Tested on Android emulator via the manual test component at `tests/local-tests/pnv/pnv-test.jsx`:
- App launches without crash
- PNV test component renders with all three action buttons
- `getVerificationSupportInfo()` returns successfully with `[{"isSupported": false}]` (expected on emulator with no SIM)
- JS → Native bridge confirmed working end-to-end via logcat (`fpnvgms: [FpnvImpl] FpnvImpl created`)

https://github.com/user-attachments/assets/a6f24d02-6808-4f74-94a3-532ae8f762d0


packages/phone-number-verification/ios/RNFBPnv/RNFBPnvModule.h
```diff
@@ -0,0 +1,23 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+#import <Foundation/Foundation.h>
+
+#import <React/RCTBridgeModule.h>
+
+@interface RNFBPnvModule : NSObject <RCTBridgeModule>
```
mikehardy: No need for the iOS code (this .h and the the .m) to even exist I think?
- the typescript layer can platform select and for anything not android (ios or other...) it can return the error directly
- it should save a bit of compile time etc not to have them, though you *may* have to add something in the module config that the react-native build system uses to indicate that there is no iOS module - I don't recall if that's required or not

packages/phone-number-verification/lib/types/namespaced.ts
```diff
@@ -0,0 +1,81 @@
+/*
```
mikehardy: For a new module, let's skip namespaced entirely since we are about to remove those APIs shortly. This should be a modular-only implementation from the start

packages/phone-number-verification/lib/web/RNFBPnvModule.android.ts
```diff
@@ -0,0 +1,4 @@
+// No-op for android.
```
mikehardy: I believe that if the main implementation does a platform check and only runs native module methods for android, with errors directly for non-android, then everything in the `lib/web` directory may be removed as these will never be loaded

packages/phone-number-verification/lib/index.ts
```diff
@@ -0,0 +1,5 @@
+export type * from './types/pnv';
+export * from './modular';
+export type { FirebasePhoneNumberVerificationTypes } from './types/namespaced';
+export * from './namespaced';
+export { default } from './namespaced';
```
mikehardy: these would all nicely go away in a modular-only future - in fact you could probably just put the implementation directly in this file

packages/phone-number-verification/lib/modular.ts
```diff
@@ -0,0 +1,99 @@
...
+/**
+ * Returns the Phone Number Verification instance for the given app.
+ * @param app - The Firebase app instance. Optional, defaults to the default app.
+ * @returns PhoneNumberVerification instance
+ */
+export function getPhoneNumberVerification(
+  app?: FirebaseApp,
+): PhoneNumberVerification {
+  const appInstance = app ? (getApp(app.name) as FirebaseApp) : (getApp() as FirebaseApp);
+  return (appInstance as any).phoneNumberVerification() as PhoneNumberVerification;
```
mikehardy: I'm surprised this works without logging deprecation warnings for use of namespaced API since there is no "call / modular deprecation argument" style here?

regardless, these shouldn't chain to namespace - just implement directly as modular only, as mentioned elsewhere

packages/phone-number-verification/lib/modular.ts
```diff
@@ -0,0 +1,99 @@
...
+}
+
+/**
+ * Enables a test session for SIM-less testing.
```
mikehardy, line 52: might be worth mentioning here that test phone numbers follow a format per docs:

> In test mode, this phone number will have a valid country code, followed by all zeros.

tests/local-tests/index.js
```diff
@@ -33,6 +33,7 @@ import { AuthMFADemonstrator } from './auth/auth-mfa-demonstrator';
 import { HttpsCallableTestComponent } from './functions/https-callable';
 import { StorageTestComponent } from './storage/storage';
 import { StreamingCallableTestComponent } from './functions/streaming-callable';
+import { PnvTestComponent } from './pnv/pnv-test';
```
mikehardy, line 36: Nice! I built this local-test harness semi-recently to allow for manual testing of items that were hard to automate. I mention it because a) I'm just happy to see it used but also b) if it is annoying to use in some way, feel free to propose some change to it

mikehardy (changes requested):
solid start - main things I see:

- should be modular only
- eliminate platforms that don't have implementations entirely, error at JS/TS layer
- no typedoc config for the package, so `yarn reference:api` won't build these types out into the reference API, looking at the generated reference docs to make sure they look good should be part of the implementation loop
- no docs.json entry for this new package, so it won't be discoverable, and no docs for it in general (README and CHANGELOG in the package itself, docs for it in the main docs tree. Should at least be a one-pager linking to the main upstream docs, embed their youtube, and show an example of how to use it from start to finish in a similar way to the upstream docs show android
- - looks like there may be a good icon for PNV (e.g. https://www.gstatic.com/mobilesdk/251124_mobilesdk/fpnv-image.png - this one is close but I bet they have one without the mobile on bottom right - we have been referencing the docs from static.invertase.io which probably doesn't have one, but a sidequest to update the icons there with fresh svgs and then reference the PNV one would be more professional if it were possible - you can see google has a suite of icons on this page https://firebase.google.com/products-build)

mikehardy:
Two other notes

- Looks like you need to sign the CLA with a user that has the same email associated with it in github as was used in your commits
- Can you run the test app on a real phone? I don't think there is any proper way to test this except with a real phone - the test app should be connected to a project where PNV will work...I think

docs/phone-number-verification/usage/index.mdx
```diff
@@ -0,0 +1,132 @@
...
+You can also access all methods through a single object:
+
+```js
+import { getPhoneNumberVerification } from '@react-native-firebase/phone-number-verification';
+
+const pnv = getPhoneNumberVerification();
+
+const supportInfo = await pnv.getVerificationSupportInfo();
+const result = await pnv.getVerifiedPhoneNumber();
+```
```
mikehardy: we're working to get everyone off namespaced API and onto modular - for native-only modules we still do our best to invent "modular-like" APIs, and these don't follow that pattern. I don't think they're really any more or less convenient but they do stray from the modular-like stye

I don't think we need them - I'd stick with surfacing only what the native SDK has, in a modular style, as a fairly strict criteria for what APIs to publish

docs/phone-number-verification/usage/index.mdx
```diff
@@ -0,0 +1,132 @@
+---
```
mikehardy, line 1: Nice! This looks good - here's how you can check the docs live (built directly from github.com) - by doing `~<PR number>` --> https://docs.page/invertase/react-native-firebase~9045

What I see is:

- the youtube embed is not working? unexpected, I thought those worked even on live previews, so this may be an issue with the change here 🤔  - other pages definitely have them working (e.g. https://rnfirebase.io/~9045/analytics/usage)

- consent presentation is one of the areas that makes or breaks an implementation and people need a hand there. I'd call out that checking support requires no consent, but for the next step call out that consent will be triggered so the app should prepare the user for what will happen, and link to upstream docs where they discuss it (and I think they have a video) showing a good example

everything else looks great though

packages/phone-number-verification/e2e/pnv.e2e.js
```diff
@@ -0,0 +1,69 @@
...
+  describe('getVerificationSupportInfo()', function () {
+    android.it('returns an array of support info', async function () {
+      const { getVerificationSupportInfo } = pnvModular;
+      const supportInfo = await getVerificationSupportInfo();
+      supportInfo.should.be.an.Array();
+    });
+  });
+
+  describe('iOS unsupported', function () {
+    ios.it('rejects on iOS', async function () {
```
mikehardy: alter to "only Android supported" and add a block for rejects on Other as well?
simply exercising it there is sometimes useful, Other blows up just trying to load modules in certain contexts sometimes so even the smallest touch to verify it fails actually adds value

packages/phone-number-verification/lib/modular.ts
```diff
@@ -0,0 +1,115 @@
...
+
+interface NativePnvModule {
+  enableTestSession(token: string): Promise<void>;
+  getVerificationSupportInfo(): Promise<VerificationSupportInfo[]>;
+  getVerifiedPhoneNumber(): Promise<VerifiedPhoneNumberResult>;
+  getDigitalCredentialPayload(nonce: string): Promise<string>;
+  exchangeCredentialResponseForPhoneNumber(dcApiResponse: string): Promise<VerifiedPhoneNumberResult>;
+}
+
+function getNativeModule(): NativePnvModule {
```
mikehardy, line 38: Nice - great way to do the common error checking whereas I was contemplating scattering it into each method, a really naive implementation. Really clean

packages/phone-number-verification/lib/modular.ts
```diff
@@ -0,0 +1,115 @@
...
+ */
+export function getPhoneNumberVerification(): PhoneNumberVerification {
+  return {
+    enableTestSession: (token: string) => enableTestSession(token),
+    getVerificationSupportInfo: () => getVerificationSupportInfo(),
+    getVerifiedPhoneNumber: () => getVerifiedPhoneNumber(),
+    getDigitalCredentialPayload: (nonce: string) => getDigitalCredentialPayload(nonce),
+    exchangeCredentialResponseForPhoneNumber: (dcApiResponse: string) =>
+      exchangeCredentialResponseForPhoneNumber(dcApiResponse),
+  };
```
mikehardy: as mentioned in the docs section - may be slightly more convenient - but definitely increases API surface area and doesn't feel like the rest of the modular API, would remove

packages/phone-number-verification/CHANGELOG.md
```diff
@@ -0,0 +1,10 @@
+# Change Log
+
+All notable changes to this project will be documented in this file.
+See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
+
+## 24.1.1 (2026-06-10)
+
+### Features
+
+- **android:** add support for Firebase Phone Number Verification (PNV)
```
mikehardy: hold off on the first entry, our release automation will automatically add it (and it'll be in 25.0.0 most likely, though I may get this in a stable release first, making it a 24.2.0... - but the release automation will decide based on conventional commits and then do all the work here - it just needs the stub to start adding to)

packages/phone-number-verification/react-native.config.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
+  dependency: {
+    platforms: {
+      ios: null,
```
mikehardy, line 4: yep - that was the config blob I was thinking for auto-linking, so  you can just ship an android-only native. ✅

packages/phone-number-verification/typedoc.json
```diff
@@ -0,0 +1,5 @@
+{
```
mikehardy, line 1: Nice, verified

Image

mikehardy (changes requested):
Really quick turn and handled all the code feedback, awesome

- [x] Bit more of a thought on constraining the API to modular-only
- [x] Augment API to expose all native SDK items
- [x] Augment docs to discuss geographic restrictions, error handling, and dev token setup
- [x] Looks like CLA is still blocked
- [x] results from testing on an emulator (showing error unsupported, then test mode success)
- [x] PR description no longer matches the implementation

just1and0:
I actually don't have a physical Android device to test it, I planned on popping out in a bit to go buy a small Android device to test. I'd be happy if you can help on this one so not to delay the work.

Thanks in advance 😁

> Really quick turn and handled all the code feedback, awesome
> 
> 
> 
> - [ ] Bit more of a thought on constraining the API to modular-only
> 
> - [ ] Thought on augmenting the doc just a touch
> 
> - [ ] Looks like CLA is still blocked
> 
> - [ ] Any results from an attempt to run the e2e app on real device? or are you iOS only locally? I can use a phone I have as a test sled if needed - I have an android with valid SIM
> 
> - [ ] PR description no longer matches the implementation

docs/phone-number-verification/usage/index.mdx
just1and0: Thanks for sharing this, that's definitely a good trick to see how the docs look before it gets merged.

docs/phone-number-verification/usage/index.mdx
```diff
@@ -0,0 +1,245 @@
...
+import { getVerifiedPhoneNumber } from '@react-native-firebase/phone-number-verification';
+
+try {
+  const result = await getVerifiedPhoneNumber();
+} catch (error) {
+  switch (error.code) {
+    case 'carrier-not-supported':
+      // Fall back to SMS verification
+      break;
+    case 'activity-context-required':
```
mikehardy: apologies for not noticing this earlier - most of the firebase-js-sdk packages prefix all error.codes with their package as a convention and we follow it - example

https://firebase.google.com/docs/reference/js/functions.md#functionserrorcodecore

> Functions error code string appended after "functions/" product prefix. See [FunctionsErrorCode](https://firebase.google.com/docs/reference/js/functions.md#functionserrorcode) for full documentation of codes.

and https://firebase.google.com/docs/reference/js/auth#autherrorcodes (etc)

so these should all have a `pnv/` prefix when emitted by the code, then it follows that the documentation should show them as `pnv/<error code>` etc

mikehardy (changes requested):
This is a thing of beauty - great demo video showing the new error handling. Docs fantastic, implementation fantastic.

The only thing I can think of now is when I actually saw the new error codes I remembered that they are usually prefixed with product (i.e. `pnv/` here) - and some typescript typing / naming stuff I'll send internal - they seem useful but nearly trivial / easy to alter now pre-publish

mikehardy (approved):
Can't think of any possible thing that would polish this more - there may be something missing, but if so I can't see it

Fantastic

PR #9058, opened 2026-06-20
chore: fix issue labeler and restrict workflow token permissions

mikehardy:
## Summary

Two CI hygiene problems on `main`:

1. **Issue labeler broken** — `issue-labels.yaml` pinned `actions-ecosystem/action-add-labels` to a non-existent SHA (typo in commit `8e53fa6b3`), so every run failed at action download since 2026-06-18. The `actions-ecosystem` actions are also unmaintained.
2. **Over-broad default token** — Most workflows inherited the org default `GITHUB_TOKEN` scope instead of declaring minimal permissions.

**Decisions:** Replace the three ecosystem actions with `github-script` and the default token (`issues: write` only); drop the admin `GH_TOKEN` dependency for this workflow. Add explicit minimal permissions per workflow (`contents: read` + `actions: write` where cache/artifacts are used). Fix a stale-bot message typo while touching `stale.yml`.

**Remediation:** Audited all failed labeler runs; only one had a missed action (closed-item comment on #8914), already posted manually.

## Test plan

- [ ] CI green on this PR
- [ ] Observe labeler behavior on real issue/PR comments over the next few days (OP reply on open → `Needs Attention` / remove `blocked: customer-response`; OP reply on closed → standard bot comment)

### DIFF
--- a/.github/workflows/create_test_patches.yml
+++ b/.github/workflows/create_test_patches.yml
@@ -19,6 +19,10 @@ on:
       - '.spellcheck.dict.txt'
       - '**/*.md'
 
+permissions:
+  contents: read
+  actions: write
+
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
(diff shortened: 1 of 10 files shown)


mikehardy:
this has already passed any of the CI checks that would be affected by the changes --> merging so labeler is fixed

Issue #8973, opened 2026-04-13
[🐛] Firestore Android: RejectedExecutionException crash in sendOnSnapshotEvent during module invalidation


christian-apollo:
## Issue

A fatal `RejectedExecutionException` crash occurs on Android when a Firestore snapshot listener callback fires after the native module's executor has been shut down during invalidation.

### Stack trace

```
Fatal Exception: java.util.concurrent.RejectedExecutionException: Task rejected from ThreadPoolExecutor [Terminated, pool size = 0, active threads = 0, queued tasks = 0, completed tasks = 0]
  at java.util.concurrent.ThreadPoolExecutor$AbortPolicy.rejectedExecution(ThreadPoolExecutor.java:2097)
  at java.util.concurrent.ThreadPoolExecutor.reject(ThreadPoolExecutor.java:851)
  at java.util.concurrent.ThreadPoolExecutor.execute(ThreadPoolExecutor.java:1386)
  at java.util.concurrent.Executors$DelegatedExecutorService.execute(Executors.java:711)
  at com.google.android.gms.tasks.Tasks.call(5)
  at io.invertase.firebase.firestore.ReactNativeFirebaseFirestoreCollectionModule.sendOnSnapshotEvent(385)
  at io.invertase.firebase.firestore.ReactNativeFirebaseFirestoreCollectionModule.lambda$handleQueryOnSnapshot$4(355)
  at com.google.firebase.firestore.Query.lambda$addSnapshotListenerInternal$3(1178)
  at com.google.firebase.firestore.core.AsyncEventListener.lambda$onEvent$0(42)
  at android.os.Handler.handleCallback(995)
```

### Root cause

The current `invalidate()` implementation correctly removes snapshot listeners before calling `super.invalidate()`. However, `AsyncEventListener` dispatches callbacks via `Handler.post()` to the main looper. If a snapshot event is already enqueued on the Handler before `invalidate()` begins, the callback will still execute after `super.invalidate()` has terminated the executor — causing `Tasks.call()` inside `sendOnSnapshotEvent` to throw `RejectedExecutionException`.

This is a classic executor-shutdown race that cannot be prevented by reordering `invalidate()` alone.

### Proposed fix

Wrap the `Tasks.call()` invocation in `sendOnSnapshotEvent` with a try-catch for `RejectedExecutionException`. The event is safe to drop because:

- The module is being torn down; no JS listener will process the event
- The snapshot data is stale/irrelevant at teardown time

[...]

PR #8940, opened 2026-03-25
fix(android, firestore): detach snapshot listeners before executor shutdown

christian-apollo:
## Summary
Removes Firestore snapshot listeners before `super.invalidate()` in `ReactNativeFirebaseFirestoreCollectionModule` and `ReactNativeFirebaseFirestoreDocumentModule`.

`ReactNativeFirebaseModule.invalidate()` shuts down `TaskExecutorService`; `sendOnSnapshotEvent` uses `Tasks.call(getTransactionalExecutor(...), ...)`. If the executor is terminated first, a late snapshot can throw `RejectedExecutionException` (terminated pool).

This matches the teardown order already used in `ReactNativeFirebaseFirestoreTransactionModule`.

## Related issue
- Fixes #8939

## Test plan
- [ ] Android: enable Firestore `onSnapshot`, trigger RN reload or scenario that invalidates native modules; confirm no crash (manual / existing e2e if any)

Made with [Cursor](https://cursor.com)

Issue #8939, opened 2026-03-25
fix(android, firestore): RejectedExecutionException in sendOnSnapshotEvent during invalidate


christian-apollo:
## Description

Production crash on Android when Firestore collection/document snapshot listeners are active during React Native bridge teardown (reload, navigation that destroys the bridge, etc.).

```
Fatal Exception: java.util.concurrent.RejectedExecutionException
Task ... rejected from java.util.concurrent.ThreadPoolExecutor@[Terminated, pool size = 0, ...]
at com.google.android.gms.tasks.Tasks.call(...)
at io.invertase.firebase.firestore.ReactNativeFirebaseFirestoreCollectionModule.sendOnSnapshotEvent(...)
at io.invertase.firebase.firestore.ReactNativeFirebaseFirestoreCollectionModule.lambda$handleQueryOnSnapshot$4(...)
```

## Root cause

`ReactNativeFirebaseFirestoreCollectionModule` and `ReactNativeFirebaseFirestoreDocumentModule` call `super.invalidate()` **before** removing Firestore `ListenerRegistration`s.

`ReactNativeFirebaseModule.invalidate()` shuts down `TaskExecutorService`, terminating the executors used by `sendOnSnapshotEvent` (`Tasks.call(getTransactionalExecutor(...), ...)`). A snapshot can still be delivered after shutdown → `RejectedExecutionException` on a terminated pool.

`ReactNativeFirebaseFirestoreTransactionModule` already uses the correct order: clean up handlers, then `super.invalidate()`.

## Proposed fix

Remove snapshot listeners (and clear the sparse arrays) **before** `super.invalidate()` in both modules, matching the transaction module.

robbie-scribepro:
We've been getting the same error in Production for some time now. Sadly, the above fix hasn't resolved it for us. Any other ideas on root causes? We're not sure where to begin. It seems quite random, but definitely related to our application coming back into focus on Android after a longer period of time.

Full trace is:

```
java.util.concurrent.ThreadPoolExecutor$AbortPolicy in rejectedExecution at line 2097
java.util.concurrent.ThreadPoolExecutor in reject at line 851
java.util.concurrent.ThreadPoolExecutor in execute at line 1386
java.util.concurrent.Executors$DelegatedExecutorService in execute at line 711
com.google.android.gms.tasks.Tasks in call at line 5
[...]

Issue #8802, opened 2025-12-10
[🐛] Bug Android[new architecture]: NullPointerException in putMetric: Attempt to invoke virtual method on a null object reference


sean5940:
## Issue

I am encountering a `NullPointerException` on Android when using the Performance Monitoring module. The app shows an error toast with the message:

> `[perf/unknown] Attempt to invoke virtual method 'void com.google.firebase.perf.metrics.Trace.putMetric(java.lang.String, long)' on a null object reference`

It seems that `putMetric` is being called on a `Trace` object that has not been initialized properly or is null at the time of execution.

I suspect this might be related to a race condition where the `Trace` object is not ready before `putMetric` is invoked, or initialization of the trace failed silently.

**This issue is reproducible in the New Architecture (Fabric) environment.**

**Potential Fix Proposal:**
I noticed that making the storage maps for traces instance variables instead of static variables in `UniversalFirebasePerfModule.java` might resolve this issue, as it ensures thread safety and proper lifecycle management per module instance.

For example, in `UniversalFirebasePerfModule.java`, changing:
```java
private static SparseArray<Trace> traces = new SparseArray<>();
private static SparseArray<ScreenTrace> screenTraces = new SparseArray<>();
private static SparseArray<HttpMetric> httpMetrics = new SparseArray<>();
```
to:
```java
private SparseArray<Trace> traces = new SparseArray<>();
private SparseArray<ScreenTrace> screenTraces = new SparseArray<>();
private SparseArray<HttpMetric> httpMetrics = new SparseArray<>();
```
could potentially fix the problem.

**Steps to reproduce:**
1. Enable New Architecture (Fabric).
2. Force quit the application and relaunch it.
3. Initialize a custom trace using `perf().newTrace('my_custom_trace')`.
4. (Potentially immediately) call `trace.putMetric('my_metric', 1)`.
5. Observe the crash/toast on Android.

---

### Javascript

<details><summary>Click To Expand</summary>
<p>

#### `package.json`:

```json
{
  "dependencies": {
    "@react-native-firebase/app": "21.7.1",
    "@react-native-firebase/perf": "21.7.1",
    "react-native": "0.81.5"
  }
}
```

</p>
</details>

### Android

<details><summary>Click To Expand</summary>
<p>

#### Have you converted to AndroidX?

- [x] my application is an AndroidX application?
- [x] I am using `android/gradle.settings` `jetifier=true` for Android compatibility?

</p>
</details>

---

## Environment

[...]

PR #9093, opened 2026-07-08
refactor: introduce thread-safe RNFBHandleMap and migrate native listener registries repo-wide

mikehardy:
### Description

Native listener and handle maps were being read and written from more than one thread (JavaScript/TurboModule entry and Firebase SDK callbacks). That race can crash or drop events when maps such as `HashMap`, `SparseArray`, or `NSMutableDictionary` are updated without synchronization.

A gap analysis identified the same class of problem in **~32 places** (functions streaming listeners, firestore snapshot/transaction maps, database transaction/query maps, storage pending tasks, auth listeners and credential/MFA caches, app-check listeners, remote-config update handlers, perf Android trace/metric maps, messaging notification map).

The fix introduces a minimal shared **`RNFBHandleMap`** helper (Android + iOS): register handles under a lock with **`put` / `get` / `take` / `takeAll`**, then perform SDK cancel/remove and JS emit **after** the handle is taken out of the map. Package-specific registries wrap HandleMap where lifecycle needs extra rules (for example Functions `StreamingHolder`, Storage take-then-cancel, Auth `putReplacing` for caches). One atomic commit per package.

While hardening tests, we found Android unit tests were pinning Robolectric **`sdk`** because of a subtle Gradle plugin interaction (`DefaultSdkPicker` vs compile SDK). This PR switches pure JVM tests to JUnit where Robolectric is not required and drops unnecessary `@Config(sdk=…)` pins (see hygiene commits).

This PR also starts **in-package iOS XCTest infrastructure** (`RNFBAppUnitTests` and per-package unit targets) so we can cover `RNFBHandleMap` and package registries at 100% on reachable native lines, including cases that full e2e does not exercise reliably.

### Related issues

- #8802
- #2899
- #2771
- #3736
- #5225
- #6798
- #8939
- #8973
- #8770
- #8771
- #8210

### Related PRs

- #2770
- #2900
- #3978
- #5230
- #5455
- #6848
- #8799
- #8603
- #8940
- #9111

### Release Summary

Fixes thread-safety crashes and listener races by routing native handle maps through a shared locked `RNFBHandleMap` across affected Firebase modules.

### Checklist

  - [x] Yes
  - [x] `Android`
  - [x] `iOS`
  - [x] `Other` (macOS, web)
- My change includes tests;
  - [x] `jest` / JVM / XCTest unit tests added or updated where native lines are touched
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

- [x] Android JVM unit tests (`yarn tests:android:unit`) + Jacoco on touched registries/HandleMap
- [x] iOS in-package XCTest (`yarn tests:ios:unit`) + LCOV merge for HandleMap and package registries
- [x] Full pre-merge e2e with coverage: iOS **1463/0**, Android **1487/0**, macOS **1283/0**
- [x] Functions e2e refactored to use shared `getE2eEmulatorHost()`

### Internal note (maintainers only)

Linear tracking for merge automation only — not a public GitHub issue closure.

Fixes [CPRN-365](https://linear.app/invertase/issue/CPRN-365/pr-9093-introduce-thread-safe-rnfbhandlemap-and-migrate-native).


packages/functions/android/src/main/java/io/invertase/firebase/functions/NativeRNFBTurboFunctions.java
```diff
@@ -285,9 +288,11 @@ public void onNext(StreamResponse streamResponse) {
...
 
+                        // Do not remove the listener here — removal only happens via the
+                        // explicit cancel path (removeFunctionsStreamingListener) or invalidate,
+                        // so the final result can still be emitted even if a cancel races in.
                         if (isFinalResult) {
                           emitStreamEvent(appName, listenerId, responseData, true, false, null);
-                          removeFunctionsStreamingListener(listenerId);
                         } else {
                           emitStreamEvent(appName, listenerId, responseData, false, false, null);
                         }
```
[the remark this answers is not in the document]
mikehardy: Valid point, fixed in 4b5de74e9 — added `unregisterFunctionsStreamingListener` for map-only cleanup after terminal emits (no `Subscription.cancel()`), preserving the cancel/emit race fix. Explicit `removeFunctionsStreaming` still cancels+removes.

packages/functions/android/src/main/java/io/invertase/firebase/functions/NativeRNFBTurboFunctions.java
```diff
@@ -297,33 +302,35 @@ public void onNext(StreamResponse streamResponse) {
                       public void onError(Throwable t) {
                         WritableMap errorMap = createErrorMap(t);
                         emitStreamEvent(appName, listenerId, null, true, true, errorMap);
-                        removeFunctionsStreamingListener(listenerId);
                       }
```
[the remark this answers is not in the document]
mikehardy: Valid point, fixed in 4b5de74e9 — added `unregisterFunctionsStreamingListener` for map-only cleanup after terminal emits (no `Subscription.cancel()`), preserving the cancel/emit race fix. Explicit `removeFunctionsStreaming` still cancels+removes.

packages/functions/android/src/main/java/io/invertase/firebase/functions/NativeRNFBTurboFunctions.java
```diff
@@ -297,33 +302,35 @@ public void onNext(StreamResponse streamResponse) {
...
-                        if (listener != null) {
+                        boolean stillRegistered;
+                        synchronized (functionsStreamingListenersLock) {
+                          stillRegistered = functionsStreamingListeners.get(listenerId) != null;
+                        }
+                        if (stillRegistered) {
                           emitStreamEvent(appName, listenerId, null, true, false, null);
-                          removeFunctionsStreamingListener(listenerId);
                         }
                       }
```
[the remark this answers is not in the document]
mikehardy: Valid point, fixed in 4b5de74e9 — added `unregisterFunctionsStreamingListener` for map-only cleanup after terminal emits (no `Subscription.cancel()`), preserving the cancel/emit race fix. Explicit `removeFunctionsStreaming` still cancels+removes.

packages/functions/android/src/main/java/io/invertase/firebase/functions/NativeRNFBTurboFunctions.java
```diff
@@ -297,33 +302,35 @@ public void onNext(StreamResponse streamResponse) {
...
                     });
               } catch (Exception e) {
                 WritableMap errorMap = createErrorMap(e);
                 emitStreamEvent(appName, listenerId, null, true, true, errorMap);
-                removeFunctionsStreamingListener(listenerId);
               }
```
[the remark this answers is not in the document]
mikehardy: Valid point, fixed in 4b5de74e9 — added `unregisterFunctionsStreamingListener` for map-only cleanup after terminal emits (no `Subscription.cancel()`), preserving the cancel/emit race fix. Explicit `removeFunctionsStreaming` still cancels+removes.

packages/functions/android/src/main/java/io/invertase/firebase/functions/NativeRNFBTurboFunctions.java
```diff
@@ -297,33 +301,53 @@ public void onNext(StreamResponse streamResponse) {
...
   private void removeFunctionsStreamingListener(int listenerId) {
-    Object listener = functionsStreamingListeners.get(listenerId);
-    if (listener != null) {
-      if (listener instanceof Subscription) {
-        ((Subscription) listener).cancel();
+    synchronized (functionsStreamingListenersLock) {
+      Object listener = functionsStreamingListeners.get(listenerId);
+      if (listener != null) {
+        if (listener instanceof Subscription) {
+          ((Subscription) listener).cancel();
```
russellwheatley, line 347: after putting this through review, it's safer to `cancel()` outside of lock, if cancel stalls or hangs then this could deadlock which makes sense I think

okf-bundle/testing/coverage-design.md
```diff
@@ -61,7 +61,39 @@ Produce after fresh e2e on every required platform, then post-process native art
 
 **Verdict line:** `100% on reachable touched lines` **or** `NOT 100%` with numbered gaps and disposition (fixed / intractable with evidence / user-accepted deferral).
 
-Reviewers treat missing or stale coverage evidence as a **blocking** finding — same bar as missing e2e counts ([change authoring § validation evidence](change-authoring-workflow.md#validation-evidence-blocking)).
+Record in `.agents/reports/<item>/coverage-evidence.md` (preferred) or work-queue notes. Orchestrators **must not** close `coverage_evidence_gate` or `review_gate` without this file when lib/native bridge is touched.
```
russellwheatley, line 64: "without this file" - could this be clarified? Review says change to "this evidence package", but it could also be "without changing this file"?

russellwheatley (changes requested):

okf-bundle/testing/coverage-design.md
mikehardy: just the sort of ambiguity that can lead to drift - good catch - I'm changing to "without recording the evidence in one of those locations"

packages/functions/android/src/main/java/io/invertase/firebase/functions/NativeRNFBTurboFunctions.java
mikehardy: The API contract for cancel on a Reactive Streams instance indicates it must be timely and non-obstructive so it would be a contract violation for to hang or stall. In my execution path analysis I couldn't suss out a deadlock path but those are seemingly always surprising, and it never hurts to completely isolate work that strictly requires the lock from any other work. I think contention is more likely and sufficient as an argument against holding the lock either way.

So I definitely take the point. I'm hanging up on this one myself as I re-think it because I feel like the locking pattern growing here (and in response to this comment) is starting to get complicated and "complicated locking" isn't something to play with. I also noticed the same pattern in Firestore on android, and in a couple places on iOS - all exhibiting the same latent / "we have been lucky so far" unlocked listener maps / concurrent teardown windows possibilities

Let me finish my analysis there and the final result will I think be an architecture decision with a generic policy on listener registry design + code patterns, then a refactor of listener registries on both platform in all modules

russellwheatley: yeah, I'm also in two minds. I feel it would never hang, and would be fixed upstream in that event probably. but then I think - is it worth the risk? 🤔

mikehardy:
will come back to this shortly

packages/functions/android/src/main/java/io/invertase/firebase/functions/NativeRNFBTurboFunctions.java
mikehardy: went for a very simple / testable locking helper that areas with this need / issue could use. Then refactored everything to use it. Callsites are now really simple uses of the helper with no locking in the callsites, and the helper does the synchronization to make sure the data structure constraints aren't violated in multi-threaded access situations

this was a huge problem repo-wide...

okf-bundle/testing/coverage-design.md
mikehardy: working separately on some OKF policy items and I know you are as well with frogging. This will improve incrementally outside this PR

mikehardy:
Updated - finally ready to go on this odyssey

please note - patch coverage is lower because I specifically accepted uncovered lines in the native code. I explored how to cover them but it required pulling in Firebase and other stuff during unit tests which wouldn't work well, or having bridge/helper intermediary objects which lowered cohesion just for some coverage. In the end, I chose to make sure the new handler object had 100% coverage but I accepted gaps in some areas of it's usage in all the packages.

packages/app/android/src/main/java/io/invertase/firebase/common/RNFBHandleMap.java
```diff
@@ -0,0 +1,88 @@
...
+   * @throws RNFBHandleCollisionException if {@code id} is already occupied
+   */
+  public void put(K id, V handle) throws RNFBHandleCollisionException {
+    synchronized (lock) {
+      if (map.containsKey(id)) {
+        throw new RNFBHandleCollisionException(id);
+      }
+      map.put(id, handle);
+    }
+  }
```
russellwheatley, line 56: Making `put` unique with no upsert pushes every replace call site into a non-atomic `take` then `put`: `RNFBAuthCacheRegistry.putReplacing`, `RNFBFirestoreTransactionRegistry.putOrSkip`, `RNFBPerfHandleRegistry.putOrDiscard`, `RNFBStorageTaskRegistry.putOrDiscard`, `RNFBMessagingNotificationRegistry.putOrDiscard`. Each of those is a read-modify-write outside the lock, which is the bug class this PR exists to remove.

Two more primitives here collapse all of them into one call:

```suggestion
  public void put(K id, V handle) throws RNFBHandleCollisionException {
    synchronized (lock) {
      if (map.containsKey(id)) {
        throw new RNFBHandleCollisionException(id);
      }
      map.put(id, handle);
    }
  }

  /** Stores {@code handle} only when {@code id} is free. Returns the existing handle, or null. */
  public V putIfAbsent(K id, V handle) {
    synchronized (lock) {
      V existing = map.get(id);
      if (existing != null) {
        return existing;
      }
      map.put(id, handle);
      return null;
    }
  }

  /** Stores {@code handle} unconditionally. Returns the displaced handle, or null. */
  public V putReplacing(K id, V handle) {
    synchronized (lock) {
      return map.put(id, handle);
    }
  }
```

Callers still cancel or stop the returned handle after the call, so "the lock only moves pointers" still holds. Same on the iOS side.

packages/storage/android/src/main/java/io/invertase/firebase/storage/RNFBStorageTaskRegistry.java
```diff
@@ -0,0 +1,78 @@
...
+
+  /** Take then cancel outside the HandleMap lock. Returns {@code false} when no mapping existed. */
+  boolean takeAndCancel(int taskId) {
+    StoragePendingHandle handle = map.take(taskId);
+    if (handle == null) {
+      return false;
+    }
+    handle.cancel();
+    return true;
+  }
```
russellwheatley: This changes what JS gets back from `StorageTask.cancel()`. The old `cancelTaskById` returned `storageTask.cancel()`, and `ReactNativeFirebaseStorageTask.cancel()` only removes itself from the map inside the `isInProgress()` branch. Here the entry is taken unconditionally and the `handle.cancel()` result is thrown away, so:

- `cancel()` resolves `true` for a task that was paused or not yet in progress, where it used to resolve `false`.
- a paused upload is dropped from `PENDING_TASKS` even though nothing was cancelled, so no later `pause`, `resume` or `cancel` can reach it. Same in the window where `storageTask` is still null (the task registers itself in the constructor, before the `StorageTask` is assigned), which leaves an upload running and untracked.

```suggestion
  /** Cancel outside the HandleMap lock. Returns the underlying cancel result. */
  boolean takeAndCancel(int taskId) {
    StoragePendingHandle handle = map.get(taskId);
    if (handle == null) {
      return false;
    }
    // A successful cancel already removes the mapping via ReactNativeFirebaseStorageTask.destroyTask().
    boolean cancelled = handle.cancel();
    if (cancelled) {
      map.take(taskId);
    }
    return cancelled;
  }
```

Worth an e2e for cancelling a paused upload, there is none today.

packages/storage/android/src/main/java/io/invertase/firebase/storage/ReactNativeFirebaseStorageTask.java
```diff
@@ -51,43 +50,11 @@ class ReactNativeFirebaseStorageTask {
...
-
-  static void destroyAllTasks() {
-    for (int i = 0, size = PENDING_TASKS.size(); i < size; i++) {
-      int key = PENDING_TASKS.keyAt(i);
-      ReactNativeFirebaseStorageTask reactNativeFirebaseStorageTask = PENDING_TASKS.get(key);
-      reactNativeFirebaseStorageTask.cancel();
+    try {
+      PENDING_TASKS.put(taskId, this);
+    } catch (RNFBHandleCollisionException collision) {
+      throw new IllegalStateException(collision.getMessage());
```
russellwheatley: Duplicate task ids used to overwrite silently, now they throw out of a constructor. `TASK_ID` in `packages/storage/lib/StorageTask.ts` is a module-level counter that resets to 0 on every JS reload, while `PENDING_TASKS` is `static` and outlives the JS context. Reload with an upload in flight and the next `putFile` throws from native.

With an atomic `putReplacing` on `RNFBHandleMap` this becomes one call: replace, then cancel the displaced task outside the lock. That keeps the old last-wins behavior without a JS-reachable throw.

packages/auth/android/src/main/java/io/invertase/firebase/auth/RNFBAuthCacheRegistry.java
```diff
@@ -0,0 +1,74 @@
...
+
+  /**
+   * Replace = take then unique put. Preserves prior HashMap {@code put} upsert (last wins). Taken
+   * value is discarded (data only). Concurrent collision after take leaves the existing mapping via
+   * {@link #putOrDiscard}.
+   */
+  void putReplacing(String key, V value) {
+    map.take(key);
+    putOrDiscard(key, value);
+  }
```
russellwheatley, line 53: `take` then `putOrDiscard` is not atomic. Two threads replacing the same key interleave as take, take, put, put, and the second value is silently dropped while the caller believes it stored. `getSession` and `getJSError` still hand the `sessionId` back to JS, so a later `resolveMultiFactorSignIn` looks up a resolver that was never cached.

```suggestion
  void putReplacing(String key, V value) {
    map.putReplacing(key, value);
  }
```

Needs the atomic `putReplacing` on `RNFBHandleMap`, see the comment there.

packages/perf/android/src/main/java/io/invertase/firebase/perf/UniversalFirebasePerfModule.java
```diff
@@ -72,17 +74,17 @@ Task<Void> startTrace(int id, String identifier) {
         () -> {
           Trace trace = FirebasePerformance.getInstance().newTrace(identifier);
           trace.start();
-
-          traces.put(id, trace);
-
+          if (!traces.putOrDiscard(id, trace, Trace::stop)) {
+            throw new IllegalStateException("perf trace id already registered: " + id);
+          }
```
russellwheatley: `Trace::stop` is not a discard, it is how a trace gets reported to Firebase. On collision this sends a bogus near-zero-duration trace under the same name and then throws. Drop the handle instead. Same shape at lines 117, 143, 195, 226 and 243.

The throw is also reachable from JS input, `startTrace` used to overwrite and now rejects. With an atomic `putReplacing` you can keep the old behavior and stop the displaced handle outside the lock if you do want it recorded.

packages/perf/android/src/main/java/io/invertase/firebase/perf/UniversalFirebasePerfModule.java
```diff
@@ -72,17 +74,17 @@ Task<Void> startTrace(int id, String identifier) {
...
           return null;
         });
   }
 
   Task<Void> stopTrace(int id, Bundle metrics, Bundle attributes) {
     return Tasks.call(
         () -> {
-          Trace trace = traces.get(id);
+          Trace trace = traces.take(id);
```
russellwheatley: The handle is taken before the attributes and metrics are applied. `putAttribute` throws on an invalid name, and if it does the trace is already out of the map so `trace.stop()` is never reached and nothing can stop or retry it. Before, the entry survived the failure. Use `get` for the work and `take` immediately before `stop()`. Same at 153, 201 and 249.

packages/messaging/android/src/main/java/io/invertase/firebase/messaging/ReactNativeFirebaseMessagingReceiver.java
```diff
@@ -45,9 +45,10 @@ public void onReceive(Context context, Intent intent) {
     RemoteMessage remoteMessage = new RemoteMessage(intent.getExtras());
     ReactNativeFirebaseEventEmitter emitter = ReactNativeFirebaseEventEmitter.getSharedInstance();
 
-    // Add a RemoteMessage if the message contains a notification payload
+    // Add a RemoteMessage if the message contains a notification payload.
+    // Unique put: duplicate message ids keep the first (no HashMap upsert).
     if (remoteMessage.getNotification() != null) {
-      notifications.put(remoteMessage.getMessageId(), remoteMessage);
+      notifications.putOrDiscard(remoteMessage.getMessageId(), remoteMessage);
```
russellwheatley: This flips duplicate message ids from last-wins to first-wins, and the `storeFirebaseMessage` call below always stores the new message, so the in-memory map and the store can now disagree for the same id. The map is static and only drained when a notification is opened, so a redelivered FCM message keeps the stale copy. `putReplacing` keeps the previous behavior.

packages/app/ios/RNFBApp/RNFBHandleMap.m
```diff
@@ -0,0 +1,74 @@
...
+- (instancetype)init {
+  self = [super init];
+  if (self) {
+    _map = [NSMutableDictionary dictionary];
+  }
+  return self;
+}
+
+- (BOOL)put:(id)key value:(id)value error:(NSError **)error {
+  @synchronized(self) {
```
russellwheatley: Non-blocking. The header tells callers never to `@synchronized` a `RNFBHandleMap` instance, but nothing enforces it and the Android side already uses a private lock object for exactly this reason. Locking on a private ivar makes the contract unbreakable rather than documented.

russellwheatley (changes requested):
Requesting changes on the storage cancel regression and the non-atomic replace pattern. The `RNFBHandleMap` approach is right, and cancelling outside the lock from the last round is resolved cleanly.

Blocking:

- `RNFBStorageTaskRegistry.takeAndCancel` changes what `StorageTask.cancel()` returns to JS and orphans paused uploads.
- `putReplacing` / `putOrSkip` / `putOrDiscard` are all `take` then `put` outside the lock, so replace is still a read-modify-write race. Two atomic primitives on `RNFBHandleMap` fix every call site at once.
- Perf `putOrDiscard(id, trace, Trace::stop)` reports the discarded trace instead of dropping it, and `stopTrace` takes the handle before the work that can throw.

Worth a decision, not blocking:

- Several call sites turned a silent overwrite into an exception reachable from JS input. Storage task ids come from a JS module counter that resets on reload while `PENDING_TASKS` is static, so it is reachable in dev.
- Messaging duplicate ids flipped to first-wins while the message store still takes the new message.

packages/app/android/src/main/java/io/invertase/firebase/common/RNFBHandleMap.java
mikehardy: This upsert decision was one I struggled with - I like the idea of extending the API even though I was trying to keep it very small - these additional primitives allow upsert to work more cleanly for those that need it

👍

packages/app/ios/RNFBApp/RNFBHandleMap.m
mikehardy: I struggled with this on as well and didn't realize ObjC had similar semantics available to do what I did on android side. Will adopt this

packages/app/android/src/main/java/io/invertase/firebase/common/RNFBHandleMap.java
mikehardy: Fixed in `e225db2a` — added `putIfAbsent` / `putReplacing` on Android+iOS HandleMap; registry replace paths use these atomically (no take-then-put outside the map lock).

packages/storage/android/src/main/java/io/invertase/firebase/storage/RNFBStorageTaskRegistry.java
mikehardy: Fixed in storage commit — `takeAndCancel` returns `handle.cancel()` and only `take`s on success; Android `cancel()` now allows paused state so JS gets the same boolean contract as iOS.

packages/storage/android/src/main/java/io/invertase/firebase/storage/ReactNativeFirebaseStorageTask.java
mikehardy: Addressed — id collision uses `registerPending()` + `putOrDiscard` with promise reject (`internal` / `Handle id already registered`), not a constructor throw.

packages/auth/android/src/main/java/io/invertase/firebase/auth/RNFBAuthCacheRegistry.java
mikehardy: Fixed — auth caches use HandleMap `putIfAbsent` / `putReplacing`; no take-then-put race outside the map lock.

packages/perf/android/src/main/java/io/invertase/firebase/perf/UniversalFirebasePerfModule.java
mikehardy: Fixed (M1) — perf start paths use `putReplacing` + stop displaced outside lock (silent last-wins, Android+iOS aligned). No throw; displaced trace is stopped, incoming is registered.

packages/perf/android/src/main/java/io/invertase/firebase/perf/UniversalFirebasePerfModule.java
mikehardy: Fixed — `stopTrace` / `stopHttpMetric` use get → apply metrics/attributes → take → stop (same on iOS).

packages/messaging/android/src/main/java/io/invertase/firebase/messaging/ReactNativeFirebaseMessagingReceiver.java
mikehardy: Fixed — messaging map uses `putReplacing` (last-wins), consistent with `storeFirebaseMessage`.

packages/app/ios/RNFBApp/RNFBHandleMap.m
mikehardy: Fixed in `e225db2a` — iOS HandleMap locks on private `_lock` ivar.

mikehardy:
@russellwheatley Re-review requested on 900cbbfbd.

Aug-24 blocking items addressed in amended commits:
- HandleMap putIfAbsent / putReplacing (atomic replace)
- Storage takeAndCancel cancel-gated + paused cancel
- Storage id collision -> promise reject (not ctor throw)
- Perf collision -> silent last-wins + stop displaced
- Perf stop ordering (get -> work -> take -> stop)
- Messaging -> putReplacing last-wins
- iOS HandleMap private lock ivar

Replied on inline threads and resolved where fixed. CI fix push: google-java-format amends, messaging JVM test, iOS upload pause/resume re-skipped (CPRN-391 — #16359 fixed cancel only).

Slot-1 R8: 1488 Android / 1463 iOS full harness green locally.

mikehardy:
@russellwheatley Re-review requested on `772346d70` (force-pushed after R10 remediations + full local gate).

Post-Aug-24 / R9 residuals addressed in amended commits:
1. Storage: SDK `cancel()` before `destroyTask()` (unmap only on success)
2. Perf: identity-checked `takeIf` on stop paths (Android + iOS)
3. Functions iOS: `done` emit only when `take` succeeds (+ XCTest)
5. Messaging: `putReplacing` returns displaced; concurrency test strengthened
6. Tip message: iOS cancel + paused-cancel only; pause/resume stays CPRN-391

Post-R10 local gate @ tip: `yarn lint` 0; Android+iOS unit green; slot-1 full e2e **1488** Android / **1465** iOS.

Evidence: `.agents/reports/pr-9093/review-r10-remediation.md`

mikehardy:
## CI fix (R11) — ObjC++ \`takeIf\` blocks

iOS CI failed compiling \`RNFBPerfModule.mm\`: bare \`^BOOL(id value)\` is invalid in ObjC++ (\`type-id cannot have a name\`).

Fixed with \`NSObject *\` + local \`expected\` captures; autosquashed into the perf HandleMap commit (\`a49440f26\`). Clean forced rebuild of \`RNFBPerfModule.mm\` verified locally.

Tip: \`acea4ce4d\`. Re-requesting review after force-with-lease push.

packages/storage/android/src/main/java/io/invertase/firebase/storage/ReactNativeFirebaseStorageTask.java
```diff
@@ -108,39 +78,47 @@ static WritableMap buildErrorSnapshotMap(
...
+        return true;
+      }
     }
 
     return false;
   }
 
   void destroyTask() {
-    PENDING_TASKS.remove(taskId);
+    PENDING_TASKS.take(taskId);
```
russellwheatley: On a taskId collision, `registerPending()` -> `putOrDiscard` cancels the incoming handle (`this`), which calls `cancel()` -> `destroyTask()` here. But `taskId` on `this` is the same id as the *existing* task already in `PENDING_TASKS` (that's the whole point of the collision), so `PENDING_TASKS.take(taskId)` evicts the existing task's entry, not the incoming one that actually got cancelled. After that, `setTaskStatus`/`invalidate()` can no longer find the original task.

The test (`registerPending_occupiedId_cancelsIncomingAndKeepsExisting`) doesn't catch this because it seeds a `FakeHandle` with no `storageTask`, so `cancel()` short-circuits on the null guard and never reaches `destroyTask()`.

```suggestion
    PENDING_TASKS.takeIf(taskId, handle -> handle == this);
```

Same identity-compare pattern already used in the perf module's `takeIf`.

packages/database/android/src/main/java/io/invertase/firebase/database/RNFBDatabaseQueryRegistry.java
```diff
@@ -0,0 +1,55 @@
...
+  DatabaseQueryHandle get(String queryKey) {
+    return map.get(queryKey);
+  }
+
+  DatabaseQueryHandle take(String queryKey) {
+    return map.take(queryKey);
+  }
+
+  void takeIfIdle(String queryKey) {
+    map.takeIf(queryKey, query -> query != null && !query.hasListeners());
```
russellwheatley: Non-blocking. This predicate runs under `RNFBHandleMap`'s lock, and `hasListeners()` acquires `RNFBDatabaseListenerRegistry`'s own `occupancyLock` underneath, so `takeIf` here nests a second lock rather than staying lightweight like the header docs ask for. Not a deadlock with the code as it stands (nothing currently calls back into this registry while holding `occupancyLock`), but it's an easy invariant to break later without noticing. Same shape in `RNFBDatabaseQueryRegistry.m` (`takeIfIdle`) on iOS.

Might be safer to read `hasListeners()` outside the map lock, then re-check inside a cheap predicate right before taking.

russellwheatley (changes requested):
Re-reviewed at acea4ce4d now that CI is green. Core HandleMap and most of the migration look solid, but found one real bug on the storage collision path, plus a fragile locking pattern on the database side.

packages/storage/android/src/main/java/io/invertase/firebase/storage/ReactNativeFirebaseStorageTask.java
mikehardy: Fixed — `destroyTask` now uses identity `takeIf(taskId, h -> h == this)`, so cancelling a colliding incoming handle no longer evicts the existing mapping. Covered by unit regression.

packages/database/android/src/main/java/io/invertase/firebase/database/RNFBDatabaseQueryRegistry.java
mikehardy: Fixed — `hasListeners` is read outside the HandleMap lock; take uses identity `takeIf`, then put-back. If put-back fails after listeners appear, we `removeAllEventListeners()` on the orphan. Unit coverage on Android + iOS.

mikehardy:
Pushed tip `9f67c7f42` with follow-ups since the Aug 28 CHANGES_REQUESTED review:

- **Storage collision (blocking):** `destroyTask` identity `takeIf` so cancel of a colliding incoming handle cannot evict an existing task
- **Database `takeIfIdle`:** read `hasListeners` outside the HandleMap lock; identity take + put-back; clear orphan listeners if put-back fails
- **Identity-gated terminals:** storage `takeAndCancel` / iOS helper success·failure·cancel, and functions streaming terminal unregister, all identity-gated (no bare `take(id)` that can steal a replacement)
- **`putIfAbsentOrSame`:** Firestore `putOrSkip` uses a single atomic helper (Android + iOS)
- **Cleanup:** messaging unused `putOrDiscard` removed; database `registerReplacing` bogus `throws` removed
- **Validation:** full call-site adversarial pass clean; slot-1 e2e Android **1488** / iOS **1465**; coverage **100%** on reachable touched lines

Ready for another look when you have a moment.

russellwheatley (approved):
LGTM pending merge conflict resolution.

PR #9095, opened 2026-07-09
fix(remote-config, ios): ensure fetchActivate() match fetch() + activate() behavior

### LINKED ISSUE #7779: [🐛]  🔥   remote config method `fetchAndActivate()` inconsistent with method `fetch()` and then `activate()` on iOS only
## Issue
If you call `await fetchAndActivate()` remote config method on iOS it always returns a `true` value, even when new values were not fetched from the remote. This value contradicts the value returned by `activate()` when values were not fetched. A code snipped below can be used to test this:

```ts
 async function fetchAndActivateCompare() {
  const lastFetch = await remoteConfig().lastFetchStatus;
  console.log('Last fetch status: ', lastFetch);
  await remoteConfig().fetch();
  await remoteConfig()
    .activate()
    .then((didFetch) => {
      if (didFetch) {
        console.log('RemoteConfigs fetched from fetch and then activate.');
      } else {
        console.log('RemoteConfigs not fetched from fetch and then activate.');
      }
    });
    
  await new Promise((res) => setTimeout(res, 1000));

  const fetchedRemotely = await remoteConfig().fetchAndActivate()
  if (fetchedRemotely) {
    console.log('RemoteConfigs fetched from fetchAndActivate.');
  } else {
    console.log('RemoteConfigs not fetched from fetchAndActivate');
  }
}

// call init in useEffect somewhere in your app
const init = async () => {

  // set config
  await remoteConfig()
        .setConfigSettings({
          minimumFetchIntervalMillis: 500, // NOTE 500 milliseconds is less than our set time out, allowing for a sufficient cache reset before calling `fetchAndActivate()`
        })
      
  await fetchAndActivateCompare();   

}

```

This results in the following output in the console:

```terminal
 LOG  Last fetch status:  success
 LOG  RemoteConfigs not fetched from fetch and then activate.
 LOG  RemoteConfigs fetched from fetchAndActivate.
```

---

### Javascript

<p>

#### `package.json`:

```json
# N/A
```

```json
# N/A
```

</details>

### iOS

<p>

#### `ios/Podfile`:

- [ ] I'm not using Pods

```ruby
# Resolve react_native_pods.rb with node to allow for hoisting
require Pod::Executable.execute_command('node', ['-p',
  'require.resolve(
    "react-native/scripts/react_native_pods.rb",
    {paths: [process.argv[1]]},
  )', __dir__]).strip

platform :ios, min_ios_version_supported
prepare_react_native_project!


flipper_config = ENV['NO_FLIPPER'] == "1" ? FlipperConfiguration.disabled : FlipperConfiguration.enabled

linkage = ENV['USE_FRAMEWORKS']
if linkage != nil
  Pod::UI.puts "Configuring Pod with #{linkage}ally linked Frameworks".green
  use_frameworks! :linkage => linkage.to_sym
end

target 'AOT' do
  config = use_native_modules!

  # Flags change depending on the env values.
  flags = get_default_flags()

  # Firebase needs to use frameworks see react-native-firebase docs
  use_frameworks! :linkage => :static

  use_react_native!(
    :path => config[:reactNativePath],
    # Hermes is now enabled by default. Disable by setting this flag to false.
    :hermes_enabled => flags[:hermes_enabled],
    :fabric_enabled => flags[:fabric_enabled],
    ## Enables Flipper.
    ##
    ## Note that if you have use_frameworks! enabled, Flipper will not work and
    ## you should disable the next line. (DISABLED FLIPPER FOR USE WITH FRAMEWORKS NEEDED FOR FIREBASE)
    # :flipper_configuration => flipper_config,
    ## An absolute path to your application root.
    :app_path => "#{Pod::Config.instance.installation_root}/.."
  )

  target 'AOTTests' do
    inherit! :complete
    # Pods for testing
  end


  post_install do |installer|

    # https://github.com/facebook/react-native/blob/main/packages/react-native/scripts/react_native_pods.rb#L197-L202
    react_native_post_install(
      installer,
      config[:reactNativePath],
      :mac_catalyst_enabled => false
    )
    __apply_Xcode_12_5_M1_post_install_workaround(installer)

  end
end

```

#### `AppDelegate.mm`:

```objc
#import "AppDelegate.h"
#import "RNBootSplash.h" // react-native-bootsplash
#import <Firebase.h> // Setup for Firebase services with react-native-firebase

// CUSTOM IMPORTS
#import <CodePush/CodePush.h>

@implementation AppDelegate

{
  [FIRApp configure]; // Setup for Firebase services with react-native-firebase
  self.moduleName = @"AOT";
  // You can add your custom initial props in the dictionary below.
  // They will be passed down to the ViewController used by React Native.
  self.initialProps = @{};

  return [super application:application didFinishLaunchingWithOptions:launchOptions];
}

{
#if DEBUG
  return [[RCTBundleURLProvider sharedSettings] jsBundleURLForBundleRoot:@"index"];
#else
  return [CodePush bundleURL]; // <--- Replaced for CodePush Previous: (return [[NSBundle mainBundle] URLForResource:@"main" withExtension:@"jsbundle"];)
#endif
}

// ⬇️ This method will need to change for react-native-bootsplash to work on 0.74+ versions. See docs
- (UIView *)createRootViewWithBridge:(RCTBridge *)bridge
                          moduleName:(NSString *)moduleName
                           initProps:(NSDictionary *)initProps {
  UIView *rootView = [super createRootViewWithBridge:bridge moduleName:moduleName initProps:initProps];
  [RNBootSplash initWithStoryboard:@"BootSplash" rootView:rootView]; // ⬅️ initialize the splash screen (RN boot splash)
  return rootView;
}
// ⬆️ This method will need to be removed and change to:
// - (void)customizeRootView:(RCTRootView *)rootView {
//   [RNBootSplash initWithStoryboard:@"BootSplash" rootView:rootView]; // ⬅️ initialize the splash screen
// }

@end

```

</p>
</details>

---

### Android

<p>

#### `android/build.gradle`:

```groovy
// N/A
```

```groovy
// N/A
```

```groovy
// N/A
```

#### `MainApplication.java`:

```java
// N/A
```

#### `AndroidManifest.xml`:

```xml
<!-- N/A -->
```

</p>
</details>

---

## Environment

<p>

```
 System:
  OS: macOS 14.4
  CPU: (10) arm64 Apple M1 Pro
  Memory: 208.44 MB / 32.00 GB
  Shell:
    version: "5.9"
    path: /bin/zsh
Binaries:
  Node:
    version: 18.17.0
    path: ~/.nvm/versions/node/v18.17.0/bin/node
  Yarn:
    version: 1.22.21
    path: ~/.nvm/versions/node/v18.17.0/bin/yarn
  npm:
    version: 9.6.7
    path: ~/.nvm/versions/node/v18.17.0/bin/npm
  Watchman:
    version: 2023.09.25.00
    path: /opt/homebrew/bin/watchman
Managers:
  CocoaPods:
    version: 1.13.0
    path: /opt/homebrew/bin/pod
SDKs:
  iOS SDK:
    Platforms:
      - DriverKit 23.4
      - iOS 17.4
      - macOS 14.4
      - tvOS 17.4
      - visionOS 1.1
      - watchOS 10.4
  Android SDK: Not Found
IDEs:
  Android Studio: 2022.1 AI-221.6008.13.2211.9619390
  Xcode:
    version: 15.3/15E204a
    path: /usr/bin/xcodebuild
Languages:
  Java:
    version: 17.0.10
    path: /usr/bin/javac
  Ruby:
    version: 2.6.10
    path: /usr/bin/ruby
npmPackages:
  "@react-native-community/cli": Not Found
  react:
    installed: 18.2.0
    wanted: 18.2.0
  react-native:
    installed: 0.72.4
    wanted: 0.72.4
  react-native-macos: Not Found
npmGlobalPackages:
  "*react-native*": Not Found
Android:
  hermesEnabled: true
  newArchEnabled: false
iOS:
  hermesEnabled: true
  newArchEnabled: false
```

  - [x] iOS
  - [ ] Android
  - [ ] Both
  - `^19.2.2`
  - `remote config`
  - `Y` & `5.0.4`

</p>
</details>

russellwheatley:
### Description

On iOS, `remoteConfig().fetchAndActivate()` always resolved `true`, even when nothing new was fetched/activated. This is inconsistent with a separate `fetch()` + `activate()` call, and inconsistent with Android's `fetchAndActivate()`, which correctly resolves `false` when there's nothing new.

**Root cause:** the iOS implementation used the Firebase iOS SDK's `-[FIRRemoteConfig fetchAndActivateWithCompletionHandler:]` and mapped its `FIRRemoteConfigFetchAndActivateStatus` result to a boolean (`SuccessFetchedFromRemote` → `true`, else → `false`). That status only reflects whether the *fetch* succeeded — it ignores the `changed` BOOL returned by the SDK's own `activateWithCompletion:`, which is the actual "did activation change anything" signal. Since `activateWithCompletion:` essentially never errors in modern SDK versions, the status was almost always `SuccessFetchedFromRemote`, so RNFB always resolved `true`.

**Fix:** `fetchAndActivate:` in `RNFBConfigModule.mm` now performs the fetch + activate itself (mirroring what the SDK does internally) and resolves with the `changed` BOOL from `activateWithCompletion:`, using the same "already activated" error-handling special case already present in RNFB's own `activate:` method. This matches Android's semantics and iOS's own `activate()` semantics.

### Related issues

- Fixes #7779

### Release Summary

fix(remote-config, ios): `fetchAndActivate()` now correctly resolves `false` when there is nothing new to activate, matching Android and matching `fetch()` + `activate()`.

### Test Plan

Added 3 new e2e tests to `packages/remote-config/e2e/config.e2e.js` (in the existing `fetchAndActivate()` describe block), designed as regression tests that would have failed against the old iOS behavior:

1. `returns false when there is nothing new to activate` — calls `fetchAndActivate()` twice in a row with no server-side template change; second call must resolve `false`.
2. `is consistent with fetch() followed by activate()` — asserts `fetchAndActivate()` agrees with a separate `fetchConfig()` + `activate()` call when nothing changed.
3. `returns true only once new remote values have actually been fetched and activated` — publishes a new remote config parameter via the existing cloud-function test helper (`FirebaseHelpers.updateRemoteConfigTemplate` / `testFunctionRemoteConfigUpdateV2`), then polls `fetchAndActivate()` for up to 60s (template publish → fetch-backend propagation isn't instant) asserting it eventually returns `true` and the new value is visible via `getValue()`. Cleans up the published parameter afterwards.

### Checklist

  - [x] Yes
  - [x] `Android` (unchanged, regression-tested for parity)
  - [x] `iOS`
  - [x] `Other` (macOS, web — validates test logic against firebase-js-sdk)
- My change includes tests;
- [x] I have updated TypeScript types that are affected by my change. (n/a — no public API/type change)
- This is a breaking change;
  - [ ] Yes
  - [x] No

---

### DIFF
--- a/packages/remote-config/ios/RNFBConfig/RNFBConfigModule.mm
+++ b/packages/remote-config/ios/RNFBConfig/RNFBConfigModule.mm
@@ -205,42 +205,41 @@ RCT_EXPORT_MODULE(NativeRNFBTurboConfig)
   // in-use config values.
   FIRApp *firebaseApp = firebaseAppForName(appName);
   __weak RNFBConfigModule *weakSelf = self;
-  FIRRemoteConfigFetchCompletion fetchCompletion =
-      ^(FIRRemoteConfigFetchStatus status, NSError *__nullable error) {
-        RNFBConfigModule *strongSelf = weakSelf;
-        if (!strongSelf) {
-          return;
-        }
+  FIRRemoteConfigFetchCompletion fetchCompletion = ^(FIRRemoteConfigFetchStatus status,
+                                                     NSError *__nullable error) {
+    RNFBConfigModule *strongSelf = weakSelf;
+    if (!strongSelf) {
+      return;
+    }
 
-        if (error) {
-          [RNFBSharedUtils
-              rejectPromiseWithUserInfo:reject
-                               userInfo:[@{
-                                 @"code" : convertFIRRemoteConfigFetchStatusToNSString(status),
-                                 @"message" :
-                                     convertFIRRemoteConfigFetchStatusToNSStringDescription(status)
-                               } mutableCopy]];
-          return;
-        }
+    if (error) {
+      [RNFBSharedUtils
+          rejectPromiseWithUserInfo:reject
+                           userInfo:[@{
+                             @"code" : convertFIRRemoteConfigFetchStatusToNSString(status),
+                             @"message" :
+                                 convertFIRRemoteConfigFetchStatusToNSStringDescription(status)
+                           } mutableCopy]];
+      return;
+    }
 
-        [[FIRRemoteConfig remoteConfigWithApp:firebaseApp]
-            activateWithCompletion:^(BOOL changed, NSError *_Nullable activateError) {
-              if (activateError) {
-                if (activateError.userInfo &&
-                    activateError.userInfo[@"ActivationFailureReason"] != nil &&
-                    [activateError.userInfo[@"ActivationFailureReason"]
-                        containsString:@"already activated"]) {
-                  resolve([strongSelf resultWithConstants:@([RCTConvert BOOL:@(NO)])
-                                              firebaseApp:firebaseApp]);
-                } else {
-                  [RNFBSharedUtils rejectPromiseWithNSError:reject error:activateError];
-                }
-              } else {
-                resolve([strongSelf resultWithConstants:@([RCTConvert BOOL:@(changed)])
-                                            firebaseApp:firebaseApp]);
-              }
-            }];
-      };
+    [[FIRRemoteConfig remoteConfigWithApp:firebaseApp]
+        activateWithCompletion:^(BOOL changed, NSError *_Nullable activateError) {
+          if (!activateError) {
+            resolve([strongSelf resultWithConstants:@([RCTConvert BOOL:@(changed)])
+                                        firebaseApp:firebaseApp]);
+            return;
+          }
+          if (activateError.userInfo && activateError.userInfo[@"ActivationFailureReason"] != nil &&
+              [activateError.userInfo[@"ActivationFailureReason"]
+                  containsString:@"already activated"]) {
+            resolve([strongSelf resultWithConstants:@([RCTConvert BOOL:@(NO)])
+                                        firebaseApp:firebaseApp]);
+            return;
+          }
+          [RNFBSharedUtils rejectPromiseWithNSError:reject error:activateError];
+        }];
+  };
 
   [[FIRRemoteConfig remoteConfigWithApp:firebaseApp] fetchWithCompletionHandler:fetchCompletion];
 }
--- a/packages/remote-config/ios/RNFBConfig/RNFBConfigModule.mm
+++ b/packages/remote-config/ios/RNFBConfig/RNFBConfigModule.mm
@@ -188,66 +188,13 @@ RCT_EXPORT_MODULE(NativeRNFBTurboConfig)
   }
 }
 
-- (void)fetchAndActivate:(NSString *)appName
-                 resolve:(RCTPromiseResolveBlock)resolve
-                  reject:(RCTPromiseRejectBlock)reject {
-  // NOTE: We deliberately do not use `-[FIRRemoteConfig fetchAndActivateWithCompletionHandler:]`
-  // here. Its `FIRRemoteConfigFetchAndActivateStatus` result only reflects whether the fetch
-  // succeeded (`SuccessFetchedFromRemote`) vs fell back to previously fetched data
-  // (`SuccessUsingPreFetchedData`) - it does NOT reflect whether activation actually changed any
-  // config values, so it reports `SuccessFetchedFromRemote` (=> we'd resolve `true`) even when the
-  // fetched values are identical to what's already active. See
-  // https://github.com/invertase/react-native-firebase/issues/7779
-  //
-  // Instead we perform the fetch + activate ourselves and resolve with the `changed` BOOL from
-  // `activateWithCompletion:`, which matches the semantics of our own `activate()` method as well
-  // as the Android implementation, both of which resolve `true` only when activation changed the
-  // in-use config values.
-  FIRApp *firebaseApp = firebaseAppForName(appName);
-  __weak RNFBConfigModule *weakSelf = self;
-  FIRRemoteConfigFetchCompletion fetchCompletion = ^(FIRRemoteConfigFetchStatus status,
-                                                     NSError *__nullable error) {
-    RNFBConfigModule *strongSelf = weakSelf;
-    if (!strongSelf) {
-      return;
-    }
-
-    if (error) {
-      [RNFBSharedUtils
-          rejectPromiseWithUserInfo:reject
-                           userInfo:[@{
-                             @"code" : convertFIRRemoteConfigFetchStatusToNSString(status),
-                             @"message" :
-                                 convertFIRRemoteConfigFetchStatusToNSStringDescription(status)
-                           } mutableCopy]];
-      return;
-    }
-
-    [[FIRRemoteConfig remoteConfigWithApp:firebaseApp]
-        activateWithCompletion:^(BOOL changed, NSError *_Nullable activateError) {
-          if (!activateError) {
-            resolve([strongSelf resultWithConstants:@([RCTConvert BOOL:@(changed)])
-                                        firebaseApp:firebaseApp]);
-            return;
-          }
-          if (activateError.userInfo && activateError.userInfo[@"ActivationFailureReason"] != nil &&
-              [activateError.userInfo[@"ActivationFailureReason"]
-                  containsString:@"already activated"]) {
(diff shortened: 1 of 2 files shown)


packages/remote-config/ios/RNFBConfig/RNFBConfigModule.mm
```diff
@@ -191,27 +191,58 @@ - (void)fetch:(NSString *)appName
...
+                  resolve([strongSelf resultWithConstants:@([RCTConvert BOOL:@(NO)])
+                                              firebaseApp:firebaseApp]);
+                } else {
+                  [RNFBSharedUtils rejectPromiseWithNSError:reject error:activateError];
+                }
+              } else {
+                resolve([strongSelf resultWithConstants:@([RCTConvert BOOL:@(changed)])
+                                            firebaseApp:firebaseApp]);
+              }
+            }];
```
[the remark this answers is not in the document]
russellwheatley: matches existing convention, leaving as is

packages/remote-config/ios/RNFBConfig/RNFBConfigModule.mm
```diff
@@ -191,27 +191,58 @@ - (void)fetch:(NSString *)appName
...
+                        containsString:@"already activated"]) {
+                  resolve([strongSelf resultWithConstants:@([RCTConvert BOOL:@(NO)])
+                                              firebaseApp:firebaseApp]);
+                } else {
+                  [RNFBSharedUtils rejectPromiseWithNSError:reject error:activateError];
+                }
+              } else {
+                resolve([strongSelf resultWithConstants:@([RCTConvert BOOL:@(changed)])
+                                            firebaseApp:firebaseApp]);
+              }
```
wkedzierawski: nit: An early return for the successful activation path could make this block slightly easier to scan.

```
if (!activateError) {
  resolve([strongSelf resultWithConstants:@([RCTConvert BOOL:@(changed)])
                              firebaseApp:firebaseApp]);
  return;
}

if (activateError.userInfo &&
    activateError.userInfo[@"ActivationFailureReason"] != nil &&
    [activateError.userInfo[@"ActivationFailureReason"] containsString:@"already activated"]) {
  resolve([strongSelf resultWithConstants:@([RCTConvert BOOL:@(NO)])
                              firebaseApp:firebaseApp]);
  return;
}

[RNFBSharedUtils rejectPromiseWithNSError:reject error:activateError];
```

russellwheatley: [b5c6b70](https://github.com/invertase/react-native-firebase/pull/9095/commits/b5c6b707cbdb09bcfa63e3d74e3fa98150cfa484)

russellwheatley:
### Why `codecov/patch` still fails

Patch coverage improved 34% → 45% after deduplicating the `activate()`/`fetchAndActivate()` activation logic (06aa617), but is still below the 58.61% target.

Cross-checking the head-commit coverage report line-by-line against what the e2e suite actually exercises shows the remaining "misses" aren't real gaps:

- `activate:` (the shared call site) shows **0% coverage** on every line, despite the existing `"activates fetched config"` e2e test calling it directly and passing.
- `setConfigSettings:` and `setDefaults:` — both **unrelated, untouched by this PR** — also show 0% across the board, despite being exercised by passing tests (e.g. `"minimumFetchIntervalMillis sets correctly"`).

Since untouched, definitely-executed methods elsewhere in the same file are also reported as 0%, this looks like a stale/corrupted iOS-native coverage report for this run rather than a genuine drop in test coverage (see the "Stale coverage data" caveats already documented for this pipeline). The two branches that remain legitimately hard to cover via e2e — generic (non-"already activated") SDK errors from `fetchWithCompletionHandler:`/`activateWithCompletion:` — are pre-existing patterns shared with `fetch:`/`activate:`, not new risk introduced here, and aren't deterministically triggerable without native fault-injection test infra that doesn't currently exist in this repo.

mikehardy (approved):
Looks great, great test coverage - annoying that the obvious native API to reach for doesn't yield consistent results with android or the disaggregated fetch+activate path, but given that, this is exactly the fix +1

PR #9099, opened 2026-07-10
test(e2e): make all contentious e2e resources configurable

mikehardy:
## Motivation

RNFB e2e today runs **one platform at a time** in a serial workflow — a single Metro, a single Firebase emulator suite, and one device identity per platform. That caps throughput at “one run, wait, next platform.”

This PR makes every contentious e2e resource **configurable via environment variables** so multiple platform runs can coexist on one host. The goal is to unlock e2e testing throughput from that serial state to **fully parallel “as many concurrent cells as your host can support via available compute”** — e.g. three worktrees each running android + ios + macos at once, bounded only by CPU/RAM/emulator capacity, not by hard-coded ports or bundle IDs.

**Default behaviour is unchanged:** unset env vars preserve serial `yarn tests:*` commands, ports, and workflows.

## Start here (most important)

**[e2e-parallel-design.md](https://github.com/invertase/react-native-firebase/blob/e2e-parallel/okf-bundle/testing/e2e-parallel-design.md)** — the architecture, slot model, and operational playbook for horizontal scale-out. Read this first.

Rebased onto current `main` after [#9192](https://github.com/invertase/react-native-firebase/pull/9192) (`tests-macos/` split). Slotted e2e is reimplemented on that layout: **two Metro roots** (`tests/` for iOS/Android, `tests-macos/` for macOS). A full slot that runs android/ios **and** macos starts **two** packagers (distinct `RNFB_{ANDROID,IOS,MACOS}_METRO_PORT`s). Same-platform scale-out uses **one git worktree per slot** (`3× android + 3× ios + 3× macos` ⇒ three worktrees).

## What landed

- **`packages/app/e2e/helpers.js`** — shared Metro, Jet, and Firebase emulator host/port resolution. Platform-prefixed `RNFB_{ANDROID,IOS,MACOS}_*` first, then `RCT_METRO_PORT` / serial defaults. `getE2eEmulatorHost()` is `10.0.2.2` on Android and `127.0.0.1` on iOS/macOS (never `localhost`).
- **Prefs reconnect** — `packages/app/e2e/config.e2e.js` reconnects Firestore through `getE2eEmulatorHost()` so Android slotted runs are not stuck on `localhost`.
- **`tests/.detoxrc.js` / `tests/.jetrc.js`** — multi-slot Android AVD / iOS simulator devices; env-driven reverse ports; Jet/Metro for **android/ios only**.
- **`tests-macos/.jetrc.js` + `tests-macos/.babelrc`** — macOS Jet spawn (`stdio: ignore`), `RNFB_MACOS_PRODUCT_NAME`, and the same `RNFB_*` Babel inlines as mobile. Packager: `yarn tests:macos:packager:*` from `tests-macos/` — never `yarn tests:packager:*` for macOS Jet.
- **macOS `PRODUCT_NAME`** — `io.invertase.testing$(RNFB_MACOS_PRODUCT_NAME_SUFFIX)` in the pbx; slotted `RNFB_MACOS_PRODUCT_NAME=io.invertase.testing.sN`. Do **not** pass `PRODUCT_NAME=` on the xcodebuild CLI.
- **`tests/e2e/firebase.test.js`** — Jet spawn/orchestration and macOS stale-app guard respect slotted ports.
- **Package e2e helpers** (`auth`, `database`, `firestore`, `functions`, `ai`) — emulator URLs through shared helpers (Functions localhost helper removed).
- **`AppDelegate.mm`** — honour `RCT_METRO_PORT` (mobile `tests/ios`, macOS `tests-macos/macos`).
- **Scripts** — `export-slot-env.sh`, `start-emulator-slotted.sh`, `run-slotted-packager.sh`, `run-slotted-test-cover.sh`, `check-e2e-resources.sh` / `release-e2e-resources.sh` (slot-scoped when env is loaded). `create-android-avds.sh` / `create-ios-simulators.sh` provision `TestingAVD-N` / `RNFB E2E iOS slot-N`.
- **Docs** — `okf-bundle/testing/running-e2e.md` (slot lifecycle + dual packager roots), `e2e-parallel-design.md`, `CONTRIBUTING.md` macOS bullets.

Slot formula (unchanged): `BASE=12000+slot*1000`; android off 0, ios +100, macos +200; metro +7, jet +10, jet-control +11. Full carry-in of all `RNFB_{ANDROID,IOS,MACOS}_*` for Babel inline.

A future orchestration layer (outside this PR) can reserve slots and export env; RNFB stays coordinator-agnostic.

## Test plan

- [x] Rebase onto `main` including the `tests-macos/` split; reimplement slotted packager/Jet/app path on the two Metro roots
- [x] Slot-0 full harness, 3 waves × `1× android ∥ 1× ios ∥ 1× macos` (1485 / 1461 / 1281 passing per wave)
- [x] Area e2e after R0 remediations (app harness): android 39 / ios 39 / macos 33; `yarn lint:js` / Jest helpers 13/13
- [x] Review threads: Metro platform-prefix before `RCT_METRO_PORT`; Functions uses `getE2eEmulatorHost()` only
- [x] **3×3 parallel:** three worktrees (slots 0/1/2) each running android+ios+macos concurrently — 3 consecutive clean waves on rewritten HEAD (1485 / 1461 / 1281 per wave)
- [ ] Serial sanity: no env overrides → `yarn tests:macos:test-cover` (and one native platform) still uses `:8081` / default emulators


packages/ai/lib/requests/request.ts
```diff
@@ -58,12 +58,21 @@ export class RequestUrl {
...
+      let fnHost = '127.0.0.1';
+      let fnPort = 5001;
+      try {
+        const { getE2eEmulatorHost, getE2eEmulatorPort } = require('@react-native-firebase/app/e2e/helpers');
+        fnHost = getE2eEmulatorHost();
+        fnPort = getE2eEmulatorPort('functions');
+      } catch (_e) {
+        const isAndroid = Platform.OS === 'android';
+        fnHost = isAndroid ? '10.0.2.2' : '127.0.0.1';
+      }
```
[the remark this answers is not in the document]
mikehardy: Fixed in 8d47ad427: removed the `require("@react-native-firebase/app/e2e/helpers")` path entirely. Test builds now resolve the functions emulator via `Platform.OS` (host) plus `RNFB_{ANDROID,IOS,MACOS}_EMULATOR_FUNCTIONS_PORT` from env (inlined by Babel in the test bundle), with a `Number.isFinite` guard on parse.

tests/.detoxrc.js
```diff
@@ -1,101 +1,188 @@
...
+  return [
+    intEnv('RNFB_ANDROID_EMULATOR_FIRESTORE_PORT', 8080),
+    intEnv('RNFB_ANDROID_METRO_PORT', intEnv('RCT_METRO_PORT', 8081)),
+    intEnv('RNFB_ANDROID_JET_PORT', intEnv('JET_REMOTE_PORT', 8090)),
+    intEnv('RNFB_ANDROID_EMULATOR_DATABASE_PORT', 9000),
+    intEnv('RNFB_ANDROID_EMULATOR_AUTH_PORT', 9099),
+    intEnv('RNFB_ANDROID_EMULATOR_STORAGE_PORT', 9199),
+    intEnv('RNFB_ANDROID_EMULATOR_FUNCTIONS_PORT', 5001),
+  ];
+}
```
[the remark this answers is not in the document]
mikehardy: Fixed: dropped the `if (!firestore)` guard so partial overrides work — `androidReversePortsFromEnv()` always builds the reverse-port list via `intEnv()` with serial defaults. Added `Number.isFinite` in `intEnv` so bad env values fall back instead of producing NaN.

scripts/e2e/create-android-avds.sh
```diff
@@ -0,0 +1,49 @@
...
+  local src_avd="${AVD_DIR}/${src}.avd"
+  local dst_ini="${AVD_DIR}/${dst}.ini"
+  local dst_avd="${AVD_DIR}/${dst}.avd"
+
+  echo "[avd] cloning $src → $dst"
+  cp -R "$src_avd" "$dst_avd"
+  cp "$src_ini" "$dst_ini"
+  sed -i.bak "s/${src}/${dst}/g" "$dst_ini" "${dst_avd}/config.ini" 2>/dev/null || \
+    sed -i '' "s/${src}/${dst}/g" "$dst_ini" "${dst_avd}/config.ini"
+  rm -f "${dst_ini}.bak"
```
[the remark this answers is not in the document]
mikehardy: Fixed: replaced the cross-platform `sed -i` dance with a small inline `node -e` that rewrites both `.ini` and `config.ini` via string split/join (no backup files). We prefer Node over perl for portability in this repo.

scripts/e2e/create-ios-simulators.sh
```diff
@@ -0,0 +1,24 @@
...
+set -euo pipefail
+
+COUNT="${1:-4}"
+BASE_NAME="${RNFB_IOS_BASE_SIMULATOR:-iPhone 17}"
+RUNTIME=$(xcrun simctl list runtimes available -j | node -e "
+  const j=JSON.parse(require('fs').readFileSync(0,'utf8'));
+  const ios=j.runtimes.filter(r=>r.isAvailable&&r.platform==='iOS').sort((a,b)=>b.version.localeCompare(a.version));
+  if(!ios.length) process.exit(1);
+  console.log(ios[0].identifier);
+")
```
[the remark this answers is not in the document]
mikehardy: Fixed: runtime selection now compares version components numerically (split on `.`, pad missing segments with 0) instead of `localeCompare`, so e.g. 17.10 sorts above 17.2.

tests/e2e/firebase.test.js
```diff
@@ -38,12 +38,65 @@ function logCloudPressureAnalysisPointer(context) {
...
+  }
+  try {
+    if (typeof detox !== 'undefined' && detox?.device?.getPlatform) {
+      return detox.device.getPlatform();
+    }
+  } catch (_e) {
+    // detox not ready
+  }
+  return 'android';
+}
```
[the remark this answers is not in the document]
mikehardy: Partially adopted: `detoxPlatformKey()` prefers `detox.device.getPlatform()` (Jest host context) with a bare-`device` fallback where available; callers use `detoxPlatformKey()` instead of unguarded `device.getPlatform()`. Bare `device` alone is not reliable in `firebase.test.js` host orchestration — we saw `device is not defined` when using it directly.

packages/app/e2e/helpers.js
```diff
@@ -8,3 +8,104 @@ exports.getE2eEmulatorHost = function getE2eEmulatorHost() {
...
+  return DEFAULT_JET_SERIAL;
+};
+
+exports.getJetRemoteUrl = function getJetRemoteUrl() {
+  const host = exports.getE2eEmulatorHost();
+  return `ws://${host}:${exports.getJetRemotePort()}`;
+};
+
+exports.getMetroPort = function getMetroPort() {
+  if (process.env.RCT_METRO_PORT) {
```
russellwheatley: priority order here is flipped compared to jetrc and firebase.test.js. those check the platform specific RNFB var before RCT_METRO_PORT, but this checks RCT_METRO_PORT first. since RCT_METRO_PORT is often already set by RN tooling, this will silently point the bundle at the wrong metro port in slotted runs. can we match the other two and check platform specific first?

packages/functions/e2e/functions.e2e.js
```diff
@@ -88,6 +88,21 @@ const SAMPLE_DATA = {
   ],
 };
 
+const { getE2eEmulatorHost, getE2eEmulatorPort } = require('../../app/e2e/helpers');
+
+function functionsEmulatorConnectHost() {
+  const host = getE2eEmulatorHost();
+  return host === '10.0.2.2' ? '10.0.2.2' : 'localhost';
```
russellwheatley: this returns localhost for non android but functionsCallableUrl below uses 127.0.0.1 from getE2eEmulatorHost directly. can we drop this helper and just use getE2eEmulatorHost everywhere so both paths agree

russellwheatley (changes requested):

mikehardy:
will come back to this shortly

packages/app/e2e/helpers.js
mikehardy: Fixed on the rebased `e2e-parallel` tree: `getMetroPort()` now prefers the platform-prefixed `RNFB_{ANDROID,IOS,MACOS}_METRO_PORT` over `RCT_METRO_PORT` (then unprefixed `RNFB_METRO_PORT` / `JET_METRO_PORT`). Jest asserts that in `packages/app/__tests__/e2eHelpers.test.js` (`prefers the platform-prefixed port over every other source`).

packages/functions/e2e/functions.e2e.js
mikehardy: Fixed: the extra localhost helper is gone. `packages/functions/e2e/functions.e2e.js` uses `getE2eEmulatorHost()` for both `connectFunctionsEmulator` and `functionsCallableUrl` (Android `10.0.2.2`, iOS/macOS `127.0.0.1` — never `localhost`).

packages/ai/lib/requests/request.ts
```diff
@@ -57,11 +57,25 @@ export class RequestUrl {
         'Running VertexAI in test environment, pointing to Firebase Functions emulator URL',
       );
       const isAndroid = Platform.OS === 'android';
+      const isMacos = (Platform.OS as string) === 'macos';
+      let fnHost = isAndroid ? '10.0.2.2' : '127.0.0.1';
```
mikehardy: should be app package helper function gete2eemulatorhost I think?

packages/ai/lib/requests/request.ts
```diff
@@ -57,11 +57,25 @@ export class RequestUrl {
...
+      // Static process.env.RNFB_* members — babel transform-inline-environment-variables
+      // does not rewrite computed keys (see okf-bundle/testing/running-e2e.md).
+      let envPort: string | undefined;
+      if (isAndroid) {
+        envPort = process.env.RNFB_ANDROID_EMULATOR_FUNCTIONS_PORT;
+      } else if (isMacos) {
+        envPort = process.env.RNFB_MACOS_EMULATOR_FUNCTIONS_PORT;
+      } else {
+        envPort = process.env.RNFB_IOS_EMULATOR_FUNCTIONS_PORT;
```
mikehardy: in all platform switch cases like this, I prefer deterministic tests that positively confirm platform, and a default case that errors indicating it was an unknown platform. In case more platforms are added in future they won't fail or collide silently. This likely applies in the whole PR, scan whole PR and update

packages/app/__tests__/e2eHelpers.test.js
```diff
@@ -0,0 +1,181 @@
...
+  'RNFB_ANDROID_JET_PORT',
+  'RNFB_IOS_JET_PORT',
+  'RNFB_MACOS_JET_PORT',
+  'RNFB_ANDROID_EMULATOR_FIRESTORE_PORT',
+  'RNFB_IOS_EMULATOR_FIRESTORE_PORT',
+  'RNFB_MACOS_EMULATOR_FIRESTORE_PORT',
+  'JET_REMOTE_PORT',
+  'JET_METRO_PORT',
+  'RCT_METRO_PORT',
+  'RNFB_METRO_PORT',
```
mikehardy, line 40: comment for other reviewers: to my knowledge this is the full set of resources that suffer contention in our e2e tests on the app side when you run e2e parallel. The whole PR is just about setting these, and wiring them through

.github/workflows/scripts/firebase.emulator.template.json
```diff
@@ -0,0 +1,26 @@
...
+  "storage": {
+    "rules": "storage.rules"
+  },
+  "emulators": {
+    "auth": { "port": 9099 },
+    "database": { "port": 9000 },
+    "firestore": { "port": 8080 },
+    "functions": { "port": 5001 },
+    "storage": { "port": 9199 },
+    "ui": { "enabled": false }
```
mikehardy: why disable the UI? It might be useful for someone running manually in a parallel slot? Any reason to disable it? Analyze if we can leave it up with another port assignment wired through

tests-macos/macos/io.invertase.testing-macOS/AppDelegate.mm
```diff
@@ -5,9 +5,18 @@
 
 static NSString *const RNFBTestingMetroHost = @"127.0.0.1";
 
+static NSUInteger RNFBTestingMetroPortNumber(void)
+{
+  NSString *envPort = [[[NSProcessInfo processInfo] environment] objectForKey:@"RCT_METRO_PORT"];
```
mikehardy, line 10: turns out metro port is baked into the test app, so we have to wire it in directly for apple platform test apps

tests-macos/.babelrc
```diff
@@ -4,7 +4,37 @@
     [
       "transform-inline-environment-variables",
       {
-        "include": ["CI"]
+        "include": [
+          "CI",
+          "RNFB_E2E_PLATFORM",
```
mikehardy, line 9: Hermes can't see system environment variables at runtime since it is on device

so we use a babel transform to interpolate them into the bundle during metro transform, carrying system env vars into the Hermes runtime

.github/workflows/scripts/resolve-ios-simulator-name.sh
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env bash
+# Resolve the iOS simulator *device name* for boot-simulator.sh (CI + Jet reboot).
+# Prefer RNFB_IOS_SIMULATOR (slot env already exports `RNFB E2E iOS slot-N`;
+# serial default is `iPhone 17`). Detoxrc fallback skips comment lines so a
```
mikehardy: putting `iPhone 17` here violates DRY and gives us a second place to update it (or forget it) when we move to iPhone 18 etc. Describe where it is defined, not what the definition is.

scripts/e2e/lib/e2e-slot-env.sh
```diff
@@ -0,0 +1,311 @@
+#!/usr/bin/env bash
+# Slotted e2e port + device identity helpers (shared by export/run scripts).
```
mikehardy, line 2: This is the heart of the user experience - "slots" are blocks of ports, with fixed offset from base

So if e2e now has all used ports resolved dynamically, this lets us think in terms of a slot where an extra e2e can run at once with no contention because it stays within one range of ports that doesn't hit another range. Proven with exhaustive sequential 3 slot / 3 platform at once testing (9 e2e all running in parallel)

scripts/e2e/create-android-avds.sh
```diff
@@ -0,0 +1,59 @@
+#!/usr/bin/env bash
+# Clone TestingAVD into TestingAVD-0 … TestingAVD-(count-1) for parallel e2e slots.
```
mikehardy, line 2: this and the simulator create below set up your dev env with multiple emulators and simulators so you can scale out horizontally - you can't run the same simulator/emulator in parallel, they require separation

scripts/e2e/lib/e2e-resource-env.sh
```diff
@@ -0,0 +1,562 @@
+#!/usr/bin/env bash
```
mikehardy, line 1: reviewer note: for reliable fully automated runs we need to control a slot deterministically - the ability to locate and terminate any stale/hung e2e processes allocated to a slot, so it may be used again safely. This whole lib / check / release script area is to do that resolution of e2e resources running per slot, then either report, or terminate them.

tests-macos/macos/io.invertase.testing.xcodeproj/project.pbxproj
```diff
@@ -272,7 +272,7 @@
 					"-lswiftCompatibilityPacks",
 				);
 				PRODUCT_BUNDLE_IDENTIFIER = "org.reactjs.native.$(PRODUCT_NAME:rfc1034identifier)";
-				PRODUCT_NAME = io.invertase.testing;
+				PRODUCT_NAME = "io.invertase.testing$(RNFB_MACOS_PRODUCT_NAME_SUFFIX)";
```
mikehardy, line 275: key insight: only one app with same bundle id at a time can run on macOS but suffixing it to have unique bundle ids works just fine. Unlocks macOS parallel execution

.github/workflows/scripts/start-firebase-emulator.sh
```diff
@@ -1,44 +1,233 @@
...
+      echo "error: cannot start emulator suite — suite ports must be free. Abort; zero flake budget." >&2
+      exit 1
+    fi
+  done
+fi
+
+CONFIG="${SCRIPTS}/.e2e-emulator-${TAG}-${SLOT}.json"
+python3 - <<PY
+import json
+tpl = json.load(open("${SCRIPTS}/firebase.emulator.template.json"))
```
mikehardy, line 148: reviewer comment: key insight is that firebase emulator port contention can be resolved, but only by templating the config and interpolating different ports into it. There is no dynamic or command line way to set what port to use, it is config only.

scripts/e2e/start-packager.sh
```diff
@@ -0,0 +1,109 @@
...
+  sleep 1
+  pids="$(e2e_listener_pids "$port")"
+  if [[ -n "${pids}" ]]; then
+    echo "[packager] SIGKILL :${port} pids=$(echo "$pids" | tr '\n' ' ')"
+    # shellcheck disable=SC2086
+    kill -9 $pids 2>/dev/null || true
+  fi
+}
+
+watchman_reset_allowlist() {
```
mikehardy, line 60: when you are scaling horizontally in the large, watchman becomes a problem, this scopes watchman and resets it to keep fileystem activity down

.yarn/patches/detox-npm-20.51.0-3e13b6e309.patch
```diff
@@ -122,15 +122,28 @@ index d6db5d5d7a4173eae88d97bf06eaeaf8b38c3b94..a85ebdf5a30607459be68463b3238b6f
...
+diff --git a/src/devices/common/drivers/android/emulator/exec/EmulatorExec.js b/src/devices/common/drivers/android/emulator/exec/EmulatorExec.js
+index 0000000000000000000000000000000000000000..0000000000000000000000000000000000000000 100644
+--- a/src/devices/common/drivers/android/emulator/exec/EmulatorExec.js
++++ b/src/devices/common/drivers/android/emulator/exec/EmulatorExec.js
+@@ -70,6 +70,8 @@ class LaunchCommand extends ExecCommand {
+       '-no-boot-anim',
+       headless ? '-no-window' : '',
+       readonly ? '-read-only' : '',
++      '-crash-report-mode',
++      'never',
```
mikehardy, line 158: android emulator will hard hang on teardown if it's crash report dialog pops up. rare but annoying flake vector

package.json
```diff
@@ -34,51 +34,57 @@
     "tests:jest-watch": "jest --watch",
     "tests:jest-coverage": "jest --coverage",
     "tests:packager:chrome": "cd tests && yarn react-native start --reset-cache",
-    "tests:packager:jet": "cd tests && yarn react-native start --client-logs",
-    "tests:packager:jet-ci": "cd tests && (mkdir $HOME/.metro || true) && cross-env TMPDIR=$HOME/.metro yarn react-native start",
-    "tests:packager:jet-reset-cache": "cd tests && yarn react-native start --reset-cache",
+    "tests:packager:jet": "bash ./scripts/e2e/start-packager.sh tests --client-logs",
+    "tests:packager:jet-ci": "bash ./scripts/e2e/start-packager.sh tests",
+    "tests:packager:jet-reset-cache": "bash ./scripts/e2e/start-packager.sh tests --reset-cache",
```
mikehardy, line 40: all the changes here have the goal of using scripts to interpolate slots - carrying env vars into their various spots before actually starting things

mellifera/src/cli.ts
```diff
@@ -0,0 +1,244 @@
+#!/usr/bin/env node
```
mikehardy: there are 2 mellifera-related commits in this PR however those should be excluded
mellifera is co-developed in-tree but I want to keep it separate, the e2e contention PR is standalone and is the "receiver" of "slot" assignments - those may be coordinated by any system so it's fully separate from mellifera or anything else. Drop these two commits, make sure the following commit still works

okf-bundle/testing/e2e-parallel-design.md
```diff
@@ -0,0 +1,570 @@
+---
+type: Reference
+title: E2e parallel execution and host coordination
```
mikehardy, line 3: reviewer note: this is the most important item to read

mikehardy (commented):
close - a couple comments but functioning well during torture testing local

packages/ai/lib/requests/request.ts
mikehardy: Done in `c1bbb1b83` (amended into the introducing commit): `request.ts` imports `getE2eEmulatorHost` and `getE2eEmulatorPort('functions')` from `@react-native-firebase/app/e2e/helpers` — no inline `Platform.OS` routing and no dynamic `require`.

packages/ai/lib/requests/request.ts
mikehardy: Done — folded into the commits that introduced each platform router, not as follow-up commits:
- `packages/app/e2e/helpers.js` (`c1bbb1b83`): explicit android/ios/macos branches with a throwing `default`; covered by `e2eHelpers.test.js`.
- `tests/.jetrc.js` (`c1bbb1b83`): same strict switch pattern.
- `tests/e2e/firebase.test.js` (`86a495654`): `detoxPlatformKey()` port helpers use explicit cases + throwing defaults.
`request.ts` delegates to helpers, so there is no second switch table in AI.

.github/workflows/scripts/firebase.emulator.template.json
mikehardy: Analyzed in `0ed86b6e0` (amended into the emulator startup commit): UI stays disabled for headless CI — enabling it adds another listener per suite and the slot model already assigns distinct suite ports. Template keeps `"ui": { "enabled": false, "port": 4000 }` as a placeholder; `start-firebase-emulator.sh` documents why and how a per-slot UI port could be wired later for manual slot debugging.

.github/workflows/scripts/resolve-ios-simulator-name.sh
mikehardy: Fixed in `2543cbea5`: comment now references `scripts/e2e/create-ios-simulators.sh` / `E2E_DEFAULT_IOS_SIMULATOR` / `RNFB_IOS_BASE_SIMULATOR` instead of hardcoding `iPhone 17`.

mellifera/src/cli.ts
mikehardy: Dropped — history rewritten to exclude both mellifera commits. Branch is now 10 commits vs `main` with no `mellifera/` tree or `scripts/e2e/mellifera-*` scripts. Mellifera remains co-developed out-of-band; this PR is the standalone contention-resolution layer.

scripts/e2e/release-e2e-resources.sh
```diff
@@ -0,0 +1,336 @@
...
+  local name
+  while IFS= read -r name; do
+    [[ -z "$name" ]] && continue
+    echo "[release] simctl shutdown ${name}"
+    xcrun simctl shutdown "$name" 2>/dev/null || true
+  done < <(e2e_ios_simulator_names_for_release)
+  # Unscoped serial: also shut down any other booted sim (same as historical iPhone 17).
+  if [[ -z "${RNFB_IOS_SIMULATOR:-}" && -z "${RNFB_E2E_SLOT:-${RNFB_E2E_HOST_SLOT:-}}" ]]; then
+    xcrun simctl shutdown booted 2>/dev/null || true
+  fi
```
russellwheatley: This unconditionally shuts down every booted simulator, not just the serial default, whenever the run is unscoped. `clear_macos_app_hard`/`clear_android_apps` above only ever touch their own serial identity when unscoped; iOS is the one path here where an unscoped release reaches out and kills a sibling slot's device.

a plain unscoped `yarn tests:e2e:release` while another terminal has a slotted iOS run in flight will shut down that run's simulator. could we this scope to the serial device name instead of `booted`, the same way android/macos do below?

russellwheatley (changes requested):
Just the below was the major issue surfaced.

scripts/e2e/release-e2e-resources.sh
mikehardy: Fixed in the rebased branch (amended into `test(e2e): add environment-aware host lifecycle controls`):

- Removed the unscoped `simctl shutdown booted` fallback from `clear_ios_sims()`. Unscoped serial `yarn tests:e2e:release --devices` now shuts down only the resolved serial simulator name (`iPhone 17` via `e2e_ios_simulator_names_for_release`), matching Android/macOS identity scoping.
- Slotted runs (`export-slot-env` / `RNFB_IOS_SIMULATOR`) were already scoped to `RNFB E2E iOS slot-N` only.

Also folded in full review remediation (platform validation, `E2E_SLOTTED_MAX=4`, etc.) — see latest force-push on `e2e-parallel` (`0c5b8bb09`).

russellwheatley (approved):

PR #9108, opened 2026-07-17
fix(ios): set IPHONEOS_DEPLOYMENT_TARGET to 15.0 across all packages

### LINKED ISSUE #8882: this `@available` annotation shouldn't be necessary unless the pbxproj IPHONEOS_DEPLOYMENT_TARGET is wrong
this `@available` annotation shouldn't be necessary unless the pbxproj IPHONEOS_DEPLOYMENT_TARGET is wrong

In fact no, it is set to 10 ! That's not correct

4 places, here is an example https://github.com/invertase/react-native-firebase/blob/13d6bb3dfa3328fe987179d6e342d1551a208a38/packages/functions/ios/RNFBFunctions.xcodeproj/project.pbxproj#L162

confirmed iOS 15 is already our minimum supported iOS version - we're issuing the merge result of this as a breaking change but it's related to functions going new architecture only, this iOS version doesn't have to be mentioned.

As a separate issue / peeled from this review (right-click on this comment and create issue - I formatted it well for that), we should raise the deploy target to 15 in the package pbxproj's to match current requirements

_Originally posted by @mikehardy in https://github.com/invertase/react-native-firebase/pull/8799#discussion_r2790000283_

just1and0:
## Summary
- Updates `IPHONEOS_DEPLOYMENT_TARGET` from `10.0` to `15.0` in all 16 package pbxproj files and the `_TEMPLATE_` pbxproj
- Aligns with the `iosTarget: "15.0"` already declared in `packages/app/package.json` `sdkVersions`
- Prevents new packages generated from the template from reintroducing the stale `10.0` value

- Fixes #8882

## Test plan
- [x] Verified zero `IPHONEOS_DEPLOYMENT_TARGET = 10.0` values remain across all pbxproj files
- [x] `git diff --check` clean — no whitespace issues
- [ ] CI iOS build passes

### DIFF
--- a/.spellcheck.dict.txt
+++ b/.spellcheck.dict.txt
@@ -3,6 +3,7 @@
 Ad
 AdMob
 AirPods
+ai
 analytics
 Analytics
 args
--- a/docs/platforms.mdx
+++ b/docs/platforms.mdx
@@ -9,11 +9,10 @@ next: /faqs-and-tips
 
 By default React Native Firebase supports multiple platforms using the native Firebase SDK for the specific platform:
 
-| Platform | Minimum Version                      |
-| -------- | ------------------------------------ |
-| Android  | 5.0 (Lollipop) without Firebase Auth |
-|          | 6.0 (Marshmallow) with Firebase Auth |
-| iOS      | 13.0                                 |
+| Platform | Minimum Version            |
+| -------- | -------------------------- |
+| Android  | 6.0 (API 23 / Marshmallow) |
+| iOS      | 15.0                       |
 
 **Important** - To compile your application for Android, you will need a minimum JDK (Java Development Kit) version of `>= 17`.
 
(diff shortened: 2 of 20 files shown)


mikehardy (changes requested):
this is a good change as a start - can you add a cleanup chore commit that scans the entire ios codebase and cleans up any stale API availability checks that were looking for iOS 15 or below?

If there is a similar requirement that we're past for tvos/visionos/ipados/maccatalyst etc - that is, a deployment target that's stale and could be updated so the compiler knows more APIs are available - allowing us to clean out availability gates for those platforms, it would be good to get everything at once while we're here

just1and0:
Addressed in 8a96a5d — scanned all `@available` guards across the iOS codebase (10 total).

**Removed (2 guards):**
- `RNFBMessaging+UNUserNotificationCenter.m` — two `@available(iOS 14, *)` checks for `UNNotificationPresentationOptionBanner` and `UNNotificationPresentationOptionList`. With iOS 15.0 as the minimum, these are always available. Removed the dead iOS 13 fallback `else` branches and the stale TODO comment.

**Kept (8 guards):**
- Analytics, Functions, App Check, and Messaging all have mixed-platform guards like `@available(iOS 15.0, macOS 12.0, *)` or `@available(iOS 14.0, macOS 11.0, macCatalyst 14.0, ...)`. These are still meaningful because `macosTarget` is `10.15` — the macOS/macCatalyst availability checks are still load-bearing even though the iOS side is now redundant.

**Other platforms:**
- No stale tvOS/visionOS/iPadOS deployment targets found. tvOS is already `15.0` through podspecs, iPadOS follows the iOS target, and there's no explicit visionOS or macCatalyst target to bump. macCatalyst availability is handled at the source level with `@available` guards which we're keeping.

mikehardy (approved):
did a final pass - a little sad we can't remove any more of those conditionals - removing conditionality is a nice background task to pursue in general as it maintains simplicity but my audit of platform requirements and support matches up with current changes

during that audit I noticed that our platforms doc was stale for iOS minimum as well though, so I put in a final commit that bumps iOS to 15 from 13 and fixed up the Android minimum and adds the ai and phone-number-verification modules to platform support table as well

Ci should go green on lint and markdown which would verify my change then I think this is good to go

PR #9117, opened 2026-07-28
fix(app-check, ios): do not default to the debug provider in release builds

### LINKED ISSUE #9116: 🔥 [iOS] App Check installs the debug provider in release builds — config plugin registers the factory before FirebaseApp.configure()
## Issue

On iOS, release builds using the Expo config plugin exchange **App Check *debug* tokens** against `firebaseappcheck.googleapis.com`, receiving `403 App attestation failed` — and `429` once the per-project `Debug Token exchange requests per minute` quota (60/min, not adjustable) saturates.

This is the defect reported in discussion [#7518](https://github.com/invertase/react-native-firebase/discussions/7518) (open since 2023-12-19) and on [StackOverflow 77673428](https://stackoverflow.com/questions/77673428/react-native-with-firebase-appcheck-uses-exchangedebugtoken-url-in-release-build). Every report there assumes a race condition. **It is not a race — it is deterministic, and it is the line order the config plugin emits.**

Android is unaffected.

### Root cause

The config plugin injects this into `AppDelegate` (`plugin/src/ios/appDelegate.ts`):

```swift
RNFBAppCheckModule.sharedInstance()
    FirebaseApp.configure()
```

The resulting sequence:

1. `+[RNFBAppCheckModule sharedInstance]` calls `[FIRAppCheck setAppCheckProviderFactory:]` with a freshly allocated `RNFBAppCheckProviderFactory` whose `providers` dictionary is **nil** (`RNFBAppCheckModule.m:36-45`).
2. `FirebaseApp.configure()` runs `[app.container instantiateEagerComponents]` (`FIRApp.m`).
3. `FIRAppCheckInterop` is registered with `FIRInstantiationTimingAlwaysEager` (`FIRAppCheckComponent.m`), so its creation block runs *inside* `configure()`.
4. `[[FIRAppCheck alloc] initWithApp:]` finds a non-nil factory and calls `createProviderWithApp:`. With `providers[app.name] == nil`, the factory installs `providerName:@"debug" debugToken:nil` (`RNFBAppCheckProviderFactory.m:34-39`) → `FIRAppCheckDebugProvider`.

`FIRAppCheckDebugProvider` with no registered token generates a random one and exchanges it. Hence `exchangeDebugToken` traffic from release builds.

**The line order is the whole mechanism.** `-[FIRAppCheck initWithApp:]` returns **nil** when no factory is registered — it fails safely and does not crash. Had `FirebaseApp.configure()` run *first*, no App Check instance would exist at that point and the JS layer's later `configureProvider` call would install the correct provider as the first and only one.

No other Firebase product can be the creator on this ordering: a successfully created instance is cached by `FIRComponentContainer` (`*isCacheable = YES`), so later consumers (`FIRAuth`, Messaging, Installations) receive the already-built instance and cannot re-trigger provider selection.

### Why it appears intermittent (~20%, not 100%)

The debug provider is installed on *every* cold start, but an exchange only happens when both hold: the eager auto-refresh reaches the provider before JS `configureProvider` swaps the delegate, and no live keychain-cached token exists. That timing dependence is why everyone in #7518 concluded "race" — the *exchange* races, but the *misconfiguration* is deterministic.

This also explains @brianGammon's finding in #7518 that `isTokenAutoRefreshEnabled: false` drops the 403 graph to ~0: it removes the eager refresh that reaches the wrongly-defaulted provider.

### Android is immune by construction

`ReactNativeFirebaseAppCheckProviderFactory.create()` throws rather than defaulting to debug. Only iOS silently substitutes a debug provider.

### Suggested fix

Make the pre-configure default safe outside `DEBUG` builds:

```objc
#if DEBUG
    NSString *defaultProviderName = @"debug";
#else
    NSString *defaultProviderName = @"appAttestWithDeviceCheckFallback";
#endif
```

This fixes the outcome regardless of who calls the factory or in what order, and keeps debug-token development working locally. `appAttestWithDeviceCheckFallback` already handles the availability fallback in `RNFBAppCheckProvider.m`.

A complete fix arguably needs both layers — the safe default **and** corrected ordering in the config plugin so the factory is not registered before `FirebaseApp.configure()`. I've opened a PR for the factory default only, since changing the plugin's emitted `AppDelegate` affects every consumer and I can't test that blast radius. Happy to follow up on the plugin if you'd prefer that approach.

The debug default was introduced in `ee7df855` ("feat(app-check): add custom factory/provider", 2023-01-19) and is unchanged on `main` today.

---

### Javascript

<p>

#### `package.json`:

```json
{
  "dependencies": {
    "@react-native-firebase/app": "22.3.0",
    "@react-native-firebase/app-check": "22.3.0"
  }
}
```

```json
# N/A — Expo config plugin used, no firebase.json
```

</p>
</details>

### iOS

<p>

#### `ios/Podfile`:

- [ ] I'm not using Pods

```ruby
# Standard Expo-generated Podfile (expo prebuild), no App Check customisation.
```

#### `AppDelegate.m`:

```objc
// AppDelegate.swift, generated by `expo prebuild` — plugin-injected block verbatim:
//
// @generated begin @react-native-firebase/app-check - expo prebuild (DO NOT MODIFY)
RNFBAppCheckModule.sharedInstance()
    FirebaseApp.configure()
// @generated end @react-native-firebase/app-check
//
// Note the ordering: sharedInstance() registers the provider factory BEFORE
// FirebaseApp.configure() runs its eager components.
```

</p>
</details>

---

### Android

<p>

#### `android/build.gradle`:

```groovy
// N/A — Android is not affected; its factory throws instead of defaulting to debug.
```

```groovy
// N/A
```

```groovy
// N/A
```

#### `MainApplication.java`:

```java
// N/A
```

#### `AndroidManifest.xml`:

```xml
<!-- N/A -->
```

</p>
</details>

---

## Environment

<p>

```
System:
  OS: macOS 26.5.2
  CPU: (14) arm64 Apple M4 Pro
  Memory: 7.30 GB / 48.00 GB
  Shell:
    version: "5.9"
    path: /bin/zsh
Binaries:
  Node:
    version: 24.16.0
    path: ~/.nvm/versions/node/v24.16.0/bin/node
  Yarn: Not Found
  npm:
    version: 8.19.4
    path: <project>/node_modules/.bin/npm
  Watchman:
    version: 2026.07.20.00
    path: /opt/homebrew/bin/watchman
Managers:
  CocoaPods:
    version: 1.17.0
    path: /opt/homebrew/bin/pod
SDKs:
  iOS SDK:
    Platforms:
      - DriverKit 25.2
      - iOS 26.2
      - macOS 26.2
      - tvOS 26.2
      - visionOS 26.2
      - watchOS 26.2
  Android SDK:
    API Levels:
      - "35"
      - "36"
      - "36"
    Build Tools:
      - 34.0.0
      - 35.0.0
      - 36.0.0
      - 36.1.0
      - 36.1.0
    System Images:
      - android-35 | Google APIs ARM 64 v8a
      - android-36.1 | Google APIs ARM 64 v8a
      - android-36.1 | Google Play ARM 64 v8a
      - android-36 | Google Play ARM 64 v8a
    Android NDK: Not Found
IDEs:
  Android Studio: 2025.2 AI-252.28238.7.2523.14688667
  Xcode:
    version: 26.2/17C52
    path: /usr/bin/xcodebuild
Languages:
  Java:
    version: 17.0.20
    path: /Library/Java/JavaVirtualMachines/zulu-17.jdk/Contents/Home/bin/javac
  Ruby:
    version: 4.0.6
    path: /opt/homebrew/bin/ruby
npmPackages:
  "@react-native-community/cli":
    installed: 20.0.0
    wanted: 20.0.0
  react:
    installed: 19.1.0
    wanted: 19.1.0
  react-native:
    installed: 0.81.5
    wanted: 0.81.5
  react-native-macos: Not Found
npmGlobalPackages:
  "*react-native*": Not Found
Android:
  hermesEnabled: Not found
  newArchEnabled: Not found
iOS:
  hermesEnabled: Not found
  newArchEnabled: Not found

info React Native v0.86.2 is now available (your project is running on v0.81.5).
info Changelog: https://github.com/facebook/react-native/releases/tag/v0.86.2
info Diff: https://react-native-community.github.io/upgrade-helper/?from=0.81.5&to=0.86.2
info For more info, check out "https://reactnative.dev/docs/upgrading?os=macos".
```

  - [ ] iOS
  - [ ] Android
  - [ ] Both
  - `22.3.0` (also present on `main` — the file is byte-identical)
  - `App Check`
  - `Y` & `5.9.3`

</p>
</details>

kolezka:
### Description

On iOS, release builds using the Expo config plugin exchange App Check **debug** tokens against `firebaseappcheck.googleapis.com`, getting `403 App attestation failed` and then `429` once the per-project debug-token quota (60/min, not adjustable) saturates.

The config plugin emits `RNFBAppCheckModule.sharedInstance()` **before** `FirebaseApp.configure()` in `AppDelegate`. `sharedInstance()` registers the provider factory with a nil `providers` dictionary; `configure()` then instantiates `FIRAppCheckInterop` eagerly (`FIRInstantiationTimingAlwaysEager`), which calls `createProviderWithApp:` while `providers[app.name]` is still nil. The factory defaults that case to `@"debug"`, so release builds get `FIRAppCheckDebugProvider` and exchange an unregistered token.

This changes the pre-configure default to `appAttestWithDeviceCheckFallback` outside `DEBUG` builds, matching what the JS layer configures moments later anyway. `DEBUG` builds keep the debug default, which local development with debug tokens relies on. `RNFBAppCheckProvider` already handles the iOS 14 / tvOS 15 availability fallback for that provider name.

Fixing the factory rather than the plugin's line order means the outcome holds regardless of who calls the factory or in what sequence, so it survives future plugin and SDK changes. **The ordering is arguably the deeper fix**, but changing the emitted `AppDelegate` affects every consumer, and I can't test that blast radius across Objective-C and Swift AppDelegates and older plugin versions — so I've left it alone and described it in the issue instead. Happy to follow up if you'd prefer that direction, or a different non-debug default.

Android is unaffected — `ReactNativeFirebaseAppCheckProviderFactory.create()` throws rather than defaulting to debug.

The debug default dates to `ee7df855` ("feat(app-check): add custom factory/provider", 2023-01-19) and is unchanged on `main`.

**On tests:** there is no existing jest or e2e coverage for `RNFBAppCheckProviderFactory` — it is native Objective-C with no test target in this package, and the behaviour needs a release-configuration build to observe. I have left the test checkboxes unticked rather than claim coverage that does not exist; happy to add one if you can point me at where it should live.

### Related issues

Fixes #9116

Previously reported, uncaused, in discussion [#7518](https://github.com/invertase/react-native-firebase/discussions/7518) (open since 2023-12-19) and [StackOverflow 77673428](https://stackoverflow.com/questions/77673428/react-native-with-firebase-appcheck-uses-exchangedebugtoken-url-in-release-build). Both threads assume a race condition; the misconfiguration is deterministic — only the exchange timing varies, which is why it presents as intermittent.

### Release Summary

iOS: App Check no longer installs the debug provider in release builds, which caused `exchangeDebugToken` 403/429 traffic and could saturate the project's debug-token quota.

### Checklist

  - [x] Yes
  - [ ] `Android`
  - [x] `iOS`
  - [ ] `Other` (macOS, web)
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### DIFF
--- a/packages/app-check/ios/RNFBAppCheck/RNFBAppCheckProviderFactory.m
+++ b/packages/app-check/ios/RNFBAppCheck/RNFBAppCheckProviderFactory.m
@@ -32,10 +32,26 @@
   }
 
   if (self.providers[app.name] == nil) {
-    DLog(@"provider initializing (with default to debug) for app %@", app.name);
+    // This pre-configure default must not be "debug" in release builds. The Expo config
+    // plugin emits `RNFBAppCheckModule.sharedInstance()` BEFORE `FirebaseApp.configure()`
+    // in AppDelegate, so FIRAppCheckInterop — registered with
+    // FIRInstantiationTimingAlwaysEager — instantiates a provider from inside configure(),
+    // while providers[app.name] is still nil. Defaulting to debug therefore installs
+    // FIRAppCheckDebugProvider on release builds, which exchanges an unregistered debug
+    // token against firebaseappcheck.googleapis.com: 403 "App attestation failed", and 429
+    // once the per-project debug-token quota saturates. See discussion #7518.
+    //
+    // DEBUG builds keep the debug default, which local development with debug tokens
+    // relies on.
+#if DEBUG
+    NSString *defaultProviderName = @"debug";
+#else
+    NSString *defaultProviderName = @"appAttestWithDeviceCheckFallback";
+#endif
+    DLog(@"provider initializing (with default to %@) for app %@", defaultProviderName, app.name);
     self.providers[app.name] = [RNFBAppCheckProvider new];
     RNFBAppCheckProvider *provider = self.providers[app.name];
-    [provider configure:app providerName:@"debug" debugToken:nil];
+    [provider configure:app providerName:defaultProviderName debugToken:nil];
   }
 
   return self.providers[app.name];


mikehardy:
Just acknowledging this is on our radar 👋 

We are just finishing up with some massive internal refactors and that will give us space to fix things like this up

For anyone else affected, and fo you @kolezka I can say this an obviously okay temporary fix - if you set the default provider to whatever you intend to configure anyway, it can't possibly be wrong, can it? So doing this while waiting for a real fix is absolutely a fine temporary fix to carry locally via patch-package or yarn patches

russellwheatley:
Thank you for raising the issue - we opted for a more comprehensive fix which you can read about on the PR: https://github.com/invertase/react-native-firebase/pull/9166

it's now been merged and will be out in the next release 🙏

PR #9129, opened 2026-07-28
fix(ios): wire prebuilt React-Core and allow non-modular React imports for use_frameworks! (RN 0.84+)

aravi365:
## Description

React Native 0.83+ ships a **prebuilt React-Core** and enables it by default on 0.84+ (`RCT_USE_PREBUILT_RNCORE=1`). This breaks `react-native-firebase` iOS builds in two ways:

1. **Header resolution** — the legacy `<React/...>` imports in RNFB's Objective-C sources (`RCTConvert.h`, `RCTBridgeModule.h`, `RCTEventEmitter.h`, …) need the prebuilt React-Core wired via React Native's `add_rncore_dependency` helper.

2. **Framework module validation** — when a consumer builds with `use_frameworks!` (Expo's default, and required by the `firebase-ios-sdk`), each RNFB pod is compiled as a **framework module**. Clang then rejects the non-modular `<React/...>` includes with:

   ```
   error: 'React/RCTConvert.h' must be imported from module 'React' before it is required
   [-Werror,-Wnon-modular-include-in-framework-module]
   ```

Both are fixed here across all 16 package podspecs:

- Call `add_rncore_dependency(s)`, guarded with `if defined?(...)` so older `react-native` versions are unaffected.
- Set `CLANG_ALLOW_NON_MODULAR_INCLUDES_IN_FRAMEWORK_MODULES = YES`, **merged** into any existing `pod_target_xcconfig` so per-package settings (e.g. generated TurboModule header search paths in analytics/auth/firestore/storage/etc.) are preserved rather than overwritten.

## Related

- Closes the same class of issue as #8883.
- Alternative to #9024, rebased on current `main`; uses a non-destructive `.merge` on `pod_target_xcconfig` to avoid clobbering the newer per-package `HEADER_SEARCH_PATHS` entries.

## Effect for consumers

Removes the need for the common workarounds:
- `ENV['RCT_USE_PREBUILT_RNCORE'] = '0'` (forces React-Core to rebuild from source — slow clean builds)
- `expo-build-properties` `forceStaticLinking` listing every RNFB package

## Test plan

- [x] `pod install` with `use_frameworks! :linkage => :static` + `$RNFirebaseAsStaticFramework = true` on RN 0.84.x, default `RCT_USE_PREBUILT_RNCORE` (prebuilt), New Architecture.
- [x] Clean iOS build of an app using `@react-native-firebase/{app,analytics,crashlytics,messaging}` — compiles without the `RCT_USE_PREBUILT_RNCORE=0` workaround.
- [x] Ruby syntax check on all podspecs (`ruby -c`) — passing.

### DIFF
--- a/packages/analytics/RNFBAnalytics.podspec
+++ b/packages/analytics/RNFBAnalytics.podspec
@@ -39,6 +39,22 @@ Pod::Spec.new do |s|
 
   install_modules_dependencies(s);
 
+  # RN 0.83+ (default on 0.84+) ships a prebuilt React-Core (RCT_USE_PREBUILT_RNCORE=1).
+  # Wire it up so the legacy <React/...> header imports resolve against the prebuilt
+  # framework. Guarded for older react-native versions where the helper is absent.
+  if defined?(add_rncore_dependency)
+    add_rncore_dependency(s)
+  end
+
+  # RNFB's Objective-C sources import <React/...> headers non-modularly. When a consumer
+  # builds with use_frameworks! (Expo's default, and required by the firebase-ios-sdk),
+  # RNFB is compiled as a framework module and Clang rejects those imports with
+  # -Wnon-modular-include-in-framework-module ("must be imported from module 'React'").
+  # Allow them so the framework module validates. Merge so any existing config is kept.
+  s.pod_target_xcconfig = (s.to_hash["pod_target_xcconfig"] || {}).merge({
+    "CLANG_ALLOW_NON_MODULAR_INCLUDES_IN_FRAMEWORK_MODULES" => "YES",
+  })
+
   if defined?(ENV["RCT_NEW_ARCH_ENABLED"]) != nil && (ENV["RCT_NEW_ARCH_ENABLED"] == '0')
      raise "#{s.name} requires New Architecture. Enable New Architecture to use this module"
   end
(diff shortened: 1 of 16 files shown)


mikehardy:
Hey 👋  thanks for this -

Quick note to say before we look at this, it is going to have to wait for the SPM support to land (it's reviewed+approved and just needs merge):

- #8933 

...which is itself waiting on v26 to launch (there will be a PR for some fixes there, it's about ready to go)

So, hopefully not long, and it's our next priority to fix after SPM lands but it will probably be a conflict bonanza once SPM does land

aravi365:
> Hey 👋 thanks for this -
> 
> Quick note to say before we look at this, it is going to have to wait for the SPM support to land (it's reviewed+approved and just needs merge):
> 
> * [feat(ios): add SPM dependency resolution support alongside CocoaPods #8933](https://github.com/invertase/react-native-firebase/pull/8933)
> 
> ...which is itself waiting on v26 to launch (there will be a PR for some fixes there, it's about ready to go)
> 
> So, hopefully not long, and it's our next priority to fix after SPM lands but it will probably be a conflict bonanza once SPM does land

Thanks @mikehardy, makes sense to wait for SPM to land first. Happy to rebase and clean up the conflicts once it's in, just give me a shout.

russellwheatley:
Thanks for this 🙏

Superseded by #9200, which shipped the Clang non-modular-includes flag. Live pods should not also call `add_rncore_dependency`: `install_modules_dependencies` already does that, and a second call duplicates VFS overlay flags.

Closing this out.

PR #9134, opened 2026-07-31
fix(auth, app): fix web auth initialization crash and missing methods

### LINKED ISSUE #7921: [🐛] Other Platforms - [Web] [Auth] Methods are said to be unsupported
### Is there an existing issue for this?

- [X] I have searched the existing issues.

### Please confirm you are aware of the 'Other' platform limitations.

- [X] I confirm that issue is not relating to a known platform limitation.

### Please confirm, this issue is NOT for Android or iOS?

- [X] I confirm that this issue is not for Android and not for iOS.

### Please describe your issue here.

When using auth module on web, usage of these methods 
Located in `auth\lib\index.js` throw exceptions :

`NativeFirebaseError: [auth/unsupported] This operation is not supported in this environment` for
```js
this.native.configureAuthDomain();
```

`NativeFirebaseError: [auth/unknown] "" is not a function` for 
```js
this.native.addAuthStateListener();
this.native.addIdTokenListener();
this.native
      .signInWithCustomToken(customToken)
      .then(userCredential => this._setUserCredential(userCredential));
```

Firebase app, firestore & database are initiated without any issues.

### Additional context and comments

_No response_

just1and0:
## Summary

Fixes #7921 — Web (Other Platforms) auth was completely broken due to three root causes:

- **`setImmediate` crash**: `setImmediate` is a Node.js/React Native global, not available in browsers. It was used unguarded in `emitEvent()` (utils.ts) and 3 event functions in `RNFBAppModule.ts`, causing `ReferenceError: Can't find variable: setImmediate` on any auth initialization.
- **Missing method stubs**: `verifyPhoneNumberWithMultiFactorInfo` and `resolveTotpSignIn` were absent from the web auth module, causing `"" is not a function` runtime crashes.
- **Misleading AsyncStorage warning**: `getCachedAuthInstance` tried to use `getReactNativePersistence(AsyncStorage)` on web, where IndexedDB is the correct persistence mechanism.

## Changes

- Guard all `setImmediate` calls with `typeof` check, falling back to `setTimeout(fn, 0)` in browsers
- Add `verifyPhoneNumberWithMultiFactorInfo` and `resolveTotpSignIn` stubs that reject with `auth/unsupported`
- Restructure `getCachedAuthInstance` to skip AsyncStorage/persistence setup when `Platform.OS === 'web'`

## Test plan

- [x] 86 new tests (unit + integration), all passing
- [x] Browser-tested via Expo web app: `initializeApp`, `getAuth`, `onAuthStateChanged`, `signInAnonymously` all work
- [x] Verified `setImmediate` fallback fires correctly in browser environment
- [x] Verified no AsyncStorage warning on web
- [x] Existing tests unaffected

Linear: CPRN-282

### DIFF
--- a/packages/app/__tests__/webAppModule.test.ts
+++ b/packages/app/__tests__/webAppModule.test.ts
@@ -1,0 +1,91 @@
+import { describe, expect, it, jest, beforeEach, afterEach } from '@jest/globals';
+
+jest.mock('react-native', () => ({
+  DeviceEventEmitter: {
+    emit: jest.fn(),
+  },
+}));
+
+jest.mock('../lib/internal/web/firebaseApp', () => ({
+  initializeApp: jest.fn(),
+  setLogLevel: jest.fn(),
+  getApp: jest.fn(),
+  getApps: jest.fn(() => []),
+  deleteApp: jest.fn(),
+}));
+
+const appModule = require('../lib/internal/web/RNFBAppModule.ts').default;
+
+describe('RNFBAppModule setImmediate guards', () => {
+  describe('with setImmediate available', () => {
+    beforeEach(() => {
+      jest.useFakeTimers();
+    });
+
+    afterEach(() => {
+      jest.useRealTimers();
+    });
+
+    it('eventsSendEvent uses setImmediate when available', () => {
+      const spy = jest.spyOn(globalThis, 'setImmediate');
+      appModule.eventsAddListener('test_event');
+      appModule.eventsNotifyReady(true);
+      appModule.eventsPing('test_event', { data: 1 });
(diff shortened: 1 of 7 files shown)


mikehardy (approved):
Nice - we have 2 "Other" targets really, a react-native "Other" and a web "Other" where the difference is the javascript runtime - either a DOM-based / typical runtime ("web other") or a react-native Hermes runtime but using firebase-js-sdk ("Hermes Other")

We test pretty well for "Hermes Other", but "Web Other" isn't something we do well historically and this whole mess slipped through. Thanks for cleaning a huge chunk of it up

PR #9144, opened 2026-08-04
docs(ios): correct SPM product linkage and document FirebaseCore sharing limits

### LINKED ISSUE #9140: [🐛] SPM mode: Firebase products link statically into each pod framework, so FirebaseApp.app() is nil in third-party pods
## Issue

In SPM mode (26.1.0, the default on RN >= 0.75), a third-party pod that consumes a Firebase Swift Package product ends up with its **own statically linked copy of FirebaseCore**. The process then contains several `FIRApp` implementations, each with its own registry of configured apps, so `FirebaseApp.configure()` performed at launch is invisible to that pod and `FirebaseApp.app()` returns `nil` there.

I hit this while building a React Native module around `FirebaseDataConnect` (SPM-only, depends on `FirebaseCore`), but nothing about it is specific to Data Connect: it should reproduce with any pod that declares an `spm_dependency` on a firebase-ios-sdk product alongside react-native-firebase.

## Environment

| | |
| --- | --- |
| `@react-native-firebase/app` | 26.1.0 |
| react-native | 0.85.3, New Architecture |
| Podfile | `use_frameworks! :linkage => :dynamic`, SPM mode (no `$RNFirebaseDisableSPM`) |
| Xcode | 26.2, iOS 26.3 simulator |
| CocoaPods | 1.15.2 |
| Third-party pod | `spm_dependency` on `firebase/data-connect-ios-sdk` 11.12.5, product `FirebaseDataConnect` |

## What happens

`pod install` succeeds and reports SPM resolution as expected:

```
[react-native-firebase] RnSqlConnect: Using SPM for Firebase dependency resolution (products: FirebaseCore)
[react-native-firebase] RnSqlConnectExample: Linking FirebaseCore directly into the app target (SPM) ...
```

Package resolution is correct, one version for the whole graph:

```
Firebase: https://github.com/firebase/firebase-ios-sdk.git @ 12.17.0
FirebaseDataConnect: https://github.com/firebase/data-connect-ios-sdk.git @ 11.12.5
```

The app builds and launches. `AppDelegate` calls `FirebaseApp.configure()` (proved by the fact that a malformed `GoogleService-Info.plist` aborts at launch from inside FirebaseInstallations). But the runtime reports duplicated classes:

```
objc[94618]: Class FIRApp is implemented in both
  .../PackageFrameworks/FirebaseCore_-5C7B465D19C2E54A_PackageProduct.framework/...
  and .../RnSqlConnect.framework/RnSqlConnect. One of the duplicates must be removed or renamed.

objc[94618]: Class FIRApp is implemented in both
  .../PackageFrameworks/FirebaseCore_-5C7B465D19C2E54A_PackageProduct.framework/...
  and .../RNFBApp.framework/RNFBApp. One of the duplicates must be removed or renamed.
```

Same for `FIROptions`, `FIRComponentContainer`, `FIRHeartbeatLogger`, `FIRFirebaseUserAgent`, `GULLoggerWrapper` and others.

Confirming that the copies are real rather than just a naming collision:

```sh
$ nm -g .../RNFBApp.framework/RNFBApp | grep -c FIRApp
12
```

Result: any call made from the third-party pod fails, because the copy of FirebaseCore it links has no configured app.

## Why I think this happens

`firebase-ios-sdk`'s `Package.swift` declares its products without an explicit type:

```swift
.library(
  name: "FirebaseCore",
  targets: ["FirebaseCore"]
),
```

```sh
$ grep -c 'type: .dynamic' Package.swift   # firebase-ios-sdk 12.17.0
0
```

Swift Package **automatic** library products are linked statically into every binary that depends on them. Under `use_frameworks! :linkage => :dynamic` each pod is its own dynamic framework, so each pod that references a Firebase product gets a private static copy, while the app target separately gets the `..._PackageProduct.framework` dynamic build.

This may also be worth checking against the comment in `packages/app/firebase_spm.rb`:

> firebase-ios-sdk SPM requires dynamic linkage. There is no upstream statement from Google that SPM+static is supported. (all products use `.library(type: .dynamic)`)

As far as I can tell that parenthetical is not accurate for 12.17.0, and it is the reasoning quoted in `rnfirebase_fail_if_spm_static_linkage!`. The guard itself may well still be correct, I am only flagging that the stated justification did not match what I found, and it sent me down the wrong path for a while when debugging.

## Impact

- Any third-party pod consuming a Firebase SPM product next to react-native-firebase.
- Beyond the `nil` app: `FirebaseAuth` splits the same way. A pod would read "no signed-in user" from its own copy even though the app signed in through react-native-firebase. That failure is silent, which is the part that worries me most: for my use case it would mean `@auth(USER)` operations quietly returning unauthorized while `@auth(PUBLIC)` ones work.

## Reproduction

Minimal, runs on a simulator with no real Firebase project (Data Connect emulator, demo project id):

https://github.com/duysolo/rn-sql-connect

```sh
npm install
cd example/ios && USE_FRAMEWORKS=dynamic bundle exec pod install && cd -
npm run emulator          # terminal 1
npm run example:ios       # terminal 2
```

The app runs a scripted smoke test at startup and shows the failure on screen:

```
FAIL mutation: not-configured: Firebase app "[DEFAULT]" is not configured.
```

The Android side of the same package works end to end (9/9 smoke steps), so the JavaScript and the general integration are not in question here.

## Questions

1. Is there a supported way, today, for a third-party pod to share the Firebase instance that react-native-firebase configures when SPM mode is on? `rnfirebase_add_spm_core_to_app_target` suggests linking at the app target is the intended shape. Should third-party pods do the same for their own Firebase-dependent products, and would you consider exposing a small public helper for that (something like `rnfirebase_add_spm_product_to_app_target(installer, url:, product:)`)?

2. If the intended answer is instead that Firebase products must be built as dynamic frameworks, is that something you would want to pursue upstream in firebase-ios-sdk rather than in this repo?

I am happy to open a PR once you tell me which direction you would accept. I did not send one with this issue because the fixes I tried locally (building my pod as a static framework, adding the product to the app target from my own `post_install`) each ran into their own linking problems, and I would rather not guess at the shape you want across your build matrix.

Thanks for the SPM work, it landed at exactly the right time for this.

russellwheatley:
Documents the SPM multi-pod FirebaseCore sharing limit from #9140.

- Correct the false claim that firebase-ios-sdk SPM products use `.library(type: .dynamic)`
- Document when FirebaseCore can be shared, and the `$RNFirebaseDisableSPM` workaround
- Align `firebase_spm.rb` comments and the static-linkage fail message with that model

Closes #9140

### DIFF
--- a/docs/ios-spm.mdx
+++ b/docs/ios-spm.mdx
@@ -177,11 +177,11 @@ links. Another pod's copy stays unconfigured, so `FirebaseApp.app()` can
 return `nil` there (and Auth can silently look signed-out even when RNFB JS
 Auth has a user).
 
-| RNFB | Other native dependency | Shared FirebaseCore? |
-| --- | --- | --- |
-| SPM | SPM (`spm_dependency`) | No: each dynamic pod embeds its own copy |
-| SPM | CocoaPods Firebase pods | No: dual resolution (SPM + CocoaPods Firebase) |
-| CocoaPods (`$RNFirebaseDisableSPM = true`) | CocoaPods Firebase pods | Yes: one CocoaPods `FirebaseCore` |
+| RNFB                                       | Other native dependency | Shared FirebaseCore?                           |
+| ------------------------------------------ | ----------------------- | ---------------------------------------------- |
+| SPM                                        | SPM (`spm_dependency`)  | No: each dynamic pod embeds its own copy       |
+| SPM                                        | CocoaPods Firebase pods | No: dual resolution (SPM + CocoaPods Firebase) |
+| CocoaPods (`$RNFirebaseDisableSPM = true`) | CocoaPods Firebase pods | Yes: one CocoaPods `FirebaseCore`              |
 
 If a third-party native module must share the same configured `FirebaseApp`
 with RNFB, opt out of SPM and resolve Firebase through CocoaPods for every
--- a/docs/ios-spm.mdx
+++ b/docs/ios-spm.mdx
@@ -90,7 +90,7 @@ The install output identifies the selected mode:
 ```
 
 RNFB also adds an `[RNFB] Embed Firebase SPM Frameworks` build phase to the app
-target. This phase embeds SPM-built Firebase frameworks that React Native's
+target. This phase embeds Firebase frameworks built by SPM that React Native's
 pod-level SPM integration does not otherwise copy into the app bundle.
 
 ### Release build and module-resolution settings
@@ -173,7 +173,7 @@ React Native attaches SPM products to each pod target via `spm_dependency`.
 Because firebase-ios-sdk products are automatic libraries, each dynamic pod
 framework that depends on Firebase embeds its own copy of `FirebaseCore`.
 `FirebaseApp.configure()` in the app configures only the copy that target
-links. Another pod's copy stays unconfigured, so `FirebaseApp.app()` can
+links. Another pod's copy is never configured, so `FirebaseApp.app()` can
 return `nil` there (and Auth can silently look signed-out even when RNFB JS
 Auth has a user).
 
--- a/docs/ios-spm.mdx
+++ b/docs/ios-spm.mdx
@@ -25,11 +25,14 @@ SPM requires dynamic framework linkage:
 use_frameworks! :linkage => :dynamic
 ```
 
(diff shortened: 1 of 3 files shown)


mikehardy (approved):
nice cleanup

PR #9147, opened 2026-08-05
feat(app, ios): support disabling SPM via Expo config plugin

huextrat:
### Description

React Native 0.75+ resolves Firebase iOS SDK dependencies through SPM by default, and opting out requires setting `$RNFirebaseDisableSPM = true` before any target block in the Podfile. In Expo projects the Podfile is regenerated on every prebuild, so today the only way to opt out is maintaining a custom config plugin.

This PR adds an `ios.disableSPM` prop to the `@react-native-firebase/app` config plugin:

```json
["@react-native-firebase/app", { "ios": { "disableSPM": true } }]
```

When enabled, the plugin injects `$RNFirebaseDisableSPM = true` into the generated Podfile right after `prepare_react_native_project!` (before any target block, as required by `firebase_spm.rb`), using `mergeContents` with a tagged block. When the prop is `false` or omitted the generated block is removed, and the flag is never emitted as `= false` (which would still select SPM and is a documented pitfall).

The implementation mirrors the existing `@react-native-firebase/analytics` config plugin, which manages its own Podfile flags the same way.

Docs updated: `docs/index.mdx`, `docs/ios-spm.mdx` (Expo section), and `packages/app/README.md`.

### Release Summary

feat(app, ios): add `ios.disableSPM` option to the Expo config plugin to opt out of SPM (`$RNFirebaseDisableSPM = true`) in generated Podfiles

### Checklist

  - [x] Yes
  - [ ] `Android`
  - [x] `iOS`
  - [ ] `Other` (macOS, web)
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

- `yarn jest packages/app/plugin` — 15 tests pass (flag insertion, placement before target blocks, idempotency, removal, default no-op)
- `tsc --build plugin` and `eslint plugin/src/*` pass
- The injected line and placement match what is already validated in production Expo apps via hand-written config plugins setting the same flag

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/index.mdx
+++ b/docs/index.mdx
@@ -111,6 +111,19 @@ If you use `@react-native-firebase/analytics` with Expo, including EAS Build, an
 
 The `withoutAdIdSupport` option adds `$RNFirebaseAnalyticsWithoutAdIdSupport = true` to opt out of iOS Ad ID support. The `googleAppMeasurementOnDeviceConversion` option adds `$RNFirebaseAnalyticsGoogleAppMeasurementOnDeviceConversion = true` to include Google Analytics on-device conversion measurement support. You may omit either option if it is not needed.
 
+If you want to opt out of Swift Package Manager for the Firebase iOS SDK and use CocoaPods instead, pass the `disableSPM` option to the `@react-native-firebase/app` config plugin. It adds `$RNFirebaseDisableSPM = true` to the generated Podfile during prebuild - see [iOS SPM Support](/ios-spm) for details:
+
+```json
+[
+  "@react-native-firebase/app",
+  {
+    "ios": {
+      "disableSPM": true
+    }
+  }
+]
+```
+
 ### Local app compilation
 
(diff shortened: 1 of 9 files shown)


mikehardy (approved):
this looks great first-shot, well done, thank you

as an aside, why are you feeling the need to disable SPM? I ask so that we can maybe fix it. Although it may be #9154  at a guess ;-). Landing the PR that fixes that one then I'll do a release ASAP though

huextrat:
> this looks great first-shot, well done, thank you
> 
> as an aside, why are you feeling the need to disable SPM? I ask so that we can maybe fix it. Although it may be #9154 at a guess ;-). Landing the PR that fixes that one then I'll do a release ASAP though

@mikehardy One of my apps uses Braze, and their Swift SDK only ships static binary xcframeworks, but the wrapper pods (`braze-react-native-sdk`, `ExpoAdapterBraze`) don't declare `static_framework = true`. So with `:linkage => :dynamic`, CocoaPods refuses the install (transitive static binaries check).

`react-native-video` bites too, different reason: it uses `RCTEventDispatcher` (lives in React-CoreModules) without declaring the dependency. Nobody notices with static linkage, but as a dynamic framework it fails to link.

We got both working with a `pre_install` hook that forces those pods static, so dynamic + SPM does build for us. But the proper fixes live upstream, so disabling SPM and staying static felt like the safest way to ship v26 for now.

PR #9155, opened 2026-08-05
fix(ios): skip static frameworks in SPM archive embed phase

### LINKED ISSUE #9154: 🔥 [🐛][ios] SPM embed phase copies static CocoaPods frameworks into the app bundle, breaking App Store submission
### Issue

With `@react-native-firebase/app@26.1.0` (SPM mode, added in #8933), `xcodebuild archive` produces an
`.ipa` that App Store Connect rejects with, for every static CocoaPods framework in the project:

```
Invalid bundle structure. The "<App>.app/Frameworks/<X>.framework/<X>" binary file is not permitted.
Your app cannot contain standalone executables or libraries, other than a valid CFBundleExecutable
of supported bundles.
```

In our project this fired for **43 frameworks** — every Expo module plus the `Pods_<target>` umbrella.
Simulator and device (non-archive) builds are unaffected, and the app runs correctly: Firebase itself
initialises fine over SPM. Only the archive is malformed.

### Root cause

`firebase_spm.rb` → `rnfirebase_spm_embed_script` ends with an Archive fallback:

```ruby
embed_frameworks_from "${OBJROOT}/UninstalledProducts/${PLATFORM_NAME}"
```

and `embed_frameworks_from` copies **every** `*.framework` it finds, with no filtering:

```bash
find "${source_dir}" -maxdepth 1 -type d -name "*.framework" -print0 | while IFS= read -r -d '' framework; do
  ...
  echo "Embedding Firebase SPM framework ${framework_name} (from ${source_dir})"
  rsync -av --delete ... "${framework}" "${app_frameworks_dir}"
```

The comment above that line assumes `UninstalledProducts` holds only Swift Package products. It does not:
Xcode's Archive action writes **every `SKIP_INSTALL=YES` product** there, and every CocoaPods pod target
has `SKIP_INSTALL=YES`. Because SPM mode requires `use_frameworks!`, pods build as `.framework` bundles
rather than `.a` archives, so the glob matches all of them.

Dynamic pods are harmless — CocoaPods' own `[CP] Embed Pods Frameworks` phase has already placed them, so
the `[ -e "${destination}" ]` guard skips them. The damage is limited to the frameworks CocoaPods
deliberately does *not* embed, i.e. the **static** ones. Those get copied into `Frameworks/`, and Apple
rejects a static archive sitting there.

### Evidence

Inspecting the rejected `.ipa` (`Info.plist` presence / binary type via `file` / `CFBundleExecutable`):

| group | count | binary type | verdict |
|---|---|---|---|
| React Native, Firebase (SPM), Sentry, reanimated, screens, webview, RNFB | ~100 | `Mach-O dynamically linked shared library` | valid, embedded by CocoaPods |
| Expo modules + `Pods_<target>` | 43 | `current ar archive` | **invalid**, embedded by this script |

Every rejected framework had a valid `Info.plist`, a matching `CFBundleExecutable` and a `_CodeSignature`.
The only defect was that the binary is a static archive. The set matches exactly what the unfiltered glob
over `UninstalledProducts` would pick up.

The build log confirms the path — the script's own `echo` names non-Firebase frameworks:

```
Embedding Firebase SPM framework ExpoCamera.framework (from .../UninstalledProducts/iphoneos)
Embedding Firebase SPM framework Pods_<target>.framework (from .../UninstalledProducts/iphoneos)
```

### Proposed fix

Only ever embed real dynamic libraries. Minimal patch against 26.1.0:

```diff
--- a/firebase_spm.rb
+++ b/firebase_spm.rb
@@ -200,6 +200,21 @@
           continue
         fi

+        # Xcode's Archive action writes EVERY SKIP_INSTALL=YES product into
+        # ${OBJROOT}/UninstalledProducts/${PLATFORM_NAME} -- not just Swift Package
+        # products. With CocoaPods + `use_frameworks!` (which SPM mode requires),
+        # that folder therefore also contains every pod's .framework, including the
+        # STATIC ones (Expo modules, the Pods_<target> umbrella, ...). Embedding a
+        # static framework into the app bundle makes App Store validation fail with
+        # "Invalid bundle structure. The <X>.framework/<X> binary file is not
+        # permitted. Your app cannot contain standalone executables or libraries".
+        # Only ever embed real dynamic libraries.
+        binary_name="${framework_name%.framework}"
+        if [ ! -f "${framework}/${binary_name}" ] || \
+           ! file -b "${framework}/${binary_name}" | grep -q 'dynamically linked'; then
+          continue
+        fi
+
         echo "Embedding Firebase SPM framework ${framework_name} (from ${source_dir})"
         rsync -av --delete \
           --filter "- Headers" \
```

The `grep` pattern was checked against macOS `file` output for single-arch dylibs, fat dylibs,
`current ar archive`, fat static archives and Mach-O object files. Static frameworks are linked into the
app binary at build time and never need embedding, which is why non-archive builds — which read
`BUILT_PRODUCTS_DIR/PackageFrameworks` and therefore only ever see genuine SPM products — work correctly.

Running with this patch applied, the archive contains no static frameworks and App Store validation passes.

### Environment

- `@react-native-firebase/app` / `@react-native-firebase/messaging`: **26.1.0** (Firebase iOS SDK 12.17.0, resolved via SPM)
- Expo SDK 57 (`expo` 57.0.10), React Native 0.86.2, New Architecture enabled
- `expo-build-properties`: `useFrameworks: "dynamic"`, `buildReactNativeFromSource: true`, `deploymentTarget: "16.4"`
- Build: EAS Build (macOS image), submitted via `eas submit` → `fastlane pilot` → `altool`
- Xcode: code 26.5 Build version 17F42

Happy to open a PR if that's easier — the change is the diff above and nothing else.

russellwheatley:
Fixes Archive builds embedding static CocoaPods frameworks (Expo modules, `Pods_<target>` umbrella, etc.) from `${OBJROOT}/UninstalledProducts/${PLATFORM_NAME}` into the app bundle, which App Store validation rejects as invalid bundle structure.

- Filter `embed_frameworks_from` to dynamically linked Mach-O only (`file -b`), with a skip log for missing/non-dynamic binaries
- Correct the call-site comment: that folder is every `SKIP_INSTALL=YES` product, not SPM-only
- Add macOS fixture coverage for the embed script
- Move SPM Ruby integration CI onto iOS (`tests_e2e_ios.yml` debug+spm) — toolchains for cocoapods/xcodeproj and Mach-O fixtures live there
- Record the invariant in `okf-bundle/ios-spm-native-imports.md`

Closes #9154

## Ruby testing strategy

SPM/CocoaPods helpers under `packages/app/*.rb` are exercised as first-class unit tests:

- **Entrypoint:** `yarn tests:ios:ruby` runs all `packages/app/__tests__/*_test.rb` (unit + shape + embed) under SimpleCov → `coverage/ios-ruby/lcov.info`
- **CI:** `tests_e2e_ios.yml` (debug+spm) runs that yarn target after `gem update cocoapods xcodeproj`, with `BUNDLE_FROZEN=true`, and uploads Codecov flag **`ios-ruby`**
- **Agent / OKF gates:** when Ruby sources or their tests change, `yarn tests:ios:ruby` is required to close `review_gate`; touched production Ruby lines need coverage evidence (same bar as JS/native)
- **Deps / supply chain:** `packages/app/__tests__/Gemfile` + lockfile (`CHECKSUMS`), Dependabot `bundler` on that directory with 7-day cooldown, frozen CI installs
- **Coverage:** `firebase_spm.rb` / `firebase_json.rb` at 100% line coverage under the Ruby suites


mikehardy:
Follow-on: Ruby unit tests for the SPM helpers with full line coverage on `firebase_spm.rb` / `firebase_json.rb`, wired through `yarn tests:ios:ruby` into OKF agent review/validation gates and iOS CI (`ios-ruby` Codecov), plus Dependabot + frozen lockfile/checksums for the SimpleCov gems.

mikehardy (approved):
assuming CI passes, I'm good to go on this
the underlying change was fine from the start but I was worried we had so much going on in ruby without treating it as a first-class language so I stole the opportunity to add the test/coverage/lint support I like for languages we're using a lot

mikehardy:
Coverage looked strange on this PR. On investigation, Android had **valid** Gradle build-cache hits for unit tests (sources unchanged), so Jacoco restored old-but-correct `*.exec` / session timestamps — and Codecov’s default `max_report_age` rejected them.

Fixed by accepting old reports (`max_report_age: off`), since Gradle will re-run when inputs change and we never check coverage into git.

While looking at that, we also noticed every Codecov upload already named exact `files:` but most still left workspace search on (only `android-native` had `disable_search`). Hardened all of those uploads so each flag only gets the file we specify.

PR #9164, opened 2026-08-06
fix(ios): link FirebaseCore into app target under SPM

### LINKED ISSUE #9161: [🐛] [iOS] Build error - confusion about useFrameworks and disable SPM
After upgrading to v26 I have a following error during build
```
[!] [react-native-firebase] SPM + static linkage is not supported (target(s): Pods-Trojmiastopl).
firebase-ios-sdk's Swift Package only ships dynamic library products, so `use_frameworks! :linkage => :static` causes every react-native-firebase pod that resolves Firebase via SPM to embed its own copy of the same Firebase frameworks -- this produces duplicate-symbol linker errors at build time instead of a clear error here.
Fix one of the following in your Podfile, then run `pod install` again:
  - Use dynamic linkage: `use_frameworks! :linkage => :dynamic`
  - Opt out of SPM: set `$RNFirebaseDisableSPM = true` before any target block, then use the static or dynamic linkage your project requires.
```
This is my expo-build-properties config for iOS that worked fine before upgrade:
```
"useFrameworks": "static",
"buildReactNativeFromSource": true,
"forceStaticLinking": [
  "RNFBAnalytics",
  "RNFBApp",
  "RNFBCrashlytics",
  "RNFBFirestore",
  "RNFBMessaging"
]
```
I checked documentation and it says in Getting started section:
> For iOS only, since firebase-ios-sdk requires use_frameworks then you want to configure [expo-build-properties](https://docs.expo.dev/versions/latest/sdk/build-properties/#pluginconfigtypeios) app.json by adding "useFrameworks": "static".

which is what I have. But iOS SPM Support section says
> Do not use RNFB's SPM mode with use_frameworks! :linkage => :static. 

and
> To use CocoaPods mode in a generated project, pass disableSPM to the @react-native-firebase/app config plugin

so I guess these two options are required together to make things work. I tried add disableSPM to my plugin config but it doesn't solve the problem. However I made custom plugin using AI that disables SPM and it works correctly (no errors during build and app seems to work):
```
const fs = require("fs");
const path = require("path");
const { withDangerousMod } = require("@expo/config-plugins");

module.exports = function withDisableSPM(config) {
  return withDangerousMod(config, [
    "ios",
    async (config) => {
      const podfilePath = path.join(config.modRequest.platformProjectRoot, "Podfile");
      let podfile = fs.readFileSync(podfilePath, "utf-8");

      if (!podfile.includes("$RNFirebaseDisableSPM = true")) {
        podfile = `$RNFirebaseDisableSPM = true\n${podfile}`;
        fs.writeFileSync(podfilePath, podfile);
      }

      return config;
    },
  ]);
};
```
So my question is how to correctly configure app to make successful build? I guess custom plugin shouldn't be required. Also I think documentation should be more clear about that.

russellwheatley:
Fixes #9161.

`rnfirebase_add_spm_core_to_app_target` declared the `FirebaseCore` package product dependency on the app target but never added a matching `PBXBuildFile` to its Frameworks build phase, so apps whose own code calls `FIRApp`/`FIROptions` directly (e.g. Expo's injected AppDelegate hookup) failed at link time with an undefined `FIRApp` symbol under dynamic SPM linking.

- Adds the missing `PBXBuildFile`, removes it symmetrically when SPM is disabled
- Heals the idempotency guard so a project built by a prior RNFB version (dependency declared but never linked) gets repaired on the next `pod install` instead of being skipped as already done

## Test Plan

- [x] `yarn lint:ruby` (RuboCop over `firebase_spm.rb`, `firebase_json.rb`, `__tests__`) — 8 files inspected, no offenses
- [x] `yarn tests:ios:ruby` — `packages/app/__tests__/firebase_json_test.rb` (7 runs, 22 assertions)
- [x] `yarn tests:ios:ruby` — `packages/app/__tests__/firebase_spm_embed_script_test.rb` (4 runs, 29 assertions)
- [x] `yarn tests:ios:ruby` — `packages/app/__tests__/firebase_spm_shape_test.rb` (14 runs, 51 assertions), including new `frameworks_build_phase`/`PBXBuildFile#product_ref` shape checks
- [x] `yarn tests:ios:ruby` — `packages/app/__tests__/firebase_spm_test.rb` (74 runs, 269 assertions), including new `test_add_core_heals_existing_dependency_missing_build_file` and `test_remove_core_removes_orphaned_build_file_from_frameworks_build_phase` regression tests
- [x] Line coverage 301/301 (100%) on `packages/app/*.rb`

### DIFF
--- a/okf-bundle/ios-spm-native-imports.md
+++ b/okf-bundle/ios-spm-native-imports.md
@@ -3,7 +3,7 @@ type: Reference
 title: iOS SPM native integration decisions
 description: Why RNFB uses dual imports, Objective-C helpers for Swift Firebase products, and an app framework-embedding phase.
 tags: [ios, spm, cocoapods, imports, firebase, cxx-modules]
-timestamp: 2026-08-05T11:48:00Z
+timestamp: 2026-08-06T16:00:00Z
 ---
 
 # iOS SPM native integration decisions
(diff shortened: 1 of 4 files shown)


mikehardy:
Closing and reopening to trigger CI after github outage

mikehardy (approved):
LGTM - still sad my work in the rnfbdemo prior to release didn't catch this as I was working through Expo version matrix for build testing. That one remains mysterious for me

PR #9192, opened 2026-08-12
test(e2e): split macOS app; bump tests RN to 0.86

russellwheatley:
Splits the e2e apps so mobile RN can move independently of `react-native-macos`, then bumps mobile to 0.86.2 with a full NewArch-AD-20 codegen regen.

Closes CPRN-236.

- Sibling `tests-macos/` stays on RN 0.78.3 + `react-native-macos` 0.78.6; shared JS harness remains under `tests/` (Metro resolves `../tests`)
- Drop blanket root RN/codegen/CLI resolutions; per-app pins; codegen resolves from `tests/`
- Keep root script names `tests:macos:*` (cwd `tests-macos`)
- Mobile `tests/` → RN 0.86.2 / CLI 20.1.0; regenerate `packages/**/generated/**`; remove NewArch-AD-21 ResultT inject (upstream emits it)
- Preceding commit refreshes stale `tests/ios` Podfile.lock for current package versions
- Jest: use `@react-native/jest-preset` (RN 0.86 moved the preset out of `react-native`)
- Android JVM unit: skip empty modules so Gradle 9 `failOnNoDiscoveredTests` does not fail packages with no `src/test`
- AsyncStorage: pin mobile native + Metro to 3.1.1; drop unused `packages/app` 2.x nest so it cannot load against 3.x native; `tests-macos` stays on 2.x
- `tests-macos` RNFB deps workspace-link at 26.2.0 (same lerna version as `tests/`, not published npm 26.1.0)
- macOS Metro: drop blanket `tests/node_modules` blocklist; React/scheduler still resolve from the macOS app (`SINGLETON_FROM_APP`)


okf-bundle/testing/test-app-dependency-pins.md
```diff
@@ -48,6 +50,14 @@ Root `package.json` must **not** use blanket `resolutions` for `react-native`, `
 
 **Agent / Dependabot rule:** leave these pins alone unless the change is an intentional dual-app or mobile-only upgrade. Reject RN / codegen / CLI bumps that only “look green” for one app while breaking the other or codegen verify.
 
+## RNFB workspace pins (both e2e apps)
+
+Both e2e apps must declare every `@react-native-firebase/*` dependency (except `app-types`) at the **current lerna / package version** so Yarn **workspace-links** them (`workspace:packages/<pkg>`) instead of fetching published npm tarballs.
+
+Today that version is **`26.2.0`**, matching `packages/*/package.json`. A stale pin (for example `26.1.0` while packages are `26.2.0`) resolves as `npm:26.1.0` and a fresh `yarn` downloads ~19 published tarballs. Metro may still remap JS to `packages/*`, so CI can look green while the lockfile is wrong.
+
+When lerna bumps the packages, bump **both** `tests/package.json` and `tests-macos/package.json` (including each app's private `"version"`) in the same change. Leave `@react-native-firebase/app-types` at `6.7.2`.
```
mikehardy: I don't think this is workable as an LLM suggestion - release process is 100% deterministic, so this would need to be scripted and be part of the lerna versioning process somehow (hopefully lerna has hooks for it's work phases...)

mikehardy (approved):
LGTM - approving after the one change to versioning so the new version pins in test apps are automated during release and the plist dep unfuddle from merging main in (something I do pre-review in my workflow, to find problems like this!)

PR #9195, opened 2026-08-13
fix(remote-config, ios): apply pending settings writes before fetching

### LINKED ISSUE #9194: [🐛] remoteConfig.settings = … then fetchAndActivate() cancels the fetch on iOS
# remoteConfig.settings = … then fetchAndActivate() cancels the fetch on iOS

## Disclaimer
I've used Claude Opus 5 to help me with the investigation, reproducer and write this issue. But important to note

- I've actually saw the error in our app after migrating to the newer version, which triggered my investigation
- I tested the reproducer and revised this. I'm also including a screenshot of the reproducer so you guys can test on your end and compare.

In summary: I'm not dropping random AI slop here without verifying, but don't be surprised to recognize Claude writing style 😄 

## Issue

Hi, and thanks for the v25/v26 work. We hit this upgrading from v23 to v26. The sequence that breaks is the one the v25 migration notes tell people to write.

**In short:** the `settings` and `defaultConfig` setters send their value to native on a **microtask**. `fetchAndActivate()` calls native **immediately**. So the fetch always gets there first, and the config write lands while the request is already in flight. On iOS that recreates the URLSession the fetch is running on, and the request is cancelled.

The **ordering inversion is deterministic** — it follows from microtask versus synchronous call, not from device speed. Whether that write then actually **cancels** the request depends on one further thing: whether the request has reached the wire yet. In production it has, on every launch after the app's first, for the reason in "if it does not reproduce on the first run" below.

Versions: `@react-native-firebase/remote-config` **26.2.0**, firebase-ios-sdk **12.17.0**, React Native 0.86.2, Expo 57. Worked fine on 23.7.0. iOS only — see the Android note at the end.

---

### What you see

```ts
remoteConfig.settings = { minimumFetchIntervalMillis: 0, fetchTimeoutMillis: 60000 };
await fetchAndActivate(remoteConfig);
```

```
[remoteConfig/failure] fetch() operation cannot be completed successfully.
```

That message is the same for every failure, because the iOS module builds the rejection from the fetch-status enum and drops the real `NSError` ([`RNFBConfigHelper.m#L180-L185`](https://github.com/invertase/react-native-firebase/blob/v26.2.0/packages/remote-config/ios/RNFBConfig/RNFBConfigHelper.m#L180-L185)). With `-FIRDebugEnabled` on the Xcode scheme you get the real one:

```
[I-RCN000046] Making config request: https://firebaseremoteconfig.googleapis.com/...   <- fetch starts
[I-RCN000067] Successfully set configSettings ...                                       <- our write lands, mid flight
[I-RCN000050] config fetch completed. Error: NSURLErrorDomain Code=-999 "cancelled"
```

`-999` is `NSURLErrorCancelled` — the app cancelled its own request. Offline would be `-1009`, a timeout `-1001`.

---

### Why it happens

**In JavaScript: the write is deferred, the fetch is not.**

The setter discards the promise it creates, because a setter cannot return one ([`lib/index.ts#L175-L189`](https://github.com/invertase/react-native-firebase/blob/v26.2.0/packages/remote-config/lib/index.ts#L175-L189)):

```ts
set settings(settings) {
  this._settings = { ... };
  void this.setConfigSettings(settings, true);   // -> _enqueueNativeMutation
}
```

`_enqueueNativeMutation` starts with `this._nativeMutationQueue.then(task, task)` ([`#L512`](https://github.com/invertase/react-native-firebase/blob/v26.2.0/packages/remote-config/lib/index.ts#L512)), so the native call runs on a **microtask** — after the current synchronous block. `fetchAndActivate()` calls `this.native.fetchAndActivate()` directly ([`#L341`](https://github.com/invertase/react-native-firebase/blob/v26.2.0/packages/remote-config/lib/index.ts#L341)).

Write the two lines in the obvious order and native receives them in the opposite order. Every time.

**In firebase-ios-sdk: applying settings tears down the network session.**

```objc
// FIRRemoteConfig.m:721-737
- (void)setConfigSettings:(FIRRemoteConfigSettings *)configSettings {
  ...
  [self->_configFetch recreateNetworkSession];
}

// RCNConfigFetch.m:117-121
- (void)recreateNetworkSession {
  if (_fetchSession) {
    [_fetchSession invalidateAndCancel];   // kills every in-flight task with -999
  }
  ...
}
```

The fetch runs on exactly that session, so it dies.

---

### It is ordering, not timing

This is the part we would most like you to look at, because it makes the fix obvious.

Take the failing code and add one line that waits for **nothing**:

```ts
remoteConfig.settings = SETTINGS;
await Promise.resolve();          // zero elapsed time
await fetchAndActivate(remoteConfig);   // now succeeds
```

`await Promise.resolve()` does not wait. It has no timer and no I/O. All it does is yield, which lets the setter's already-queued microtask deliver the write to native before the fetch is started.

Under identical conditions, that one line is the entire difference between failing and passing. Nothing is given more time to finish — the JS layer is simply handing native the two operations in the wrong order, and the yield puts them back in the right one.

---

### It is not only `settings`

`defaultConfig` cancels fetches too (case 3 in the reproducer). That one is more surprising, because firebase-ios-sdk's `setDefaults:` never touches the network session ([`FIRRemoteConfig.m#L644-L658`](https://github.com/firebase/firebase-ios-sdk/blob/12.17.0/FirebaseRemoteConfig/Sources/FIRRemoteConfig.m#L644)). The cancellation comes from this library instead:

Every bridge method resolves with a constants payload, and building that payload **reads** native `configSettings` ([`RNFBConfigHelper.m#L112-L114`](https://github.com/invertase/react-native-firebase/blob/v26.2.0/packages/remote-config/ios/RNFBConfig/RNFBConfigHelper.m#L112)):

```objc
double minimumFetchInterval = [RCTConvert double:@([remoteConfig configSettings].minimumFetchInterval)];
double fetchTimeout = [RCTConvert double:@([remoteConfig configSettings].fetchTimeout)];
```

And the **getter** recreates the session as well ([`FIRRemoteConfig.m#L702-L719`](https://github.com/firebase/firebase-ios-sdk/blob/12.17.0/FirebaseRemoteConfig/Sources/FIRRemoteConfig.m#L702)):

```objc
- (FIRRemoteConfigSettings *)configSettings {
  dispatch_sync(_queue, ^{
    ...
    [_configFetch recreateNetworkSession];   // same teardown, on a read
  });
```

So the general statement is broader than the title suggests: **any remote-config call that returns constants can cancel an in-flight fetch** — `setDefaults`, `activate`, `ensureInitialized`, `fetch`, `reset`.

---

### Why it matters

A cancelled fetch never updates `lastFetchTime`, so the minimum fetch interval never starts throttling. Every cold start makes a real network request, and every one is cancelled the same way.

**On a fresh iOS install, Remote Config never arrives at all.** Not stale — never.

For an app that has activated config before, the failure lands specifically on the launches where the interval *had* elapsed, which are exactly the launches that would have refreshed. Launches inside the interval are served from cache, make no request, and have nothing to cancel — which is probably why this has not been reported more often.

---

### What could probably fix it

Each one of these could independently fix it. So these are all potential approaches. 

**1. Stop reading native `configSettings` when building the constants payload.** Cheapest fix, and it covers every method at once. You already keep settings in JS (`this._settings`), and you already distrust the native values — `setConfigSettings` deliberately passes `fetchTimeout: undefined, minimumFetchInterval: undefined` into `_updateFromConstants` with the comment *"Native can report stale settings constants for this call."* If they are not trusted there, reading them at the cost of destroying the network session is hard to justify.

**2. Have `fetchAndActivate()` / `fetch()` / `activate()` wait for the mutation queue before calling native.** This fixes the ordering inversion for the normal startup sequence:

```ts
fetchAndActivate(): Promise<boolean> {
  return this._nativeMutationQueue.then(() =>
    this._promiseWithConstants(this.native.fetchAndActivate()));
}
```

Wait for the queue, do not join it — the four methods on it today are all short native writes, while a fetch can take up to `fetchTimeoutMillis`, so putting it *on* the queue would block property setters behind it.

This works with no timing assumption, because `_enqueueNativeMutation` reassigns `this._nativeMutationQueue` **synchronously**. By the time `fetchAndActivate` runs, right after `remoteConfig.settings = …`, the queue already points at a promise that settles after the pending write.

**3. Export `setConfigSettings()` and `setDefaults()` as awaitable functions.**

Fixes 1 and 2 stop the cancellation. Neither gives a caller any way to know *when* a config write has actually been applied — and we hit a case where that is the only thing that helps.

**`defaultConfig` cannot be ordered ahead of a fetch using public API at all.** The harm lands when the bridge call *resolves*, because that is when the constants payload is built — and a property setter gives you no handle on its own resolution. Awaiting a round trip in front of the fetch only buys the write a head start; it is a cushion, not an ordering guarantee. Our two options were to move the defaults write until after the fetch had finished, or to cast and await the instance method. There is no third answer that is both public and deterministic.

**I know the idea of not making this part of the public interface explicitly was to have web API parity**. But I don't think exposing these cuts against the web parity goal, because the situations genuinely differ. On web, `remoteConfig.settings = …` is a synchronous write to local memory: there is nothing to await, and the property shape is exactly right. On native the identical line is an asynchronous bridge call **with side effects**. The shape came across in the migration; the asynchrony it has to carry could not. An awaitable form is not a behavioural divergence from web — it exposes something native has that web does not.

It is also cheap. Both methods already exist, still return their promises, and are already declared on `RemoteConfigInternal` in `packages/remote-config/lib/types/internal.ts` — that interface just is not exported, so a cast is currently the only way to reach them. Exporting them is additive, breaks nothing, and the property setters can stay exactly as they are. It would simply mean apps that need a real ordering guarantee have a supported way to get one.

For what it is worth we are unblocked today, but only through implementation details we found by reading the source. Every affected app has to rediscover them independently.

---

### Reproducer

Drop-in `App` component. No setup beyond a project that already has `@react-native-firebase/app` and `@react-native-firebase/remote-config` working. Tap **"Run all three"**.

(It also imports `SafeAreaView` from `react-native-safe-area-context`. That is layout only — swap it for the one in `react-native` if you do not have that package.)

Expected on iOS: **1 and 3 fail, 2 passes.** On the very first run after launch case 1 reports ⚠️ instead — see below.

<details><summary>App.tsx</summary>
<p>

```tsx
/**
 * Minimal reproducer — @react-native-firebase/remote-config v26, iOS.
 *
 * The documented v25/v26 way to configure Remote Config is to assign properties:
 *
 *   remoteConfig.settings = { ... };
 *   remoteConfig.defaultConfig = { ... };
 *
 * Both setters send the value to native, but they defer that call to a
 * microtask. `fetchAndActivate()` calls native immediately. So the fetch always
 * reaches native first, and the config write lands while the request is already
 * in flight — which on iOS recreates the URLSession the fetch is running on and
 * cancels it with NSURLErrorCancelled (-999).
 *
 * Three cases:
 *   1. settings, then fetch                     -> FAILS (the documented shape)
 *   2. settings, await Promise.resolve(), fetch -> PASSES
 *   3. defaults assigned during a fetch         -> FAILS (same bug, other setter)
 *
 * Cases 1 and 2 are the point, and they run back to back with nothing in
 * between so the only difference is that added line. `await Promise.resolve()`
 * waits for nothing — zero elapsed time. All it does is yield, which lets the
 * setter's already-queued microtask deliver the write to native before the
 * fetch is started. So this is an ordering problem in the JS layer, not a
 * timing window on a slow device.
 *
 * ⚠️ RUN IT TWICE. The very first fetch in a freshly launched process has
 * one-time setup to do (Installations token and so on) before it can put a
 * request on the wire. While that is pending there is nothing to cancel, so
 * case 1 passes the first time and fails every time after. The log says so
 * rather than reporting a misleading green tick. In a real app this does not
 * help you: that setup is cached across launches, so from the second launch
 * onward the request goes out immediately and the fetch is cancelled every
 * time.
 *
 * Drop this in as your App component in any project that already has
 * @react-native-firebase/app and @react-native-firebase/remote-config working,
 * and tap "Run all three". The only other import is `SafeAreaView` from
 * react-native-safe-area-context — swap it for the one in react-native if you
 * do not have that package; it is only layout and has nothing to do with the
 * bug.
 */
import React, { useState } from 'react';
import { ScrollView, StyleSheet, Text, TouchableOpacity } from 'react-native';

import { SafeAreaProvider, SafeAreaView } from 'react-native-safe-area-context';
import {
  fetchAndActivate,
  getRemoteConfig,
  RemoteConfig,
} from '@react-native-firebase/remote-config';

const SETTINGS = { minimumFetchIntervalMillis: 0, fetchTimeoutMillis: 60000 };
const DEFAULTS = { repro_flag: true };

const sleep = (ms: number): Promise<void> =>
  new Promise((resolve) => setTimeout(resolve, ms));

/**
 * Whether anything has been run yet in this process. Case 1 is expected to pass
 * on the very first run after launch, so a pass is only meaningful once this is
 * true — see the note at the top of the file.
 */
let hasRunBefore = false;

const describeError = (error: unknown): string => {
  const { code, message } = (error ?? {}) as {
    code?: string;
    message?: string;
  };

  return `${code ?? 'no code'} | ${message ?? String(error)}`;
};

const styles = StyleSheet.create({
  screen: { flex: 1, backgroundColor: '#fff', padding: 16 },
  title: { fontSize: 18, fontWeight: '700', marginBottom: 12 },
  button: {
    backgroundColor: '#1b1b1f',
    borderRadius: 8,
    paddingVertical: 10,
    paddingHorizontal: 12,
    marginBottom: 8,
  },
  buttonDisabled: { opacity: 0.4 },
  buttonText: { color: '#fff', fontSize: 14, fontWeight: '600' },
  clear: { paddingVertical: 6, marginBottom: 8 },
  clearText: { color: '#777', fontSize: 13 },
  log: { flex: 1, backgroundColor: '#f4f4f6', borderRadius: 8, padding: 10 },
  logLine: { fontFamily: 'Menlo', fontSize: 11, marginBottom: 3 },
});

const App: React.FC = () => {
  const [lines, setLines] = useState<string[]>([]);
  const [isRunning, setIsRunning] = useState(false);

  const write = (line: string): void => {
    // eslint-disable-next-line no-console
    console.log(`[repro] ${line}`);
    setLines((previous) => [...previous, line]);
  };

  /**
   * Setup, not part of what is being tested. Two jobs.
   *
   * 1. Apply minimumFetchInterval 0, so the fetches below really go to the
   *    network instead of being served from cache. This uses the unexported
   *    instance method on purpose: awaiting it is the one way to be certain the
   *    write landed before the cases run, so they are not measuring leftover
   *    state from the setup.
   *
   * 2. Do one throwaway fetch to warm the Firebase Installations token.
   *
   * The warm-up matters more than it looks. `fetchAndActivate()` cannot put a
   * request on the wire until it has an Installations token, and obtaining that
   * token for the first time in a process is itself a network round trip. While
   * it is pending there is no request to cancel, so the bug does not show — the
   * very first fetch after launching the app tends to pass.
   *
   * Firebase persists that token across launches, so in a real app it is warm
   * from the second launch onward. That is the state reproduced here, and it is
   * why the cancellation happens on every cold start in production.
   *
   * Nothing is in flight by the time this returns, so the setup itself cannot
   * be what gets cancelled below.
   */
  const prepare = async (): Promise<void> => {
    const remoteConfig = getRemoteConfig() as RemoteConfig & {
      setConfigSettings(s: typeof SETTINGS): Promise<void>;
    };

    await remoteConfig.setConfigSettings(SETTINGS);

    await fetchAndActivate(remoteConfig).catch(() => undefined);
    await sleep(1000);
  };

  /**
   * `passIsInconclusive` marks a case where resolving does not mean the bug is
   * absent — currently only case 1 on the very first run after launch.
   */
  const runFetch = async (
    label: string,
    passIsInconclusive = false,
  ): Promise<void> => {
    const remoteConfig = getRemoteConfig();

    try {
      const activated = await fetchAndActivate(remoteConfig);

      if (passIsInconclusive) {
        write(
          `⚠️ ${label}: resolved — but this was the FIRST run since launch`,
        );
        write('   Not a pass. The first fetch in a process is slow to reach');
        write('   the network, so there was no request to cancel yet.');
        write('   👉 Tap again — from now on it should FAIL.');
      } else {
        write(`✅ ${label}: resolved (activated=${activated})`);
      }
    } catch (error) {
      write(`❌ ${label}: REJECTED — ${describeError(error)}`);
    }

    write(`   lastFetchStatus=${remoteConfig.lastFetchStatus}`);
  };

  /**
   * 1. Exactly what the v25 migration notes tell you to write.
   *
   * Deliberately has no setup. This case is the plain documented shape, and
   * anything that touches Remote Config immediately beforehand perturbs it —
   * so it runs on its own and simply needs to be run twice (see the note at the
   * top of the file). Its own `settings` write applies minimumFetchInterval 0,
   * so from the second run onward the fetch always goes to the network.
   */
  const caseSettingsThenFetch = async (): Promise<void> => {
    write('--- 1. settings, then fetch ---');

    const isFirstRun = !hasRunBefore;

    const remoteConfig = getRemoteConfig();

    remoteConfig.settings = SETTINGS;

    await runFetch('settings then fetch', isFirstRun);
  };

  /**
   * 2. Identical to case 1, with one added line that waits for nothing.
   *
   * `await Promise.resolve()` costs zero elapsed time. It only yields, which
   * lets the setter's queued microtask deliver the write to native before the
   * fetch starts. That single line is the whole difference between failing and
   * passing.
   *
   * No setup here either, on purpose: this only means something if it runs in
   * the same conditions as case 1, which is why they run back to back.
   */
  const caseSettingsYieldFetch = async (): Promise<void> => {
    write('--- 2. settings, await Promise.resolve(), then fetch ---');

    const remoteConfig = getRemoteConfig();

    remoteConfig.settings = SETTINGS;
    await Promise.resolve();

    await runFetch('settings, yield, fetch');
  };

  /**
   * 3. The same problem through the other setter.
   *
   * The fetch gets 50ms to reach the network first, so the assignment
   * definitely lands while a request is live. That makes this one independent
   * of process warm-up: it fails on the first run too.
   *
   * Worth noting: firebase-ios-sdk's `setDefaults:` never touches the network
   * session, so this cancellation cannot come from there. It comes from this
   * library — the bridge resolves with a constants payload, and building that
   * payload reads native `configSettings`, whose getter also recreates the
   * session.
   */
  const caseDefaultsDuringFetch = async (): Promise<void> => {
    write('--- 3. defaults assigned during a fetch ---');
    await prepare();

    const remoteConfig = getRemoteConfig();

    const pendingFetch = fetchAndActivate(remoteConfig).then(
      (activated) => ({ ok: true as const, activated }),
      (error) => ({ ok: false as const, error }),
    );

    await sleep(50);
    remoteConfig.defaultConfig = DEFAULTS;

    const result = await pendingFetch;

    if (result.ok) {
      write(
        `✅ defaults during fetch: resolved (activated=${result.activated})`,
      );
    } else {
      write(
        `❌ defaults during fetch: REJECTED — ${describeError(result.error)}`,
      );
    }

    write(`   lastFetchStatus=${remoteConfig.lastFetchStatus}`);
  };

  const run = (scenario: () => Promise<void>) => async (): Promise<void> => {
    setIsRunning(true);

    try {
      await scenario();
    } finally {
      hasRunBefore = true;
      setIsRunning(false);
    }
  };

  const buttons: { label: string; onPress: () => Promise<void> }[] = [
    {
      label: 'Run all three',
      onPress: run(async () => {
        // 1 and 2 run back to back with nothing in between, so the only
        // difference between them is the added `await Promise.resolve()`.
        await caseSettingsThenFetch();
        await sleep(2000);
        await caseSettingsYieldFetch();
        await sleep(2000);
        await caseDefaultsDuringFetch();
      }),
    },
    { label: '1. settings, then fetch', onPress: run(caseSettingsThenFetch) },
    {
      label: '2. settings, yield, then fetch',
      onPress: run(caseSettingsYieldFetch),
    },
    {
      label: '3. defaults during a fetch',
      onPress: run(caseDefaultsDuringFetch),
    },
  ];

  return (
    <SafeAreaProvider>
      <SafeAreaView style={styles.screen}>
        <Text style={styles.title}>{'RNFB v26 Remote Config repro'}</Text>

        {buttons.map(({ label, onPress }) => (
          <TouchableOpacity
            key={label}
            style={[styles.button, isRunning && styles.buttonDisabled]}
            disabled={isRunning}
            onPress={onPress}
          >
            <Text style={styles.buttonText}>{label}</Text>
          </TouchableOpacity>
        ))}

        <TouchableOpacity style={styles.clear} onPress={() => setLines([])}>
          <Text style={styles.clearText}>{'Clear log'}</Text>
        </TouchableOpacity>

        <ScrollView style={styles.log}>
          {lines.map((line, index) => (
            // eslint-disable-next-line react/no-array-index-key
            <Text key={`${index}:${line}`} style={styles.logLine}>
              {line}
            </Text>
          ))}
        </ScrollView>
      </SafeAreaView>
    </SafeAreaProvider>
  );
};

export default App;
```

</p>
</details>

<details>
<summary>Reproducer screenshot</summary>

Image

</details>

<details><summary>Android is not affected, and why that supports the diagnosis</summary>
<p>

All three cases pass on Android (firebase-bom 34.16.0). The cancellation needs something only iOS has.

iOS keeps one long-lived `NSURLSession`, and the fetch timeout lives in its configuration — so changing the timeout means throwing the session away, taking in-flight tasks with it. Android builds a connection per fetch (`ConfigFetchHttpClient.createHttpURLConnection()`), so there is nothing to tear down. And Android's settings write is a SharedPreferences write that never touches the network layer at all.

</p>
</details>

<details><summary>Things we ruled out</summary>
<p>

| Suspected | Why not |
|---|---|
| Throttling | The code is `failure`, not `throttled`. Firebase reports those separately. |
| Installations not ready | The log shows `Success to get iid: ...` before the fetch begins. |
| Bad network | `-999` is a local cancel. Offline is `-1009`, timeout `-1001`. |
| Not initialized yet | `ensureInitialized()` resolves in ~50ms and the fetch still fails. |
| A timing window | Case 2 adds zero elapsed time and passes. |
| Performance URLSession swizzling | Disabled via `firebase_performance_instrumentation_enabled = false`. Failure persisted. |
| Project or credentials | The same fetch succeeds seconds later in the same process. |

</p>
</details>

<details><summary>Source references</summary>
<p>

**react-native-firebase [`v26.2.0`](https://github.com/invertase/react-native-firebase/tree/v26.2.0)**

| Claim | Source |
|---|---|
| `set settings` discards the promise | [`lib/index.ts#L175-L189`](https://github.com/invertase/react-native-firebase/blob/v26.2.0/packages/remote-config/lib/index.ts#L175-L189) |
| `set defaultConfig` does the same | [`lib/index.ts#L143-L165`](https://github.com/invertase/react-native-firebase/blob/v26.2.0/packages/remote-config/lib/index.ts#L143-L165) |
| Mutations run on a microtask; the queue pointer is reassigned synchronously | [`lib/index.ts#L512-L521`](https://github.com/invertase/react-native-firebase/blob/v26.2.0/packages/remote-config/lib/index.ts#L512) |
| `fetchAndActivate` calls native directly | [`lib/index.ts#L341`](https://github.com/invertase/react-native-firebase/blob/v26.2.0/packages/remote-config/lib/index.ts#L341) |
| The constants payload reads native `configSettings` | [`RNFBConfigHelper.m#L106-L131`](https://github.com/invertase/react-native-firebase/blob/v26.2.0/packages/remote-config/ios/RNFBConfig/RNFBConfigHelper.m#L106) |
| The real `NSError` is discarded | [`RNFBConfigHelper.m#L180-L185`](https://github.com/invertase/react-native-firebase/blob/v26.2.0/packages/remote-config/ios/RNFBConfig/RNFBConfigHelper.m#L180-L185) |

**firebase-ios-sdk [`12.17.0`](https://github.com/firebase/firebase-ios-sdk/tree/12.17.0)**

| Claim | Source |
|---|---|
| Setting config settings recreates the session | [`FIRRemoteConfig.m#L721-L737`](https://github.com/firebase/firebase-ios-sdk/blob/12.17.0/FirebaseRemoteConfig/Sources/FIRRemoteConfig.m#L721) |
| **Reading** config settings recreates it too | [`FIRRemoteConfig.m#L702-L719`](https://github.com/firebase/firebase-ios-sdk/blob/12.17.0/FirebaseRemoteConfig/Sources/FIRRemoteConfig.m#L702) |
| Recreating cancels every in-flight request | [`RCNConfigFetch.m#L117-L121`](https://github.com/firebase/firebase-ios-sdk/blob/12.17.0/FirebaseRemoteConfig/Sources/RCNConfigFetch.m#L117-L121) |
| `setDefaults:` never touches the session | [`FIRRemoteConfig.m#L644-L658`](https://github.com/firebase/firebase-ios-sdk/blob/12.17.0/FirebaseRemoteConfig/Sources/FIRRemoteConfig.m#L644) |
| The interval is in-memory and resets to 12h each launch, so it must be set after creation | [`RCNConfigSettings.m#L83`](https://github.com/firebase/firebase-ios-sdk/blob/12.17.0/FirebaseRemoteConfig/Sources/RCNConfigSettings.m#L83) |

</p>
</details>

---

## Environment

<p>

- **Platform:**
  - [x] iOS
  - [ ] Android
  - [ ] Both

  (Android was tested and all three cases pass — see the Android note above.)
- **`react-native-firebase` version:** `26.2.0` (worked on `23.7.0`)
- **Firebase module:** Remote Config (firebase-ios-sdk 12.17.0)
- **TypeScript:** `Y` & `5.8.3`

```json
{
  "dependencies": {
    "@react-native-firebase/app": "^26.2.0",
    "@react-native-firebase/remote-config": "^26.2.0",
    "expo": "~57.0.12",
    "react": "19.2.3",
    "react-native": "0.86.2"
  }
}
```

`ios/Podfile` — using CocoaPods rather than SPM (`$RNFirebaseDisableSPM = true`), static frameworks, Expo bare workflow. Firebase is auto-initialised by `RNFBAppModule` from `GoogleService-Info.plist`.

**`react-native info`:**

```
System:
  OS: macOS 26.6.1
  CPU: (10) arm64 Apple M5
Binaries:
  Node: 24.15.0
  Yarn: 4.4.1
Managers:
  CocoaPods: 1.16.2
IDEs:
  Xcode: 26.6/17F113
npmPackages:
  react: 19.2.3
  react-native: 0.86.2
iOS:
  hermesEnabled: true
  newArchEnabled: true
```

</p>
</details>

giaBaoJS:
### Description

Assigning `remoteConfig.settings` and then immediately awaiting `fetchAndActivate()` cancels the fetch on iOS with `NSURLErrorCancelled (-999)`.

**Why.** The `settings` and `defaultConfig` setters cannot return a promise, so they hand their native write to `_enqueueNativeMutation`, which only reaches native on a later microtask. The reads — `activate()`, `fetch()`, `fetchAndActivate()`, `ensureInitialized()` — called native synchronously:

```ts
fetchAndActivate(): Promise<boolean> {
  return this._promiseWithConstants(this.native.fetchAndActivate());
}
```

So native received the read **first**, and the queued `setConfigSettings` write landed while the fetch was already in flight. On iOS, firebase-ios-sdk's `setConfigSettings:` recreates the SDK's URL session via `invalidateAndCancel`, which cancels that in-flight request. The result is that the documented pattern below fails on its first call:

```js
const remoteConfig = getRemoteConfig();
remoteConfig.settings = { minimumFetchIntervalMillis: 0, fetchTimeoutMillis: 60000 };
await fetchAndActivate(remoteConfig); // rejects: NSURLErrorCancelled (-999)
```

**Fix.** Route those four reads through a new `_afterPendingNativeMutations`, which waits for the mutation queue before touching native.

Two properties make this safe:

- `_enqueueNativeMutation` reassigns `_nativeMutationQueue` **synchronously** (only the `.then` callback is deferred), so by the time a read runs the queue already covers every write issued before it. There is no timing assumption and no added delay when nothing is queued.
- Reads **wait for** the queue rather than **joining** it — they do not reassign `_nativeMutationQueue`. A fetch can run for `fetchTimeoutMillis`, and must not hold up property writes made while it is still in flight. There is a test covering exactly this.

Argument validation in `fetch()` remains synchronous.

This removes the need for the "yield once before fetching" workaround.

**Known remaining piece (deliberately not in this PR).** There is a second, native-side contributor on iOS: the constants payload in `RNFBConfigHelper.m` reads `configSettings`, so the session can be recreated on calls that were not meant to mutate settings. That is a separate change in native code with a different risk profile, so this PR is scoped to the JS ordering bug only.

### Related issues

Fixes #9194

### Release Summary

Fixed `fetchAndActivate()` / `fetch()` / `activate()` / `ensureInitialized()` being cancelled on iOS with `NSURLErrorCancelled (-999)` when called right after assigning `settings` or `defaultConfig`.

### Checklist

  - [x] Yes
  - [x] `Android`
  - [x] `iOS`
  - [x] `Other` (macOS, web)
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

The fix is in shared JS, so all platforms get the correct ordering; the crash it resolves is iOS-specific.

No new e2e tests: `packages/remote-config/e2e/config.e2e.js` already exercises this exact sequence (`remoteConfig.settings = { minimumFetchIntervalMillis: 0 }` immediately followed by `await fetchAndActivate(...)`) at lines 71, 85 and 103, so the existing suite covers the path against a real device.

No public type changed — `_afterPendingNativeMutations` is private and the four method signatures are unchanged.

### Test Plan

New jest suite: `packages/remote-config/__tests__/nativeCallOrdering.test.ts` (9 tests). The native module is exposed through a `Proxy` that refuses `set`, so `jest.spyOn` cannot be used; ordering is read off `mock.invocationCallOrder` on the `jest.fn()`s that `jest.setup.ts` already installs.

Coverage:

- a queued `settings` write reaches native before each of `fetchAndActivate`, `fetch`, `activate`, `ensureInitialized`
- a queued `defaultConfig` write reaches native before `fetchAndActivate`
- two queued writes stay in order relative to each other and to the fetch
- **non-vacuity:** with nothing queued, all four reads still reach native exactly once and resolve with the native result
- `fetch()` still validates its argument synchronously
- a `settings` write still reaches native while a fetch is in flight (guards the wait-for vs. join-the-queue choice)

Counterfactual — with the source change reverted and the new tests kept:

```
Test Suites: 1 failed, 2 passed, 3 total
Tests:       6 failed, 29 passed, 35 total
```

Failing with, for example, `Expected: < 9, Received: 10` — native received `fetchAndActivate` one call ahead of `setConfigSettings`.

With the fix applied:

```
Test Suites: 3 passed, 3 total
Tests:       35 passed, 35 total
```

Baseline before this PR was 26/26 in 2 suites, so all 26 pre-existing tests still pass and 9 are added. `tsc --noEmit`, `eslint` and `prettier --check` are clean on both changed files.

### DIFF
--- a/packages/remote-config/__tests__/nativeCallOrdering.test.ts
+++ b/packages/remote-config/__tests__/nativeCallOrdering.test.ts
@@ -1,0 +1,182 @@
+/*
+ * Copyright (c) 2016-present Invertase Limited & Contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this library except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+import { beforeEach, describe, expect, it } from '@jest/globals';
+import { TurboModuleRegistry } from 'react-native';
+
+import {
+  activate,
+  ensureInitialized,
+  fetchAndActivate,
+  fetchConfig,
+  getRemoteConfig,
+} from '../lib';
+
+// The native module is exposed through a Proxy that refuses `set`, so `jest.spyOn` cannot be used
+// here. The jest setup already installs `jest.fn()`s for every method, so ordering is read straight
+// off `mock.invocationCallOrder`, which is a process-wide monotonic counter shared by all mocks.
+const native = TurboModuleRegistry.get('NativeRNFBTurboConfig') as any;
+
+const MOCKED_METHODS = [
+  'setConfigSettings',
+  'setDefaults',
+  'activate',
+  'fetch',
+  'fetchAndActivate',
+  'ensureInitialized',
+];
+
+function firstCallOrderOf(method: string): number {
+  const calls: number[] = native[method].mock.invocationCallOrder;
+  if (calls.length === 0) {
+    throw new Error(`expected the native \`${method}\` to have been called, but it never was`);
+  }
+  return calls[0];
+}
+
+/** Resolves after every pending microtask has run. */
+function flushMicrotasks(): Promise<void> {
+  return new Promise(resolve => setTimeout(resolve, 0));
+}
+
+const CONSTANTS = {
+  lastFetchTime: Date.now(),
+  lastFetchStatus: 'success',
+  fetchTimeout: 60,
+  minimumFetchInterval: 43200,
+  values: {},
+};
+
+describe('remoteConfig() native call ordering', function () {
+  let config: any;
+
+  beforeEach(async function () {
+    config = getRemoteConfig();
+    // `getRemoteConfig()` is a singleton, so drain any mutation left queued by a previous test
+    // before clearing the recorded call order.
+    await flushMicrotasks();
+    MOCKED_METHODS.forEach(method => native[method].mockClear());
+  });
+
+  describe('deferring reads behind queued property writes', function () {
+    // The `settings` and `defaultConfig` setters cannot return a promise, so they hand the native
+    // write to a microtask queue. Reads used to call native synchronously and therefore overtook
+    // that queued write. On iOS, `setConfigSettings:` landing mid-fetch tears down and recreates
+    // the SDK's URL session, cancelling the in-flight request with NSURLErrorCancelled (-999).
+    // https://github.com/invertase/react-native-firebase/issues/9194
+
+    it('sends a queued `settings` write before `fetchAndActivate`', async function () {
+      config.settings = { minimumFetchIntervalMillis: 0, fetchTimeoutMillis: 60000 };
+      await fetchAndActivate(config);
+
+      expect(native.setConfigSettings).toHaveBeenCalledTimes(1);
+      expect(native.fetchAndActivate).toHaveBeenCalledTimes(1);
+      expect(firstCallOrderOf('setConfigSettings')).toBeLessThan(
+        firstCallOrderOf('fetchAndActivate'),
+      );
+    });
+
+    it('sends a queued `defaultConfig` write before `fetchAndActivate`', async function () {
+      config.defaultConfig = { some_flag: true };
+      await fetchAndActivate(config);
+
+      expect(native.setDefaults).toHaveBeenCalledTimes(1);
+      expect(native.fetchAndActivate).toHaveBeenCalledTimes(1);
+      expect(firstCallOrderOf('setDefaults')).toBeLessThan(firstCallOrderOf('fetchAndActivate'));
+    });
+
+    it('sends a queued `settings` write before `fetch`', async function () {
+      config.settings = { minimumFetchIntervalMillis: 0 };
+      await fetchConfig(config);
+
+      expect(firstCallOrderOf('setConfigSettings')).toBeLessThan(firstCallOrderOf('fetch'));
+    });
+
+    it('sends a queued `settings` write before `activate`', async function () {
+      config.settings = { minimumFetchIntervalMillis: 0 };
+      await activate(config);
+
+      expect(firstCallOrderOf('setConfigSettings')).toBeLessThan(firstCallOrderOf('activate'));
+    });
+
+    it('sends a queued `settings` write before `ensureInitialized`', async function () {
+      config.settings = { minimumFetchIntervalMillis: 0 };
+      await ensureInitialized(config);
+
+      expect(firstCallOrderOf('setConfigSettings')).toBeLessThan(
+        firstCallOrderOf('ensureInitialized'),
+      );
+    });
+
+    it('keeps the writes in order when both setters run before a fetch', async function () {
+      config.defaultConfig = { some_flag: true };
+      config.settings = { minimumFetchIntervalMillis: 0 };
+      await fetchAndActivate(config);
+
+      expect(firstCallOrderOf('setDefaults')).toBeLessThan(firstCallOrderOf('setConfigSettings'));
+      expect(firstCallOrderOf('setConfigSettings')).toBeLessThan(
+        firstCallOrderOf('fetchAndActivate'),
+      );
+    });
+  });
+
+  describe('reads with nothing queued', function () {
+    it('still reaches native and resolves with the native result', async function () {
+      await expect(fetchAndActivate(config)).resolves.toBe(true);
+      await expect(activate(config)).resolves.toBe(true);
+      await expect(ensureInitialized(config)).resolves.toBe(true);
+      await expect(fetchConfig(config)).resolves.toBe(true);
+
+      expect(native.fetchAndActivate).toHaveBeenCalledTimes(1);
+      expect(native.activate).toHaveBeenCalledTimes(1);
+      expect(native.ensureInitialized).toHaveBeenCalledTimes(1);
+      expect(native.fetch).toHaveBeenCalledTimes(1);
+      expect(native.setConfigSettings).not.toHaveBeenCalled();
+      expect(native.setDefaults).not.toHaveBeenCalled();
+    });
+
+    it('keeps validating `fetch` arguments synchronously', function () {
+      expect(() => config.fetch('not a number')).toThrow(
+        "firebase.remoteConfig().fetch(): 'expirationDurationSeconds' must be a number value.",
+      );
+      expect(native.fetch).not.toHaveBeenCalled();
+    });
+  });
+
+  describe('reads do not hold up later writes', function () {
+    it('lets a `settings` write through while a fetch is still in flight', async function () {
+      // A fetch may run for up to `fetchTimeoutMillis`. Reads wait for the queue but must not join
+      // it, otherwise every later property write would be stuck behind an in-flight fetch.
+      let releaseFetch: () => void = () => {};
+      native.fetchAndActivate.mockImplementationOnce(
+        () =>
+          new Promise(resolve => {
+            releaseFetch = () => resolve({ result: true, constants: CONSTANTS });
+          }),
+      );
+
+      const pendingFetch = fetchAndActivate(config);
+      config.settings = { minimumFetchIntervalMillis: 0 };
+
+      await flushMicrotasks();
+      expect(native.fetchAndActivate).toHaveBeenCalledTimes(1);
+      expect(native.setConfigSettings).toHaveBeenCalledTimes(1);
+
+      releaseFetch();
+      await expect(pendingFetch).resolves.toBe(true);
+    });
+  });
+});
--- a/packages/remote-config/lib/index.ts
+++ b/packages/remote-config/lib/index.ts
@@ -317,7 +317,9 @@ class FirebaseConfigModule extends FirebaseModule<typeof nativeModuleName> {
    * @returns {Promise<boolean>}
    */
   activate(): Promise<boolean> {
-    return this._promiseWithConstants(this.native.activate());
+    return this._afterPendingNativeMutations(() =>
+      this._promiseWithConstants(this.native.activate()),
+    );
   }
 
   /**
@@ -333,17 +335,23 @@ class FirebaseConfigModule extends FirebaseModule<typeof nativeModuleName> {
       );
     }
 
-    return this._promiseWithConstants(
-      this.native.fetch(expirationDurationSeconds !== undefined ? expirationDurationSeconds : -1),
+    return this._afterPendingNativeMutations(() =>
+      this._promiseWithConstants(
+        this.native.fetch(expirationDurationSeconds !== undefined ? expirationDurationSeconds : -1),
+      ),
     );
   }
 
   fetchAndActivate(): Promise<boolean> {
-    return this._promiseWithConstants(this.native.fetchAndActivate());
+    return this._afterPendingNativeMutations(() =>
+      this._promiseWithConstants(this.native.fetchAndActivate()),
+    );
   }
 
   ensureInitialized(): Promise<void> {
-    return this._promiseWithConstants(this.native.ensureInitialized());
+    return this._afterPendingNativeMutations(() =>
+      this._promiseWithConstants(this.native.ensureInitialized()),
+    );
   }
 
   /**
@@ -518,6 +526,21 @@ class FirebaseConfigModule extends FirebaseModule<typeof nativeModuleName> {
       () => undefined,
     );
     return next;
+  }
+
+  private _afterPendingNativeMutations<T>(task: () => Promise<T>): Promise<T> {
+    // The `settings` and `defaultConfig` setters cannot be awaited, so they hand their native write
+    // to `_enqueueNativeMutation`, which only reaches native on a later microtask. Reads have to
+    // wait for that write, otherwise native sees the read first and the write lands mid-flight — on
+    // iOS `setConfigSettings:` then recreates the SDK's URL session and cancels the running fetch
+    // with NSURLErrorCancelled (-999). See #9194.
+    //
+    // `_enqueueNativeMutation` reassigns `_nativeMutationQueue` synchronously, so by the time a read
+    // runs the queue already covers every write issued before it — no timing assumption needed.
+    //
+    // Reads wait for the queue rather than joining it: a fetch can run for `fetchTimeoutMillis`, and
+    // must not hold up property writes made while it is still in flight.
+    return this._nativeMutationQueue.then(task, task);
   }
 }
 


russellwheatley:
Hey, thanks for the PR, we already have one in the pipeline with an e2e test: https://github.com/invertase/react-native-firebase/pull/9196

PR #9198, opened 2026-08-14
docs(auth): use PhoneAuthProvider in MFA sign-in examples

### LINKED ISSUE #9197: [📚] Outdated documentation or wrong type definition of Auth module
The [Multi-factor Auth ](https://rnfirebase.io/auth/multi-factor-auth#sign-in-flow-using-phone-multi-factor) documentation says to call the method `verifyPhoneNumberWithMultiFactorInfo` of the auth module, but that method is not exposed by the Auth type.

The method exists in the type definition of the auth module (found [here](https://github.com/invertase/react-native-firebase/blob/de8dbc08b69e6b4aac7421fefec400a589852044/packages/auth/lib/index.ts#L513), but the [`getAuth` function](https://github.com/invertase/react-native-firebase/blob/de8dbc08b69e6b4aac7421fefec400a589852044/packages/auth/lib/index.ts#L1014) is set to return a [stripped down version](https://github.com/invertase/react-native-firebase/blob/de8dbc08b69e6b4aac7421fefec400a589852044/packages/auth/lib/types/auth.ts#L40) of it, that does not include the `verifyPhoneNumberWithMultiFactorInfo` method.

Either the return type of the `getAuth` function should include the method, or the documentation should be changed to reflect the new way of doing multi factor authentication (if one exists).

russellwheatley:
Fixes #9197

The MFA sign-in snippet called `getAuth().verifyPhoneNumberWithMultiFactorInfo`, which is not on the public `Auth` type. Use `PhoneAuthProvider.verifyPhoneNumber` instead.

Also pass `auth` into `getMultiFactorResolver` in the combined example.

### DIFF
--- a/docs/auth/multi-factor-auth.mdx
+++ b/docs/auth/multi-factor-auth.mdx
@@ -123,9 +123,11 @@ verification code to the user:
 
 ```js
 const hint = resolver.hints[0];
-const sessionId = resolver.session;
 
-const verificationId = await getAuth().verifyPhoneNumberWithMultiFactorInfo(hint, resolver.session); // Triggers message to user
+const verificationId = await new PhoneAuthProvider(getAuth()).verifyPhoneNumber({
+  multiFactorHint: hint,
+  session: resolver.session,
+}); // Triggers message to user
 ```
 
 Once the user has entered the verification code you can create a multi-factor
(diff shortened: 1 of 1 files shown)


mikehardy (approved):
good fix, thanks

PR #9208, opened 2026-08-20
build(release): update CocoaPods lockfiles during release

mikehardy:
## Summary

Why? After every release, the test apps Podfile.lock files are churning in our local workspaces because installs cause them to bump their versions to match the release. Why don't we just bump them during release?

This PR does that: it eliminates post-release Pod lockfile churn by bumping the file during release

It fixes everything else I saw in the area at the same time

Atomic commits

- Pin CocoaPods 1.17.0 and xcodeproj 1.28.1 in a root Gemfile; CI/Tart use frozen bundle install (no gem update).
- Stop deleting Podfile.lock on yarn pod install; SDK updater removes the iOS lockfile only when firebase-ios-sdk changes.
- Refresh iOS/macOS lockfiles on the pinned gems (iOS RNFB pods match lerna 26.3.1 after the tokenless publish).
- Run publish on macos-26 and refresh CocoaPods lockfiles in the Darwin lerna version hook (fmt gate, two yarn pod installs, git diff --exit-code). Timeout 60 minutes.

Fixes CPRN-332

### Related issues

Fixes CPRN-332

### Release Summary

Keep CocoaPods lockfiles in the release commit so worktrees stay clean after lerna version.

### Checklist

  - [x] Yes
  - [ ] `Android`
  - [x] `iOS`
  - [x] `Other` (macOS publish / CocoaPods)
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

- [x] `yarn tests:ios:ruby` after the root Gemfile pin
- [x] `yarn tests:ios:pod:install` and `yarn tests:macos:pod:install` (idempotent second run)
- [x] Independent review vs origin/main after rebase: yarn, lerna:prepare, bundle install, lint, version.js syntax, both pod installs, lockfile git diff --exit-code
- [x] CI green on this PR


package.json
```diff
@@ -68,11 +68,11 @@
     "tests:ios:test-cover-reuse": "cd tests && SIMCTL_CHILD_GULGeneratedClassDisposeDisabled=1 yarn detox test --configuration ios.sim.debug --reuse --loglevel warn",
     "tests:ios:test:process-coverage": "node tests/scripts/process-ios-native-coverage.js",
     "tests:ios:test-cover-and-process": "yarn tests:ios:test-cover && yarn tests:ios:test:process-coverage",
-    "tests:ios:pod:install": "rm -f tests/ios/Podfile.lock && rm -rf tests/ios/ReactNativeFirebaseDemo.xcworkspace && bundle exec pod install --project-directory=tests/ios",
+    "tests:ios:pod:install": "cd tests/ios && bundle exec pod install",
     "lint:ruby": "bundle exec rubocop --config packages/app/__tests__/.rubocop.yml packages/app/firebase_spm.rb packages/app/firebase_json.rb packages/app/__tests__",
     "tests:ios:ruby": "yarn lint:ruby && bundle exec ruby packages/app/__tests__/run_with_coverage.rb",
     "tests:macos:build": "cd tests-macos && yarn build:macos",
-    "tests:macos:pod:install": "rm -f tests-macos/macos/Podfile.lock && bundle exec pod install --project-directory=tests-macos/macos",
+    "tests:macos:pod:install": "bundle exec pod install --project-directory=tests-macos/macos",
```
mikehardy: - not sure why these 2 have slightly different patterns - one does a cd and one uses --project-directory
- what's the institutional knowledge on the rm -rf of the xcworkspace? git history may be informative - maybe we're losing something here?

scripts/version.js
```diff
@@ -92,3 +184,18 @@ packages.forEach(package => {
...
+if (process.platform === 'darwin') {
+  assertTestsMacosFmtAtLeastFloor();
+
+  execSync('yarn tests:ios:pod:install', { stdio: 'inherit' });
+  execSync('yarn tests:macos:pod:install', { stdio: 'inherit' });
+  execSync(`git add -- ${POD_LOCKFILE_PATHS.join(' ')}`, { stdio: 'inherit' });
+
+  execSync('yarn tests:ios:pod:install', { stdio: 'inherit' });
+  execSync('yarn tests:macos:pod:install', { stdio: 'inherit' });
+  execSync(`git diff --exit-code -- ${POD_LOCKFILE_PATHS.join(' ')}`, { stdio: 'inherit' });
```
mikehardy, line 201: add comment indicating the second run is to verify it the update is idempotent and won't continue to drift after staging/committing the update

tests/ios/Podfile.lock
```diff
@@ -2536,7 +2536,7 @@ PODS:
     - React-Core
   - RNDeviceInfo (15.0.2):
     - React-Core
-  - RNFBAnalytics (26.3.0):
+  - RNFBAnalytics (26.3.1):
```
mikehardy, line 2539: Amend this entire commit into the original 26.3.0 bump and force push the branch back out, no need to have an "intermediate" bump to 26.3.0 in the commit stream with 26.3.1 already out

mikehardy (commented):
few quick items to fix

package.json
russellwheatley: not even sure what that is for `rm -rf tests/ios/ReactNativeFirebaseDemo.xcworkspace` - must be legacy from a while ago 🤔 

I don't think it is normal practice to remove xcworkspace i.e. its fine to remove.

russellwheatley (approved):
We'll probably need to refresh Podifle.lock if/when gem dependencies are updated (thinking dependabot) but other than that, LGTM

package.json
mikehardy: I had the same question - git history spelunking indicates it was from 2019, hasn't actually done anything in ages, and is not best practice anyway, no

mikehardy:
I'm going to merge this then specifically do a release - it was one "fix" commit that should trigger it, and the release pipeline is otherwise empty / has no other commits in it so it's a safe time to test

PR #9211, opened 2026-08-21
fix(ios): link FirebaseCore into Expo app target under SPM

### LINKED ISSUE #9158: useFrameworks: static + forceStaticLinking rejected with "SPM + static linkage is not supported" on RNFBApp/RNFBMessaging 26.1.0 (Expo 54.0.36)
Environment:
- Expo SDK 54.0.36
- React Native 0.81.5
- @react-native-firebase/app 26.1.0
- @react-native-firebase/messaging 26.1.0
- expo-build-properties (latest via npx expo install)
- EAS Build, iOS

I'm seeing a direct conflict between current documentation and actual build behavior on this version.

Following the documented pattern (and this recent guide: https://hwangrolee.github.io/blog/expo-firebase-non-modular-header-error/), I configured:

"plugins": [
  "@react-native-firebase/app",
  "@react-native-firebase/messaging",
  [
    "expo-build-properties",
    {
      "ios": {
        "useFrameworks": "static",
        "forceStaticLinking": ["RNFBApp", "RNFBMessaging"]
      }
    }
  ]
]

This fails during pod install with:

[!] [react-native-firebase] SPM + static linkage is not supported (target(s): Pods-FDBeacon).
firebase-ios-sdk's Swift Package only ships dynamic library products, so `use_frameworks! :linkage => :static` causes every react-native-firebase pod that resolves Firebase via SPM to embed its own copy of the same Firebase frameworks -- this produces duplicate-symbol linker errors at build time instead of a clear error here.

This appears to be a hard validation check in the plugin itself, not a generic Xcode/CocoaPods error — it explicitly names "SPM + static linkage" as unsupported.

Switching to useFrameworks: "dynamic" (with forceStaticLinking pointed at an unrelated native module we also use, react-native-nfc-manager, which needed static linking for a separate reason) gets past this error and successfully builds and links every individual Firebase pod/framework. However, the final app target then fails at the very last link step with:

Undefined symbols for architecture arm64
Symbol: _OBJC_CLASS_$_FIRApp
Referenced from: in AppDelegate.o

along with these warnings just before it:
ld: Could not find or use auto-linked framework 'CoreAudioTypes': framework 'CoreAudioTypes' not found
ld: Could not find or use auto-linked framework 'UIUtilities': framework 'UIUtilities' not found

Firebase.framework itself builds successfully earlier in the same log — this only fails when the main app target tries to link against it.

Questions:
1. Is the static + forceStaticLinking pattern still expected to work on 26.1.0 specifically, or did a recent version introduce the stricter SPM check that's now blocking it?
2. If dynamic linking is the correct path going forward for this version, what's needed to get the main app target to correctly find FIRApp — is this a known missing step in the RNFB config plugin's handling of "dynamic" mode specifically?

Happy to share the full build log if useful.

### LINKED ISSUE #9219: [🔥] [iOS][SPM] FirebaseCore is not linked into the app target until a second pod install — breaks single-pass CI (EAS, Expo prebuild)
## Issue

Under SPM resolution, RNFB does not link `FirebaseCore` into the app target on the **first** `pod install` of a clean tree. It links on the second. Since `AppDelegate` calls `[FIRApp configure]`, the first-pass project fails to link:

```
Undefined symbols for architecture arm64:
  "_OBJC_CLASS_$_FIRApp", referenced from:
       in AppDelegate.o
ld: symbol(s) not found for architecture arm64
```

This matters beyond local machines: EAS Build (and any CI that runs `expo prebuild` then one `pod install`) never gets a second pass, so the build simply fails. There is no hook between `pod install` and `xcodebuild` to insert one.

The skip is **silent** — no `[react-native-firebase]` message, no warning that automatic linking could not be applied. The only symptom is the linker error much later.

## Reproduction

Expo CNG project (no committed `ios/`), each row starting from a clean tree:

| what was run | `FirebaseCore in Frameworks` on app target? |
| --- | --- |
| `expo prebuild -p ios --clean` (runs one `pod install`) | no |
| `expo prebuild -p ios --clean --no-install`, then one `pod install` | no — and no warning emitted |
| same, plus explicit `rnfirebase_add_spm_core_to_app_target(installer)` in `post_install` | no |
| a second `pod install` | **yes** — logs `Linking FirebaseCore directly into the app target (SPM) …` |

Checked with `rg -c "FirebaseCore in Frameworks" ios/<App>.xcodeproj/project.pbxproj` after each.

## Cause

`packages/app/firebase_spm.rb`, in `rnfirebase_add_spm_core_to_app_target`:

```ruby
next unless target.shell_script_build_phases.any? { |phase| phase.name == '[CP] Embed Pods Frameworks' }
```

On a clean first install that phase does not yet exist on the in-memory user project — CocoaPods writes it during user-project integration, after `post_install` hooks run. On the second install it is already on disk, so the guard passes.

Because the documented fallback calls the same function from `post_install`, it is subject to the same guard. That is why row 3 above changes nothing, and it means **no `post_install`-based fix can work in one pass**. The docs currently present that fallback as the remedy for exactly this symptom:

> **Archive fails with `Undefined symbols ... _OBJC_CLASS_$_FIRApp`:** … call `rnfirebase_add_spm_core_to_app_target(installer)` in your `post_install` block as a fallback

## Possible directions

- Drop the `[CP] Embed Pods Frameworks` precondition, or fall back to another signal for "this is the app target" when the phase is not present yet.
- Hook later than `post_install` (e.g. after the user project is written) so the phase is visible on the first pass.
- At minimum, emit the "could not apply automatically" warning when the guard skips every target — right now a first-pass install looks completely clean.

## Second, smaller issue in the same area

The SPM branch of the Crashlytics dSYM script phase hardcodes:

```bash
"${SPM_UPLOAD_SYMBOLS}" -gsp "${PROJECT_DIR}/GoogleService-Info.plist" …
```

Expo prebuild writes the plist into the app-target folder (`ios/<App>/GoogleService-Info.plist`), not `${PROJECT_DIR}`, so the build fails with:

```
error: Unable to read Google Service plist at path /…/ios/GoogleService-Info.plist
```

The CocoaPods branch is unaffected because `${PODS_ROOT}/FirebaseCrashlytics/run` auto-discovers the plist. Happy to open this separately if you'd prefer.

## Environment

- `@react-native-firebase/{app,analytics,crashlytics}`: `26.2.0`
- React Native: `0.83.2`, New Architecture enabled, Hermes
- Expo SDK 55 (CNG — `ios/` generated by prebuild, not committed)
- Linkage: `use_frameworks! :linkage => :dynamic` via `expo-build-properties`
- SPM mode (no `$RNFirebaseDisableSPM`); `pod install` confirms `Using SPM for Firebase dependency resolution` for all three pods
- iOS deployment target 17.0; Xcode 26.5; simulator, Debug

Tested on 26.2.0 rather than 26.3.0 because 26.3.0 does not compile for us (#9206).

Happy to test a patch.

russellwheatley:
- Fixes #9158
- Fixes #9219

Expo `prebuild --clean` still failed iOS linking with undefined `_OBJC_CLASS_$_FIRApp` after #9164. The app-target helper ran during CocoaPods `post_install`, before `integrate_user_project` added `[CP] Embed Pods Frameworks`, so the generated app target was skipped.

- Run user-project SPM helpers through the automatic `post_integrate` hook so the app target receives the FirebaseCore package product and Frameworks link after CocoaPods integration
- Save `Pods.xcodeproj` when those later hooks change build settings, preserving the explicit-modules workaround used by the existing iOS test app
- Add `yarn test-expo:ios:link` as the documented-path closer and keep the fixture's RNFB versions synced during releases

The manual `post_integrate` Podfile snippet is a fallback only when the automatic hook reports that it could not be installed. Users do not need to add it normally.

The separate Crashlytics plist-path report in #9219 is outside this change. I've asked the OP to create a separate issue for it.

---
Maintainer note: Fixes CPRN-301


mikehardy (approved):
LGTM - there was one thing I saw where the frontmatter changes in recent OKF fixup:

- #9223 

...wasn't taken into account during the addition of the new packages/app/index.md file

I amended the obvious fix into this and once CI is green this should be good to go?

PR #9215, opened 2026-08-22
build: install gems via bundler via yarn hook

mikehardy:
## Summary
- Add `yarn ruby:install` (`bundle check || bundle install`) and run it from `postinstallDev` so a root `yarn` installs root Gemfile gems after the move to `bundle exec pod`.
- Reorder CI to run `BUNDLE_FROZEN=true bundle install` **before** yarn; postinstall `ruby:install` is then a no-op via `bundle check`, keeping frozen lockfile enforcement intact.
- Update OKF install docs so agents use `yarn ruby:install` / root `yarn`, not bare `bundle install`.

Noticed while validating e2e after rebasing onto `main`: fresh worktrees that only ran `yarn` hit `Bundler::GemNotFound` on the first Apple `pod:install`. This closes that bootstrap hole for local/dev paths analyzed while keeping CI supply-chain hardening.

- Related: #9208

---
Maintainer note: internal tracking Fixes CPRN-367


package.json
```diff
@@ -3,7 +3,8 @@
   "private": true,
   "scripts": {
     "version": "node ./scripts/version.js",
-    "postinstallDev": "yarn prepare && yarn lerna:prepare",
+    "postinstallDev": "yarn prepare && yarn lerna:prepare && yarn ruby:install",
+    "ruby:install": "bundle check || bundle install",
```
russellwheatley: `bundle check || bundle install` fails the whole yarn when `bundle` is not on PATH. Yarn runs both sides of `||`, you get two `command not found: bundle`, postinstallDev exits 127.

That's Lint / Jest / TypeScript / attw / Android on this PR. Those jobs never run setup-ruby. Same for a JS- or Android-only local checkout.

Skip if bundle isn't there. Don't add setup-ruby to every workflow just to make this succeed.

.github/workflows/tests_e2e_ios.yml
```diff
@@ -245,8 +236,11 @@ jobs:
 
       # Pinned cocoapods/xcodeproj + SimpleCov/RuboCop from the root Gemfile.
       # Shape/embed suites need those gems; yarn tests:ios:pod:install uses
-      # `bundle exec pod`. BUNDLE_FROZEN fails CI on Gemfile.lock drift
-      # (do not use deprecated --frozen; it would persist into .bundle/config).
+      # `bundle exec pod`. Run frozen bundle install BEFORE yarn so CI remains
+      # supply-chain attack-hardened via the freeze; after yarn, postinstallDev →
+      # yarn ruby:install is a no-op (bundle check succeeds).
```
russellwheatley, line 241: This reorder is correct here. Frozen bundle first, then yarn, hook is a no-op.

It does not cover tests.yml / Android / lint / types. Those still `yarn` with no Ruby, and that's where CI is red. Fix the script, not those workflows.

russellwheatley (changes requested):
Local hole after the root Gemfile move is real, and hanging gem install off `postinstallDev` is the right place. As written, `yarn ruby:install` makes Bundler a hard dependency of every root `yarn`.

JS/Android CI never runs setup-ruby. Lint already died with `command not found: bundle` twice, then postinstallDev 127. Same for Jest, types, attw, spelling, patch-package, Android. iOS/macOS are green only because those jobs already install Ruby first.

Skip the hook when `bundle` is not on PATH. Don't add setup-ruby to every workflow just to make this succeed.

package.json
mikehardy: good catch, obviously worked for me locally since I have ruby installed, and I didn't check CI post-push

russellwheatley (approved):

PR #9221, opened 2026-08-24
docs(okf): keep gated work queues on Linear, not in git

russellwheatley:
Gated / multi-item work queues live on a Linear project document, not in `okf-bundle/`. The issue body points at that document from both Work Completed and Work To Be Completed.

Existing `*work-queue.md` files stay until next pickup, then migrate and delete. No dual-write.

Companion docs: Cross Platform Issue Authoring & Agent Workflow Guide (Step 10), How the Work Queue Orchestrator Works.


mikehardy (approved):
I think this needs to go in still or there's drift between linear and rnfb - let's go

PR #9224, opened 2026-08-26
docs(okf): add e2e startup fail-fast markers

russellwheatley:
`main` already has § startup fail-fast poll. This adds extra abort matches to that table: Metro bundle URL, missing `Detox.framework`, Android ABI mismatch, stale Jet client, and exhausted `launchApp` timeout. `launchApp failure reason=` during in-process retry is not an abort.

Also adds the `#startup-fail-fast-markers` alias and a stalled-run pointer at the poll section.

Keeps the existing Do not match rules: `currentStatus` / `APP_STATUS` timeout is latency, and Metro `:12007` is only a serial leftover when paired with `emulator-5554`.

---
Maintainer note: Related to internal CPRN-334


mikehardy (approved):
LGTM

PR #9229, opened 2026-08-27
fix(messaging, ios): emit onMessage for foreground pushes with content-available: 0

### LINKED ISSUE #9228: [🐛] iOS: onMessage never fires in foreground when APNs payload has `content-available: 0`
### Summary

On iOS, `messaging().onMessage()` is **never called** for a foreground push whose APNs payload explicitly sets `aps.content-available: 0`.

- The same push is delivered and displayed correctly in the **background** and **quit** states, and tapping it works.
- The same app receives `onMessage` normally for foreground pushes that **omit** the `content-available` key.
- Nothing arrives at JS at all: no event, no error, nothing in the notification center either (since the app never gets a chance to present it).

`content-available: 0` is valid APNs usage — it is an explicit declaration that the push is *not* a silent push. Any sender that builds its own `aps` dictionary and serializes that field as `0` hits this. We first saw it after integrating a third‑party marketing push provider that sets `mutable-content: 1` and assembles `apns.payload.aps` itself; our own backend pushes (FCM `notification` block only, no `content-available` key) were unaffected, which is why it went unnoticed.

### Usage

```ts
import { getMessaging, onMessage } from '@react-native-firebase/messaging';

useEffect(() => {
  const unsubscribe = onMessage(getMessaging(), async remoteMessage => {
    console.log('foreground message', remoteMessage); // never logged for content-available: 0
  });
  return unsubscribe;
}, []);
```

### Reproduction

Send two foreground pushes to the same iOS device (physical device — the simulator cannot receive remote pushes) and observe `onMessage`:

| # | Payload | `onMessage` |
|---|---|---|
| 1 | `{"aps":{"alert":{"title":"a","body":"b"}}}` | fires ✅ |
| 2 | `{"aps":{"alert":{"title":"a","body":"b"},"content-available":0}}` | **never fires** ❌ |

Both payloads carry `gcm.message_id`, so both pass the outer guard in `willPresentNotification`; only the inner `contentAvailable` check differs.

Via FCM HTTP v1 the second payload is produced by:

```json
{
  "message": {
    "token": "<device token>",
    "apns": {
      "payload": {
        "aps": {
          "alert": { "title": "a", "body": "b" },
          "content-available": 0
        }
      }
    }
  }
}
```

### Root cause

`RNFBMessagingSerializer` stores `contentAvailable` whenever the key is **present**, regardless of its value:

```objc
// packages/messaging/ios/RNFBMessaging/RNFBMessagingSerializer.m
// message.contentAvailable
if (apsDict[@"content-available"] != nil) {
  message[@"contentAvailable"] = @([RCTConvert BOOL:apsDict[@"content-available"]]);
}

// message.mutableContent  <- note: this one compares the value
if (apsDict[@"mutable-content"] != nil && [apsDict[@"mutable-content"] intValue] == 1) {
```

So `content-available: 0` yields `@(NO)` — a non-nil `NSNumber`.

`willPresentNotification` then guards event emission on **key presence** rather than on the value:

```objc
// packages/messaging/ios/RNFBMessaging/RNFBMessaging+UNUserNotificationCenter.m:122
// Don't send an event if contentAvailable is true -
// application:didReceiveRemoteNotification will send the event for us, we
// don't want to duplicate them
if (!notificationDict[@"contentAvailable"]) {
  [[RNFBRCTEventEmitter shared] sendEventWithName:@"messaging_message_received"
                                             body:notificationDict];
}
```

`@0` is a non-nil object, so `!notificationDict[@"contentAvailable"]` is false and `messaging_message_received` is never emitted. The comment directly above states the intent in terms of the **value** ("if contentAvailable is true"), so this is an implementation/intent mismatch rather than deliberate behaviour.

This guard is older than PR #5604 — that PR actually *removed* it and it was restored by the revert in #5641 (Aug 2021). It has not been revisited since, including through the `willPresentNotification` changes this year (#8786, #9094, #8945). The code path is unchanged on `main` at the time of writing.

### Fix

Compare the value, matching the stated intent. One line:

```diff
-    if (!notificationDict[@"contentAvailable"]) {
+    if (![notificationDict[@"contentAvailable"] boolValue]) {
```

Genuine silent pushes (`content-available: 1`) are still skipped, preserving the de‑duplication with `application:didReceiveRemoteNotification:`.

PR with this one-line fix follows (linked below).

Verified on a physical device before/after the change by streaming the native delegate with `xcrun devicectl device process launch --console`: before, the `content-available: 0` payload reaches `willPresentNotification` but no event is emitted; after, `onMessage` fires and the serialized message carries `"contentAvailable": false` — confirming the serializer had been producing a correct dictionary all along and only the emission guard was wrong.

### Workaround (until released)

Patch the one line above with `patch-package` / pnpm `patchedDependencies`. There is no JS-side workaround: the decision not to emit lives inside the native bridge, so `onMessage` subscribers never see an event that was never sent.

### Possibly related

Same symptom string, but those threads never identified a payload-level cause and some describe a different failure mode (provisional authorization / token registration timing), so I am **not** claiming this explains them: #7772, #6107, #4513.

---

### Javascript

<p>

#### `package.json`:

```json
{
  "dependencies": {
    "react": "19.0.0",
    "react-native": "0.78.x",
    "@react-native-firebase/app": "19.3.0",
    "@react-native-firebase/messaging": "19.3.0"
  }
}
```

(Trimmed to the relevant packages. The bug is independent of app configuration — it is a payload-value check in the native bridge, reproducible with the two payloads above alone.)

```json
# N/A — defaults; no messaging-related keys set
```

</details>

### iOS

<p>

#### `ios/Podfile`:

- [ ] I'm not using Pods

```ruby
# Standard RN 0.78 Podfile; use_frameworks! :linkage => :static
# No RNFB-specific overrides. Not relevant to this issue (native bridge value check).
```

#### `AppDelegate.m`:

```objc
// Standard RN 0.78 AppDelegate (Swift). No custom UNUserNotificationCenterDelegate —
// RNFBMessaging's own delegate is the one handling willPresentNotification.
```

</p>
</details>

---

### Android

<p>

#### `android/build.gradle`:

```groovy
// N/A — iOS-only issue (the affected code is in the iOS bridge)
```

```groovy
// N/A
```

```groovy
// N/A
```

#### `MainApplication.java`:

```java
// N/A
```

#### `AndroidManifest.xml`:

```xml
<!-- N/A -->
```

</p>
</details>

---

## Environment

<p>

```
System:
  OS: macOS 26.x
Binaries:
  Node: 22.x
IDEs:
  Xcode: 26.x
npmPackages:
  react: 19.0.0
  react-native: 0.78.x
Device: iPhone (physical), iOS 26.5.1
JS engine: JSC, old architecture (newArchEnabled=false) — not relevant, the guard runs before the bridge
```

  - [ ] iOS
  - [ ] Android
  - [ ] Both
  - `19.3.0` (reproduced); the affected code is unchanged on `main` (`26.x`)
  - `Messaging`
  - `Y` & `5.x`

</p>
</details>

---

lnuvy:
### Description

On iOS, `onMessage` is never emitted for a foreground push whose payload sets
`aps.content-available: 0`. `willPresentNotification` guards emission on **key presence**
(`!notificationDict[@"contentAvailable"]`), but the serializer stores `@(NO)` for `0`, which is
non-nil — so the event is skipped as if it were a silent push. The comment above the guard
describes the intent in terms of the value ("if contentAvailable is true").

Compare the value instead:

```diff
-    if (!notificationDict[@"contentAvailable"]) {
+    if (![notificationDict[@"contentAvailable"] boolValue]) {
```

Genuine silent pushes (`content-available: 1`) are still skipped, preserving de-duplication
with `application:didReceiveRemoteNotification:`.

Full analysis, reproduction payloads and history in #9228.

### Related issues

Fixes #9228

### Release Summary

iOS: `onMessage` is now emitted for foreground pushes that explicitly set `content-available: 0`.

### Checklist

  - [x] Yes
  - [ ] `Android`
  - [x] `iOS`
  - [ ] `Other` (macOS, web)
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

Manual on a physical iOS 26.5.1 device (`@react-native-firebase/messaging` 19.3.0), native
delegate streamed via `xcrun devicectl device process launch --console`.

- Before: `content-available: 0` payload reaches `willPresentNotification`, no
  `messaging_message_received` emitted, JS `onMessage` never called. Same payload without the
  key fires normally.
- After: the same payload fires `onMessage`; serialized message carries `"contentAvailable": false`.

No ObjC test harness exists for this delegate path; happy to add a payload-variant e2e case if
maintainers point at the preferred shape.

### DIFF
--- a/packages/messaging/ios/RNFBMessaging/RNFBMessaging+UNUserNotificationCenter.m
+++ b/packages/messaging/ios/RNFBMessaging/RNFBMessaging+UNUserNotificationCenter.m
@@ -122,7 +122,7 @@ struct {
     // Don't send an event if contentAvailable is true -
     // application:didReceiveRemoteNotification will send the event for us, we
     // don't want to duplicate them
-    if (!notificationDict[@"contentAvailable"]) {
+    if (![notificationDict[@"contentAvailable"] boolValue]) {
       [[RNFBRCTEventEmitter shared] sendEventWithName:@"messaging_message_received"
                                                  body:notificationDict];
     }


russellwheatley (approved):
Confirmed against `main`: the serializer sets `contentAvailable` to `@(NO)` whenever `content-available` is present at all, and the emit guard was checking key presence instead of value, so `content-available: 0` got treated as a silent push and swallowed. The comment above the guard already states the intent in terms of the value, so this is a straightforward mismatch, not deliberate behavior.

Also checked the nil case: when `content-available` is absent, `notificationDict[@"contentAvailable"]` is `nil`, and `[nil boolValue]` returns `NO` per Objective-C's nil-messaging semantics (verified this by compiling and running it standalone). So `onMessage` still fires for pushes without the key, same as before. No crash risk there.

We don't have an XCTest harness for this delegate path yet, so this is verified by code review plus your manual device testing rather than an automated test. Given how small and low-risk the change is, that's enough for me here.

Approving, want to get this into the next release.

lnuvy:
@russellwheatley 
Thanks for the quick review !
I've signed the CLA, so that check is green now.
The PR still shows as blocked. From the outside it looks like it might need a second approval, but I can't see the branch protection rules so I may be misreading it. Is there anything else I can do on my end?

mikehardy (approved):
nice catch!
internally tracking unit testing for messaging etc / messaging improvements in general in CPRN-263 - this does not resolve that, but it is related

PR #9234, opened 2026-08-29
fix(analytics): validate hashed conversion inputs

OskarEichler:
### Description

Validate and safely decode the SHA-256 values passed to the hashed on-device conversion APIs.

React Native exposes the hashes as hex strings, while Firebase's Apple SDK requires the decoded 32-byte SHA-256 value. The previous decoder accepted any string, parsed non-hex pairs as lossy zero-like bytes, and indexed past the end of odd-length input. It could therefore pass malformed data to Firebase or raise an Objective-C exception.

This change:

- requires exactly 64 hexadecimal characters at the public JavaScript boundary;
- validates the same contract again at the native boundary;
- decodes directly into a bounded 32-byte buffer;
- documents the React Native hex-string representation.

Firebase's [on-device conversion guide](https://firebase.google.com/docs/tutorials/ads-ios-on-device-measurement/step-3#use-hashed-credentials) specifies that the SDK value must be a 32-byte SHA-256 digest rather than a hexadecimal string; this bridge decodes the validated 64-character representation before calling it.

### Breaking changes

No public API or type changes.

This is an observable input-validation correction: malformed, non-hex, or non-64-character hash strings now reject instead of being silently decoded into invalid bytes or triggering an out-of-bounds exception. Valid lowercase and uppercase SHA-256 hex strings are unchanged.

### Related issues

No matching open issue found.

### Release Summary

Validated and safely decoded hashed Analytics on-device conversion inputs on iOS.

### Checklist

  - [x] Yes
  - [x] `Android` (shared JavaScript validation/no-op contract)
  - [x] `iOS`
  - [x] `Other` (shared validation and macOS compilation)
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

- Analytics compile and existing Jest suites: 59/59 passing.
- Focused temporary boundary control: 7/7 passing across valid lower/uppercase hashes, empty/short/long/odd/non-hex input; no test file is included in the patch.
- iOS formatting and `git diff --check`: passing.
- Fresh iOS and macOS pod installs: passing.
- Android, iOS simulator, and macOS builds: passing.
- Android native unit suite: passing.
- Focused app + Analytics iOS E2E with valid hashed email and phone calls reaching native: 104 passing, 5 pending.

### DIFF
--- a/packages/analytics/__tests__/analytics.test.ts
+++ b/packages/analytics/__tests__/analytics.test.ts
@@ -270,12 +270,96 @@ describe('Analytics', function () {
       );
     });
 
+    it('`initiateOnDeviceConversionMeasurementWithHashedPhoneNumber` throws if not a string before E.164', function () {
+      expect(() =>
+        // @ts-ignore
+        initiateOnDeviceConversionMeasurementWithHashedPhoneNumber(getAnalytics(), true),
+      ).toThrow(
+        "firebase.analytics().initiateOnDeviceConversionMeasurementWithHashedPhoneNumber(*) 'hashedPhoneNumber' expected a string value.",
+      );
+    });
+
     it('`initiateOnDeviceConversionMeasurementWithHashedPhoneNumber` should throw if the value is in E.164 format', function () {
       expect(() =>
         initiateOnDeviceConversionMeasurementWithHashedPhoneNumber(getAnalytics(), '+1234567890'),
       ).toThrow(
-        "firebase.analytics().initiateOnDeviceConversionMeasurementWithHashedPhoneNumber(*) 'hashedPhoneNumber' expected a sha256-hashed value of a phone number in E.164 format.",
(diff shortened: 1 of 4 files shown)


mikehardy:
Rebased onto current `main` and added coverage for the SHA-256 hashed on-device conversion contract.

**Tests**
- Jest: empty / short / odd / long / non-hex reject, plus valid lowercase and uppercase 64-hex accept, for both hashed email and hashed phone. Hashed-phone non-strings are rejected as strings *before* the E.164 check.
- E2e (analytics area): the same JS matrix on Android, iOS, and macOS. On iOS, NativeModules bypass hits the native decoder reject paths (empty / odd / non-hex) so the 32-byte hex decode is actually exercised.

**Also in this update**
- Hashed-phone validation order is now string → E.164 → 64-character hex.
- E.164 error text mentions the 64-character SHA-256 hex contract.

**Local verification** (Android ∥ iOS ∥ macOS)
- Jest 74/74
- Android 117 passing, iOS 123, macOS 110
- 100% on reachable touched JS validation lines and the iOS native decoder

mikehardy (approved):
Hey @OskarEichler 👋

Cheers

PR #9236, opened 2026-08-29
fix(analytics, ios): serialize session completion

OskarEichler:
### Description

Serialize the iOS Analytics session-ID completion path on the main queue.

Firebase documents that `sessionIDWithCompletion:` invokes its completion handler on a system-defined global concurrent queue, while this bridge's timeout already runs on the main queue. Both paths read and mutate the same `__block BOOL completed`, so they can race at the timeout boundary and resolve the React Native promise twice.

The Firebase callback now marshals its completion work to the main queue before checking or updating `completed`. This makes the timeout and SDK callback mutually ordered while preserving the existing timeout, `null`, error, and successful session-ID behavior.

Firebase reference: https://firebase.google.com/docs/reference/ios/firebaseanalytics/api/reference/Classes/FIRAnalytics

### Breaking changes

None. Public APIs, types, timeout behavior, and resolved values are unchanged. The fix only prevents a race between the SDK completion and timeout paths.

### Related issues

No matching open issue found.

### Release Summary

Fixed a race that could resolve iOS Analytics `getSessionId()` more than once at the timeout boundary.

### Checklist

  - [x] Yes
  - [ ] `Android`
  - [x] `iOS`
  - [ ] `Other`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

- iOS Objective-C/Objective-C++ lint: passing.
- Fresh iOS and macOS pod installs: passing.
- iOS simulator build: passing.
- macOS build: passing.
- Focused App + Analytics iOS E2E: 102 passing, 5 pending.

No test file is changed: the synchronization boundary is native implementation detail, while the existing `getSessionId()` E2E coverage verifies successful native behavior.

### DIFF
--- a/packages/analytics/ios/RNFBAnalytics/RNFBAnalyticsModule.mm
+++ b/packages/analytics/ios/RNFBAnalytics/RNFBAnalyticsModule.mm
@@ -171,26 +171,28 @@ RCT_EXPORT_MODULE(NativeRNFBTurboAnalytics)
   });
 
   [FIRAnalytics sessionIDWithCompletion:^(int64_t sessionID, NSError *_Nullable error) {
-    if (completed) {
-      return;
-    }
-    completed = YES;
+    dispatch_async(dispatch_get_main_queue(), ^{
+      if (completed) {
+        return;
+      }
+      completed = YES;
 
-    // Occasionally sessionID is 0 despite nil error, reject as if it were an error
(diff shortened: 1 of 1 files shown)


mikehardy (approved):
Hey @OskarEichler 👋

Cheers

PR #9237, opened 2026-08-29
fix(analytics, web): serialize queue processing

OskarEichler:
### Description

Serialize web Analytics queue drains and client-ID initialization.

The queue currently uses a 250 ms interval without tracking an in-flight drain. When client-ID storage or an Analytics request takes longer than that interval, later batches are removed and sent concurrently. This can reorder events, and the initial concurrent batches can independently read/create different client IDs.

This change:

- permits only one queue drain at a time, preserving batch order;
- shares one in-flight client-ID load/create promise across callers;
- uses the cached client ID after initialization;
- derives queue-running state from the timer itself and clears its reference on stop;
- avoids creating a permanent idle interval when collection is enabled with no queued events.

### Breaking changes

None. Public APIs, types, batching size, 250 ms cadence, payloads, and fire-and-forget behavior are unchanged. Event batches are now sent in queue order with one stable client ID.

### Related issues

No matching open issue found.

### Release Summary

Fixed overlapping web Analytics queue processing and concurrent client-ID initialization.

### Checklist

  - [x] Yes
  - [ ] `Android`
  - [ ] `iOS`
  - [x] `Other` (web)
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

- Exact temporary fake-timer controls: baseline starts two drains while the first remains pending, performs two concurrent client-ID reads, and retains one idle interval; the fixed source reports one drain, one client-ID read, and zero idle timers. All three controls pass after the fix and were removed before commit.
- Analytics compile: passing.

### DIFF
--- a/packages/analytics/lib/web/api.ts
+++ b/packages/analytics/lib/web/api.ts
@@ -58,12 +58,13 @@ class AnalyticsApi implements IAnalyticsApi {
   private userProperties: AnalyticsUserProperties;
   private consent: AnalyticsConsent;
   private analyticsCollectionEnabled: boolean;
-  private started: boolean;
+  private processingQueue: boolean;
   private installationId: string | null;
   private debug: boolean;
   private currentScreen: string | null;
   private sessionId?: number;
   private cid?: string | null;
+  private cidPromise?: Promise<string>;
 
   constructor(appName: string, measurementId: string) {
     this.appName = appName;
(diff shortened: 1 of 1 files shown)


mikehardy (approved):
Hey @OskarEichler 👋

Cheers

PR #9238, opened 2026-08-29
fix(analytics, web): preserve session lifetime

OskarEichler:
### Description

Preserve a web Analytics session across queue runs until the documented 30-minute inactivity timeout.

The current implementation creates `sessionId` whenever queue processing starts and clears the timer whenever the queue drains. As a result, two events more than 250 ms apart receive different session IDs even when the user remains active, inflating session counts and breaking session attribution.

The bridge now derives the session from event activity:

- consecutive events retain the current session ID;
- 30 minutes of inactivity starts a new session, matching Google Analytics' default;
- each queued event captures its session ID when logged, so delayed network processing cannot relabel an older event.

Google Analytics session behavior: https://support.google.com/analytics/answer/12798876

Firebase Analytics default timeout: https://firebase.google.com/docs/reference/android/com/google/firebase/analytics/FirebaseAnalytics#setSessionTimeoutDuration(long)

### Breaking changes

None. Public APIs, types, payload format, and queue timing are unchanged.

Observable reporting correction: events logged within one active 30-minute window now share a session instead of creating a new session whenever the 250 ms queue drains. Session totals and attribution may therefore become more accurate and lower than the previously inflated web values.

### Related issues

No matching open issue found.

### Release Summary

Fixed web Analytics session IDs resetting whenever the event queue briefly drained.

### Checklist

  - [x] Yes
  - [ ] `Android`
  - [ ] `iOS`
  - [x] `Other` (web)
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

- Exact temporary fake-clock/fetch control: baseline emits different `sid` values for events one second apart; the fixed source keeps them equal and rotates `sid` after exactly 30 minutes of inactivity. The control was removed before commit.
- Analytics compile: passing.

### DIFF
--- a/packages/analytics/lib/web/api.ts
+++ b/packages/analytics/lib/web/api.ts
@@ -45,7 +45,10 @@ function generateGAClientId(): string {
 interface AnalyticsEvent {
   name: string;
   params: AnalyticsEventParameters;
+  sessionId: number;
 }
+
+const DEFAULT_SESSION_TIMEOUT_MS = 1_800_000;
 
 class AnalyticsApi implements IAnalyticsApi {
   public readonly appName: string;
@@ -63,6 +66,7 @@ class AnalyticsApi implements IAnalyticsApi {
   private debug: boolean;
   private currentScreen: string | null;
   private sessionId?: number;
+  private lastEventTime?: number;
   private cid?: string | null;
   private cidPromise?: Promise<string>;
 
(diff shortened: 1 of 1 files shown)


mikehardy (approved):
Hey @OskarEichler 👋

Cheers

PR #9239, opened 2026-08-29
fix(analytics): enforce documented input limits

OskarEichler:
### Description

Enforce Firebase Analytics' documented event-name and session-timeout constraints at the JavaScript boundary.

The existing event-name check allows names beginning with digits/underscores and names using the reserved `firebase_`, `google_`, and `ga_` prefixes. Firebase drops those events. The existing session-timeout check also treats `NaN` and infinities as valid numbers, allowing them to become invalid or misleading native durations.

This change:

- uses one exact event-name pattern requiring an alphabetic first character, 1–40 total alphanumeric/underscore characters;
- rejects all three documented reserved prefixes in addition to the existing reserved-name list;
- rejects non-finite session timeout values before they reach native code;
- removes the unnecessary mutable cast used to check the readonly reserved-name list.

Firebase event-name contract: https://firebase.google.com/docs/reference/android/com/google/firebase/analytics/FirebaseAnalytics.Event

### Breaking changes

None for valid inputs. Public APIs and types are unchanged.

Observable validation correction: previously accepted invalid event names and non-finite timeout values now throw synchronously instead of being dropped by Firebase or converted into invalid native durations. Existing errors for non-number and negative timeout values are unchanged.

### Related issues

No matching open issue found.

### Release Summary

Validated Analytics event-name prefixes/first characters and finite session timeout durations.

### Checklist

  - [x] Yes
  - [x] `Android`
  - [x] `iOS`
  - [x] `Other`
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

- Exact temporary public-API controls: baseline accepts names beginning with `1`/`_`, all three reserved prefixes, `NaN`, and both infinities; the fixed source rejects all eight cases with the intended validation errors. The controls were removed before commit.
- Analytics compile: passing.

### DIFF
--- a/packages/analytics/lib/index.ts
+++ b/packages/analytics/lib/index.ts
@@ -18,10 +18,10 @@
 import type { FirebaseApp } from '@react-native-firebase/app';
 import { Platform } from 'react-native';
 import {
-  isAlphaNumericUnderscore,
   isE164PhoneNumber,
   isIOS,
   isBoolean,
+  isFinite,
   isNull,
   isNumber,
   isObject,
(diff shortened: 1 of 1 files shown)


mikehardy (approved):
Hey @OskarEichler 👋

Cheers

PR #9241, opened 2026-08-29
fix(messaging, ios): preserve notification settings handler

OskarEichler:
## Fix

When an application UNUserNotificationCenter delegate implements openSettingsForNotification, React Native Firebase now forwards the callback to that delegate and still emits its own messaging_settings_for_notification_opened event. Previously, forwarding used an either/or branch that silently disabled setOpenSettingsForNotificationsHandler and quit-state tracking.

## Breaking changes

None. This restores the documented React Native Firebase handler while preserving the application delegate callback.

## Verification

- Messaging ESLint passed
- Full iOS debug test-app build passed after a fresh pod install
- Full macOS test-app build passed
- git diff --check passed


mikehardy (approved):
Hey @OskarEichler 👋

Cheers

PR #9242, opened 2026-08-29
fix(messaging, android): preserve stored message integrity

OskarEichler:
## Fixes and observable correction

- Restored notification maps now preserve their serialized `to` recipient instead of replacing it with the message ID.
- Clearing a notification now removes only the exact stored ID instead of every matching substring.

The deprecated API still receives the historical synthetic `to` fallback when a stored message has no recipient, so it can continue reconstructing a Firebase RemoteMessage.

## Breaking changes

No API or type changes. Apps that read `RemoteMessage.to` from an Android initial/opened notification will now receive the original Firebase recipient rather than the incorrect message ID. This is an observable correctness change for code that depended on the previous erroneous value.

## Why

The comma-delimited ID list previously used unrestricted String.replace. For IDs `123` and `23`, clearing `23` transformed `123,23,` into `1`, corrupting retention bookkeeping and leaving the `123` payload orphaned.

## Verification

- Exact temporary Robolectric control: baseline produced `1`; fixed source preserves `123,`
- All existing Android native unit checks passed
- Messaging Android assemble and lint passed with zero errors
- Full Android debug app/test assembly and lint passed: 1,749 tasks
- Google Java Format and git diff --check passed

Temporary test scaffolding was removed before commit.


mikehardy (approved):
Hey @OskarEichler 👋

Cheers

PR #9243, opened 2026-08-29
fix(messaging, android): preserve stored JSON values

OskarEichler:
## Fixes and observable correction

Android Messaging persistence now preserves numeric precision and restores JSON booleans in maps and arrays.

React Native exposes JS numbers as doubles. The previous serializer called getInt first, which deliberately truncates doubles and never throws on current React Native. Consequently a notification sentTime such as 1750000000000 was persisted as 2147483647. The reverse converter also narrowed parsed integral values to 32-bit and silently omitted booleans.

## Breaking changes

No API or type changes. Restored Android initial/opened notifications now expose the correct millisecond sentTime and any future/persisted boolean or fractional fields instead of truncated or missing values. Code that relied on the previous incorrect sentTime may observe a changed value.

## Verification

- Exact temporary map/array controls reproduced baseline truncation for 1.5 and 1750000000000
- Fixed controls preserve fractional and wide numbers in maps and arrays and restore nested true/false values
- Messaging Android assemble and lint passed with zero errors
- Full Android debug app/test assembly and lint passed: 1,749 tasks
- Google Java Format and git diff --check passed

Temporary test dependencies and code were removed before commit.

### DIFF
--- a/packages/messaging/android/src/main/java/io/invertase/firebase/messaging/JsonConvert.java
+++ b/packages/messaging/android/src/main/java/io/invertase/firebase/messaging/JsonConvert.java
@@ -27,11 +27,7 @@ public abstract class JsonConvert {
           jsonObject.put(key, readableMap.getBoolean(key));
           break;
         case Number:
-          try {
-            jsonObject.put(key, readableMap.getInt(key));
-          } catch (Exception e) {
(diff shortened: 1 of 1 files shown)


mikehardy (approved):
Hey @OskarEichler 👋

Cheers

PR #9244, opened 2026-08-29
fix(messaging): validate APNs token input

OskarEichler:
## Observable behavior / breaking changes

- Invalid Apple APNs token strings now throw synchronously at the JavaScript boundary and reject at the native boundary instead of being silently mangled. An empty token previously reached an unsigned length underflow and could raise an Objective-C range exception.
- Android remains a documented no-op and continues accepting any string without crossing the native APNs path.

## What changed

- Require Apple APNs tokens to be non-empty, even-length hexadecimal strings without hard-coding a token byte length.
- Defend the Objective-C bridge independently for old bundles and direct native calls.
- Decode into one pre-sized data buffer instead of repeatedly appending bytes.
- Add focused Apple E2E coverage for empty, odd-length, and non-hex values.

## Verification

- Messaging ESLint and 32 Jest tests
- Messaging package compile, TypeScript, and dependency-cruiser
- Google clang-format and git diff --check
- Fresh iOS pod install and full iOS simulator build
- Focused iOS App + Messaging E2E: 70 passing, 23 platform-pending
- Full macOS build
- Full Android debug app/test assembly and lint: 1,749 tasks, zero Messaging lint errors

Environment-only CocoaPods checksum drift was restored before commit. No application dependency or lockfile changed.

### DIFF
--- a/packages/messaging/e2e/messaging.e2e.js
+++ b/packages/messaging/e2e/messaging.e2e.js
@@ -252,6 +252,20 @@ describe('messaging()', function () {
         }
       });
 
+      it('requires a non-empty, even-length hexadecimal token on Apple platforms', function () {
+        const { getMessaging, setAPNSToken } = messagingModular;
+        if (!Platform.ios && !Platform.macos) {
+          this.skip();
+        }
+
+        for (const token of ['', '0', 'xyz', '001z']) {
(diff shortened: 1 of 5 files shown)


mikehardy (approved):
Hey @OskarEichler 👋

Cheers

PR #9246, opened 2026-08-29
fix(messaging): nest iOS string sounds correctly

### LINKED ISSUE #6375: [🔥 ] [🐛] IOS sound - incorrect TS / incorrect data structure in messaging onMessage() fn
## Issue

Hello,

I am using react-native-firebase/messaging for listening to push notifications on IOS and when it is received on foreground I am showing push notification via notifee.

We are using custom sounds for push notification so I need to pass sound name from message received via onMessage() to nofifee... my code looks like:

```ts
const onReceiveNotificationIOS = async (message: FirebaseMessagingTypes.RemoteMessage) => {
  const notification = message.notification

  const body = notification?.body
  const title = notification?.title
  const sound = (notification as any)?.sound || 'default'

  void notifee.displayNotification({
    id: v4(),
    body: body,
    title: title,
    ios: {
      sound: sound,
    },
  })
}
```

The issue is with this line ` const sound = (notification as any)?.sound || 'default'` where I need to type cast notification to `any` because [notification type does not contain sound field](https://github.com/invertase/react-native-firebase/blob/main/packages/messaging/lib/index.d.ts#L165) however the data are there.

I believe this is bug. We are sending sound as string with file name from our BE so it is filled into notification object directly as sound field as a string.

If we would send object from our BE instead of string then it would be appended to `notification.ios.sound` However in this instance types are also wrong because [it says](https://github.com/invertase/react-native-firebase/blob/main/packages/messaging/lib/index.d.ts#L222) it can be `string | NotificationIOSCriticalSound` which I believe is also wrong because it cannot be string.

It is pretty obvious from code that it is bug: https://github.com/invertase/react-native-firebase/blob/main/packages/messaging/ios/RNFBMessaging/RNFBMessagingSerializer.m#L182 
As you can see, good intention is there as the comment above that line is pointing to right location but data are written to wrong location....

So fix is actually easy, I could create PR for that but it would be breaking change - if we fix location where we store the data... Other solution is just add sound field to type as optional string and remove string as possible type from `ios` subfield... And once the new major release will be ready we can introduce breaking change and fix the location?

What do you think?
---

### Javascript

<p>

#### `package.json`:

```json
"@react-native-firebase/messaging": "14.11.0",
```

```json
# N/A
```

</details>

### iOS

<p>

#### `ios/Podfile`:

- [ ] I'm not using Pods

```ruby
# N/A
```

#### `AppDelegate.m`:

```objc
// N/A
```

</p>
</details>

---

### Android

<p>

#### `android/build.gradle`:

```groovy
// N/A
```

```groovy
// N/A
```

```groovy
// N/A
```

#### `MainApplication.java`:

```java
// N/A
```

#### `AndroidManifest.xml`:

```xml
<!-- N/A -->
```

</p>
</details>

---

## Environment

<p>

```
 OUTPUT GOES HERE
```

  - [ ] iOS
  - [ ] Android
  - [ ] Both
  - `14.11.0`
  - `messaging`
  - `Y` & `4.6.3`

</p>
</details>

---

OskarEichler:
## Fix

String-valued APNs sounds are now serialized at the documented `message.notification.ios.sound` path. Previously, only string sounds were placed at the undocumented `message.notification.sound` path, while critical-sound dictionaries already used the correct iOS nesting.

A direct iOS native regression test now exercises the serializer and verifies the complete notification shape.

## Breaking behavior correction

Consumers that worked around the bug by reading `message.notification.sound` for a string-valued APNs sound must switch to `message.notification.ios.sound`. This aligns runtime behavior with the public `RemoteMessage` type and documentation.

Fixes #6375.

## Validation

- Reproduced the regression against the previous implementation: the new assertion received `{ sound: "tone.aiff" }` instead of `{ ios: { sound: "tone.aiff" } }`
- iOS Messaging/App E2E: 70 passing, 23 pending
- Android Messaging/App E2E: 86 passing, 7 pending
- iOS native coverage confirms the changed serializer line executed once
- iOS test application build succeeded
- Android test application build and lint succeeded (0 errors)
- `yarn lint:ios:check`
- `yarn lint:js`
- Messaging Jest suite: 32 passing
- `git diff --check`


mikehardy:
marking this as do not merge temporarily until next breaking release

PR #9247, opened 2026-08-29
fix(messaging, ios): normalize sent time

OskarEichler:
## Fix

iOS now exposes `RemoteMessage.sentTime` as a numeric epoch-millisecond value, matching the public contract and Android behavior. FCM supplies `google.c.a.ts` as epoch seconds (commonly a string); the serializer now validates the numeric representation, converts seconds to milliseconds, and rejects non-positive or overflowing values.

A direct iOS native regression test exercises the serializer with the same timestamp shape used by the Firebase iOS SDK tests.

## Breaking behavior correction

On iOS, `sentTime` changes from a string containing epoch seconds (for example, `"1522880044"`) to a number containing epoch milliseconds (`1522880044000`). Consumers that parsed or compared the accidental string/seconds value must use the documented number/milliseconds contract. Android behavior is unchanged.

## Validation

- Reproduced on the previous implementation: the assertion received string `"1522880044"` instead of numeric `1522880044000`
- iOS App + Messaging E2E: 70 passing, 23 pending
- Android App + Messaging E2E: 86 passing, 7 pending
- Processed iOS LLVM coverage confirms every changed production line executed
- iOS test application build succeeded
- Android test application build and lint succeeded (0 errors)
- Messaging Jest suite: 32 passing
- `yarn lint:ios:check`
- `yarn lint:js`
- `git diff --check`

### DIFF
--- a/packages/messaging/e2e/messaging.e2e.js
+++ b/packages/messaging/e2e/messaging.e2e.js
@@ -72,6 +72,19 @@ describe('messaging()', function () {
     message.notification.should.eql({ ios: { sound: 'tone.aiff' } });
   });
 
+  it('serializes iOS sent times as epoch milliseconds', async function () {
+    if (!Platform.ios) {
+      this.skip();
+    }
+
+    const { getRNFBTesting } = require('../../app/e2e/helpers');
+    const message = await getRNFBTesting().serializeMessagingUserInfo({
+      'google.c.a.ts': '1522880044',
+    });
+
+    message.sentTime.should.equal(1522880044000);
+  });
+
   before(async function () {
     // our device registration tests require permissions. Set them up
     const { getMessaging, requestPermission } = messagingModular;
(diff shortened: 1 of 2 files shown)


OskarEichler:
I re-reviewed the serializer and found that `longLongValue` accepted malformed values such as `1.5` and `1522880044invalid` by parsing their numeric prefix. I pushed a follow-up that now accepts only non-empty decimal integer strings (or numeric values converted to that representation) before applying the existing positivity and overflow checks.

I also expanded the focused e2e coverage for string and numeric timestamps plus empty, negative, fractional, and trailing-garbage inputs.

Fresh validation on commit `03d519be`:

- `yarn lint:ios:check` — passed
- `yarn lint:js` — passed
- `yarn tests:ios:build` — passed
- iOS area-focused e2e (`app` + `messaging`, `RNFBDebug: true`) — 71 passing, 23 pending
- iOS native coverage processing — passed; every touched serializer line executed
- `yarn tests:android:build` — passed
- Android area-focused e2e (`app` + `messaging`, `RNFBDebug: true`) — 86 passing, 8 pending
- `git diff --check` — passed

The local harness/device overrides were removed before the commit.

mikehardy:
Nice. Your update made it before this one went into my local review / validate queue

mikehardy (approved):
looks like a real issue, looks like a good fix - going to queue this one for next breaking change release

PR #9248, opened 2026-08-29
fix(messaging, android): support disabled notification storage

OskarEichler:
## What this fixes

- Treats `messaging_max_stored_notifications <= 0` as disabled storage.
- Clears notifications retained before storage was disabled.
- Prevents the `IndexOutOfBoundsException` currently raised when the configured limit is zero or negative.
- Preserves the existing any-number configuration contract without silently clamping values.

## Behavior

Positive values retain their existing meaning. Zero and negative values now disable notification persistence instead of crashing while pruning an empty list. The schema documents this observable correction.

## Breaking changes

None. This defines behavior for a configuration that currently crashes; positive limits are unchanged.

## Verification

- Baseline regression: Android e2e fails with `IndexOutOfBoundsException: Failed to disable notification storage`.
- Android native build: 1,749 tasks, successful.
- Android Messaging e2e: 87 passing, 6 platform-pending.
- Android unit suite and merged JaCoCo report: successful; every changed reachable store line covered.
- iOS native build: successful.
- iOS Messaging e2e: 69 passing, 24 platform-pending.
- `yarn lint:js`
- `yarn lint:android`
- `yarn tsc:compile`
- `yarn tsc:compile:consumer`
- `yarn reference:api`
- `yarn tests:jest` (103 suites, 1,479 tests)
- `yarn compare:types`
- `git diff --check`

### DIFF
--- a/okf-bundle/packages/messaging/index.md
+++ b/okf-bundle/packages/messaging/index.md
@@ -5,7 +5,7 @@ Knowledge for FCM messaging native behavior, primarily the iOS `UNUserNotificati
 ## Documents
 
 * [iOS UNUserNotificationCenter delegate forwarding](ios-notification-delegate-forwarding.md) — `completionHandler` exactly-once contract, delegate chaining design, regression history (#8754 → #8786 → #9049 / #9050)
-* Native integration probes (`completesNonFCMRemoteNotification`, `messagingPreservesExistingDelegate`) — [running e2e § test-app native modules](../../testing/running-e2e.md#test-app-native-modules); product-line hits in `coverage/ios-native/lcov.info` — [coverage design § two test native modules](../../testing/coverage-design.md#test-native-modules)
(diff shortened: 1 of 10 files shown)


tests/android/app/src/main/java/com/invertase/testing/RNFBTestingCoverageModule.kt
```diff
@@ -42,7 +45,62 @@ class RNFBTestingCoverageModule(reactContext: ReactApplicationContext) :
     }
   }
 
+  @ReactMethod
```
mikehardy: 🤔 

I'm extracting this coverage module to a fully separate project so this isn't the right home for this

Let's let this PR sit for just a moment while that resolves (the work is in progress right now) and while more generic Android unit test infrastructure arrives (that is also in process right now)

mikehardy (changes requested):
No need to take action now because it will naturally be ordered after two other changes that are in-flight in the repo, but we need to put the unit test in a dedicated unit test object for the app module, vs slip-streaming in on the coverage module

mikehardy:
Will use same style of e2e-driven test-app-hosted native probe as we are attempting to pioneer in #9255 (once that has approval and lands)

mikehardy (approved):
re-homed the native probe to our new test-hosted e2e native probe module for special tough-to-probe cases like this

the original fix was always fine but this makes it maintainable as well, ready to go

PR #9249, opened 2026-08-29
fix(messaging, ios): preserve existing FIRMessaging delegate

OskarEichler:
## What this fixes

- Preserves an existing `FIRMessagingDelegate` when React Native Firebase installs its token-refresh interceptor.
- Forwards registration-token callbacks to that original delegate after emitting `messaging_token_refresh`.
- Retains the existing AppDelegate fallback and avoids invoking it twice when it is also the original Firebase delegate.

Firebase documents assigning `FIRMessaging.delegate` after `FIRApp configure` as the supported registration-token callback path. React Native Firebase currently replaces a separately assigned delegate, so it silently stops receiving token refreshes.

https://firebase.google.com/docs/cloud-messaging/ios/get-started#set_the_messaging_delegate

## Breaking changes

None. This restores callbacks to the application delegate object that Firebase was already configured to notify. React Native Firebase events and the existing AppDelegate fallback are unchanged.

## Verification

- Exact native baseline control: failed (`false`) when a separate Firebase delegate was replaced.
- Fixed iOS App + Messaging E2E: 70 passing, 23 platform-pending.
- Fixed Android App + Messaging E2E: 86 passing, 7 platform-pending.
- iOS simulator, Android debug/test/lint (1,749 tasks), and macOS builds: successful.
- iOS LLVM coverage: every changed reachable production line executed.
- Android native unit and coverage processing: successful.
- `yarn lint:ios:check`
- `yarn lint:js`
- `yarn tsc:compile`
- `yarn tsc:compile:consumer`
- `yarn reference:api`
- `yarn tests:jest` (103 suites, 1,479 tests)
- `yarn compare:types`
- `git diff --check`

### DIFF
--- a/packages/messaging/e2e/messaging.e2e.js
+++ b/packages/messaging/e2e/messaging.e2e.js
@@ -288,6 +288,20 @@ describe('messaging()', function () {
       });
     });
 
+    describe('native messaging delegate', function () {
+      it('preserves an existing iOS Firebase Messaging delegate', async function () {
+        if (!Platform.ios) {
+          this.skip();
+        }
+
+        const { NativeModules } = require('react-native');
+        should.equal(
+          await NativeModules.RNFBTestingCoverage.messagingPreservesExistingDelegate(),
+          true,
+        );
+      });
+    });
+
     describe('deleteToken()', function () {
       it('generate a new token after deleting', async function () {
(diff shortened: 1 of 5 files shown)


mikehardy (approved):
Hey @OskarEichler 👋

Cheers

PR #9251, opened 2026-08-29
fix(messaging): update cached settings after native success

OskarEichler:
## What this fixes

Rejected native setting updates currently mutate the JavaScript cache first. After a rejection, these public getters can therefore report values that were never applied:

- `isAutoInitEnabled`
- `isDeliveryMetricsExportToBigQueryEnabled`
- `isNotificationDelegationEnabled`

This PR updates each cache only after the corresponding native promise resolves.

It also preserves the documented iOS notification-delegation behavior: delegation is Android-only, so the iOS getter remains `false` after the no-op setter resolves.

## Breaking changes

None. Successful Android/iOS setting updates are unchanged. Rejected updates now preserve their previous values, and iOS delegation now matches its existing documented `false`-only contract.

## Verification

- Exact baseline unit controls: 3 failures; all three caches incorrectly changed after native rejection.
- Fixed focused Jest: 35 passing.
- Full Jest: 103 suites, 1,482 tests, 31 snapshots passed.
- Focused iOS App + Messaging E2E: 69 passing, 23 pending.
- Focused Android App + Messaging E2E: 86 passing, 6 pending.
- Android unit tests and merged native coverage passed.
- `yarn lerna:prepare`
- `yarn lint:js`
- `yarn lint:deps`
- `yarn tsc:compile`
- `yarn tsc:compile:consumer`
- `yarn reference:api`
- `yarn compare:types`
- `git diff --check`

### DIFF
--- a/packages/messaging/__tests__/messaging.test.ts
+++ b/packages/messaging/__tests__/messaging.test.ts
@@ -28,31 +28,35 @@ import {
   subscribeToTopic,
   unsubscribeFromTopic,
   isDeliveryMetricsExportToBigQueryEnabled,
+  isNotificationDelegationEnabled,
   isSupported,
   experimentalSetDeliveryMetricsExportedToBigQueryEnabled,
+  setNotificationDelegationEnabled,
   AuthorizationStatus,
   NotificationAndroidPriority,
   NotificationAndroidVisibility,
 } from '../lib';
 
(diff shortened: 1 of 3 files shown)


mikehardy (approved):
Hey @OskarEichler 👋

Cheers

PR #9252, opened 2026-08-29
fix(messaging, ios): serialize notification badges as strings

OskarEichler:
## What changed

- Serialize the iOS `aps.badge` value as a string, matching the public `RemoteMessage.notification.ios.badge?: string` contract.
- Add a direct native serializer E2E regression check.

- A numeric APNs badge such as `7` now reaches JavaScript as `"7"` instead of the previously undocumented numeric value `7`.
- There is no public API or type change. Consumers that relied on the accidental numeric runtime value should parse the documented string when they need a number.

## Why

The iOS serializer forwarded the native `NSNumber` directly across the bridge while the public TypeScript API documents a string. Converting at the serializer boundary keeps runtime behavior consistent with the published contract.

## Verification

- Baseline native E2E reproduced the mismatch (`7` versus `"7"`).
- iOS Messaging E2E: 70 passing, 23 pending.
- Android Messaging E2E: 86 passing, 7 pending.
- Changed iOS native lines are covered by the generated LLVM coverage report.
- JS lint, iOS formatting, dependency-cruiser, library and consumer TypeScript compilation, API-reference generation, type-parity comparison, Android unit/coverage, and `git diff --check` all pass.

### DIFF
--- a/packages/messaging/e2e/messaging.e2e.js
+++ b/packages/messaging/e2e/messaging.e2e.js
@@ -102,6 +102,19 @@ describe('messaging()', function () {
     }
   });
 
+  it('serializes iOS notification badges as strings', async function () {
+    if (!Platform.ios) {
+      this.skip();
+    }
+
+    const { getRNFBTesting } = require('../../app/e2e/helpers');
+    const message = await getRNFBTesting().serializeMessagingUserInfo({
+      aps: { badge: 7 },
+    });
+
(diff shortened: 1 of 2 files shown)


mikehardy (approved):
looks good thanks! this is a small break but, it is a break that could be non-silent, queueing this for next breaking change

PR #9254, opened 2026-08-29
fix(messaging): pass sent message IDs to listeners

OskarEichler:
## What changed

- Unwrap the native `{ messageId }` event body before calling `onMessageSent` listeners.
- Add unit and Android native-bridge regressions for the public callback contract.

- `onMessageSent` listeners now receive the documented message-ID string instead of the accidental `{ messageId: string }` object.
- Consumers that worked around the incorrect runtime shape should remove the object unwrap. The public API and TypeScript signature are unchanged.

## Why

Android intentionally emits a map so the React Native event bridge can carry the message ID. Unlike the token-refresh listener, `onMessageSent` forwarded that transport wrapper directly to application code even though its public callback has always been typed and documented as `(messageId: string) => any`.

## Verification

- Exact baseline shared-event regression received `{ messageId: "message-id" }`; the fixed listener receives `"message-id"`.
- Focused Messaging Jest: 33 passing.
- Android native-bridge Messaging E2E: 87 passing, 6 pending.
- Full Jest: 103 suites, 1,480 tests, 31 snapshots.
- Android unit checks, `lerna:prepare`, JS lint, dependency-cruiser, both TypeScript compiles, API-reference generation, type-parity comparison, and `git diff --check` pass.

### DIFF
--- a/packages/messaging/__tests__/messaging.test.ts
+++ b/packages/messaging/__tests__/messaging.test.ts
@@ -38,6 +38,7 @@ import {
 } from '../lib';
 
 import remoteMessageOptions from '../lib/remoteMessageOptions';
+import { SharedEventEmitter } from '@react-native-firebase/app/dist/module/internal';
 
 describe('remoteMessageOptions', function () {
   it('serializes array data values as JSON', function () {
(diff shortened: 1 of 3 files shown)


mikehardy (approved):
Another good fix - but also something that could break consumers, going to queue this for a breaking change release

PR #9255, opened 2026-08-29
fix(messaging, ios): complete non-FCM notifications

OskarEichler:
## What changed

- Invoke the iOS background-fetch completion handler with `UIBackgroundFetchResultNoData` when the AppDelegate donor receives a non-FCM remote notification.
- Add a direct native regression through the iOS test bridge.

## Behavioral impact

- Non-FCM APNs deliveries no longer leave Apple's fetch completion contract unsettled when RNFB supplied the AppDelegate implementation.
- FCM foreground/background handling, emitted events, public APIs, and types are unchanged. There are no breaking changes.

## Why

RNFB's donor method is installed when the application delegate does not implement `application:didReceiveRemoteNotification:fetchCompletionHandler:`. The method completed every FCM path but returned without invoking the callback when `gcm.message_id` was absent, despite then owning the system completion contract.

## Verification

- Exact baseline native regression returned `false` because the callback was not invoked.
- Fixed iOS App + Messaging E2E: 70 passing, 23 pending.
- Android App + Messaging E2E: 86 passing, 7 pending.
- LLVM coverage executes the new production fallback branch.
- iOS build and native formatting pass.
- JS lint, dependency-cruiser, both TypeScript compiles, API-reference generation, type-parity comparison, and `git diff --check` pass.


mikehardy (approved):
Hey @OskarEichler 👋

Cheers

mikehardy:
Holding this one for a small follow-up before merge.

The product fix (complete non-FCM remote notifications on iOS) looks good, but the probe lives in RNFBTestingCoverageModule / the coverage native module. Same pattern as the other messaging PRs we parked: that module is for coverage plumbing, not feature unit probes.

We'll move the unit test to a proper spot (messaging-owned native/unit test surface, not the coverage module) and then bring this back for merge.

mikehardy:
I've done what I want to do here (unless I need to fix some CI fail/nit) - asked for internal review since I've hijacked a small product fix to "grow" a new test capability.

russellwheatley (approved):

PR #9256, opened 2026-08-29
fix(messaging): make FCM options optional

OskarEichler:
## What changed

- Make `RemoteMessage.fcmOptions` optional.
- Update the compile-time `sendMessage` fixture to verify a message without web-only FCM options.

## Compatibility

- This is a type-only loosening with no runtime behavior change and no breaking changes.
- Existing messages that include `fcmOptions` continue to type-check unchanged.

## Why

`fcmOptions` contains web-only features and is absent from normal native incoming messages. The Android `sendMessage` serializer also does not require or consume it. Requiring the property forced TypeScript users to add an empty object that had no runtime effect; Firebase's corresponding message payload field is optional as well.

## Verification

- Exact baseline TypeScript compile failed because the fixture omitted `fcmOptions`.
- `lerna:prepare`, JS lint, dependency-cruiser, library and consumer TypeScript compiles, API-reference generation, type-parity comparison, and `git diff --check` pass.

### DIFF
--- a/packages/messaging/lib/types/messaging.ts
+++ b/packages/messaging/lib/types/messaging.ts
@@ -95,7 +95,7 @@ export interface RemoteMessage {
   /**
    * Options for features provided by the FCM SDK for Web.
    */
-  fcmOptions: FcmOptions;
+  fcmOptions?: FcmOptions;
 
   /**
    * Priority - android-specific, undefined on non-android platforms, default PRIORITY_UNKNOWN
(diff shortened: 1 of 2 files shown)


mikehardy (approved):
Hey @OskarEichler 👋

Thanks for the PR — this looks like a real problem to me. I ran local Jest/unit plus the android/ios/macos e2e suite against it here and it came out clean. Pending CI, then I'll merge.

I looked especially at whether this breaks type consumers or needed API / type parity changes vs firebase-js-sdk. As it is an RN-specific type, no parity issues apply and it won't break consumers.

Cheers

PR #9258, opened 2026-08-29
fix(storage): decode data URLs with custom content types

OskarEichler:
## Compatibility

- No breaking changes.
- Observable correction: data URL uploads with an explicit metadata contentType are now decoded before reaching the native base64-only upload path. Previously the complete data URL and unsupported data_url format were forwarded.

## What changed

- Always replace a validated data URL with its base64 payload and base64 format.
- Continue inferring contentType only when callers do not provide one.
- Add a focused unit regression and native Storage emulator coverage.

## Verification

- Regression test fails on upstream main and passes with this change.
- Full Jest: 104 suites, 1,480 tests, 31 snapshots.
- iOS App + Storage E2E: 141 passing, 7 pending.
- Android App + Storage E2E: 143 passing, 5 pending.
- Android assembleDebug, assembleAndroidTest, and lintDebug.
- lerna:prepare, JS lint, dependency-cruiser, Storage TypeScript, repository/consumer TypeScript, API reference, compare-types, Prettier, and diff checks.

### DIFF
--- a/packages/storage/__tests__/StorageReference.test.ts
+++ b/packages/storage/__tests__/StorageReference.test.ts
@@ -1,0 +1,20 @@
+import { describe, expect, it } from '@jest/globals';
+
+import Reference from '../lib/StorageReference';
+import { StringFormat } from '../lib/StorageStatics';
+import type { StorageInternal } from '../lib/types/internal';
+
+describe('StorageReference string uploads', () => {
+  it('decodes a data URL when metadata already provides a content type', () => {
+    const reference = new Reference({} as StorageInternal, '/file.txt');
+    const metadata = { contentType: 'text/custom' };
+
+    expect(
+      reference._updateString('data:text/plain;base64,aGVsbG8=', StringFormat.DATA_URL, metadata),
+    ).toEqual({
+      _format: StringFormat.BASE64,
+      _metadata: metadata,
+      _string: 'aGVsbG8=',
+    });
+  });
+});
--- a/packages/storage/e2e/StorageTask.e2e.js
+++ b/packages/storage/e2e/StorageTask.e2e.js
@@ -171,6 +171,21 @@ describe('storage() -> StorageTask', function () {
         uploadTaskSnapshot.metadata.should.be.an.Object();
       });
 
+      it('uploads a data_url with an explicit content type', async function () {
+        const { getStorage, ref, uploadString, StringFormat, TaskState } = storageModular;
+
+        const uploadTaskSnapshot = await uploadString(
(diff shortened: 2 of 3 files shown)


mikehardy (changes requested):
Hey @OskarEichler 👋

Thanks for the PR — the real-problem analysis looks right to me, and I like the unit coverage. I ran local Jest plus the android/ios/macos e2e suite here. Android and iOS are green on the new case, but macOS failed, so I'm requesting changes until we understand that path.

macOS e2e failure from our run:

```
storage() -> StorageTask
  StorageTask modular
    putString()
      uploads a data_url with an explicit content type:
AssertionError: expected 8 to equal 5
  at packages/storage/e2e/StorageTask.e2e.js:185:51
```

The test uploads `data:text/plain;base64,aGVsbG8=` with `{ contentType: 'text/custom' }` and expects `metadata.size === 5` (decoded `hello`). macOS reported **8**, which looks like the undecoded base64 payload length rather than the decoded bytes.

We have a lot of trouble on the firebase-js-sdk / HermesVM target when it comes to base64 and Web APIs in general — `atob`, `btoa`, `Blob`, `FileReader`, data URLs, and friends have all been edge-case magnets for us. I think this may be a similar edge case on macOS rather than a simple one-line fix being wrong on mobile.

Could you dig into why macOS still uploads the undecoded payload when an explicit `contentType` is set, and whether the fix needs a Hermes-specific path (or a different assertion strategy on macOS)? Happy to pair on it if useful.

Cheers

OskarEichler:
I traced the macOS path. The PR’s `_updateString` result is correct (`aGVsbG8=`, format `base64`), and the macOS/web bridge then decodes it in `RNFBStorageModule.ts` before passing a `Uint8Array` to firebase-js-sdk `uploadBytesResumable`. The remaining 8-byte upload is the existing Hermes `Blob` / firebase-js-sdk binary-upload interaction already isolated in #9006 (`@mikehardy/storage-upload-bytes`), not the data-URL/content-type branch. That branch replaces this path with `uploadBytes` plus the Hermes Blob fallback and should make the same assertion report 5. I do not think changing the macOS assertion to 8 is correct because it would codify the corrupted upload; the clean path is to keep this regression and re-run/rebase it after #9006 lands (or stack #9006 underneath it for verification).

mikehardy:
Tracking note: this PR is no longer on the Oskar review-stack umbrella. Follow-up (Other/Hermes Blob + `#9006` dependency, keep asserting size `5`) lives on Linear **CPRN-421**.

russellwheatley:
Blocking this until #9006 is merged.

PR #9259, opened 2026-08-29
fix(app): handle empty Android Firebase JSON

OskarEichler:
## Compatibility

- No breaking changes.
- Observable correction: jsonGetAll now returns an empty map when no firebase.json React Native settings exist instead of crashing on Android.

## What changed

- Iterate JSONObject keys directly, which handles an empty object and avoids allocating a JSONArray of key names.
- Return the already-created empty bridge map if parsing was unavailable.
- Add a focused Robolectric regression for the default empty-object configuration.

## Verification

- Exact regression fails on upstream main with NullPointerException and passes with this change.
- Full App Android unit suite passes.
- App Android debug AAR assembly and lint pass.
- Repository Android formatting and whitespace checks pass.

### DIFF
--- a/packages/app/android/src/test/java/io/invertase/firebase/common/ReactNativeFirebaseJSONTest.java
+++ b/packages/app/android/src/test/java/io/invertase/firebase/common/ReactNativeFirebaseJSONTest.java
@@ -1,0 +1,108 @@
+package io.invertase.firebase.common;
+
+/*
+ * Copyright (c) 2016-present Invertase Limited & Contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this library except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


packages/app/android/src/test/java/io/invertase/firebase/common/ReactNativeFirebaseJSONTest.java
```diff
@@ -0,0 +1,54 @@
...
+import java.lang.reflect.Field;
+import org.json.JSONObject;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.MockedStatic;
+import org.robolectric.RobolectricTestRunner;
+import org.robolectric.annotation.Config;
+
+@RunWith(RobolectricTestRunner.class)
+@Config(sdk = 34)
```
mikehardy: I've got a PR in flight that drops this - the sdk = 34 robolectric pin is idiomatic in the codebase but is unnecessary and will result in unnecessary robolectric implementation jars in the future

mikehardy (changes requested):
this looks good but needs to drop the robolectric pin using some work I've got in-flight elsewhere - queueing this for a quick edit after that lands, to rebase the other work in and remove the pin

mikehardy (approved):
test verifies correct behavior not just crash now, plain JVM, robolectric not needed in this case

PR #9260, opened 2026-08-29
fix(app): preserve web Promise contracts

OskarEichler:
## What this fixes

The web App shim returned plain values from methods whose native specs and public contract are asynchronous. Code using `.then()` or `.catch()` therefore failed on web even though `await` appeared to work.

This makes the web implementations return real Promises for metadata, JSON config, preferences, listener diagnostics, and event pings. `eventsPing` now also resolves with the event body, matching the native implementations. Existing `await` behavior is unchanged.

## Verification

- Added a focused regression covering every corrected Promise contract
- Focused App Jest: 7/7 passed
- Full Jest: 104 suites passed
- `yarn lerna:prepare`
- `yarn tsc:compile`
- `yarn tsc:compile:consumer`
- `yarn lint:deps`
- `yarn reference:api`
- `yarn compare:types`
- Focused ESLint and `git diff --check`

## Compatibility

No public API or type changes. This is a web-only runtime compatibility fix; callers already using `await` retain the same behavior, while Promise chaining now works consistently with native.

### DIFF
--- a/packages/app/__tests__/webAppModule.test.ts
+++ b/packages/app/__tests__/webAppModule.test.ts
@@ -16,6 +16,30 @@ jest.mock('../lib/internal/web/firebaseApp', () => ({
 
 const appModule = require('../lib/internal/web/RNFBAppModule.ts').default;
 
+describe('RNFBAppModule Promise contracts', () => {
+  it('matches native async configuration and event methods', async () => {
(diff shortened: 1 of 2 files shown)


mikehardy (approved):
Hey @OskarEichler 👋

Cheers

PR #9261, opened 2026-08-30
fix(app): retain global logger configuration

OskarEichler:
## What this fixes

`setLogLevel()` and `onLog()` only updated Logger instances that already existed. A Firebase package loaded after either call therefore started with the default INFO level and no user callback, silently ignoring the application-wide configuration. This is observable with lazy or dynamic loading of the AI package.

The App logger now retains the current global level and user handler so future Logger instances inherit both. Existing instances are still updated immediately. Level normalization is performed once per global update.

## Verification

- Added exact late-construction regressions for both the global level and callback threshold
- Baseline: both regressions failed (INFO instead of ERROR; callback received zero calls)
- Fixed focused App Jest: 12/12 passed
- Full Jest: 103 suites / 1,481 tests / 31 snapshots passed
- `yarn lerna:prepare`
- `yarn lint:deps` and focused ESLint
- `git diff --check`

## Compatibility

No public API/type or breaking change. This restores the documented application-wide behavior for Logger instances created after configuration; already-created Logger behavior is unchanged.

### DIFF
--- a/packages/app/__tests__/app.test.ts
+++ b/packages/app/__tests__/app.test.ts
@@ -8,7 +8,7 @@ import {
   getApp,
   setLogLevel,
 } from '../lib';
-import { Logger } from '../lib/internal/logger';
+import { Logger, LogLevel } from '../lib/internal/logger';
 import { NativeFirebaseError } from '../lib/internal';
 
 describe('App', function () {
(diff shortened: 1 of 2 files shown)


mikehardy (approved):
Hey @OskarEichler 👋

Cheers

PR #9262, opened 2026-08-30
fix(app, ios): clear stale custom auth domains

OskarEichler:
## What this fixes

The iOS App module recorded a secondary app's custom `authDomain` before Firebase accepted the app configuration and never removed it when the app was deleted. A later app with the same name could therefore receive an auth domain from a failed or deleted predecessor.

This update:

- records the domain only after Firebase configuration succeeds
- removes it when initialization omits the domain or deletion succeeds
- synchronizes access to the shared domain map
- reads the map once when serializing app options

## Verification

- Added direct native lifecycle regressions for delete/recreate and failed-initialize/recover
- Baseline: both returned `react-native-firebase-testing.firebaseapp.com` instead of `undefined`
- Fixed iOS App E2E: 41/41 passed
- Clean iOS simulator build passed
- Native LLVM coverage collected for the changed lifecycle/serialization paths
- Full Jest: 103 suites / 1,479 tests / 31 snapshots passed
- `yarn lerna:prepare`
- `yarn lint:deps`, focused ESLint, and iOS formatting
- `git diff --check`

Temporary harness configuration and pod-install-only lockfile checksum drift were removed before commit.

## Compatibility

No public API/type or breaking change. Correctly initialized apps retain their configured auth domain; only stale domains from failed or deleted app instances are removed.

### DIFF
--- a/packages/app/e2e/app.e2e.js
+++ b/packages/app/e2e/app.e2e.js
@@ -15,6 +15,8 @@
  *
  */
 
+const APP_MODULE = NativeModules.NativeRNFBTurboApp;
+
 describe('modular', function () {
   describe('firebase v9 modular', function () {
     it('it should allow read the default app from native', function () {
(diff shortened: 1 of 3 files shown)


mikehardy (commented):
queuing this for a look after #9093 lands - the auth domains here may be a case for the HandleMap functionality landing there

mikehardy (approved):
ok to merge as-is, doesn't intersect enough with #9093 to warrant HandleMap use

PR #9263, opened 2026-08-30
fix(app): make debug native proxies behavior-safe

OskarEichler:
## What this fixes

RNFB debug logging could change otherwise valid native-module calls:

- circular or BigInt arguments/results made `JSON.stringify` throw
- a replaced/throwing `console.debug` prevented calls from running
- any object with a `then` property was treated as a Promise, even when `then` was not callable
- web proxy methods were called without their module receiver, breaking methods that use `this`
- web returned a fresh proxy on every resolution

The two platform implementations now share one behavior-safe debug proxy. It serializes circular/BigInt/Error values defensively, contains console failures, preserves synchronous errors, recognizes only callable Promise-like results, binds methods to their module, and memoizes/invalidate web proxies consistently.

## Verification

- Added focused regressions for circular + BigInt values, non-callable `then`, console failures, web receiver binding, and web proxy identity
- Baseline failed on circular serialization and unbound web methods
- Fixed focused App Jest: 5/5 passed
- Full Jest: 103 suites / 1,483 tests / 31 snapshots passed
- App E2E with `RNFBDebug` enabled: iOS 39/39 and Android 39/39 passed
- `yarn lerna:prepare`
- `yarn lint:deps` and focused ESLint
- `git diff --check`

Temporary debug harness configuration was removed before commit.

## Compatibility

No public API/type or breaking change. Non-debug behavior is untouched. Debug mode now observes calls without changing their receiver, return value, or failure behavior.

### DIFF
--- a/packages/app/__tests__/nativeModuleResolver.test.ts
+++ b/packages/app/__tests__/nativeModuleResolver.test.ts
@@ -1,6 +1,10 @@
 import { describe, expect, it, jest } from '@jest/globals';
 import { TurboModuleRegistry } from 'react-native';
 import { getReactNativeModule } from '../lib/internal/nativeModuleAndroidIos';
+import {
+  getReactNativeModule as getWebReactNativeModule,
+  setReactNativeModule as setWebReactNativeModule,
+} from '../lib/internal/nativeModuleWeb';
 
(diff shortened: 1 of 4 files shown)


mikehardy (approved):
Hey @OskarEichler 👋

Cheers

PR #9264, opened 2026-08-30
fix(app, expo): add missing Firebase imports

OskarEichler:
## What this fixes

- Ensures the Expo config plugin imports Firebase in custom Objective-C AppDelegates that do not import `AppDelegate.h`.
- Ensures the plugin imports FirebaseCore in custom Swift AppDelegates that do not import Expo.
- Preserves existing import placement for standard templates, newline style, and idempotence.

Without the import, the plugin could add `[FIRApp configure]` or `FirebaseApp.configure()` while leaving the generated native source uncompilable.

## Implementation

A shared import insertion helper now:

- recognizes an existing exact import;
- inserts after the conventional AppDelegate/Expo anchor when present;
- otherwise inserts before the first import;
- falls back to the start of an import-free file.

## Verification

Regression tests reproduce both failures on upstream `main` and pass with this change. The complete repository gates also pass:

- `yarn lerna:prepare`
- `yarn tests:jest --runInBand` — 103 suites, 1,481 tests, 31 snapshots
- `yarn tsc:compile`
- `yarn tsc:compile:consumer`
- `yarn lint:deps`
- `yarn reference:api`
- `yarn compare:types`
- focused ESLint
- `git diff --check`

## Compatibility

No public API or breaking changes.

### DIFF
--- a/packages/app/plugin/__tests__/iosPlugin.test.ts
+++ b/packages/app/plugin/__tests__/iosPlugin.test.ts
@@ -110,4 +110,36 @@ describe('Config Plugin iOS Tests', function () {
     expect(twiceModifiedAppDelegate).toContain(singleImport);
     expect(twiceModifiedAppDelegate).not.toContain(doubleImport);
   });
+
+  it('adds the Firebase import to an Objective-C AppDelegate without the standard header import', function () {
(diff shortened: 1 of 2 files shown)


mikehardy (approved):
Hey @OskarEichler 👋

Cheers

PR #9265, opened 2026-08-30
fix(app): support arbitrary web storage keys

OskarEichler:
## What this fixes

- Prevents web event names such as `hasOwnProperty` from replacing the listener map method and crashing subsequent listener operations.
- Preserves event and preference keys such as `__proto__` instead of silently losing them through JavaScript prototype behavior.
- Stops `eventsGetListeners()` callers from mutating the internal listener registry.

These names are valid strings at the bridge boundary and native maps already support them.

## Implementation

The web shim now stores string-keyed data in null-prototype dictionaries, performs explicit own-property checks, and returns detached null-prototype snapshots.

## Verification

Exact baseline regressions failed with a TypeError and a lost preference value. All permanent focused cases now pass. Complete repository gates also pass:

- `yarn lerna:prepare`
- `yarn tests:jest --runInBand` — 103 suites, 1,482 tests, 31 snapshots
- `yarn tsc:compile`
- `yarn tsc:compile:consumer`
- `yarn lint:deps`
- `yarn reference:api`
- `yarn compare:types`
- focused ESLint
- `git diff --check`

## Compatibility

No public API/type or breaking change. Ordinary event and preference keys retain their existing behavior.

### DIFF
--- a/packages/app/__tests__/webAppModule.test.ts
+++ b/packages/app/__tests__/webAppModule.test.ts
@@ -113,3 +113,34 @@ describe('RNFBAppModule setImmediate guards', () => {
     });
   });
 });
+
+describe('RNFBAppModule arbitrary keys', () => {
+  it('tracks event names that collide with Object prototype properties', async () => {
+    appModule.eventsAddListener('hasOwnProperty');
+    appModule.eventsAddListener('hasOwnProperty');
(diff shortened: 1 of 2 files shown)


mikehardy (approved):
Hey @OskarEichler 👋

Cheers

PR #9266, opened 2026-08-30
fix(app): reject FileReader startup failures

OskarEichler:
## What this fixes

`Base64.fromData(blob)` now rejects its returned Promise when `FileReader.readAsDataURL()` cannot start. Previously that failure escaped synchronously even though the method declares and is consumed as a Promise, bypassing Storage upload rejection handling.

## Implementation

The Blob path now uses a native Promise executor, installs handlers before starting the read, and relies on the executor to convert synchronous FileReader exceptions into rejections. This also removes the unnecessary deferred-Promise bookkeeping.

## Verification

The exact baseline regression throws synchronously; the permanent fixed case observes the expected rejected Promise. Complete repository gates pass:

- `yarn lerna:prepare`
- `yarn tests:jest --runInBand` — 103 suites, 1,480 tests, 31 snapshots
- `yarn tsc:compile`
- `yarn tsc:compile:consumer`
- `yarn lint:deps`
- `yarn reference:api`
- `yarn compare:types`
- focused ESLint
- `git diff --check`

## Compatibility

Observable correction: FileReader startup failures now reject asynchronously instead of throwing synchronously. Successful Blob, ArrayBuffer, and Uint8Array conversions are unchanged; there is no type change.

### DIFF
--- a/packages/app/__tests__/app.test.ts
+++ b/packages/app/__tests__/app.test.ts
@@ -10,6 +10,7 @@ import {
 } from '../lib';
 import { Logger } from '../lib/internal/logger';
 import { NativeFirebaseError } from '../lib/internal';
+import Base64 from '../lib/common/Base64';
 
 describe('App', function () {
   describe('modular', function () {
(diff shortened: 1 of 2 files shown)


mikehardy (approved):
Hey @OskarEichler 👋

Cheers

PR #9267, opened 2026-08-30
fix(app): support prototype-named apps

OskarEichler:
## What this fixes

- Allows valid secondary Firebase app names that match JavaScript Object prototype properties, including `toString` and `__proto__`.
- Prevents per-app modular instance caching from traversing or writing through inherited objects for those names.

Previously both names were rejected as already registered before native initialization. The modular cache could additionally attach module entries to inherited prototype objects.

## Implementation

The app registry and both levels of the modular-instance registry now use null-prototype dictionaries. Lookup, memoization, deletion, and `Object.values()` behavior otherwise remain unchanged.

## Verification

Permanent focused cases initialize, resolve, memoize modules for, and delete both names while asserting no Object prototype pollution. Both fail on upstream `main`. Complete repository gates pass:

- `yarn lerna:prepare`
- `yarn tests:jest --runInBand` — 103 suites, 1,481 tests, 31 snapshots
- `yarn tsc:compile`
- `yarn tsc:compile:consumer`
- `yarn lint:deps`
- `yarn reference:api`
- `yarn compare:types`
- focused ESLint
- `git diff --check`

## Compatibility

No public API/type or breaking change. This expands valid app-name handling to match the existing non-empty-string contract.

### DIFF
--- a/packages/app/__tests__/app.test.ts
+++ b/packages/app/__tests__/app.test.ts
@@ -1,4 +1,5 @@
 import { describe, expect, it, jest } from '@jest/globals';
+import { NativeModules } from 'react-native';
 import {
   deleteApp,
   registerVersion,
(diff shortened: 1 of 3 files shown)


mikehardy (approved):
Hey @OskarEichler 👋

Cheers

PR #9268, opened 2026-08-30
fix(app): preserve empty memory storage values

OskarEichler:
## What this fixes

The in-memory AsyncStorage fallback now returns a stored empty string as `""` instead of converting it to `null`. This restores the AsyncStorage contract: only a missing key returns `null`.

## Implementation

- Use nullish fallback on the single Map lookup, preserving all defined strings including `""`.
- Return the underlying prefixed storage Promises directly instead of adding redundant async/await wrappers.

## Verification

The exact round-trip regression returns `null` on upstream `main` and `""` with this change. Complete repository gates pass:

- `yarn lerna:prepare`
- `yarn tests:jest --runInBand` — 103 suites, 1,480 tests, 31 snapshots
- `yarn tsc:compile`
- `yarn tsc:compile:consumer`
- `yarn lint:deps`
- `yarn reference:api`
- `yarn compare:types`
- focused ESLint
- `git diff --check`

## Compatibility

Observable correction: an explicitly stored empty string now round-trips instead of appearing absent. Non-empty and missing values are unchanged; there is no public type break.

### DIFF
--- a/packages/app/__tests__/app.test.ts
+++ b/packages/app/__tests__/app.test.ts
@@ -15,6 +15,12 @@ import Base64 from '../lib/common/Base64';
 import FirebaseModule from '../lib/internal/FirebaseModule';
 import { getOrCreateModularInstance } from '../lib/internal/registry/modular';
 import type { ModuleConfig } from '../lib/types/internal';
+import {
+  getItem,
+  removeItem,
+  setItem,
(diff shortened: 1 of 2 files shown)


mikehardy (approved):
Hey @OskarEichler 👋

Cheers

PR #9269, opened 2026-08-30
fix(app): isolate native routing keys

OskarEichler:
## What this fixes

Reading an inherited JavaScript property such as `toString` from a wrapped native module no longer gets mistaken for a routed native member and crashes while dereferencing a nonexistent route. This can occur during ordinary inspection or implicit coercion of the composite wrapper.

## Implementation

The composite router, method cache, and constant cache now use null-prototype dictionaries. Actual enumerable native members—including inherited TurboModule host methods—continue to be collected and routed unchanged.

## Verification

The exact baseline regression throws inside the wrapper getter; the permanent fixed assertion resolves the absent property as `undefined` while all existing single-host and merged-host routing checks pass. Complete repository gates pass:

- `yarn lerna:prepare`
- `yarn tests:jest --runInBand` — 103 suites, 1,479 tests, 31 snapshots
- `yarn tsc:compile`
- `yarn tsc:compile:consumer`
- `yarn lint:deps`
- `yarn reference:api`
- `yarn compare:types`
- focused ESLint
- `git diff --check`

## Compatibility

No public API/type or breaking change. Routed native methods and constants retain their existing behavior.

### DIFF
--- a/packages/app/__tests__/nativeModuleContract.test.ts
+++ b/packages/app/__tests__/nativeModuleContract.test.ts
@@ -50,6 +50,7 @@ describe('TurboModule wrapper contract (NewArch-AD-17.1)', function () {
     expect(methodB).toHaveBeenCalledTimes(1);
     expect(Object.keys(wrapped)).toContain('methodA');
     expect(Object.keys(wrapped)).toContain('methodB');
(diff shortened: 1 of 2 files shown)


mikehardy (approved):
Hey @OskarEichler 👋

Cheers

PR #9271, opened 2026-08-30
fix(app, android): clear stale custom auth domains

OskarEichler:
## Fixes

- Clear a secondary app’s cached custom `authDomain` after successful deletion, preventing a later app with the same name from inheriting stale configuration.
- Make the shared auth-domain cache thread-safe for bridge and serialization access.
- Read each cached domain once while serializing app options.

## Verification

- Added a regression assertion to the existing Android delete-app unit test; it failed before the fix and passes after it.
- `:packages:app:testDebugUnitTest`
- `:packages:app:assembleDebug`
- `:packages:app:lintDebug` (only the four existing generated-spec `StringFormatTrivial` warnings)
- `yarn lint:android` formatter completed with no additional formatting changes; its diff guard sees only this intentional native patch.
- `yarn tests:jest --runInBand` (104 suites, 1,480 tests)
- `yarn tsc:compile`
- `yarn tsc:compile:consumer`
- `yarn lint:deps`
- `yarn reference:api`
- `yarn compare:types`
- `git diff --check`

## Compatibility

No public JavaScript, TypeScript, native module, or generated-spec API changes.

### DIFF
--- a/packages/app/android/src/reactnative/java/io/invertase/firebase/app/NativeRNFBTurboApp.java
+++ b/packages/app/android/src/reactnative/java/io/invertase/firebase/app/NativeRNFBTurboApp.java
@@ -35,11 +35,12 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 
 public class NativeRNFBTurboApp extends NativeRNFBTurboAppSpec implements LifecycleEventListener {
   private static final String TAG = "App";
 
-  public static Map<String, String> authDomains = new HashMap<>();
+  public static final Map<String, String> authDomains = new ConcurrentHashMap<>();
 
   NativeRNFBTurboApp(ReactApplicationContext reactContext) {
     super(reactContext);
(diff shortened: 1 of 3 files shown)


OskarEichler:
I re-reviewed this PR against the current `main` branch and found that its branch had accidentally included the preceding PR's commit. I have force-updated the branch so the PR now contains only its intended three-file change:

- make the Android auth-domain cache concurrency-safe;
- clear the cached auth domain after a successful app deletion;
- avoid the duplicate cache lookup while serializing app options;
- cover both the cached and absent serializer paths, as well as deletion cleanup.

The final implementation is intentionally small: 31 insertions and 3 deletions across the two production files and one existing JVM test file. I did not find further simplifications that would preserve the lifecycle behavior and thread safety.

Validation completed on the frozen production diff:

- canonical Android Java formatter: pass
- Android JVM unit suite: pass
- merged JaCoCo report: 100% on all reachable touched executable lines and both touched branches
- Android build: pass
- Android app-area e2e: 39 passing
- iOS build and app-area e2e: 39 passing
- macOS build and app-area e2e: 33 passing
- `git diff --check`: pass

The branch is now a single commit directly on current `main`.

mikehardy:
with 41 PRs to process I made a temporary parallel queue processor local for review / triage, gotten through a big chunk already and still chewing on the rest. Hadn't gotten to this one yet, so no worries :-). It'll only pull this updated version. Cheers

mikehardy (commented):
this one also needs queuing for a look after #9093 lands because it's working in the same area of thread safety for stored handles in maps

mikehardy (approved):
ok to merge as-is, doesn't intersect enough with #9093 to warrant HandleMap use

PR #9275, opened 2026-09-01
refactor(ai)!: remove Imagen API

russellwheatley:
Removes the Imagen API from `@react-native-firebase/ai` now that the Imagen service has shut down. Matches the removals already landed in the [JS](https://github.com/firebase/firebase-js-sdk/pull/10267), [iOS](https://github.com/firebase/firebase-ios-sdk/pull/16483), and [Android](https://github.com/firebase/firebase-android-sdk/pull/8512) SDKs.

- Deletes the full vertical slice: `ImagenModel` / `TemplateImagenModel`, the public and internal Imagen types, `ImagenImageFormat`, the predict-only request/response helpers, and the `PREDICT` / `TEMPLATE_PREDICT` task members
- Gemini image generation via `getGenerativeModel()` and the non-Imagen template request paths are untouched
- Adds `docs/migrating-to-v27.mdx` mapping all 14 removed exports to their Gemini replacements, with before/after code

One thing worth a look during review: `compare:types` passes at 0 undocumented / 0 stale, but only because the 14 Imagen exports are documented as `missingInRN`. The repo pins `firebase@12.17.1`, resolving `@firebase/ai@2.14.0`, which still ships Imagen upstream. Those entries go stale automatically once the pin reaches 12.18.0 and `compare:types` will fail then, which is the intended tripwire. The removal condition is recorded in `configs/ai.ts`.

iOS and Android e2e were not run. Both fail at host infrastructure before any JS executes, so they'd reproduce a known failure rather than test this change. macOS e2e ran at full unfocused harness: 1281 passing, 0 failing, and "Imagen" appears nowhere in the log.

### Also included

Two smaller commits picked up while writing the migration guide, both tooling rather than product:

- `fix(docs)`: `lint:spellcheck` exited 0 when a page's YAML frontmatter failed to parse, silently skipping that page's frontmatter. Usual trigger is an unquoted colon in a value. `scripts/spellcheck.mjs` wraps the existing invocation with identical flags and makes it fatal. Body misspellings already exited 1 and still do.
- `docs(okf)`: records the docs-lint table padding gotcha, plus two agent traps (bash `PIPESTATUS` is empty in zsh, and a trailing `&` does not keep a service alive).

Happy to split those out if you'd rather keep this to the AI removal.

Closes CPRN-323


mikehardy:
I've got a huuuuge stack o' fixes I want to get in for a v26 fix release before doing a major, so without reviewing this at all, just marking it as do not merge (sorry)

mikehardy:
batch with 4 deferred / breaking changes from Oskar in CPRN-402 currently approved but waiting for a break batch

PR #9279, opened 2026-09-02
fix(firestore, android): avoid Tasks.await wedging collection gets

### LINKED ISSUE #9278: Firestore: unbounded Tasks.await on a single-thread executor permanently wedges all collection reads after sign-out
## Summary

`Tasks.await(task)` is called with no timeout inside a body that runs on a **single-thread executor with an unbounded queue**, held in a `static` map for the process lifetime. One task that never completes blocks that thread permanently, and every subsequent read on that module's executor queues behind it and **never starts** — silently, with no error and no rejection.

In practice this is triggered by **signing out**: a `get()` in flight against a Firestore client being torn down at the credential change never completes, and from that moment every `getDocs` in the process hangs until the app is killed.

## Versions

- `@react-native-firebase/firestore` **26.3.3** (also reproduced on 23.8.6 — the code path is unchanged across those majors)
- `@react-native-firebase/app` 26.3.3
- React Native 0.81.5, New Architecture enabled (bridgeless)
- Firebase Android BoM: Firestore native SDK 26.6.0
- Android 15, physical device

## The mechanism

`ReactNativeFirebaseFirestoreQuery.java`:

```java
public Task<WritableMap> get(Executor executor, Source source) {
  return Tasks.call(
      executor,
      () -> {
        QuerySnapshot querySnapshot = Tasks.await(query.get(source));  // unbounded
        ...
      });
}
```

`NativeRNFBTurboFirestoreCollection.handleQueryGet` passes `turboSupport.getExecutor()`. In `TaskExecutorService`:

```java
public ExecutorService getExecutor() {
  boolean isTransactional = maximumPoolSize <= 1;   // default 1 -> true
  return getExecutor(isTransactional, "");
}

private ExecutorService getNewExecutor(boolean isTransactional) {
  if (isTransactional) {
    return Executors.newSingleThreadExecutor();      // one thread, unbounded queue
  }
  ...
}

private static final Map<String, ExecutorService> executors = new HashMap<>();  // process lifetime
```

So the Firestore call is issued **from** the single executor thread, and `Tasks.await` blocks that thread for its entire duration. An unbounded queue never rejects, so `RejectedExecutionException` — which `handleQueryGet` does catch — never fires.

`NativeRNFBTurboFirestoreDocument.java` has the same unbounded await on its own module executor.

## Why it presents as "some reads hang, others don't"

Executors are keyed `name + "TransactionalExecutor" + identifier`, where `name` is the module name:

| call | executor key | effect |
|---|---|---|
| `collectionGet` | `RNFBCollectionTransactionalExecutor` | **wedges** |
| `collectionOnSnapshot` | `RNFBCollectionTransactionalExecutor` + `listenerId` | unaffected |
| `documentGet` | `RNFBDocumentTransactionalExecutor` | unaffected |

Because `getTransactionalExecutor(identifier)` gives listeners their own executor and the document module constructs `FirestoreTurboModuleSupport("RNFBDocument")`, `onSnapshot` and `getDoc` keep working while every `getDocs` hangs. The same collection can appear on both sides — a `users/{uid}` `getDoc` succeeds while a `users` collection query hangs — which makes it look like a data or rules problem rather than a threading one.

## Evidence that the query never reaches Firestore

With `setLogLevel('debug')`, comparing two windows of a **single process**, 60 seconds apart, same user, same rules:

| window | `Allocate target` events | timed-out `getDocs` |
|---|---|---|
| before sign-out | **25** | 0 |
| after sign-out → sign-in | **5** (only the document gets and one listener) | **10** |

Firestore allocates no target for the hanging queries, so they are not failing inside Firestore — the task never runs. Meanwhile the Watch stream is demonstrably healthy in the same window: `Stream is open`, gRPC `READY`, `ListenRequest` sent and `ListenResponse` received, `FirestoreCallCredentials: Successfully fetched AppCheck token`.

## Reproduction

1. Sign in (Google), let the app perform some collection reads.
2. `signOut()`.
3. Sign in again — same or different user.
4. Every `getDocs` from then on never resolves. `getDoc`, `onSnapshot` and writes continue to work.
5. Survives foreground/background, network loss and recovery, and time (observed >50 minutes). Only killing the process clears it.

A restored session is unaffected, because no in-flight `get()` is orphaned — which is why this does not reproduce by force-stop-and-relaunch and looks intermittent.

## Why client-side timeouts do not help

A JS-side timeout rejects the promise but cannot cancel the native task. The thread stays blocked, so each "handled" timeout leaves the executor exactly as stuck as it found it — and each retry queues one more task that will never run.

## Suggested fix

Bound the awaits:

```java
Tasks.await(query.get(source), timeoutSeconds, TimeUnit.SECONDS);
```

`TimeoutException` propagates through `Tasks.call`, so the existing `addOnCompleteListener` / `rejectPromiseFirestoreException` path already surfaces it correctly. Bounding fails the queued task rather than draining a backlog, which matters because a drained backlog would execute stale queries against a credential that has since changed.

Alternatively, move gets off a shared single-thread executor, or use a bounded queue so `RejectedExecutionException` can actually fire.

We are running this locally as a patch with a 30s bound — deliberately above our 12s client-side read bound, so the caller still fails fast while the thread frees rather than never.

Happy to supply the full instrumented logs or test against a fix.

mikehardy:
- Fixes #9278

## Summary
Android Firestore query/document `get` used unbounded `Tasks.await` on the shared single-thread module executor, so one never-completing get could permanently stall later `getDocs`/`getDoc`. This switches those paths to `Task.continueWith` (same non-blocking style as aggregates / iOS).

Somewhat speculative without a reporter repro for the sign-out trigger; the executor wedge mechanism and the regression unit test are solid either way, which is why this is a draft.

## Test plan
- [x] JVM: `FirestoreAsyncTaskMapTest` (pending upstream does not block single-thread executor; failed-upstream exception shape)
- [x] Slot-3 Android firestore-narrowed e2e (763) + Jacoco 100% on touched lines
- [ ] Reporter repro / regression once available

---
Maintainer note: Fixes internal CPRN-406


mikehardy:
Closed by a bad force-push during squash (branch briefly pointed at main). Replaced by draft https://github.com/invertase/react-native-firebase/pull/9296 (single squashed commit).

PR #9284, opened 2026-09-03
chore: set npm homepage to rnfirebase.io on published packages

russellwheatley:
Adds `homepage` so npm (and scanners that key off it) can bind `@react-native-firebase/*` to https://rnfirebase.io. Repository URLs are unchanged. Private workspace and test apps are untouched.

Does not show on the registry until the next publish.

---
Maintainer note: Fixes internal CPRN-337

### DIFF
--- a/packages/ai/package.json
+++ b/packages/ai/package.json
@@ -11,6 +11,7 @@
     "compile": "bob build",
     "prepare": "yarn run build && yarn compile"
   },
+  "homepage": "https://rnfirebase.io",
   "repository": {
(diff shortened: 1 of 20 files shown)


mikehardy (approved):
I know it's draft but I was just in this area for Trusted Publish / OIDC so I had to look to see what happened

> Repository URLs are unchanged

That's the critical part for OIDC to work, and looking at diff I don't think this will break publish. So if it helps some other aspect (discoverability in this case), awesome, +1

PR #9285, opened 2026-09-03
fix(e2e): gate Detox -crash-report-mode flag on emulator -help support

russellwheatley:
Detox's `EmulatorExec.js` LaunchCommand always appended `-crash-report-mode never`. On emulator 36.2.12.0 that flag doesn't exist, so the emulator spawn fails outright (`SPAWN_FAIL`) and Android e2e never reaches Jest.

`EmulatorExec.js` now probes `emulator -help` once (memoized) via a new `emulatorSupportsFlag(helpOutput, flag)` helper and only appends `-crash-report-mode never` when the flag is actually listed. Hosts that omit it skip the args, and a one-time `log.warn` fires so it's visible instead of silent. On hosts where the probe can't add the flag, the crash-report modal still needs to be suppressed manually via macOS `defaults write com.android.Emulator set.crashReportPreference -int 2`, documented as a wave-operator fallback in `running-e2e.md`.

- `tests/e2e/androidAdbRange.js`: add `emulatorSupportsFlag` pure helper
- `.yarn/patches/detox-npm-20.51.0-3e13b6e309.patch`: gate the flag behind the probe in `EmulatorExec.js`
- `packages/app/__tests__/androidAdbRange.test.js`: unit tests for the helper + updated patch-shape pinning
- `okf-bundle/ci-workflows/detox-patches.md`: inventory row updated to describe the probe
- `okf-bundle/testing/running-e2e.md`: manual fallback step for hosts where the probe omits the flag


russellwheatley:
Fresh-eyes review (independent pass, zero prior context on this work): PASS.

- Diffed the full Detox patch against the prior 11 hunks (idling resources, timeouts, ADB install timeout, pinned console port, early-ready buffering, lock staleness) \u2014 nothing dropped or regressed, only the crash-report-mode hunk changed as intended.
- Manually verified `emulatorSupportsFlag`'s regex against real `emulator -help`-style output: correct on flag-present, flag-absent, prefix-only (rejected), `=` separator, end-of-string, CRLF, and malformed input.
- Unit tests exercise the helper meaningfully (not just superficial regex pinning), no `.only` in the changed test files.
- Docs (`detox-patches.md`, `running-e2e.md`) match what the code actually does.

mikehardy (approved):
sorry about that, didn't realize it was version specific

I'm here locally

> INFO         | Android emulator version 37.1.11.0 (build_id 15917651) (CL:N/A)

and on CI

> 13:14:03.035 detox[24254] i device Detected emulator binary version { major: 37, minor: 1, patch: 11, toString: [Function: toString] }


So this was just a silent "works for me" type assumption. I keep my android studio pretty up to date locally and just assumed everyone did ;-)


The flag is a real e2e timesaver though since the emulator borks out semi-frequently and just hands an e2e slot

russellwheatley:
@mikehardy - kinda torn on this. I could just update android studio and this would be resolved hypothetically, but also thinking future devs might run into same problem. But maybe the out of this is to log emulator  >= v37 requirement in okf-bundle? 🤔

mikehardy:
less code is better code IMHO, and I'm usually a fan of rules for developers that say "if you're a developer the general expectation is you should have up to date toolchains or you can expect to be required to update them at any time"

russellwheatley:
Agreed, going the less-code route. Closing this in favor of a doc-only note instead of the runtime probe: the flag genuinely needs emulator >= 37, and the toolchain-currency expectation applies generally (Android and iOS).

PR #9287, opened 2026-09-03
docs(e2e): note Android Emulator >=37 requirement for -crash-report-mode

russellwheatley:
Detox's `EmulatorExec.js` LaunchCommand passes `-crash-report-mode never` to suppress the Google crash-report modal during unattended e2e waves. That flag needs Android Emulator **>= 37** (confirmed present on 37.1.11.0); emulator 36.2.12.0 lacks it and `SPAWN_FAIL`s at launch.

Instead of adding a runtime capability probe around the flag, this documents the minimum version in `detox-patches.md` and adds a general toolchain-currency note to `running-e2e.md` (Android Studio/emulator, Xcode/CocoaPods) so hosts stay current rather than the codebase growing version-compat shims.

Replaces #9285.


mikehardy (approved):
👍

PR #9288, opened 2026-09-03
fix(auth): [LOW] handle malformed action links

OskarEichler:
🔥

## Summary

Malformed percent-encoding in an Authentication action link currently escapes the documented invalid-link handling because the deep-link parser runs before the existing `try` block. An externally supplied link such as one containing `link=%` or `deep_link_id=%` can therefore throw a `URIError` into application code instead of returning `null`.

This moves deep-link parsing into the existing error boundary so every malformed or invalid action link follows the documented `null` result.

## Risk

**LOW:** this is an availability and robustness issue at an external deep-link boundary. It does not bypass authentication or disclose data, but an unhandled exception can interrupt an application flow if the caller assumes the documented nullable result.

## Tests

- Added regression cases for malformed `link` and `deep_link_id` values.
- Focused auth Jest suite: 100 tests passed.
- Full Jest suite: 105 suites / 1,522 tests / 31 snapshots passed.
- `yarn tsc:compile`
- `yarn tsc:compile:consumer`
- `yarn lint:js`
- `yarn lint:deps`

### DIFF
--- a/packages/auth/__tests__/auth.test.ts
+++ b/packages/auth/__tests__/auth.test.ts
@@ -247,6 +247,8 @@ describe('Auth', function () {
     it('`parseActionCodeURL` returns null for invalid action links', function () {
       expect(parseActionCodeURL('https://example.com/not-an-action-link')).toBeNull();
       expect(ActionCodeURL.parseLink('https://example.com/not-an-action-link')).toBeNull();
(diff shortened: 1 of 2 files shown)


mikehardy (approved):
Nice - LGTM - thank you!

PR #9289, opened 2026-09-03
fix(ci): [MEDIUM] verify yeetd package checksum

OskarEichler:
🔥

## Summary

The iOS E2E workflow and Tart provisioning script download the privileged `yeetd` installer from a release URL and execute it with `sudo` without verifying the downloaded bytes. A replaced or corrupted release asset would therefore run before the test build with host-level privileges.

This verifies the SHA-256 digest of the published 1.0 package before either installer invocation. A mismatch fails immediately and prevents installation.

## Risk

**MEDIUM:** exploitation requires compromise or replacement of the external release asset or its delivery path, but successful exploitation reaches a GitHub Actions runner or golden CI image with elevated privileges.

## Verification

- Re-downloaded the official 1.0 asset and independently confirmed SHA-256 `adb04eb720581a32d94daef3ea61108c74494083a79f00836556a36214f354be`.
- Confirmed the valid package passes verification.
- Confirmed a modified package is rejected.
- `bash -n scripts/tart/provision-toolchain.sh`
- ShellCheck passed with only the script's pre-existing exclusions.
- Workflow YAML parse and Prettier checks passed.
- `git diff --check`

### DIFF
--- a/.github/workflows/tests_e2e_ios.yml
+++ b/.github/workflows/tests_e2e_ios.yml
@@ -312,6 +312,7 @@ jobs:
       - name: Install yeetd and fix iOS perf issues
         run: |
           wget https://github.com/biscuitehh/yeetd/releases/download/1.0/yeetd-normal.pkg
+          echo "adb04eb720581a32d94daef3ea61108c74494083a79f00836556a36214f354be  yeetd-normal.pkg" | shasum -a 256 -c -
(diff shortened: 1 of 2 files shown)


mikehardy (approved):
nice - supply chain hardening is an ongoing process and a current goal so this fits well, thanks

PR #9290, opened 2026-09-03
fix(ci): [MEDIUM] pin npm in privileged workflows

OskarEichler:
🔥

## Summary

The npm publish and API-reference deployment workflows install `npm@latest` immediately before dependency installation and OIDC-enabled publication/deployment. The mutable dist-tag makes the privileged workflow's executable dependency change without repository review.

This pins npm to the currently verified `12.0.2` release in both privileged workflows. Future npm upgrades remain straightforward, but now arrive through an auditable repository change.

## Risk

**MEDIUM:** this is supply-chain hardening at two workflows with `id-token: write`; the publish workflow also handles a repository-admin token. Pinning limits an unexpected or compromised `latest` update from being selected automatically.

## Verification

- Confirmed `npm@12.0.2` from the npm registry and its declared Node.js compatibility.
- Installed the exact package into an isolated prefix and confirmed it reports version `12.0.2`.
- Workflow YAML parse and Prettier checks passed.
- `git diff --check`

### DIFF
--- a/.github/workflows/deploy-api-reference.yml
+++ b/.github/workflows/deploy-api-reference.yml
@@ -27,11 +27,11 @@ jobs:
         with:
           node-version: 24
           registry-url: 'https://registry.npmjs.org'
-      - name: Install latest NPM
-        run: |
-          npm --version
-          npm install -g npm@latest
-          npm --version
+      - name: Install pinned NPM
+        # Content-pin via hardcoded dist.integrity (see script).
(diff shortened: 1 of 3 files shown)


mikehardy (approved):
same as the other supply-chain hardening - definitely a current goal of ours and *especially* in our release workflows. I overlooked this one in previous passes - thank you

mikehardy (approved):
this cannot be tested until publish, but worked locally during some testing

PR #9291, opened 2026-09-03
test(coverage): adopt react-native-coverage@0.2.0

mikehardy:
## Summary
- Migrate the RNFB test app onto published `react-native-coverage@0.2.0` for native coverage flush and JS pull/report.
- Delete in-tree Android/iOS coverage TurboModule sources and duplicated host scripts; keep thin RNFB wrappers for Jet transport, Istanbul instrumentation, and Jacoco/XCTest merge.
- Extract the standalone iOS messaging delegate probe and update OKF coverage ownership docs to match.
- Document RNFB coverage migration rollback playbook in OKF.

## Test plan
- [x] Slot-2 narrowed iOS `:test-cover` (app + messaging) — 74 passing; `tests:ios:test:process-coverage` exit 0
- [x] Slot-2 narrowed Android `:test-cover` — 88 passing; `tests:android:post-e2e-coverage` exit 0
- [x] `yarn tests:coverage:assert-presence` exit 0 (non-empty `coverage/ios-native/lcov.info`, Jacoco XML, `coverage/js/{ios,android}/lcov.info`)
- [x] `yarn lint:js` exit 0 on frozen cutover
- [x] CI e2e Codecov uploads for iOS native LCOV + Android Jacoco XML (debug jobs: Upload Codecov *native* steps succeeded on tip)

---
Maintainer note: Fixes internal CPRN-363


.github/workflows/tests_e2e_ios.yml
```diff
@@ -414,9 +414,10 @@ jobs:
           yarn tests:ios:test-cover 2>&1 | tee detox-step.log
           exit ${PIPESTATUS[0]}
 
+      # Presence guard: empty packagesHits or missing profraw exits non-zero (prefer 2).
+      # Do not continue-on-error — silent empty coverage must fail CI.
       - name: Process iOS native coverage
         if: always() && contains(matrix.buildmode, 'debug')
-        continue-on-error: true
         run: yarn tests:ios:test:process-coverage
```
russellwheatley: Android's guard in `android-e2e-detox.sh` only fails the job on a coverage error when `TEST_EXIT -eq 0`, so a real Detox failure doesn't also surface a coverage failure. This step runs on `always()` with no equivalent check now that `continue-on-error` is gone, so a test failure that leaves no profraw will show as two red steps instead of one. Might be worth mirroring the Android guard here, or is the asymmetry intentional?

russellwheatley (changes requested):
This has a merge conflict with `main` right now (in `okf-bundle/packages/messaging/index.md` and `okf-bundle/testing/coverage-design.md`), both from main's newer `messagingStoreSupportsDisabledStorage` probe doc landing after this branch was cut. Needs a rebase before merge.

One inline note below on the iOS coverage guard. Otherwise the migration itself checked out fine against the published `react-native-coverage@0.2.0` source: Podfile/pbxproj wiring, per-image dynamic-framework flush config, Gradle resolution, the Android presence guard, and the deleted in-tree TurboModule cleanup all look consistent, no stale references left behind.

Separately, not blocking: `tests/scripts/capture-native-coverage-baseline.js` (653 lines) plus the committed `coverage-baseline.json` (the T=1% variance harness) reads as independent scope from the package swap itself. Might be worth its own PR, but not asking you to split it out at this point.

.github/workflows/tests_e2e_ios.yml
mikehardy: Hey Russell 👋

Not intentional. I mirrored the Android guard: process still runs on `always()`, but the step only fails the job when Detox succeeded. Empty coverage after a green Detox run still fails CI; a Detox failure that leaves no profraw no longer paints a second red X.

mikehardy:
Thanks for the careful pass.

1. Rebased onto `main`. Kept the `messagingStoreSupportsDisabledStorage` probe docs and folded them into the post-cutover module map.
2. iOS presence guard: see the inline reply. Same policy as `android-e2e-detox.sh` now.
3. Baseline harness: agree it reads as separable from the package swap. Leaving it in for this PR since you said not blocking / not asking to split. Happy to peel it later if it still bothers the review surface.

russellwheatley (approved):
LGTM pending green CI

mikehardy:
Oh no, it looks like https://github.com/invertase/react-native-firebase/commit/8776b2878803419bf6fffc363635eab242c7e4e3 is causing problems

- ref #9283

That's going to start flaking CI and that's not okay, I don't think it's related to this PR at all but it needs a fix, will track separately

mikehardy:
Flake fix in here:

- #9297

PR #9293, opened 2026-09-04
fix(ai): avoid shell execution for fixture tags

russellwheatley:
The AI test-fixture fetcher (`scripts/fetch_ai_mock_responses.ts`) interpolated the latest git tag from the external `vertexai-sdk-test-data` repo directly into an `execSync` shell string. A tag containing shell metacharacters (e.g. `v17.0;id`) would have the non-tag portion executed by the shell.

- Switch both git invocations (`ls-remote`, `clone`) to `execFileSync` with argument arrays
- Add a `--` separator before positional repo/path args
- Wrap the top-level logic in an exported `fetchAiMockResponses()` behind a `require.main === module` guard so it's testable
- Add a regression test asserting the malicious tag never reaches shell interpretation

Reported by Oskar Eichler, who also supplied the original patch this is based on.

---
Maintainer note: Fixes internal CPRN-418

### DIFF
--- a/packages/ai/__tests__/fetchAiMockResponses.test.ts
+++ b/packages/ai/__tests__/fetchAiMockResponses.test.ts
@@ -1,0 +1,104 @@
+import { execFileSync } from 'child_process';
+import { existsSync, readdirSync } from 'fs';
+import { rimrafSync } from 'rimraf';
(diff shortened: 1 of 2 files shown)


mikehardy (approved):

PR #9294, opened 2026-09-04
fix(app): stop evaluating firebase.json content as Ruby/Python source in ios_config.sh

russellwheatley:
`packages/app/ios_config.sh` spliced `firebase.json` content straight into `ruby -e` / `python3 -c` source strings when reading config keys during the iOS Xcode build phase. A single quote in a string value breaks out of the Ruby string literal, so the rest of the value runs as Ruby with the build user's privileges (CWE-94). Reported privately through our security disclosure process, credit to the reporter.

Exploiting this needs attacker control over `firebase.json` content (a malicious project contribution, a copied config artifact, or a compromised config-supply chain), it isn't reachable from the mobile runtime.

- Pass `firebase.json`'s path and the requested key(s) through `ARGV`/`sys.argv` instead of interpolating file content into interpreter source, for both the root parse and `getFirebaseJsonKeyValue`
- Regenerate the CocoaPods-embedded copy of the same script in `tests/ios/testing.xcodeproj/project.pbxproj` to match (this is the copy Xcode actually runs for the e2e test app)
- Apply the same fix to the identical copy in `test-rn-bare/ios/testrnbare.xcodeproj/project.pbxproj`, which carried the same vulnerable pattern
- Add `packages/app/__tests__/iosConfig.test.js`, a regression test that runs the script against a crafted `firebase.json` containing the injection payload and asserts it doesn't execute


mikehardy (approved):

PR #9295, opened 2026-09-04
fix(app): reject unsafe deepSet path segments

russellwheatley:
Rejects `__proto__`, `prototype`, and `constructor` path segments in `deepSet` before traversal, closing a low-severity prototype-pollution primitive (CWE-1321).

`deepSet({}, 'constructor.prototype.polluted', true, false)` previously walked the inherited `constructor` -> `prototype` chain and wrote onto `Object.prototype`, so any unrelated object created afterward inherited `polluted`.

- Add the guard in `packages/app/lib/common/deeps.ts`
- Add regression tests for both the `__proto__` and `constructor.prototype` attack forms

Reported privately through the OSS security inbox. `deepSet` is currently only reachable internally via `MutatableParams`, no remotely-controlled call path was identified.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -30,4 +30,12 @@ module.exports = {
   transformIgnorePatterns: [
     'node_modules/(?!((jest-)?react-native|@firebase|@react-native(-community)?))',
   ],
+  // Packages cross-import each other's built output (e.g. database imports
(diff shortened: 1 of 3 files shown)


mikehardy (approved):

PR #9303, opened 2026-09-08
fix(firestore): fail iOS transaction timeout without SDK retry

### LINKED ISSUE #9298: [🐛] [iOS] [firestore] Hard abort in transactionGetDocument after the 15s bridge timeout: "A transaction object cannot be used after its update callback has been invoked"
**Repo:** invertase/react-native-firebase · **Package:** `@react-native-firebase/firestore` 23.8.6 (the code is unchanged on `main` as of 2026-09, now `RNFBFirestoreTransactionModule.mm`) · **Platform:** iOS only · **RN:** 0.86.2

## Summary

`runTransaction` can crash the app with a Firestore hard assertion when the JS update function answers late. The iOS bridge waits up to 15 s for JS on a semaphore, returns `FIRFirestoreErrorCodeDeadlineExceeded` from the update block on timeout, and the Firestore SDK treats that error as **retryable** and re-invokes the block. The bridge keeps one `transactionState` per transaction id across attempts and `transactionGetDocument` calls `[transaction getDocument:ref error:]` directly on whatever transaction that state holds, with no guard for "the update block already returned". When JS eventually answers both the stale and the retry `update` events for the same id, the second read lands on an already-committed transaction and the SDK aborts the process:

```
NSInternalInconsistencyException
FIRESTORE INTERNAL ASSERTION FAILED: A transaction object cannot be used after its update callback has been invoked. (expected !committed_)
  firebase::firestore::core::Transaction::EnsureCommitNotCalled()   transaction.cc:241
  firebase::firestore::core::Transaction::Lookup(...)
  -[FIRTransaction getDocument:error:]                                FIRTransaction.mm:169
  -[RNFBFirestoreTransactionModule transactionGetDocument::::::]      RNFBFirestoreTransactionModule.m:82
```

When the late read lands a few milliseconds earlier, before `Commit()` is called but after the buffered writes were applied, the SDK instead rejects the promise with `firestore/invalid-argument` ("Firestore transactions require all reads to be executed before all writes"). We see both, in that order, seconds apart, on the same device.

## Where it comes from (bridge code, current)

- `transactionBlock`: `dispatch_semaphore_wait(semaphore, dispatch_time(DISPATCH_TIME_NOW, 15 * NSEC_PER_SEC))`; on timeout `*errorPointer = [NSError errorWithDomain:FIRFirestoreErrorDomain code:FIRFirestoreErrorCodeDeadlineExceeded ...]`.
- Firestore core: `TransactionRunner::HandleTransactionError` retries when `IsRetryableTransactionError(status)`, which is true for anything `!Datastore::IsPermanentError(error)`, and `kErrorDeadlineExceeded` is listed as **not** permanent in `datastore.cc`. So the bridge's own timeout error triggers a retry of the block.
- The retried block stores a new semaphore and transaction into the **same** `transactionState` dictionary and sends a second `update` event for the same `transactionId`.
- `transactionGetDocument` looks the state up by id and calls the SDK directly. Nothing records that the previous block invocation has returned or committed, so a read issued for the earlier attempt reaches the SDK on the wrong transaction.
- `methodQueue` is the module's serial queue, so a slow `getDocument` (transaction reads go to the server) from one transaction delays the reads of another one queued behind it. Two transactions overlapping on a slow connection is enough to push the second one past 15 s without any JS stall.

Android has the same design (`transactionHandler.await()` with a 15 s timeout, `DEADLINE_EXCEEDED` thrown from the block) but the equivalent late read throws a catchable Java exception and surfaces as a rejected promise rather than a crash.

## Reproduction shape (what we observe in production)

1. A read-merge-write `runTransaction` on one document, fired on a debounce and again from the `AppState` background handler, so two can overlap.
2. A slow or flaky connection (most of our failures carry `[firestore/unavailable]` / `deadline-exceeded` in the minutes before).
3. The second transaction's `get` waits behind the first on the serial queue, its block times out at 15 s, the SDK retries, JS then runs the update function twice for that id.
4. Crash. 27 crashes / 13 installs in 11 days on ~150–220 iOS DAU, only on the builds that introduced the transaction; zero before.

A deterministic repro is harder because it needs the second update event to arrive after the first attempt's buffer was applied; a `runTransaction` whose update function `await`s something slower than 15 s before calling `transaction.get()` should hit the same assertion once the SDK retries.

## Suggested fix

In `transactionBlock`, after the semaphore wait, mark the state (`transactionState[@"blockReturned"] = @YES`, or store the attempt's semaphore identity) before returning. In `transactionGetDocument` / `transactionApplyBuffer`, reject with a `firestore/aborted`-style error when the stored transaction is not the live attempt (semaphore mismatch or the returned flag set) instead of calling into the SDK. That turns the hard abort into a rejected promise, which the SDK already retries anyway.

## Environment

- `@react-native-firebase/app` / `firestore` 23.8.6 (Firebase iOS SDK per the pod lock of that release)
- React Native 0.86.2, Expo SDK 57, new architecture
- iOS 17.6 – 26.6, iPhone 15 and 16 family, production App Store builds

russellwheatley:
- Fixes #9298

iOS `runTransaction` was returning a retryable Firestore `DeadlineExceeded` after the 15s JS-bridge wait. The SDK retried, and a leftover `transactionGetDocument` hit `EnsureCommitNotCalled`.

Fail that wait with a non-`FIRFirestoreErrorDomain` timeout so the SDK does not retry. That is required with this design: `prepare` resets get eligibility, so a retry would make a leftover JS get look live on the next `FIRTransaction`. Android still throws retryable `DEADLINE_EXCEEDED`, so the same slow `updateFunction` fails once on iOS and keeps retrying on Android.

- Isolate leftover get and applyBuffer. Leftover get rejects `deadline-exceeded` (or `aborted` if disposed). A missing id still rejects `internal-error`. Leftover applyBuffer is a no-op.
- JS skips `applyBuffer` / a second reject once the id is already finished

---
Maintainer note: Fixes internal CPRN-426

### DIFF
--- a/packages/firestore/RNFBFirestore.podspec
+++ b/packages/firestore/RNFBFirestore.podspec
@@ -37,6 +37,7 @@ Pod::Spec.new do |s|
     'ios/RNFBFirestore/RNFBFirestoreTransactionModule.h',
     'ios/RNFBFirestore/RNFBFirestoreListenerRegistry.h',
     'ios/RNFBFirestore/RNFBFirestoreTransactionRegistry.h',
+    'ios/RNFBFirestore/RNFBFirestoreTransactionAttempt.h',
     'ios/generated/**/*.h',
   ]
   s.exclude_files       = 'ios/generated/RCTThirdPartyComponentsProvider.*', 'ios/generated/RCTAppDependencyProvider.*', 'ios/generated/RCTModuleProviders.*', 'ios/generated/RCTModulesConformingToProtocolsProvider.*', 'ios/generated/RCTUnstableModulesRequiringMainQueueSetupProvider.*', 'ios/*UnitTests/**'
--- a/packages/firestore/__tests__/runTransaction.test.ts
+++ b/packages/firestore/__tests__/runTransaction.test.ts
@@ -42,3 +42,167 @@ describe('runTransaction options passthrough', function () {
     expect(transactionBegin).not.toHaveBeenCalled();
   });
 });
+
+describe('runTransaction dead-id guard', function () {
+  function nativeMocks(handler: FirestoreTransactionHandler) {
+    return handler._firestore.native as unknown as {
+      transactionApplyBuffer: jest.Mock;
+      transactionDispose: jest.Mock;
+    };
+  }
+
+  it('does not applyBuffer or reject again after the id is already finished', async function () {
+    const handler = createHandler(jest.fn());
+    const { transactionApplyBuffer } = nativeMocks(handler);
+
+    let releaseUpdate: (value?: unknown) => void = () => undefined;
+    const pending = handler._add(
+      () =>
+        new Promise(resolve => {
+          releaseUpdate = resolve;
+        }),
+    );
+
+    const id = Number(Object.keys(handler._pending)[0]);
+    const updateWork = handler._handleUpdate({ listenerId: id, body: { type: 'update' } });
+
+    handler._handleError({
+      listenerId: id,
+      body: {
+        type: 'error',
+        error: { code: 'deadline-exceeded', message: 'timeout' },
+      },
+    });
+
+    await expect(pending).rejects.toMatchObject({
+      code: 'firestore/deadline-exceeded',
+    });
+
+    releaseUpdate('late');
+    await updateWork;
+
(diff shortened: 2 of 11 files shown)


mikehardy (approved):

PR #9318, opened 2026-09-11
fix(app, ios): pin SPM Firebase SDK and retain Measurement ObjC categories

russellwheatley:
SPM was resolving firebase-ios-sdk upToNextMajor from 12.18.0 and landing on 12.19.0, which hung or crashed live Remote Config e2e. CocoaPods stayed on exact 12.18.0.

This pins the SPM package to the declared SDK version and keeps `-ObjC` on RNFB pods so Measurement categories are not stripped.

---
Maintainer note: Fixes internal CPRN-447


mikehardy (approved):
LGTM, no review findings

PR #9321, opened 2026-09-14
fix(app,app-check,auth): detect UIScene lifecycle in iOS config plugins

gabrieldonadel:
### Description

Expo SDK 58 adopts the UIKit scene life cycle, which Xcode 27 makes mandatory. The generated `AppDelegate.swift` no longer sets `moduleName`, no longer calls `factory.startReactNative`, and no longer declares `application(_:open:options:)`. All three moved into a new `SceneDelegate.swift`, and the SDK 58 template also writes `internal import Expo` instead of a bare `import Expo`.

Every iOS insertion point these config plugins relied on disappeared. Against an SDK 58 project, `expo prebuild` produced:

```
» @react-native-firebase/app-check: Unable to determine correct insertion point in AppDelegate.swift. Skipping App Check addition.
» @react-native-firebase/app: Unable to determine correct Firebase insertion point in AppDelegate.swift. Skipping Firebase addition.
✖ Prebuild failed
Error: Failed to apply iOS openURL fix because no 'openURL' method was found
```

**app / app-check.** Both plugins gain a `didFinishLaunchingWithOptions` fallback anchor. The offset into the method body is computed from the file instead of hardcoded, so a signature wrapped across several lines still resolves. The `factory.startReactNative` anchor stays first, which keeps app-delegate projects on the exact insertion point they have today. `FirebaseApp.configure()` deliberately stays in the app delegate rather than moving to the scene delegate: it must run before any scene connects, so the GoogleUtilities app and scene delegate swizzlers are installed in time.

**auth.** The plugin now detects the life cycle and patches the matching file. On scene-based projects the reCAPTCHA redirect filter moves to `SceneDelegate.swift`, because `ExpoAppSceneDelegate` forwards every incoming URL to `RCTLinkingManager` and would otherwise let `firebaseauth://` redirects reach the router. Detection uses two independent signals, since either can be missing depending on which mods have run: a generated `SceneDelegate.swift`, or `UIApplicationSceneManifest` in the app `Info.plist`. The two openURL fixes are mutually exclusive and resolve at prebuild time, so the app-delegate path now defers silently instead of warning or throwing once it sees a scene-based project.

One known limit, called out here so it is not mistaken for an oversight: the scene override covers the warm path only. Cold-start URLs are routed by `ExpoAppSceneDelegate.scene(_:willConnectTo:)` through a static helper that a subclass cannot intercept. This is not a gap in practice, because the reCAPTCHA redirect always returns to an app that is already running.

Projects on the app-delegate life cycle are unaffected. Every pre-existing plugin snapshot is byte-identical.

### Release Summary

Support the Expo SDK 58 UIScene life cycle in the iOS config plugins for `app`, `app-check`, and `auth`.

### Checklist

  - [x] Yes
  - [ ] `Android`
  - [x] `iOS`
  - [ ] `Other` (macOS, web)
- My change includes tests;
- This is a breaking change;
  - [ ] Yes
  - [x] No

### Test Plan

Automated:

```
yarn tests:jest packages/app/plugin packages/app-check/plugin packages/auth/plugin
```

New `AppDelegate_sdk58.swift` and `SceneDelegate_sdk58.swift` fixtures drive snapshot tests for `app` and `app-check` against the SDK 58 template. `packages/auth/plugin/__tests__/iosPlugin_sceneDelegate.test.ts` covers scene detection (nested project layout, the `apps/bare-expo` root layout, `Pods` and `build` exclusion, the `Info.plist` signal, and the negative case), the `SceneDelegate.swift` patch and its idempotency, and the app-delegate plugin deferring silently on a scene-based project.

The existing app-delegate snapshots are unchanged, which is the regression check for projects on SDK 57 and earlier.

Manual:

1. Create an Expo SDK 58 project and install `@react-native-firebase/app`, `app-check`, and `auth` with `captchaOpenUrlFix` enabled.
2. Run `npx expo prebuild --clean -p ios`. It completes with no skip warnings and no openURL error.
3. Confirm `ios/<Project>/AppDelegate.swift` contains `FirebaseApp.configure()` and the App Check provider inside `didFinishLaunchingWithOptions`, and `ios/<Project>/SceneDelegate.swift` contains the `@react-native-firebase/auth-openURL` block.
4. Run the same steps on an SDK 57 project and confirm the generated files are unchanged from the current release.
5. Sign in with a provider that uses the reCAPTCHA flow and confirm the redirect returns to the app without the router handling the `firebaseauth://` URL.


packages/auth/plugin/src/ios/openUrlFix.ts
```diff
@@ -64,6 +65,11 @@ export function withOpenUrlFixForAppDelegate({
 
   const newContents = modifyAppDelegate(contents, language);
   if (newContents === null) {
+    // Under the UIScene life cycle (Xcode 27+) UIKit stops calling `application(_:open:options:)` and the template drops the method
+    // `withIosCaptchaSceneDelegateFix` patches `SceneDelegate.swift` for those projects.
+    if (usesSceneLifecycle(config.modRequest.platformProjectRoot)) {
```
russellwheatley: This returns silently for Info.plist-only projects too, not just when `SceneDelegate.swift` is present.

`withIosCaptchaSceneDelegateFix` no-ops if `findSceneDelegateFile` is null, so both plugins can miss. I think silent defer should require the file. Plist-only can still skip the throw, but it should warn, or the scene plugin should warn when the manifest is there and the file isn't.

packages/auth/plugin/__tests__/iosPlugin_sceneDelegate.test.ts
```diff
@@ -0,0 +1,195 @@
...
+      const result = withOpenUrlFixForAppDelegate({
+        config: makeConfig(appDelegate, tmpDir),
+        props: undefined,
+      });
+
+      expect(result.modResults.contents).toBe(appDelegate);
+      expect(spy).not.toHaveBeenCalled();
+    });
+
+    it('does not throw when captchaOpenUrlFix is forced on and the project uses scenes', async () => {
```
russellwheatley: This always creates `SceneDelegate.swift`, so it doesn't cover the Info.plist-only signal.

Can you add a case with the manifest, no scene file, no `openURL`, and `captchaOpenUrlFix: true`? Right now that will not throw and will not warn. Also worth asserting `contents` is unchanged here, not only `not.toThrow()`.

russellwheatley (changes requested):

packages/auth/plugin/src/ios/openUrlFix.ts
gabrieldonadel: Sure, I changed it so silent defer requires `SceneDelegate.swift`:

- `findSceneDelegateFile` non-null → return silently, `withIosCaptchaSceneDelegateFix` owns it.
- Manifest in Info.plist but no scene file → warn and skip the throw. Neither plugin can patch anything, so the user hears about it.
- Neither signal → unchanged: throw when forced on, warn otherwise.

Also dropped `usesSceneLifecycle` given nothing needs the combined boolean anymore

packages/auth/plugin/__tests__/iosPlugin_sceneDelegate.test.ts
gabrieldonadel: Added a new case: manifest present, no `SceneDelegate.swift`, no `openURL`, `captchaOpenUrlFix: true`. 
It asserts that contents are unchanged and that only one warning fires with the "no SceneDelegate.swift was found" message.

Also tightened the forced-on scene test to assert contents are unchanged and that no warning fires, instead of only `not.toThrow()`.

Issue #9322, opened 2026-09-15
[🐛] SPM + dynamic frameworks (Expo SDK 57): GoogleUtilities classes duplicated in RNFBAnalytics / RNFBMessaging


LorenzHi:
## Issue

With SPM mode and `ios.useFrameworks: "dynamic"` (Expo SDK 57), the app builds and launches, but the Objective‑C runtime reports duplicate classes at startup: several GoogleUtilities classes are implemented both in the SPM package‑product framework and statically inside `RNFBAnalytics.framework` (and, for Reachability, also inside `RNFBMessaging.framework`).

`FIRApp` is not duplicated (so the FirebaseCore part of #9140 looks fixed), but GoogleUtilities pulled in via FirebaseAnalytics/FirebaseMessaging still seems to be linked statically into the RNFB pod frameworks. We observed no crash in a short smoke run, but the runtime warns about "spurious casting failures and mysterious crashes".

**Console (excerpt)**

```
Class GULNetwork is implemented in both …/Frameworks/GULAppDelegateSwizzler_-24B6F7B546001A5C_PackageProduct.framework/GULAppDelegateSwizzler_-24B6F7B546001A5C_PackageProduct and …/RNFBAnalytics.framework/RNFBAnalytics. This may cause spurious casting failures and mysterious crashes. One of the duplicates must be removed or renamed.
```

Affected classes in `RNFBAnalytics`:
`GULMutableDictionary`, `GULNetwork`, `GULNetworkURLSession`, `GULNetworkURLSessionWeakHolder`, `GULReachabilityChecker`, `GULSessionDeallocTracker`, `GoogleUtilities_GoogleUtilities_Network_SWIFTPM_MODULE_BUNDLER_FINDER`, `GoogleUtilities_GoogleUtilities_Reachability_SWIFTPM_MODULE_BUNDLER_FINDER`

In `RNFBMessaging`:
`GULReachabilityChecker`, `GoogleUtilities_GoogleUtilities_Reachability_SWIFTPM_MODULE_BUNDLER_FINDER`

**Steps to reproduce**

1. Expo SDK 57 app with the RNFB modules listed below; `@react-native-firebase/app` config plugin without `disableSPM`; `expo-build-properties` with `ios.useFrameworks: "dynamic"` (no RNFB entries in `forceStaticLinking`).
2. `npx expo prebuild --clean`, `pod install`, build Release for the iOS Simulator, launch and watch the console.

[...]

PR #9326, opened 2026-09-18
fix(ios): resolve duplicate GUL/Firebase classes under Expo SPM dynamic frameworks

russellwheatley:
Demonstrates a fix for duplicate `GUL*`/Firebase classes compiled into both `RNFBAnalytics.framework` and `RNFBMessaging.framework` under Expo's SPM + `useFrameworks: dynamic` path (#9322), using the local dynamic umbrella a firebase-ios-sdk maintainer suggested (since firebase-ios-sdk itself can't ship one, firebase/firebase-ios-sdk#8945).

Once Firebase goes through the umbrella, the individual Firebase headers are hidden from Analytics's and Messaging's `.mm`/`.m` files, so their `__has_include` guards fall back to `@import FirebaseAnalytics`/`@import FirebaseMessaging`, illegal in Objective-C++ without `-fcxx-modules` (which in turn breaks React Native's own build).

To get a clean build under the umbrella:
- Move every direct Firebase Analytics/Messaging call out of `.mm`/`.m` and behind a small Swift facade (`RNFBAnalyticsFacade`, `RNFBMessagingFacade`), the same pattern `RNFBFunctionsCallHandler` already used.
- Rewrite the FCM token-refresh delegate in Swift, split into an internal implementation (conforms to Firebase's `MessagingDelegate`) and a thin public wrapper, since a public Swift type conforming to a hidden framework's protocol forces Swift to add its own illegal `@import` to the generated header.
- Add a same-pod Objective-C bridge (`RNFBMessagingEventEmitter`) so the Swift delegate can still reach `RNFBApp`'s `RNFBRCTEventEmitter`, since Swift can't `import` a sibling RNFB pod directly (every RNFB pod marks nearly all its headers private).
- Give `RNFBAnalytics` an explicit framework search path and `-framework GoogleAppMeasurement`: `FirebaseAnalytics` resolves to a prebuilt XCFramework rather than plain source, so it doesn't link automatically under the umbrella.

`yarn test-expo:ios:link` now builds clean, and a new `assert_dynamic_umbrella_graph` check confirms none of `RNFBApp`/`RNFBAnalytics`/`RNFBMessaging` define their own `FIRApp`/`GUL*` symbols, and all three link the shared dynamic `RNFBFirebase` umbrella rather than embedding their own copy.

This is a draft for discussion, not ready to merge: `RNFBApp` has its own separate fix and isn't reworked here, and this only covers the Expo/SPM + dynamic frameworks path.


russellwheatley:
Not the path for #9322.

We're going with explicit GoogleUtilities SPM products on the RNFB podspecs that still compile those classes in, plus an nm closer on the Expo documented-path graph. See #9327.

The first-party FIRApp duplicate (app target + RNFBApp) is a separate problem from #9140, not this spike.

